PR #7, opened 2021-11-11
Add Build system components to Staging

sbose78:
### Components of the App Studio Build System

In *production*, this should probably be on tenant clusters only.

The App Studio Build System is composed of the following 
components:

1. OpenShift Pipelines. 
2. AppStudio-specific Pipeline Definitions in `build-templates` for building images. <-- there would be multiple ways of getting these in here.
3. AppStudio-specific `ClusterTasks`.

Long-term, (2) and (3) would be pulled from some remote catalog. 

As a non-admin user, one would have access to the Pipeline definitions in `build-templates`. 

To be able to use them in one's personal namespace, the UI or the client would need to run the equivalent of the following command:

### Usage

```
oc get pipelines -n build-templates -o yaml | sed 's/namespace: .*/namespace: YOUR-NAMESPACE/' | oc apply -f -
```

### Testing out this PR

Install OpenShift GitOps and create the following `Application`
 
```
kind: Application
metadata:
  name: build
  
spec:
  project: default

  source:
    path: components/build
    repoURL: https://github.com/sbose78/infra-deployments.git
    targetRevision: build-service-init ## <-- points to my fork's branch

  destination:
    namespace: default # not relevant yet since every object has an explicit namespace.

  syncPolicy:

    # Comment this out if you want to manually trigger deployments (using the 
    # Argo CD Web UI or Argo CD CLI), rather than automatically deploying on
    # every new Git commit to your directory.
    automated: 
      prune: true
      selfHeal: true

    syncOptions:
    - CreateNamespace=true

    retry:
      limit: 50 # number of failed sync attempt retries; unlimited number of attempts if less than 0
      backoff:
```

Since this doesn't use sync waves yet, the subscription.yaml might need a nudge since it needs to be sync'd first.


jgwest (approved):
LGTM

PR #9, opened 2021-11-16
Add "view" in relevant namespaces to build and gitops members

sbose78:
## Why

* Devs/Leads needs to be able to "view" resources in control plane namespaces ( builds, gitops..). This could be slimmed down further in future.

## What

If a user is a member of `redhat-appstudio-sre/stage`, they would be able to login and create their own projects. If explicit rolebindings are created, they would get to 'view' resources in specific namespaces ( build , HAS, gitops control planes ).

## How to test

0. Install OpenShift GitOps, and switch to `openshift-gitops` namespace.
1. Apply https://github.com/redhat-appstudio/infra-deployments/blob/main/argo-cd-apps/base/build.yaml on your cluster.
2. Apply https://github.com/redhat-appstudio/infra-deployments/blob/main/argo-cd-apps/base/gitops.yaml on your cluster.
3. Apply the following Application in `openshift-gitops` namespace
```
kind: Application
metadata:
  name: authentication
  
spec:
  project: default

  source:
    path: components/authentication
    repoURL: https://github.com/sbose78/infra-deployments.git
    targetRevision: staging-auth #<----- from my fork

  destination:
    namespace: default # not relevant yet since every object has an explicit namespace.

  syncPolicy:

    # Comment this out if you want to manually trigger deployments (using the 
    # Argo CD Web UI or Argo CD CLI), rather than automatically deploying on
    # every new Git commit to your directory.
    automated: 
      prune: true
      selfHeal: true

    syncOptions:
    - CreateNamespace=true

    retry:
      limit: 50 # number of failed sync attempt retries; unlimited number of attempts if less than 0
      backoff:


```


components/authentication/view-build-service.yaml
```diff
@@ -0,0 +1,29 @@
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: read-build
+  namespace: openshift-pipelines
+subjects:
```
sbose78, line 6: @guillaumerose and @vdemeester has access to `openshift-pipelines` namespace.

components/authentication/view-build-service.yaml
```diff
@@ -0,0 +1,29 @@
...
+  name: view
+---
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: read-build-templates
+  namespace: build-templates
+subjects:
+  - kind: User
+    name: jduimovich
```
sbose78, line 23: @jduimovich and I will have access to the pipeline catalog.

components/authentication/view-gitops-service.yaml
```diff
@@ -0,0 +1,16 @@
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: read-gitops-service
+  namespace: gitops
+subjects:
```
sbose78, line 6: @jgwest , @samyak-jn and I will have access to gitops control plane namespace.

components/authentication/github-client.yaml
```diff
@@ -0,0 +1,34 @@
...
+  namespace: openshift-gitops
+---
+apiVersion: config.openshift.io/v1
+kind: OAuth
+metadata:
+  name: cluster
+spec:
+  identityProviders:
+    - github:
+        clientID: 6eb113c35c61ae10ccd5
```
sbose78, line 25: This will need to change before merge.

sbose78:
Please squash and merge - whoever merges this :)

jgwest (approved):
LGTM, thanks @sbose78!

PR #11, opened 2021-11-17
windows scripts for bootstrap and destroy

jduimovich:
I ported the bootstrap and destroy scripts to Windows for use with CRC. 

The scripts match the original scripts and were tested on:
```
OpenShift version: 4.9.5 (bundle installed at C:\Program Files\CodeReady Containers\crc_hyperv_4.9.5.crcbundle)
```

### DIFF
--- a/hack/bootstrap-cluster.cmd
+++ b/hack/bootstrap-cluster.cmd
@@ -1,0 +1,41 @@
+@echo off 
+
+set ROOT=%~dp0\..
+
+echo "Installing the OpenShift GitOps operator subscription:"
+kubectl apply -f %ROOT%/openshift-gitops/subscription-openshift-gitops.yaml
+
+echo.
+echo "Waiting for default project (and namespace) to exist:" 
+:L1
+choice /D N /T 1  > nul
+kubectl get appproject/default -n openshift-gitops 2>nul 
+if %ERRORLEVEL% == 1 goto :L1 
+
+echo.
+echo "Waiting for OpenShift GitOps Route:"
+:L2
+choice /D N /T 1  > nul
+kubectl get route/openshift-gitops-server -n openshift-gitops 2>nul 
+if %ERRORLEVEL% == 1 goto :L2
+ 
+echo.
+echo "Add Role/RoleBindings for OpenShift GitOps:"
+kustomize build %ROOT%/openshift-gitops/cluster-rbac | kubectl apply -f -
+ 
+echo.
+echo "Add parent Argo CD Application:"
+kubectl apply -f %ROOT%/argo-cd-apps/app-of-apps/all-applications-staging.yaml
+ 
+echo.
+echo "========================================================================="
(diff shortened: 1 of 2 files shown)


sbose78 (approved):

jgwest:
@jduimovich I'm hesitant to include this, as it create an expectation that this is fully supported and would require us to maintain two sets of scripts (unless you are willing to add your name to the file as the maintainer and to keep them up to date yourself). 

But also, can't Windows users can just use the existing scripts with WSL?

sbose78:
I'd say, it would be good to have @jduimovich as the maintainer of the windows scripts. 🚨

jduimovich:
> But also, can't Windows users can just use the existing scripts with WSL?

I tried WSL and I was getting errors with weird paths and filenames mixing G:\path and /mnt/g/path but perhaps we could fix by carefully writing the scripts. I can investigate because having two scripts is a pain. It may be that windows and linux binaries are being mixed as you can directly run kubectl.exe from shell.    The other issue is cr/lf messes in WSL, if you use windows / vscode and run in WSL you get a bunch or "\r" errors and the scripts fail. 
If I can get WSL to work we can fix.

jgwest:
@jduimovich My guess is if the Linux x64 version of the kubectl binary is installed in WSL, and that is what is first on your PATH,  that should resolve the issues you are seeing re: supported path formats.

And yes, CR/LF issues with Git were always terrible when I was on Windows, and `autocrlf` never seemed to do what I wanted in all cases, so I fully sympathize :upside_down_face:. If you git clone from inside WSL that should give you the line endings you need, but there's no guarantee that editing the text from within Windows won't mess that up again.

jduimovich:
I tested it and the kubectl was fine after many tries to find which one was
running :)
Turns out the kustomize prints an error about checking symlinks which come
up as "g:\mnt" and messes from there.
I deleted all my binaries and reloaded and it seems to work so we may be
able to get away with WSL only.
The CR/LF really only seems to hurt the scripts for now, and yes, checkout
in windows breaks WSL, checkout in WSL and is ok.
The issue is depending on where you check it out, it may show all (24+)
files changed if you look from the "other side"
The simple fix is rm, and checkout again and it doesn't show modified :) If
you dos2unix the files they also claim to be changed ala cr/lf differences

I think I'm ok without the windows scripts for now.
I have a new script which I'm experimenting with local dev which redirects
the gitops repositories via a patch and a  kustomize directory overlay. It
means that we can leave the originals that target sandbox in place to
minimize the risk that someone sends in changes which includes references
to their private repo when testing components.

jduimovich:
I have verified WSL is proven to work (with messy cr/lf and other "nits") but the scripts work.  The solution was to ensure recent versions of kubectl, kustomize. some random version of kustomize had issues with symlinks for some reason. I deleted all and found fresh, reinstalled and scripts started to work.  I am closing without merge

PR #16, opened 2021-11-19
development mode scripts

jduimovich:
This pr has a set of scripts and configurations that help with using this repository with your own cluster. 

Once you bootstrap, the root ArgoCD "app-of-apps", each of the component applications will each point to the upstream repository.  To enable development  test changes on your own cluster using the same gitops used to create staging, you replace the references to `https://github.com/redhat-appstudio/infra-deployments.git` with references to your own fork. The use of gitops means you can also easily share your config with others using the same bootstrap mechanism as staging. 

This PR includes a set of scripts that help with this, and minimize the changes needed in a fork. 
See the main README for usage instructions.
The usage of this PR required yq to update manifests see https://github.com/mikefarah/yq
 
The PR  
1) enables use of forks to test via gitops in dev environments
2) A script `hack/development-mode.sh` will update a single file in your fork to redirect gitops to that fork. This file must be committed to your fork. This uses kustomize overlays to map the repositories per component to your fork. You can also just replace a reference to your own component and keep the rest from upstream. As this file is committed back to your fork so care is needed not to include it in pull requests upstream however it accidentally included upstream, it will be "harmless" (default forks of upstream will use your fork until replace in dev mode) 
3) users can switch from dev mode back to upstream mode
4) Use of a separate overlay directory is intended to minimize the risk of a PR including accidental mods to staging and only one file is modified.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,31 @@ Steps:
 3) Log-in to the Web UI using your OpenShift credentials (using 'Login with OpenShift' button).
 3) View the Argo CD UI to see the status of deployments.
 
+## Development mode for your own clusters
+
+Once you bootstrap a cluster above, the root ArgoCD Application and all of the component applications will each point to the upstream repository.
+
+To enable development for a team or individual to test changes on your own cluster, you need to replace the references to `https://github.com/redhat-appstudio/infra-deployments.git` with references to your own fork.
+
+There are a set of scripts that help with this, and minimize the changes needed in your forks.
+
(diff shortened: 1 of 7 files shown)


jgwest (approved):
LGTM, thanks @jduimovich!

amfred (commented):
These are a good idea, but I haven't had a chance to test them yet; I'm still debugging my CodeReady Workspaces bootstrapping process.

amfred:
I see that Jonathan beat me to it. :-)

PR #17, opened 2021-11-22
added CRC tips

amfred:
Added a couple of details that are needed to get this running in a clean CRC install.
Tested with:
OpenShift version: 4.9.5


jduimovich (approved):
The memory limits and configuration are similar to my CRC config on macOS.
LGTM

PR #21, opened 2021-11-24
Create a pipline maintainer role

sbose78:
This role allows users to elevate themselves to the role of a CI maintainer without granting unnecessary privileges:
* create secrets ( eventually would be replaced by SPI generated content )
* modify the 'pipeline' service account : would enable specification of pull secrets.

This PR grants @sbose78 pipeline-maintainer in the following namespaces:
* application-service
* gitops

PR #27, opened 2021-11-24
Add ci for HAS

sbose78:
## Why?
We got to dogfood our current tekton service and while doing so "dry run" what app studio would be creating for users' projects. 

## What
On push to any branch in the https://github.com/redhat-appstudio/application-service repo, a new tekton build would be triggered in the staging cluster itself. An image tagged with the git sha would be pushed to Quay. **This wouldn't conflict with the existing tags/tag format.**

## How

* Emulated what App Studio would do: Pull a pipeline definition from the catalog into one's own namespace: components/has/.tekton/pipeline.yaml .  This would be same for every component.

* Created the "webhook" related objects.

## dependencies

For this to work, 
* I would be creating the image push `secret` on the cluster and linking the same from the pipeline sa. I've created a robot account in Quay for the same.
* To do the above, we need the https://github.com/redhat-appstudio/infra-deployments/pull/21 role.
This only builds the controller image, nothing OLM yet.

### DIFF
--- a/components/has/.tekton/event-listener.yaml
+++ b/components/has/.tekton/event-listener.yaml
@@ -1,0 +1,12 @@
+apiVersion: triggers.tekton.dev/v1alpha1
+kind: EventListener
+metadata:
+  name: application-service
+spec:
+  serviceAccountName: pipeline
+  triggers:
+    - bindings:
+        - kind: ClusterTriggerBinding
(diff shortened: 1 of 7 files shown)


sbose78:
Self-approving because this is non-interfering change.

PR #29, opened 2021-11-26
feat: Initial SPI components deployment

skabashnyuk:
## Details 
This is a initial deployment if SPI components
 - `service-provider-integration-operator` https://github.com/redhat-appstudio/service-provider-integration-operator
 - `service-provider-integration-oauth` https://github.com/redhat-appstudio/service-provider-integration-oauth

Знімок екрана 2021-12-30 о 10 42 52
Знімок екрана 2021-12-30 о 12 10 39
Знімок екрана 2021-12-30 о 12 11 10

Now both components are at the initial stage of **SPI V2** implementation  https://docs.google.com/document/u/1/d/1Agp8WMn5uSyzJBin1JfbgCIlKbGt5ziHV_aF2wT5eKc/edit?usp=drive_web&ouid=110694458140179268052
 - `service-provider-integration-operator` provides CRDs https://docs.google.com/document/d/1ymo3DgBKdKAq0z1SNgM6PcOqpJyi4EDjO0HVBWPrYn4/edit
 - `service-provider-integration-oauth` implemented OAuth2 flow for Github.

## NOTICE
In this pr OAuth service configuration is represented as ` spi-oauth-config-xxx` and `spi-shared-secret` contains dummy data. In the future, it has to be replaced with some production-grade provider of secrets. See also https://github.com/redhat-appstudio/infra-deployments/pull/46

### DIFF
--- a/components/spi/kustomization.yaml
+++ b/components/spi/kustomization.yaml
@@ -1,0 +1,16 @@
+resources:
+  - https://github.com/redhat-appstudio/service-provider-integration-operator/config/default?ref=d7a6475c6d948452cdc00941616024c9ef6607cc
+
+
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+images:
+  - name:  quay.io/redhat-appstudio/service-provider-integration-operator
+    newName: quay.io/redhat-appstudio/service-provider-integration-operator
+    newTag: v0.2.0
+  - name: quay.io/redhat-appstudio/service-provider-integration-oauth
+    newName: quay.io/redhat-appstudio/service-provider-integration-oauth
+    newTag: v0.2.0
+
+namespace: spi-system
--- a/components/spi/kustomization.yaml
+++ b/components/spi/kustomization.yaml
@@ -1,5 +1,5 @@
 resources:
-  - https://github.com/redhat-appstudio/service-provider-integration-operator/config/default?ref=d7a6475c6d948452cdc00941616024c9ef6607cc
+  - https://github.com/redhat-appstudio/service-provider-integration-operator/config/default?ref=4f8d0b4d4e9980518f9a759f7683639e025bc32c
 
 
 apiVersion: kustomize.config.k8s.io/v1beta1
--- a/argo-cd-apps/base/kustomization.yaml
+++ b/argo-cd-apps/base/kustomization.yaml
@@ -3,6 +3,7 @@ resources:
 - build.yaml
 - authentication.yaml
 - has.yaml
+- spi.yaml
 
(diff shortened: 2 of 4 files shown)


jduimovich (commented):
The name seems to be in a strange font.  Unclear this is intentional or whether it will even cause issues so this is an FYI  See this
 image

skabashnyuk:
> The name seems to be in a strange font. Unclear this is intentional or whether it will even cause issues so this is an FYI See this image

@jduimovich I'm still experimenting with vault deployment. After a few tries, I've cleaned up these charts.

skabashnyuk:
I've added a **draft** mark for this pr. The reason is that this pr has to be reworked in the section of operator deployment. It has to use regular deployment instead of OLM

skabashnyuk:
@sbose78 @jduimovich pr un-drafted and ready to review. The thing that I'm not sure about is about the cert-manager. Maybe it's better to rework it this way https://docs.openshift.com/container-platform/4.6/security/certificates/service-serving-certificate.html#add-service-certificate-validating-webhook_service-serving-certificate

sbose78 (commented):
Could you check if you've done the equivalent of https://github.com/redhat-appstudio/application-service/pull/23 ?

skabashnyuk:
>Could you check if you've done the equivalent of redhat-appstudio/application-service#23 ?

@sbose78 done in https://github.com/redhat-appstudio/service-provider-integration-operator/pull/51

sbose78:
Thanks, I will review!

components/spi/kustomization.yaml
```diff
@@ -0,0 +1,16 @@
+resources:
+  - https://github.com/redhat-appstudio/service-provider-integration-operator/config/default?ref=d7a6475c6d948452cdc00941616024c9ef6607cc
```
sbose78, line 2: We'll need https://github.com/redhat-appstudio/service-provider-integration-operator/commit/4f8d0b4d4e9980518f9a759f7683639e025bc32c ?

sbose78 (changes requested):

sbose78 (approved):

flacatus:
I see the e2e are failing with an ?unexpected? error in has component:

https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/redhat-appstudio_infra-deployments/29/pull-ci-redhat-appstudio-infra-deployments-main-appstudio-e2e-deployment/1481492363904815104/artifacts/appstudio-e2e-deployment/gather-extra/artifacts/pods/application-service_application-service-controller-manager-64bbb845f-rchss_manager.log

PR #31, opened 2021-11-30
Build updates

jduimovich:
This PR includes a set of build pipeline updates

    - working nodejs builds, java may also build some repos
    - more test scripts to run builds, determine buildable repos
    - a new simple noop pipeline for quick test runs 
    - utils for prune and cleanup of pipelines and pvc
    - util to determine build type for a repo (temporary)
    - doc README updates

### DIFF
--- a/components/build/build-templates/docker-build.yaml
+++ b/components/build/build-templates/docker-build.yaml
@@ -15,6 +15,10 @@ spec:
     - description: 'Fully Qualified Output Image'
       name: output-image
       type: string
+    - default: .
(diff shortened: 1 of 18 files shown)


components/build/build-templates/docker-build.yaml
```diff
@@ -15,6 +15,10 @@ spec:
     - description: 'Fully Qualified Output Image'
       name: output-image
       type: string
+    - default: .
```
sbose78, line 18: 👍 thanks

PR #33, opened 2021-12-01
Update version pinning of tasks and new scripts 

jduimovich:
- Openshift Pipelines 1.6 caused pinned version 1-5-0 missing in 1.6 bug
-  update test to build all three container, node.js (container and nodejs) and springboot java  example
- update scripts to manage build, autobuild and cleanup of pipelineruns and storage
- update build and cleanup scripts to use pv- name 
- new deploy.sh to deploy build image examples to prove working builds
- WIP for bundling of tekton tasks from build templates instead of scraping from NS


sbose78 (approved):
Lovely!

PR #36, opened 2021-12-06
pipelines into bundles and devfile builder

jduimovich:
This pull request supports the Build Contract API

- supports Build Contract annotations for build definitions
- adds devfile builder
- uses bundles to deliver App Studio pipelines  
- adds test scripts to build all milestone 2 repos including deployment 
- adds deployment scripts to test containers
- adds sample for pruning and pvc maintenance
- adds doc to README on new scripts


sbose78 (changes requested):
components/build/build-templates/analyze-devfile.yaml

Shouldn't this be in ClusterTasks ?

jduimovich:
> components/build/build-templates/analyze-devfile.yaml
> 
> Shouldn't this be in ClusterTasks ?

There is no reason to move it yet. You end up adding three new files and we still need to refactor this directory so that the OCI image is created in an external repo (redhat-appstudio/build-service ?) because the pipelines no-longer need to be pre-installed into the cluster. The redhat-appstudio/build-service  is where the ClusterTasks images are being created and perhaps the default pipelines move there too . I was waiting to do the cleanup once we move the pipelines.

sbose78 (approved):

sbose78:
> There is no reason to move it yet. 

There is a folder precisely for cluster tasks, that's why 🤷‍♂️ . Sure, we can remove them in the next commit but it doesn't hurt to have a historical record of how things were laid out initially.

PR #39, opened 2021-12-10
Consistently uses ./hack

amfred:
There were a couple of places where cutting-and-pasting the examples didn't work, such as:
```
% .hack/build/build.sh https://github.com/jduimovich/single-container-app noop
zsh: no such file or directory: .hack/build/build.sh
```
Consistently using ./hack is safe.

Tested on 
OpenShift version: 4.9.5 (bundle installed at /Applications/CodeReady Containers.app/Contents/Resources/crc_hyperkit_4.9.5.crcbundle)

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ If you don't already have a test OpenShift cluster available, CodeReady Containe
 
 ### Bootstrap App Studio
 Steps:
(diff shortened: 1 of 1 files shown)


jduimovich (approved):
LGTM 
Thanks, I was planning a doc update as some scripts as the build definitions mechanisms evolve.  Also noticed some wording needs fixing so when the doc is updated I'll do that.

PR #47, opened 2021-12-16
Patch the pipeline sa to use the appropriate secret

sbose78:
.. for the gitops service


sbose78:
self-approving to be able to get CI working.

PR #48, opened 2021-12-16
ensure pipeline sa always has the push secret specified

sbose78:
The pipline sa was manually patched previously to contain the quay.io push secret


sbose78:
Self approving as this is causing build failures.

PR #52, opened 2022-01-06
Refresh HAS in Staging

johnmcollier:
Refreshes HAS in staging to pick up https://github.com/redhat-appstudio/application-service/commit/a7d0420d5c5a45f1240bf51618256194f2f5dfc4


components/has/kustomization.yaml
```diff
@@ -1,6 +1,6 @@
 resources:
 - allow-argocd-to-manage.yaml
-- https://github.com/sbose78/application-service/config/default?ref=2a012d849c344153e3884bf1a37d30339e114fac
+- https://https://github.com/redhat-appstudio/application-service/tree/main/config/default?ref=a7d0420d5c5a45f1240bf51618256194f2f5dfc4
```
ralphbean, line 3:
```suggestion
- https://github.com/redhat-appstudio/application-service/tree/main/config/default?ref=a7d0420d5c5a45f1240bf51618256194f2f5dfc4
```

johnmcollier: D'oh, thanks for catching that.

ralphbean (approved):

components/has/kustomization.yaml
```diff
@@ -1,6 +1,6 @@
 resources:
 - allow-argocd-to-manage.yaml
-- https://github.com/sbose78/application-service/config/default?ref=2a012d849c344153e3884bf1a37d30339e114fac
+- https://github.com/redhat-appstudio/application-service/config/default?ref=a7d0420d5c5a45f1240bf51618256194f2f5dfc4
```
sbose78, line 3: So embarrassing 😮‍💨 , does imply I definitely testing things though haha

sbose78 (approved):

PR #55, opened 2022-01-12
feat: add scripts for installing Toolchain (Sandbox) operators

MatousJobanek:
Changes introduced in this PR:
* Two scripts for installing Toolchain (Sandbox) operators in two kinds of modes:
    * dev
    * e2e
*  Added part in `destroy-cluster.sh` script for cleaning up the Sandbox stuff
* New `reduce-gitops-cpu-requests.sh` which sets all `resource.request.cpu` values to `50m` in `argocd/openshift-gitops` resource. The reason is that with the default size of the dev cluster, I wasn't able to install Sandbox operators along with the GitOps stuff - there wasn't enough of CPU in the cluster. 
Both of the sandbox-related scripts call this "reduce" script automatically.
* Changes in the README - feel free to fix my poor grammar :sweat_smile:

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,28 @@ Both of the scripts will:
 3. Print:
     - The landing-page URL that you can use for signing-up for the Sandbox environment that is running in your cluster.
     - Proxy URL. 
+    
+#### SSO
+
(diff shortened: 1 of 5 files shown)


jduimovich (approved):
These all look good to me. 
Added a minor readme update to include `yq`  as it was another dependency I had forgotten to document.

PR #56, opened 2022-01-13
Build: Automate build of build-templates-bundle and appstudio-utils

Michkov:
- Add trigger-template which triggers pipelines for building images:
  - redhat-appstudio/build-templates-bundle
  - redhat-appstudio/appstudio-tasks
  - redhat-appstudio/appstudio-utils

Configuration needed on stage:
- [x] Create secret redhat-appstudio-staginguser-pull-secret in build-templates
- [x] Check/add write permission for robot account in quay.io for redhat-appstudio/build-templates-bundle
- [x] Check/add write permission for robot account in quay.io for redhat-appstudio/appstudio-tasks 
- [x] Check/add write permission for robot account in quay.io for redhat-appstudio/appstudio-utils
- [x] Set webhook in build-definitions to trigger the build workflow

### DIFF
--- a/components/build/.tekton/serviceaccount.yaml
+++ b/components/build/.tekton/serviceaccount.yaml
@@ -1,0 +1,6 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: pipeline
+secrets:
+  - name: redhat-appstudio-staginguser-pull-secret
--- a/components/build/.tekton/event-listener.yaml
+++ b/components/build/.tekton/event-listener.yaml
@@ -1,0 +1,12 @@
+apiVersion: triggers.tekton.dev/v1alpha1
+kind: EventListener
+metadata:
+  name: build-service
+spec:
+  serviceAccountName: pipeline
+  triggers:
+    - bindings:
+        - kind: ClusterTriggerBinding
+          ref: github-push
+      template:
+        ref: build-service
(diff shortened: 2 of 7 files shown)


jduimovich (commented):
This looks good but I noticed just now that build definitions repo is incorrect and I had already merged the PR. The incorrect image had the version bump. Note, the pipelines bundle version needs to be updated to v0.1.3 (and built) but the PR included a bump of the utils image version instead (also to v0.1.3)
The can be independently updated but I think it would be better if we just kept them in sync to prevent this. 
Note the build definitions creates two images
Pipelines are in a bundle -> `quay.io/redhat-appstudio/build-templates-bundle:v0.1.x`   (note, when we have 10, we'll need a better mechanism)
Base image for multiple cluster tasks -> `quay.io/redhat-appstudio/appstudio-utils:v0.1.x ` -- basically only really needs update if new tools or scripts added to it. 
We can merge this when build-definitions is updated

Michkov:
https://github.com/redhat-appstudio/build-definitions/pull/14 should fix it, but it's temporary this way is hard to manage.

components/build/.tekton/serviceaccount.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: pipeline
+secrets:
+  - name: redhat-appstudio-staginguser-pull-secret
```
sbose78, line 6: I can add this secret to the cluster

Michkov: Could you also check quay robot permission from description? Or even that webhook?

jduimovich (approved):
LGTM

PR #58, opened 2022-01-14
Development mode: Custom GitHub organization

Michkov:
* Allow to set custom GitHub organization in development mode.
* Add arguments to ./hack/development-mode.sh for setting custom git
  repository.

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,7 +119,26 @@ One option to prevent accidentally including this modified file, you can run the
 After you commit your changes you can rerun to `./hack/development-mode.sh` and reset your repo to point back to the fork. 
 
 Note running these scripts in a clone repo will have no effect as the repo will remain `https://github.com/redhat-appstudio/infra-deployments.git`
- 
+
+### Optional: Configure HAS GitHub Organization
+
+After deployment `has` application is failing to start. It's trying to connect to default github organization and credentials are not set.
+
+To run HAS in development mode, you need to set custom GitHub organization and token.
+
+Steps:
+1) Create organization in GitHub
+2) Create user token with permissions:
+    - `repo`
+    - `delete_repo`
+3) Set environment variables:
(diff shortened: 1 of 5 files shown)


Michkov:
@johnmcollier @sbose78 @jduimovich Please take a look. Way how to setup custom github org in development mode.

jduimovich (approved):
Nice patch - the the changes work (tested) and has now syncs.
image

Some updates we could do now (or in future) .
1. The name of the util script is `./hack/util-set-gitlab-org` but it sets the github org - confusing, especially if we later add gitlab.
2. I would also prefer that this be a little more automatic and repeatable.  Some of the steps are manual. My suggestion is we configure some env vars similar to `./hack/build/utils/install-secrets.sh` which uses the env vars MY_GITHUB_USER, MY_GITHUB_TOKEN,  MY_QUAY_USER, MY_QUAY_TOKEN  which we could mandate for dev mode.
This would allow us to create the secrets automatically in the dev mode scripts
such that dev mode always "works" without any additional steps.
ie add something like this to dev mode 
 ./hack/util-set-gitlab-org $MY_GITHUB_USER
`oc create secret generic has-github-token --from-literal=token="$MY_GITHUB_TOKEN"`\
Also, adding a line to kill the application-service-controller-manager automatically would be nice so no additional steps needed in dev mode ... 

This is ok to merge, as it does work and allows local dev as-is and we can do a follow-on with those updates to simplify dev mode.

Michkov:
@jduimovich 
1) not confusing, just regular bug, renamed
2) now using MY_GITHUB_TOKEN, do not know way how to remove pod or restart deployment after configmap is changed. It cannot be done immediately with execution of `hack/development-mode.sh`, it has to be deleted after changes are applied by gitops.

johnmcollier (approved):

PR #62, opened 2022-01-18
add sbose78 to spi

sbose78:


sbose78:
Self approving so that I can debug..

PR #63, opened 2022-01-18
fix: remapped spi-system view role to github's username of skabashnyuk

skabashnyuk:
remapped spi-system view role to github's username of skabashnyuk

Context :
- https://github.com/redhat-appstudio/infra-deployments/pull/61
- https://github.com/redhat-appstudio/infra-deployments/pull/60


sbose78 (approved):
Hope subjects are

PR #69, opened 2022-01-21
fix(hack): do not fail setting secret if it doesn't exists yet

goldmann:
@Michkov - could you please review? Any specific reason for using `replace` here?


Michkov:
Ah, I thought that I fixed that, probably included only in testing commit which was dropped. 

LGTM, thanks for fix.

AndrienkoAleksandr:
+1 to merge

sbose78 (approved):

PR #78, opened 2022-01-28
Use single PVC per namespace for pipelineRun build service components

AndrienkoAleksandr:
Use single PVC per namespace for pipelineRun build service components

Small demo:

https://www.youtube.com/watch?v=zAPaqh4FH-M

Signed-off-by: Oleksandr Andriienko <oandriie@redhat.com>


components/gitops/.tekton/trigger-template.yaml
```diff
@@ -31,12 +31,6 @@ spec:
...
-                accessModes:
-                  - ReadWriteOnce
-                resources:
-                  requests:
-                    storage: 10Mi
-                #storageClassName: gp2
-                volumeMode: Filesystem
+            persistentVolumeClaim:
+              claimName: app-studio-default-workspace
+            subPath: gitops-service-$(tt.params.git-revision)
```
sbose78, line 36: Can we do a `gitops-service/$(tt.params.git-revision)` ?

components/has/.tekton/trigger-template.yaml
```diff
@@ -31,12 +31,6 @@ spec:
...
-                  - ReadWriteOnce
-                resources:
-                  requests:
-                    storage: 10Mi
-                #storageClassName: gp2
-                volumeMode: Filesystem
\ No newline at end of file
+            persistentVolumeClaim:
+              claimName: app-studio-default-workspace
+            subPath: application-service-$(tt.params.git-revision)
```
sbose78, line 36: Can we do `application-service/$(tt.params.git-revision)`

sbose78: though we can leave it as is since this workspace only has a single component anyway.

sbose78:
@jduimovich is it fine if we merge this and then refactor the sub path in a follow-up PR ? I'd like to clean up the PVs lying around today manually and then have this go in. :) 

> The build-tag is a timestamp

Curious, how would the timestamp's placeholder be passed onto the resource template's workspace subpath?

jduimovich:
Yes, this is fine to merge. The cleanup task can be updated in a follow-on PR.

Not sure we can generate the timestamp in a trigger, but the scripts I did make a pretty name, as I find the generateName: to be awful

jduimovich (approved):
LGTM
- follow on update todo for cleanup task to compute the subPath in a GC

components/has/.tekton/pvc.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: app-studio-default-workspace
+spec:
+  accessModes:
+  - ReadWriteOnce
+  resources:
+    requests:
+      storage: 1Gi
```
goldmann, line 10: How do we want control the size of the workspace? The default 1GB can be sufficient for many projects, but some will require much more space, especially if there are a few runs left on the workspace.

sbose78: Good point @goldmann .
We could come up with a guidance ( which we'll need anyway for our users ) - but at the moment, I would leave it to the component owners to make the right call.

sbose78 (approved):

PR #80, opened 2022-01-30
Add bash scripts to demo chains signed taskrun verification

simonbaird:
Following the example set by @jduimovich in hack/build, here are two scripts in hack/chains that demonstrate how to verify signed taskruns.

See commit messages for further explanations.

As mentioned in the first commit, this doesn't work unless chains is installed, but it could be useful while working on the chains configuration, so I think there's no harm in merging early and often.

See also https://docs.engineering.redhat.com/x/zzd-Dw which shows how it works.

JIRA: [HACBS-37](https://issues.redhat.com/browse/HACBS-37)

### DIFF
--- a/hack/chains/check-pipelinerun.sh
+++ b/hack/chains/check-pipelinerun.sh
@@ -1,0 +1,25 @@
+#!/bin/bash
+
+SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+
+PIPELINERUN_NAME=$1
+
+# Preserve sanity while hacking
+set -ue
+
+if [[ -z $PIPELINERUN_NAME ]]; then
+  # Use the most recently created pipelinerun
+  # (Fixme: Would be better to exclude running pipelines)
+  PIPELINERUN_NAME=$(
+    kubectl get pipelinerun -o name --sort-by=.metadata.creationTimestamp |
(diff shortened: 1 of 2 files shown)


simonbaird:
https://docs.engineering.redhat.com/x/zzd-Dw shows how it works (RH internal link)

sbose78:
Sidenote: 
We'll need to update the `buildah` cluster task( upstream?) to adhere to the `result` naming pattern enforced by Chains.

Michkov:
Currently the folder are named by teams, eg. `hack/build` are managed by build team. Unfortunatelly `build` and `release` workstreams/teams are slightly confusing (when we use them as folder names). So not sure if this should not be moved under `hack/release/chains/`.
@jduimovich @sbose78 WDYT?

simonbaird:
If you want to match the workstream names, hack/contract or hack/enterprise-contract could work.

I like the shorter version, and it matches the label in Jira.

jduimovich:
I don't think it needs to be moved. I agree folders called "build" and "release" would be confusing.
The chains work is part of the devsecops work.

simonbaird:
Let me know if I should do anything about the tests. It looks like failures are due to resourcing problems and timeouts.

Also this I expect this patch to have no impact on the tests, so perhaps we can merge it anyway.

simonbaird:
Also team names are ephemeral and might change, so better not to put them in code. +1 to leaving it in hack/chains.

PR #90, opened 2022-02-02
Increase memory and CPU limits for HAS

johnmcollier:
Increases resource limits for HAS in staging.

### DIFF
--- a/components/has/kustomization.yaml
+++ b/components/has/kustomization.yaml
@@ -19,3 +19,6 @@ configMapGenerator:
   behavior: replace
 
 namespace: application-service
+
+patches:
+  - ./manager_resources_patch.yaml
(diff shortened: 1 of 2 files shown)


maysunfaisal (approved):

sbose78 (approved):

sbose78:
Ah system had slipped in

johnmcollier:
I ended up needed to add `namespace: system` back in, otherwise the patch doesn't work correctly (I assume because the resource it's trying to patch, has `namespace: system` set in it too?)

I verified that by running `kustomize build`, the actualy namespace HAS is deployed in is still `application-service` though.

sbose78:
taking one final look before merging.

sbose78 (approved):

PR #91, opened 2022-02-02
Add role for pruning old HAS resources

johnmcollier:
🪓 

Discussed on slack - HAS seems to be crashing on start up, as it tries to reconcile hundreds of resources at once, Shoubhik suggested clearing out any old HAS CRs that have been left behind, as a one-time measure.

### DIFF
--- a/components/authentication/prune-has.yaml
+++ b/components/authentication/prune-has.yaml
@@ -1,0 +1,43 @@
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: has-pruner
+rules:
+  - verbs:
+      - list
+      - get
+      - delete
+      - update
+    apiGroups:
+      - 'appstudio.redhat.com'
+    resources:
+      - applications
+      - components
+      - componentdetectionqueries
+  - verbs:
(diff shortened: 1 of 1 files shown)


components/authentication/prune-has.yaml
```diff
@@ -0,0 +1,43 @@
...
+      - delete
+      - update
+    apiGroups:
+      - 'appstudio.redhat.com'
+    resources:
+      - applications
+      - components
+      - componentdetectionqueries
+  - verbs:
+      - create # allows addition of credentials only.
```
sbose78, line 18: throw in `list` ?

sbose78: oh this is for tekton

sbose78: nope, secrets. 

You don't need secrets and tekton in the ClusterRole, do you?

sbose78 (approved):

components/authentication/prune-has.yaml
johnmcollier: Done.

johnmcollier:
Discussed on slack with @sbose78, merging in now.

PR #94, opened 2022-02-02
Use the kubeadmin user when bootstrapping the cluster/Wait for Argo CD route to be up

Roming22:
The goal of this PR is to streamline the deployment by:

- ensuring the user is logged with the right account (as by default they'll be logged as `developer` if they followed the CRC install documentation)
- Give the user the URL to connect to instead of the route and wait for the URL to be reachable before exiting the script.

### DIFF
--- a/hack/bootstrap-cluster.sh
+++ b/hack/bootstrap-cluster.sh
@@ -44,15 +44,21 @@ echo
 echo "Add parent Argo CD Application:"
 kubectl apply -f $ROOT/argo-cd-apps/app-of-apps/all-applications-staging.yaml
 
-ARGO_CD_ROUTE="$(kubectl get route/openshift-gitops-server -n openshift-gitops)"
+ARGO_CD_URL="https://$(kubectl get route/openshift-gitops-server -n openshift-gitops -o template --template={{.spec.host}})"
 
 echo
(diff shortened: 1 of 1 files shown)


jduimovich (approved):
LGTM

hack/bootstrap-cluster.sh
```diff
@@ -2,6 +2,14 @@
 
 ROOT="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"/..
 
+OC_WHOAMI="$(oc whoami)"
+if [ "$OC_WHOAMI" != "kubeadmin" ]; then
```
Michkov, line 6: I would not connect this with specific user, any user with cluster-admin role can bootstrap the cluster. You can check if the user can do anything on cluster using `oc auth can-i '*' '*' --all-namespaces`. If not then I prefer to stop script with proper message that user should login as user with cluster-admin role and run the script again.

Roming22: Fixed. I briefly mention the kubeadmin user in the help message so that newcomers don't have to guess which user they should be connected with.

Michkov (approved):

PR #96, opened 2022-02-04
Configure tekton PipelineRuns pruner

Michkov:
Use job to configure pruner in TektonConfig.


jduimovich (approved):
Two nits - `oc get TektonConfig config`  does not need a namespace as its clusterscope and `patameters` 
Do not need to be changed to merge.

PR #98, opened 2022-02-04
fix: test-known-build used a hard-coded user

Roming22:
The hardcoded user prevent anyone but that user to run the test
successfully because of the unauthorized access to the test repository.

### DIFF
--- a/hack/build/m2-builds.sh
+++ b/hack/build/m2-builds.sh
@@ -17,7 +17,12 @@ else
     fi 
 fi
 
-CMD=$SCRIPTDIR/build$DEPLOY.sh 
+CMD=$SCRIPTDIR/build$DEPLOY.sh
+
+if [ -z "${MY_GITHUB_USER}" ]; then
+    MY_GITHUB_USER=$(git ls-remote --get-url | sed 's|^git@github.com:|https://github.com/|' | cut -d/ -f4)
+fi
+
 
 # devfile based builds 
 $CMD https://github.com/devfile-samples/devfile-sample-java-springboot-basic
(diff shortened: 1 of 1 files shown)


hack/build/m2-builds.sh
```diff
@@ -26,8 +31,8 @@ $CMD https://github.com/devfile-samples/devfile-sample-code-with-quarkus
 $CMD https://github.com/devfile-samples/devfile-sample-python-basic 
 
 # auto-detect base builds
-$CMD https://github.com/jduimovich/single-container-app
-$CMD https://github.com/jduimovich/single-nodejs-app
-$CMD https://github.com/jduimovich/spring-petclinic  java-builder
+$CMD https://github.com/${MY_GITHUB_USER}/single-container-app
```
Michkov, line 34: This is not a fix, that are testing repositories, maybe not ideal that they are under John's account but switching it under custom user requires to fork all of these repositories. For basic validation it seems to me as overkill to fork everything. It's fine to have it as variable but the default I would keep jduimovich.

jduimovich: Yes, you would only need write access if you want to make a change and demo that change.
I think in my "demo" mode, I ran the polling detection for a change to force a rebuild and deploy.
I suggest two things.
1) leave jduimovich for now, OR 
1a) move the samples  redhat-appstudio --- the container and nodejs apps are essentially the same just with different builds and the java was trying to get to an early java samples. The official set are driven by devfiles but we should be able to build non-devfile repos without artifically adding a devfiel. 
2) for testing changes to a repo and showing a demo, the polling (pollbuild.sh) can be run on a fork which can be used to demo an full change. As this was really temporary until webhooks were in, its really a local only use. 
For devmode, I would like webhooks to work even to local but for that you need to use a redirector for webhooks (smee or ngrok, or webhookrelay) or we write a cron-job poll which triggers the webhook locally ... I would vote for that and would work on it when I have the time

jduimovich:
@Roming22 closing for now - can revisit when we refactor the examples or add to e2e tests

PR #99, opened 2022-02-04
fix: Make development-mode agnostic to CWD

Roming22:
Running development-mode.sh from any other dir than the
root of the workspace was generating errors.

### DIFF
--- a/hack/util-update-app-of-apps.sh
+++ b/hack/util-update-app-of-apps.sh
@@ -24,7 +24,7 @@ else
 fi
 
 PATCHREPO="$(printf '.spec.source.repoURL="%q"' $GITURL)" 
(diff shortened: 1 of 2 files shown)


hack/util-update-app-of-apps.sh
```diff
@@ -24,7 +24,7 @@ else
 fi
 
 PATCHREPO="$(printf '.spec.source.repoURL="%q"' $GITURL)" 
-PATCHOVERLAY="$(printf '.spec.source.path="%q"' $OVERLAYDIR)"  
+PATCHOVERLAY="$(printf '.spec.source.path="%q"' "$ROOT/$OVERLAYDIR")"
```
Michkov, line 27: this is the issue which breaks argocd

PR #101, opened 2022-02-07
Assign PVC by onetime job

Michkov:
Use newly created PVC in job to force PV assignement. On some cloud
providers the PV is assigned to PVC after PVC is being used.


sbose78:
In a separate PR, could you also add yourself to https://github.com/redhat-appstudio/infra-deployments/blob/main/components/authentication/view-has.yaml so that you can see what's happening ?

jduimovich (approved):
LGTM

PR #102, opened 2022-02-07
Deployment of SPI's simple demo UI

skabashnyuk:
This is a simple UI made for demoing and testing purposes.
Знімок екрана 2022-02-07 о 15 44 12

Reference issue https://issues.redhat.com/browse/SVPI-49.

## Noticeable detail 
- To be able to protect from unauthorized access in front of UI backend I've added kube-rbac-proxy which is doing tokenreviews and subjectaccessreviews. See more https://github.com/redhat-appstudio/service-provider-integration-scm-file-retriever/blob/main/server/config/server/kube-rbac-proxy.yaml
- To allow https://github.com/redhat-appstudio/service-provider-integration-scm-file-retriever to communicate with SPI. I've added necessary roles for UI backend server-side SA https://github.com/redhat-appstudio/service-provider-integration-scm-file-retriever/blob/main/server/config/server/rbac.yaml


sparkoo:
I'm seeing `Status:403 Message:Forbidden (user=user7, verb=get, resource=SPIAccessTokenBinding, subresource=)` even though user should have permissions in the namespace. I'm investigating what's wrong

sparkoo (changes requested):
I was failing to pass kube-rbac-proxy with real user. Following what is defined in `admin` ClusterRole, I had to update kube-rbac-proxy configuration like this https://github.com/redhat-appstudio/service-provider-integration-scm-file-retriever/pull/16

skabashnyuk:
@sparkoo your changes have been applied and tested
Знімок екрана 2022-02-09 о 12 07 13
Also, I've added https://github.com/redhat-appstudio/service-provider-integration-oauth/pull/34 changes to this update.

sparkoo (approved):
works for me now

PR #65, opened 2022-01-18
spi config to secret

sparkoo:
spi config moved to secret. you have to manually create it because it contains secret data so we can't commit it.

It also creates Route with generated host.

Issue: https://issues.redhat.com/browse/SVPI-48

to test:
1. deploy everything with `./hack/bootstrap-cluster.sh`
2. you should see spi deployed but pods failing to start due to missing mount
3. update `components/spi/config.yaml` with your values (you need to create GitHub oauth app first)
4. create actual secret in cluster with `kubectl create secret generic oauth-config --from-file=components/spi/config.yaml -n spi-system`
5. in few moments, you should see pods running
6. follow https://github.com/redhat-appstudio/service-provider-integration-operator#go-through-the-oauth-flow-manually to test oauth flow

❗  Links to my fork branch are there for testing purpose. I'll revert them back to infra-deployment repo before the merge

PR #105, opened 2022-02-10
Set integration with SPI in preview mode

Michkov:
- Allow to configure SPI integration using preview mode
- Add script hack/spi/oauth-flow.sh to proceed oauth workflow


skabashnyuk:
@Michkov Can you explain the purpose of this pr. For what and how it should be used?

Michkov:
Preview mode is for local testing

skabashnyuk:
@Michkov I'm sorry I don't understand the relation between preview mode and `hack/spi/oauth-flow.sh` script.

CC @sbose78

Michkov:
Let's split it into two parts:

1. I want to provide developers easy way to setup SPI. That's done via variables, they will set them once and it will work for all deployments.
2. In #65 there was mention how to test, so I automated it so developers can verify that it's set properly.

Taking point 2) as nice to have, but if you have different opinion, then I can remove it from the PR.

skabashnyuk:
`hack/spi/oauth-flow.sh` It is a real hack that is not suppose to be widely used. I prefer to keep it only in spi repository. Can we remove it from this pr?

Michkov:
@skabashnyuk 'a real hack' was removed from hack folder :-)

hack/preview-template.env
```diff
@@ -11,3 +11,10 @@ export MY_GIT_FORK_REMOTE=
 export MY_GITHUB_ORG=
 ### Personal API token with repo and delete_repo permission
 export MY_GITHUB_TOKEN=
+
+## SPI integration
+### Based on https://github.com/redhat-appstudio/service-provider-integration-operator#configuration
+export SHARED_SECRET= # Random string
+export SPI_TYPE=
```
skabashnyuk, line 18: # Github

skabashnyuk (approved):

PR #107, opened 2022-02-10
Make hack/bootstrap-cluster.sh wait more calmly

simonbaird:
Avoid looking at a screen like this while we wait:

  error: the server doesn't have a resource type "appproject"
  error: the server doesn't have a resource type "appproject"
  error: the server doesn't have a resource type "appproject"
  ...
  Error from server (NotFound): namespaces "openshift-gitops" not found
  Error from server (NotFound): namespaces "openshift-gitops" not found
  Error from server (NotFound): namespaces "openshift-gitops" not found
  ...

Follow the style introduced by @Roming22 when waiting for the Argo
CD URl to work, and show an OK when it's done.

### DIFF
--- a/hack/bootstrap-cluster.sh
+++ b/hack/bootstrap-cluster.sh
@@ -14,19 +14,24 @@ echo "Installing the OpenShift GitOps operator subscription:"
 kubectl apply -f $ROOT/openshift-gitops/subscription-openshift-gitops.yaml
 
 echo
-echo "Waiting for default project (and namespace) to exist:"
+echo -n "Waiting for default project (and namespace) to exist: "
 while : ; do
-  kubectl get appproject/default -n openshift-gitops && break
+  kubectl get appproject/default -n openshift-gitops >/dev/null 2>&1 && break
+  echo -n .
   sleep 1
 done
+echo "OK"
 
 echo
-echo "Waiting for OpenShift GitOps Route:"
+echo -n "Waiting for OpenShift GitOps Route: "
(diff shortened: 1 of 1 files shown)


simonbaird:
Before: http://pastebin.test.redhat.com/1028837

simonbaird:
After: http://pastebin.test.redhat.com/1028953

hack/bootstrap-cluster.sh
```diff
@@ -14,19 +14,24 @@ echo "Installing the OpenShift GitOps operator subscription:"
 kubectl apply -f $ROOT/openshift-gitops/subscription-openshift-gitops.yaml
 
 echo
-echo "Waiting for default project (and namespace) to exist:"
+echo -n "Waiting for default project (and namespace) to exist: "
 while : ; do
-  kubectl get appproject/default -n openshift-gitops && break
```
Michkov, line 19: I do not like infinity loops with break, rather `while ! kubectl get appproject/default -n openshift-gitopsa &>/dev/null; do`

jduimovich: We can leave it as is for now, your change was essentially the minimal update to the existing logic.

jduimovich: Oh, and I tested this morning on a new cluster and worked so if you revised now I'd have to retest :(

simonbaird: Ack, sounds good to me.

PR #115, opened 2022-02-15
Infra for PR checks / Add Pipelines As Code for internal use

sbose78:
This will enable App Studio Github Repos to easily configure Tekton as their CI for PR checks.

Example: 
As a consume of this service, I'll need to add the following:
1. https://github.com/redhat-appstudio/application-service/pull/54
2. https://github.com/redhat-appstudio/infra-deployments/pull/115/files#diff-ec798dfb834a5a1748e53d815d44fd0970c6e7bf04197269f7d4f44e0f1f03e6

- [x] Create GitHub App.
- [x] Ensure rbac is setup for @chmouel and @sbose78 to debug staging cluster

After merge:

- [ ] Create `Secret` in pipelines-as-code namespace on staging.
- [ ] Ensure Pipelines are being used from the AppStudio bundle


components/build/openshift-pipelines/allow-argocd-create-sa.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: allow-argo-cd-sa
```
sbose78, line 4: not really associated with this PR

components/build/pipelines-as-code/kustomization.yaml
```diff
@@ -0,0 +1,6 @@
+resources:
+- install.yaml
```
Michkov, line 2: What is difference between this and https://github.com/openshift-pipelines/pipelines-as-code/tree/main/config ? Why not to include it directly from git?

sbose78: This is a dump of the release.yaml . Right, I could probably use https://github.com/openshift-pipelines/pipelines-as-code/tree/release-0.5.2/config as well 🤷‍♂️

sbose78: Ah this won't work because https://github.com/openshift-pipelines/pipelines-as-code/tree/main/config  isn't a kustomize directory.

components/build/openshift-pipelines/allow-argocd-create-sa.yaml
sbose78: This enables us to reduce the blanket cluster-admin which the cluster config argocd has.

components/build/pipelines-as-code/kustomization.yaml
chmouel: yeah it's a ko one

sbose78:
Leaving the functional aspect of aside, @Michkov , do you see any obvious issues in this PR ?  If not, may I request an approval please :)

We'll need to do a bit of post merge verification: 

1. Create the `Secret` containing Github App credentials in the `pipelines-as-code` ns.
2. Create a PR https://github.com/redhat-appstudio/application-service  to see if a PiplineRun gets triggered and see if Github checks show up.

sbose78:
As a Pipelines As Code SME, I'd love a 🟢 signal from you too, @chmouel :)

chmouel:
LGTM :)

components/build/pipelines-as-code/kustomization.yaml
Michkov: In that case use - `https://raw.githubusercontent.com/openshift-pipelines/pipelines-as-code/release-0.5.2/release-0.5.2.yaml`

Michkov (approved):

components/build/pipelines-as-code/kustomization.yaml
sbose78: @Michkov As a kustomize reference, or a copy-paste? :)

Michkov: Reference to the yaml file, it's clear what is installed and in which version.
```
resources:
- https://raw.githubusercontent.com/openshift-pipelines/pipelines-as-code/release-0.5.2/release-0.5.2.yaml
```

sbose78: Ah gotcha.
How does this look ? https://github.com/redhat-appstudio/infra-deployments/pull/115/commits/4effd7c8e036b766c534629b5feb4d4ad2d148b3

PR #117, opened 2022-02-16
Tekton ci for gitops service repo

sbose78:


sbose78:
Self approved because this is harmless while hooking up the gitops service repo with the Tekton CI.

PR #119, opened 2022-02-17
A few updates to bootstrap script.

parasense:
* Use linux REALPATH(1) to reduce sub-shells in $ROOT assignment
* Use case/switch statements instead of repeated if/elif on $MODE
* Break ARGO_CD_URL into multi-lines to avoid long line side-scrolling
* Remove kustomize dependency, it's now part of `kubectl apply -k`

Signed-off-by: Jon Disnard <jdisnard@redhat.com>


parasense:
Any feedback is highly appreciated. :smiley:

hack/bootstrap-cluster.sh
```diff
@@ -2,9 +2,9 @@
 
 MODE=$1
 
-ROOT="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"/..
+ROOT=$( realpath "${BASH_SOURCE[0]%/*}/.." )
```
Michkov, line 5: Having one issue:
```
cd hack/
bash -ex bootstrap-cluster.sh 
+ MODE=
++ realpath bootstrap-cluster.sh/..
realpath: bootstrap-cluster.sh/..: Not a directory
+ ROOT=
```

hack/bootstrap-cluster.sh
```diff
@@ -45,21 +43,26 @@ kubectl patch argocd/openshift-gitops -n openshift-gitops -p '{"spec":{"rbac":{"
...
+esac
 
-ARGO_CD_URL="https://$(kubectl get route/openshift-gitops-server -n openshift-gitops -o template --template={{.spec.host}})"
+printf -v ARGO_CD_URL "https://%s" \
+    $(
+        kubectl get route/openshift-gitops-server \
+        -n openshift-gitops \
+	-o template \
+	--template={{.spec.host}}
+    )
```
Michkov, line 65: Much harder for me to read, but it's ok.

Michkov (commented):
Added few comments, mostly neutral. I like the `while` update and kustomize removal.

hack/bootstrap-cluster.sh
```diff
@@ -2,9 +2,9 @@
 
 MODE=$1
 
-ROOT="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"/..
+ROOT=$( realpath "${BASH_SOURCE[0]%/*}/.." )
 
-if [ "$(oc auth can-i '*' '*' --all-namespaces)" != "yes" ]; then
+if [[ "$(oc auth can-i '*' '*' --all-namespaces)" != "yes" ]]; then
```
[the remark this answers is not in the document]
parasense: Removed double bracket, it wasn't really doing anything besides changing style.
However, double bracket is probably the better way to evaluate test statements.
(resolving)

hack/bootstrap-cluster.sh
parasense: This could be refactored, or even dropped.
If it burns your eyes I'd rather just drop it?

Michkov: Compromise?
```
ARGO_CD_ROUTE=$(kubectl get \
                 -n openshift-gitops \
                 -o template \
                 --template={{.spec.host}} \
                 route/openshift-gitops-server \
               )
ARGO_CD_URL="https://$ARGO_CD_ROUTE"
```

hack/bootstrap-cluster.sh
```diff
@@ -1,8 +1,9 @@
 #!/bin/bash
 
 MODE=$1
+shift $#
 
-ROOT="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"/..
+ROOT="$(realpath --canonicalize-missing --quiet -- ${BASH_SOURCE[0]}/../..)"
```
Michkov, line 6: Now it's more complicated then original solution. I prefer original solution since it's oneliner and also realpath is behaving slightly differently on MacOS (not sure if also in this case). That `shift` is also slightly ugly (not much in this script, but other scripts are having ROOT as well, so it could be confusing).

hack/bootstrap-cluster.sh
parasense: I like this. Updated the PR to include the ARGO_CD_ROUTE as shown.

hack/bootstrap-cluster.sh
parasense: It's not more complicated, and actually It's much less complicated.

* one command, instead of three.
* no conditionals
* no output redirection
* one sub-shell instead of two.
* no dangling unquoted strings

Regardless, I've removed the long-args to significantly shorten the line length. That way line length isn't conflated with complexity, and I think it is much nicer on the eyes. ;-)

For MacOS, that's an interesting point. 
I'm not sure people are running the script on MacOS, but if they are then they should probably consider installing GNU coreutils prior. I'm told that's a common task for FOSS developers, and goes something like `brew install coreutils` (unconfirmed, not tested, just what I'm told)

Michkov: Ok with removal of swift it's better. LGTM, but please rebase first.

parasense:
Re-opening this accidentally closed PR.

parasense:
I believe all the open issues were resolved?
If yes, then please merge this PR at your earliest convenience.
TYIA :smiley:

Michkov (approved):

PR #127, opened 2022-02-25
Added automated pre-requisite installation for fedora and ubuntu

bamachrn:
This PR sets up scripts for installing  the pre-requisites for setting up app-studio in local machine.
* mainly setting up pip3,yq, kubectl, kustomize
* Firstly it checks which type of system is running this
* Then it installs the dependencies one by one


setup/debian-based.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+#This script sets up the pre-requisites for setting up app-studio in redhat based machines.
+# mainly setting up pip3,yq, kubectl, kustomize
+# Firsly it checks which type of system is running this
+# Then it installs the dependencies one by one
+sudo apt-get update -y
+sudo apt-get -y install pip3
```
Michkov, line 7: why pip3?

setup/debian-based.sh
```diff
@@ -0,0 +1,15 @@
...
+sudo pip3 install --upgrade setuptools
+sudo snap install yq
+curl -LO https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/amd64/kubectl
+chmod +x ./kubectl
+sudo mv ./kubectl /usr/local/bin
+curl -s "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh"  | bash
+sudo cp kustomize /usr/bin/
+sudo chmod ug+x /usr/bin/kustomize
```
Michkov, line 15: kustomize should not be needed anymore after merge of #119

setup/redhat-based.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+#This script sets up the pre-requisites for setting up app-studio in redhat based machines.
+# mainly setting up pip3,yq, kubectl, kustomize
+# Firsly it checks which type of system is running this
+# Then it installs the dependencies one by one
+
+sudo yum -y install pip3
+sudo pip3 install --upgrade setuptools
+sudo snap install yq
```
Michkov, line 9: snap is not usually installed on fedora.

setup/install-pre-req.sh
```diff
@@ -0,0 +1,25 @@
...
+#This script sets up the pre-requisites for setting up app-studio in local machine.
+# mainly setting up pip3,yq, kubectl, kustomize
+# Firsly it checks which type of system is running this
+# Then it installs the dependencies one by one
+
+OSNAME=$(grep '^NAME=' /etc/os-release)
+echo $OSNAME
+ROOT="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"/..
+echo $ROOT
+
```
Michkov, line 11: Since it's installing crc there should be a check that there is enough RAM to proceed.

Michkov (commented):
Please move this into `hack` folder, the name of the folder `setup` is not clear what it is setting.

setup/install-pre-req.sh
```diff
@@ -0,0 +1,25 @@
...
+case $OSNAME in
```
amfred, line 12: I think $OSTYPE will work better here: https://stackoverflow.com/questions/394230/how-to-detect-the-os-from-a-bash-script

This calculation of $OSNAME doesn't work on a Mac.

setup/prepare-crc.sh
```diff
@@ -0,0 +1,28 @@
...
+echo $CRCBINARY
+#Get CRC VMs configured in local machine
+$CRCBINARY setup
+#Set memory and CPU allowance for CRC
+$CRCBINARY config set memory 16384
+$CRCBINARY config set cpus 6
+#enable netmetrices addons, to make sure member cluster has proper resources
+$CRCBINARY config set enable-cluster-monitoring true
+#Delete existing CRC cluster to apply the config updates
+$CRCBINARY delete
```
amfred, line 18: Let's see if we can take a parameter for this script that will run $CRCBINARY delete -y if we're in a CI mode where we can't take interactive input.

setup/debian-based.sh
amfred: I know that on Macs, it's necessary to specify pip3 instead of pip because the default Python for the OS is an older version.  I'm not sure if it's necessary for Debian-based systems.  Is there a problem with using pip3, @Michkov?

setup/install-pre-req.sh
amfred: After determining $OSNAME is looks like you still need the $OSTYPE to distinguish between different Linux distributions.

setup/prepare-crc.sh
amfred: Ideally we should make the memory and CPU configurable via parameters.  However I would be OK with getting a hardcoded version in first and then adding that feature.

setup/install-pre-req.sh
bamachrn: That's a good point. We need to check for the RAM as we are setting the default MEMORY config to 16GB for running app-studio

amfred (changes requested):

setup/prepare-crc.sh
```diff
@@ -0,0 +1,28 @@
...
+#It also configures CRC to have required addons.
+
+#TODO: Check for CRC installation and binary location and set it to CRC-BINARY
+CRCBINARY=$(readlink -f ~/.crc/bin/crc)
+
+echo $CRCBINARY
+#Get CRC VMs configured in local machine
+$CRCBINARY setup
+#Set memory and CPU allowance for CRC
+$CRCBINARY config set memory 16384
```
amfred, line 13: These are recommended minimums.  I'm not sure what the actual minimums are.
You can certainly go higher.

setup/debian-based.sh
Michkov: no problem is that we do not need python at all, so no pip.

setup/install-pre-req.sh
amfred: CodeReady Containers requires the following system resources:

4 physical CPU cores
9 GB of free memory
35 GB of storage space

But we'll definitely need more than that.  The question is how much.  I know that some of our developers are happy with these values and others (with beefier machines) set them higher.

setup/prepare-crc.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+#This script configures CRC for running app-studio in it
+#It cofigures the CRC to have better memory and CPUs 
+#It also configures CRC to have required addons.
+
+#TODO: Check for CRC installation and binary location and set it to CRC-BINARY
+CRCBINARY=$(readlink -f ~/.crc/bin/crc)
```
Michkov, line 7: If I remember well, `readlink -f` does not work on macOS. @amfred could you check?

setup/debian-based.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+#This script sets up the pre-requisites for setting up app-studio in redhat based machines.
+# mainly setting up pip3,yq, kubectl, kustomize
+# Firsly it checks which type of system is running this
+# Then it installs the dependencies one by one
+sudo apt-get update -y
+sudo apt-get -y install pip3
+sudo pip3 install --upgrade setuptools
+sudo snap install yq
```
Michkov, line 9: snap is not always available. Rather download it :
```
wget https://github.com/mikefarah/yq/releases/download/v4.20.2/yq_linux_amd64 -O /usr/local/bin/yq && \
    chmod +x /usr/local/bin/yq
```

setup/debian-based.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
```
Michkov, line 1: it's almost the same as fedora-based.sh, would be better to merge to avoid code duplicity

setup/install-pre-req.sh
```diff
@@ -0,0 +1,25 @@
...
+  *"Fedora"*)
+    echo "I am in Fedora"
+    $ROOT/redhat-based.sh
+    $ROOT/prepare-crc.sh
+    ;;
+
+  *"Ubuntu"*)
+    echo "I am in Ubuntu"
+    $ROOT/debian-based.sh
+    $ROOT/prepare-crc.sh
```
Michkov, line 22: put prepare-crc.sh after the `case` -> it should be applied for all OS.

setup/debian-based.sh
amfred: Oh. When did we get rid of the python dependency? It was needed for one of the pre-reqs when we initially wrote the setup scripts, a couple of months ago.  Now if I could only remember which one...

Michkov: there are two versions of `yq` -> pip one and go lang one, we are using the golang one.

setup/prepare-crc.sh
amfred: That is correct.  On Macs, the default CRC installation also links the CRC command to `/usr/local/bin/crc`

hack/setup/prepare-crc.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+#This script configures CRC for running app-studio in it
+#It cofigures the CRC to have better memory and CPUs 
+#It also configures CRC to have required addons.
+
+ROOT="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+echo $ROOT
+CRCBINARY=$(readlink -f ~/.crc/bin/crc)
+
```
amfred, line 9: The CRCBINARY line doesn't work on MacOS, but if you add this line right after, it works on MacOS (it just uses the pre-configured crc command in case readlink doesn't work):
[[ -z "$CRCBINARY" ]] && CRCBINARY="crc"

Definitely want to fix that before merging.

hack/setup/prepare-crc.sh
```diff
@@ -0,0 +1,33 @@
...
+#Start CRC with modified configs
+$CRCBINARY start
+
+#Point local environment clients (kubectl and oc) to the CRC cluster
+eval $($CRCBINARY oc-env)
+kubectl config use-context crc-admin
+
+#Reduce cpu resource request for each AppStudio Application
+#TODO: Set the file path properly
+../../hack/reduce-gitops-cpu-requests.sh
```
amfred, line 33: I think you want a newline at the end of this line.

hack/setup/redhat-based.sh
```diff
@@ -0,0 +1,12 @@
...
+# mainly setting up pip3,yq, kubectl, kustomize
+# Firstly it checks which type of system is running this
+# Then it installs the dependencies one by one
+
+sudo yum -y install pip3 snapd
+sudo pip3 install --upgrade setuptools
+sudo snap install yq
+curl -LO https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/amd64/kubectl
+chmod +x ./kubectl
+sudo mv ./kubectl /usr/local/bin
```
amfred, line 12: I think you want a newline at the end of this line.

hack/setup/prepare-crc.sh
amfred: This line doesn't work if you run `hack/setup/install-prereq.sh` from the infra-deployments directory:
`/Users/afred/git/infra-deployments/hack/setup/prepare-crc.sh: line 34: ../../hack/reduce-gitops-cpu-requests.sh: No such file or directory`

Do these need to be in a setup sub-folder, or can they go directly into the hack folder? If they were all in the same folder, you wouldn't need to include path information.

hack/setup/debian-based.sh
```diff
@@ -0,0 +1,12 @@
...
+# mainly setting up pip3,yq, kubectl, kustomize
+# Firstly it checks which type of system is running this
+# Then it installs the dependencies one by one
+sudo apt-get update -y
+sudo apt-get -y install pip
+sudo pip install --upgrade setuptools
+sudo snap install yq
+curl -LO https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/amd64/kubectl
+chmod +x ./kubectl
+sudo mv ./kubectl /usr/local/bin
```
amfred, line 12: I think you want a newline at the end of this line.

components/has/kustomization.yaml
```diff
@@ -22,4 +22,4 @@ configMapGenerator:
 namespace: application-service
 
 patches:
-  - ./manager_resources_patch.yaml
+  - ./manager_resources_patch.yaml
```
amfred, line 25: I think you want a newline at the end of this line.

amfred (changes requested):

setup/install-pre-req.sh
amfred: I think we can rely on the 'crc setup' and 'crc start' commands to do the checks.

amfred: Moved to https://github.com/redhat-appstudio/infra-deployments/issues/214

setup/prepare-crc.sh
amfred: The force-delete is moved to https://github.com/redhat-appstudio/infra-deployments/issues/214

amfred: The configurable memory and CPU requirements is moved to https://github.com/redhat-appstudio/infra-deployments/issues/215

hack/setup/prepare-crc.sh
amfred: This is fixed in the latest commit, thanks.

hack/setup/prepare-crc.sh
amfred: This is fixed in the latest commit.

hack/setup/redhat-based.sh
amfred: Fixed.

hack/setup/debian-based.sh
amfred: Fixed.

components/has/kustomization.yaml
amfred: Fixed.

setup/debian-based.sh
amfred: These scripts are now using the correct version of yq and not using snap or pip.

amfred (approved):
I've tested this on MacOS using the latest CRC and it's working.
infra-deployments % crc version
CodeReady Containers version: 2.0.1+f0fcd50e
OpenShift version: 4.10.3
Podman version: 3.4.4

amfred (approved):
LGTM

PR #131, opened 2022-02-25
create/delete only for secrets

sbose78:


components/authentication/user-ci-maintainer.yaml
```diff
@@ -20,7 +20,8 @@ rules:
     resources:
       - secrets
   - verbs:
-      - '*' # needed till we figure out how to cleanup workspaces.
+      - create
+      - delete
```
Michkov, line 24: you are limiting pipelineruns

sbose78: ouch sorry

PR #134, opened 2022-03-01
Remove kustomize command dependency

Michkov:
Kustomize command is integrated to kubectl/oc, it's not used as
standalone command anymore.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ Simply update the files under `components/(team-name)`, and open a PR with the c
 
 ### Required prerequisites
 The prerequisites are:
-- You must have `kubectl`, `oc`, `jq`, `yq` and `kustomize` installed. 
+- You must have `kubectl`, `oc`, `jq and `yq` installed.
 - You must have `kubectl` and `oc` pointing to an existing OpenShift cluster, that you wish to deploy to.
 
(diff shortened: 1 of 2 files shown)


jduimovich (commented):
I understand removing the use because of "built in" to kubectl however not sure I like removing it. 
1. testing your kustomize changes via kustomize build is useful and the kubectl way is verbose and annoying (dry-run and -o yaml)
2. piping kustomize builds to kubectl allows easier debug by adding "| tee some-file | " to the scripts to capture the yaml.

Michkov:
@jduimovich you can use `oc kustomize`, nobody blocks you to use kustomize but it's not in the scripts so it's not needed to be installed.

Unfortunately the validation is failing for some reason, maybe it's used in some other project with testing scripts.

jduimovich:
Yeah
```
command -v kustomize >/dev/null 2>&1 || { echo "kustomize is not installed. Aborting."; exit 1; }
```

the start script in the e2e tests seems to check for kustomize.
I think we should just remove the uses of it in our scripts now, and then remove the kustomize install later.

jduimovich (approved):
This looks good. I tested just to be sure, worked fine.

PR #136, opened 2022-03-04
Install tekton results

Michkov:
- Install tekton results
- Add permission to pipeline-maintainers to list and search in results
- Create helper script for setting `tkn results` command

TODO:

- [x] Move postgres from my personal quay account -> issue is that the main image is on dockerhub, it has features which are not available in redhat postgres images - execution of sql script during start
- [ ] Generate tekton-results-postgres secret on stage cluster
- [ ] Generate tekton-results-tls secret on stage cluster
- [x] Update README

### DIFF
--- a/hack/bootstrap-cluster.sh
+++ b/hack/bootstrap-cluster.sh
@@ -45,6 +45,30 @@ echo
 echo "Add Role/RoleBindings for OpenShift GitOps:"
 kubectl apply --kustomize $ROOT/openshift-gitops/cluster-rbac
 
+echo "Setting secrets for Tekton Results"
+if ! kubectl get namespace tekton-pipelines &>/dev/null; then
+  kubectl create namespace tekton-pipelines
+fi
+if ! kubectl get secret -n tekton-pipelines tekton-results-tls &>/dev/null; then
+  ROUTE=$(oc whoami --show-console | sed 's|https://console-openshift-console|api-tekton-pipelines|')
+  openssl req -x509 \
+    -newkey rsa:4096 \
+    -keyout key.pem \
+    -out cert.pem \
+    -days 3650 \
+    -nodes \
+    -subj "/CN=tekton-results-api-service.tekton-pipelines.svc.cluster.local" \
+    -addext "subjectAltName = DNS:tekton-results-api-service.tekton-pipelines.svc.cluster.local, DNS:$ROUTE"
+  kubectl create secret tls -n tekton-pipelines tekton-results-tls --cert=cert.pem --key=key.pem
+  rm cert.pem key.pem
+fi
+if ! kubectl get secret -n tekton-pipelines tekton-results-postgres &>/dev/null; then
+  kubectl create secret generic tekton-results-postgres \
+    --namespace="tekton-pipelines" \
+    --from-literal=POSTGRES_USER=results \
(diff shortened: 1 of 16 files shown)


components/build/tekton-results/patch-postgres-statefulset.yaml
```diff
@@ -0,0 +1,21 @@
...
+  template:
+    spec:
+      volumes:
+        - name: run
+          emptyDir: {}
+        - name: data
+          emptyDir: {}
+      containers:
+        - name: postgres
+          image: quay.io/mkovarik/postgres:13
```
goldmann, line 16: Have you made any changes to official image or is it a way to pin/control it?

Michkov: No change, but wanted to avoid using dockerhub since there is pull limitation, when you have testing machine in the company network then it's usually failing. Wanted to switch to RH image but there is missing feature of applying sql files during initialization. Having this in TODO.

sbose78: > Wanted to switch to RH image but there is missing feature of applying sql files during initialization. Having this in TODO.

Is this a TODO to fix the postgresql image? :)

jduimovich: I read it as follows
1. He is using an `unmodified container` copied from dockerhub to mkovarik to avoid the docker unauthenticated throttle. I assume this is from [ https://hub.docker.com/_/postgres ](https://hub.docker.com/_/postgres) and you are using using version 13 ? something like  .... 

```
docker pull postgres:13
docker tag docker.io/library/postgres:13 quay.io/mkovarik/postgres:13
docker push  quay.io/mkovarik/postgres:13
```

if we need to do this for a while, we should add it to the quay app-studio account so we can keep track and add to the build maintainers should we need to update

3. I see there is an an[ official v13 image from RH ](https://catalog.redhat.com/software/containers/rhel8/postgresql-13/5ffdbdef73a65398111b8362) which we could use but I assume from your comment that despite being same versions they are different in how they do initialization which means we cannot currently use it

Is this a bug or fearture of the RH v13 ?

Michkov: Fixed by creating additional configmap for execution of sql script.

goldmann (approved):
I tested it locally, works. LGTM.

hack/bootstrap-cluster.sh
```diff
@@ -45,6 +45,30 @@ echo
...
+if ! kubectl get secret -n tekton-pipelines tekton-results-tls &>/dev/null; then
+  ROUTE=$(oc whoami --show-console | sed 's|https://console-openshift-console|api-tekton-pipelines|')
+  openssl req -x509 \
+    -newkey rsa:4096 \
+    -keyout key.pem \
+    -out cert.pem \
+    -days 3650 \
+    -nodes \
+    -subj "/CN=tekton-results-api-service.tekton-pipelines.svc.cluster.local" \
+    -addext "subjectAltName = DNS:tekton-results-api-service.tekton-pipelines.svc.cluster.local, DNS:$ROUTE"
```
Michkov, line 61: test image does contain old openssl (RHEL7) which does not have this option.

PR #144, opened 2022-03-08
Move HAS configuration closer to the others

ralphbean:
There are a series of optional post-bootstrap sections for gitops
and SPI. This section for HAS was further down in the document,
seemingly out of order.

Here, move it up closer to the others so that they make more sense
when read together.


Michkov:
This is related only for `Development modes for your own clusters`. It should not be needed on production/staging cluster that's why it's under  `Development modes for your own clusters`

PR #150, opened 2022-03-09
Create MR when new build-definitions pipelines are created

Michkov:
Clone infra-deployments, update the code, push and create MR.

Testing MR produced by the pipeline: #149 
Secret is set on the stage instance.

### DIFF
--- a/components/build/.tekton/trigger-template.yaml
+++ b/components/build/.tekton/trigger-template.yaml
@@ -1,3 +1,13 @@
+# This trigger template creates pipelines for building images:
+#      - redhat-appstudio/build-templates-bundle
+#      - redhat-appstudio/appstudio-tasks
+#      - redhat-appstudio/appstudio-utils
+# MR request to redhat-appstudio/infra-deployments is created.
+# There has to be fork repository for pushing changes and token to create MR.
+# Fork repository has to have set Deploy key
+# Secret has to be created:
+# oc create -n build-templates secret generic build-definitions-deploy-key --from-file=id_rsa=$PATH_TO_KEY --from-literal=token=$GITHUB_TOKEN
+
 apiVersion: triggers.tekton.dev/v1alpha1
 kind: TriggerTemplate
 metadata:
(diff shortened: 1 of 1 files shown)


components/build/.tekton/trigger-template.yaml
```diff
@@ -63,14 +73,53 @@ spec:
...
+                - name: BASE_IMAGE
+                  value: registry.redhat.io/openshift-pipelines/pipelines-git-init-rhel8@sha256:af7dd5b3b1598a980f17d5f5d3d8a4b11ab4f5184677f7f17ad302baa36bd3c1
+                - name: GIT_USER_NAME
+                  value: Build deployer
+                - name: GIT_SCRIPT
+                  value: |
+                    ssh-keyscan -t rsa github.com >> ~/.ssh/known_hosts
+                    git clone https://github.com/redhat-appstudio/infra-deployments
+                    cd infra-deployments
+                    git remote add push git@github.com:Michkov/infra-deployments.git
```
mmorhun, line 89: We shouldn't use personal accounts for such tasks

components/build/.tekton/trigger-template.yaml
```diff
@@ -63,14 +73,53 @@ spec:
...
+                    done
+                    git commit -a -m "Build definitions update" -m "from $(params.git-url) $(params.revision)"
+                    git push -f --set-upstream push build-definitions-update
+
+                    set +x
+                    curl -H "Authorization: token $(cat /root/.ssh/token)" \
+                      -X POST \
+                      -H "Accept: application/vnd.github.v3+json" \
+                      https://api.github.com/repos/redhat-appstudio/infra-deployments/pulls \
+                      -d '{"head":"Michkov:build-definitions-update","base":"main","title":"Build definitions update"}'
```
mmorhun, line 103: Same as above

mmorhun (changes requested):

components/build/.tekton/trigger-template.yaml
Michkov: Updated to new org redhat-appstudio-mr-creation

components/build/.tekton/trigger-template.yaml
Michkov: Updated to new org redhat-appstudio-mr-creation

mmorhun (approved):
LGTM

mmorhun:
Also it would be good to grant access to the MR org for, at least, leads and architects.

PR #152, opened 2022-03-09
dynamic support for openssl.cnf location

joejstuart:
The config is in a different location on a mac

### DIFF
--- a/hack/bootstrap-cluster.sh
+++ b/hack/bootstrap-cluster.sh
@@ -49,6 +49,9 @@ echo "Setting secrets for Tekton Results"
 if ! kubectl get namespace tekton-pipelines &>/dev/null; then
   kubectl create namespace tekton-pipelines
 fi
+
+OPENSSLDIR=`openssl version -d | cut -f2 -d'"'`
+ 
 if ! kubectl get secret -n tekton-pipelines tekton-results-tls &>/dev/null; then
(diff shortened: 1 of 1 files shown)


hack/bootstrap-cluster.sh
```diff
@@ -49,6 +49,13 @@ echo "Setting secrets for Tekton Results"
 if ! kubectl get namespace tekton-pipelines &>/dev/null; then
   kubectl create namespace tekton-pipelines
 fi
+
+OPENSSLDIR=`/usr/bin/openssl version -a |grep OPENSSLDIR |awk '{print $2}' |tr -d '"'`
```
Michkov, line 53: `openssl version -d | cut -f2 -d'"'`

Michkov: Please move before calling `openssl req` command

hack/bootstrap-cluster.sh
```diff
@@ -49,6 +49,13 @@ echo "Setting secrets for Tekton Results"
...
+# make sure something didn't go wrong and a directory comes back
+if [ ! -d $OPENSSLDIR ]; then
```
Michkov, line 55: does not work, `$OPENSSLDIR` must be `"$OPENSSLDIR"`. In this case I would do -> `${OPENSSLDIR:-/etc/pki/tls}` in `openssl req` comand

Michkov (approved):

PR #155, opened 2022-03-10
Allow Pending PVC caused by WaitForFirstConsumer strategy

Michkov:
PVC which are used by CI are not used until CI is triggered.
PVCs are created but not Bound due toStorageClass configuration
WaitForFirstConsumer. Allow Pending PVC to be marked as Healthy instead
of simulating first customer by job.

### DIFF
--- a/hack/bootstrap-cluster.sh
+++ b/hack/bootstrap-cluster.sh
@@ -41,7 +41,33 @@ kubectl patch argocd/openshift-gitops -n openshift-gitops -p '{"spec": {"server"
 # - Once we have a proper access policy in place, this should be updated to be consistent with that policy.
 kubectl patch argocd/openshift-gitops -n openshift-gitops -p '{"spec":{"rbac":{"policy":"g, system:authenticated, role:admin"}}}' --type=merge
 
-echo
+# Mark Pending PVC as Healthy, workaround for WaitForFirstConsumer StorageClasses.
+# If the attachment will fail then it will be visible on the pod anyway.
+kubectl patch argocd/openshift-gitops -n openshift-gitops -p '
+spec:
+  resourceCustomizations: |
+    PersistentVolumeClaim:
+      health.lua: |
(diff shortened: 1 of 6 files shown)


hack/bootstrap-cluster.sh
```diff
@@ -41,6 +41,32 @@ kubectl patch argocd/openshift-gitops -n openshift-gitops -p '{"spec": {"server"
...
+            if obj.status.phase == "Bound" then
+              hs.status = "Healthy"
+              hs.message = obj.status.phase
+              return hs
+            end
+          end
+        end
+        hs.status = "Progressing"
+        return hs
+' --type=merge
```
Michkov, line 68: @sbose78 This line should be applied on stage cluster before merging.

mmorhun:
Could you update the description with instructions on how to test the MR?

jduimovich:
To test this, you should simply be able to 
1. get a clusterbot cluster
2. run  bootstrap.sh 
3. all of the unbound PVC components will show as healthy instead of pending due to the patch in the script.

Note - this cannot be demonstrated on CRC as the PVCs are bound immediately on CRC. 

> @sbose78 This line should be applied on stage cluster before merging.
As the bootstrap is not run on staging this needs to be done manually by someone with access.
Alternatively we can run as a job in the gitops repo to ensure it gets configured this way like we do for the tektonconfig

Michkov:
This is tested via ci/prow/appstudio-e2e-deployment  since it creates new cluster and deploy there applications, waiting for build and application-service to be synced -> if this would not work then it would never be synced and it would wait for Bound state of PVC. But test passed so it did not wait forever for bound of the PVC.

mmorhun (approved):

hack/bootstrap-cluster.sh
Michkov: @jgwest Could you run this command on stage cluster?

Issue #157, opened 2022-03-11
grant explicit permissions to Argo CD to deal with tekton results


sbose78:
When the build component's Application CR is created in isolation - without making Argo CD cluster-admin, the sync fails because ArgoCD hasn't been granted the permissions.

 

```
one or more objects failed to apply, reason: services is forbidden: User "system:serviceaccount:openshift-gitops:openshift-gitops-argocd-application-controller" cannot create resource "services" in API group "" in the namespace "tekton-pipelines",deployments.apps is forbidden: User "system:serviceaccount:openshift-gitops:openshift-gitops-argocd-application-controller" cannot create resource "deployments" in API group "apps" in the namespace "tekton-pipelines",statefulsets.apps is forbidden: User "system:serviceaccount:openshift-gitops:openshift-gitops-argocd-application-controller" cannot create resource "statefulsets" in API group "apps" in the namespace "tekton-pipelines",routes.route.openshift.io is forbidden: User "system:serviceaccount:openshift-gitops:openshift-gitops-argocd-application-controller" cannot create resource "routes" in API group "route.openshift.io" in the namespace "tekton-pipelines",serviceaccounts is forbidden: User "system:serviceaccount:openshift-gitops:openshift-gitops-argocd-application-controller" cannot create resource "serviceaccounts" in API group "" in the namespace "tekton-pipelines". Retrying attempt #10 at 3:19AM
```

The fix needs to be along the lines of https://github.com/redhat-appstudio/infra-deployments/pull/67

PR #158, opened 2022-03-11
Allow argoCD manage tekton-pipelines namespace

Michkov:
Fix of #157 

@sbose78 is this enough? Or do you expect better fencing? Currently tekton-results define in tekton-pipelines these resources:
ClusterRole
ClusterRoleBinding
ConfigMap
Deployment
Job
Namespace
RoleBinding
Route
Service
ServiceAccount
StatefulSet

### DIFF
--- a/components/build/tekton-results/allow-argocd-to-manage.yaml
+++ b/components/build/tekton-results/allow-argocd-to-manage.yaml
@@ -1,0 +1,13 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: grant-argocd
+  namespace: tekton-pipelines
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: admin
+subjects:
+- kind: ServiceAccount
+  name: openshift-gitops-argocd-application-controller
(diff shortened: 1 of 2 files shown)


sbose78:
This is good enough 👍

components/build/tekton-results/allow-argocd-to-manage.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: grant-argocd
+  namespace: tekton-pipelines
```
lcarva, line 5: If Tekton is installed via a `Subscription` resource, it's deployed under `openshift-pipelines`. The namespace `tekton-pipelines` is only relevant if you're installing tekton directly from upstream. Is my understanding correct?

Michkov: Yes, we are installing `tekton-results` which are not part of `openshift-pipelines`. `tekton-results` are installed into `tekton-pipelines` and also having pointers to `tekton-pipelines` so that's why there are kept in `tekton-pipelines` namespace.

PR #160, opened 2022-03-11
Fix: Add missing route for e2e-tests event listener.

rhopp:
Signed-off-by: Radim Hopp <rhopp@redhat.com>


rhopp:
(This was missed, because my testing deployment of appstudio was deployed on cluster behind VPN, so I was using smee.io to redirect github webhooks to my cluster - directly to the service endpoint instead of using public route)

Michkov (approved):

PR #165, opened 2022-03-12
Support builds for private git repos

sbose78:
https://github.com/redhat-appstudio/application-service/commit/2976522a9052887a66a537c5e502cb7ac3fb8848


sbose78:
Self approving since the CI / e2e tests are green, and the only change going in was previously approved.

/approve

PR #166, opened 2022-03-14
Preview mode: set docker.io globally

Michkov:
Avoid pull limit issue with docker.io by allow to set global image pull
secret.

### DIFF
--- a/hack/preview.sh
+++ b/hack/preview.sh
@@ -53,6 +53,18 @@ if [ -n "$SHARED_SECRET" ] && [ -n "$SPI_TYPE" ] && [ -n "$SPI_CLIENT_ID" ] && [
     rm $TMP_FILE
 fi
 
+if [ -n "$DOCKER_IO_AUTH" ]; then
+    AUTH=$(mktemp)
+    # Set global pull secret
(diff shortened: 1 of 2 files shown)


hack/preview.sh
```diff
@@ -53,6 +53,18 @@ if [ -n "$SHARED_SECRET" ] && [ -n "$SPI_TYPE" ] && [ -n "$SPI_CLIENT_ID" ] && [
     rm $TMP_FILE
 fi
 
+if [ -n "$DOCKER_IO_AUTH" ]; then
+    AUTH=$(mktemp)
+    # Set global pull secret
+    oc get secret/pull-secret -n openshift-config --template='{{index .data ".dockerconfigjson" | base64decode}}' > $AUTH
+    oc registry login --registry=docker.io --auth-basic=$DOCKER_IO_AUTH --to=$AUTH
```
mmorhun, line 60: Does it makes sense to make registry configurable?

Michkov: In this case it's targeted to docker.io due to pull limitation. Otherwise we do not use private repositories or at least we are not blocked by them.

mmorhun (approved):
LGTM

goldmann (approved):

PR #172, opened 2022-03-15
Enable Tekton chains' folks to view logs

sbose78:


sbose78:
/lgtm 
/approve

Self-approving because I need to check logs

simonbaird:
Lgtm. I might add other contract team members later.

PR #183, opened 2022-03-16
Signature verification for in-toto via verify-blob

zregvart:
For in-toto the data being signed is in the DSSE protocol format[1], so
when verifying using `cosign verify-blob` we need to present the data in
that format.

[1] https://github.com/secure-systems-lab/dsse/blob/master/protocol.md

### DIFF
--- a/hack/chains/check-taskrun.sh
+++ b/hack/chains/check-taskrun.sh
@@ -42,7 +42,8 @@ get-resourcesresult() {
 # Fetch task run signature and payload
 TASKRUN_UID=$( get-jsonpath metadata.uid )
 SIGNATURE=$( get-chainsval signature-taskrun-$TASKRUN_UID )
-PAYLOAD=$( get-chainsval payload-taskrun-$TASKRUN_UID | base64 --decode )
+RAW_PAYLOAD=$( get-chainsval payload-taskrun-$TASKRUN_UID | base64 --decode )
+PAYLOAD=$RAW_PAYLOAD
 
 # For a task that builds an image, image digest should be available
 # in the task results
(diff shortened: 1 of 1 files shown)


lcarva (approved):
Verified on my local setup. LGTM

hack/chains/check-taskrun.sh
```diff
@@ -114,29 +120,17 @@ say $SIGNATURE
...
 
 # For debugging...
 title "To view taskrun"
 say " env EDITOR=view kubectl edit $TASKRUN_NAME"
 
 # Clean up
 rm $SIG_FILE $PAYLOAD_FILE
-
-# Use the exit code from cosign
-exit $COSIGN_EXIT_CODE
```
simonbaird, line 142: The reason for the set +e and exit code hoop jumping was to ensure the `rm $SIG_FILE ..` clean up happened if the verify failed.

Probably not that important, but I remember now why I did that...

simonbaird:
One nitpick: Could mention the Jira in the commit, it's https://issues.redhat.com/browse/HACBS-89 .

sbose78 (approved):

hack/chains/check-taskrun.sh
lcarva: Ah ok. In that case, maybe `trap` right after the files are created is a good idea:
```bash
# Cosign needs files on disk to do a verify-blob afaict
SIG_FILE=$( mktemp )
PAYLOAD_FILE=$( mktemp )
trap "rm -f ${SIG_FILE} ${PAYLOAD_FILE}" EXIT
echo -n "$PAYLOAD" > $PAYLOAD_FILE
echo -n "$SIGNATURE" > $SIG_FILE
```

simonbaird: Yeah that's sensible, but imo it doesn't particularly need to be in this PR.

zregvart: I've added a [trap](https://github.com/redhat-appstudio/infra-deployments/pull/183/files#diff-0aaca8058d25ea8b1a3abd31c7d3eb390726d1ab896ef5829ad0b67b5e187390R114-R117) to cleanup the temp files.

zregvart:
I've modified the commit message to mention [HACBS-89](https://issues.redhat.com/browse/HACBS-89)

simonbaird:
Looks good to me - nice patch!

PR #189, opened 2022-03-17
enable management of gitopsdeployments

sbose78:


sbose78:
Required to create resources during the demo.

PR #188, opened 2022-03-16
Misc small chains improvements

simonbaird:
A handful of miscellaneous fixes and cleanups related to chains. It's mostly tweaks to the demo scripts, but there's one configuration change intended to allow argocd admin permissions in the tekton-chains namespace.

See commits for explanations.

PR #191, opened 2022-03-17
kaniko demo using build-definitions script

joejstuart:


hack/chains/kaniko-demo.sh
```diff
@@ -1,7 +1,6 @@
 #!/bin/bash
 
 source $(dirname $0)/_helpers.sh
-set -ue
```
lcarva, line 4: Why remove this?

hack/chains/kaniko-demo.sh
```diff
@@ -40,10 +39,13 @@ tkn task start kaniko-chains \
 title "Wait a few seconds for chains finalizers to complete"
 sleep 10
 
-# This will use cosign to verify the new build
-$SCRIPTDIR/kaniko-cosign-verify.sh
+title "Create and run the verification task"
+oc create -f verify-attestation-task.yaml
```
lcarva, line 43: Consider using `oc apply` instead so we can run the script multiple times.

hack/chains/kaniko-demo.sh
joejstuart: I think I did while debugging. Adding it back.

hack/chains/verify-attestation-task.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: verify-attestation-signature
```
simonbaird, line 4: I'm wondering if it is too soon to define this task over in https://github.com/redhat-appstudio/build-definitions/tree/main/tasks ?

hack/chains/README.md
```diff
@@ -33,10 +33,6 @@ demonstrate chains.)
     cd hack/chains
     ./create-signing-secret.sh
 
-### Apply some CA cert related hacks to make SSL work for the internal registry
-
-    ./trust-local-cert.sh
-
```
simonbaird, line 39: I reworded this in #188 FYI so we might get a conflict. See if you like my change there and if it makes it worthwhile to keep mentioning this script in the readme.

simonbaird:
Looks cool! I still have the wish to support run the "old" demo as well as the new one using a taskrun, but maybe I don't care so much now.

hack/chains/kaniko-demo.sh
```diff
@@ -40,10 +40,13 @@ tkn task start kaniko-chains \
...
+oc apply -f verify-attestation-task.yaml
+tkn task start verify-attestation-signature \
+  --use-param-defaults \
+  --showlog
 
 pause
 
 # This will use rekor-cli to verify the new build
-$SCRIPTDIR/rekor-verify-taskrun.sh
+#$SCRIPTDIR/rekor-verify-taskrun.sh
```
simonbaird, line 52: Are you planning to add a taskrun for this part as well?

simonbaird:
See also @zregvart's  #197 and the related task definition PR which I think is a stronger version of this.

PR #192, opened 2022-03-18
HACBS-60 - Add support for local cluster rekor

robnester-rh:
This adds support for deploying rekor server in a local cluster.

Signed-off-by: Rob Nester <rnester@redhat.com>

### DIFF
--- a/hack/chains/README.md
+++ b/hack/chains/README.md
@@ -28,6 +28,19 @@ Wait a while until you see mostly healthy/synced at [Argo CD](https://openshift-
 matters for the basic pipeline functionality needed to do builds and
 demonstrate chains.)
 
+### (Optional) Deploy local cluster rekor server
+
+    cd hack/chains
+    ./deploy-local-rekor.sh
+
+To enable the use of a locally deployed rekor instance:
+
+    ./config.sh rekor-local
+
+To return to using the default, public rekor instance:
+
+   ./config.sh rekor-default
+
 ### Create a key-pair signing secret for chains
 
     cd hack/chains
--- a/hack/chains/deploy-local-rekor.sh
+++ b/hack/chains/deploy-local-rekor.sh
@@ -1,0 +1,92 @@
+#!/bin/bash
+#
+# A wrapper for deploying rekor into a local cluster.
+# Uses the deploy-local-rekor.tpl file, which has a placeholder for
+# {domain} which is replaced by getting the domain from the openshift
+# cluster. 
+
+# Make sure we have our template file
+if [[ ! -f $PWD/deploy-local-rekor.tpl ]];then
+  echo "Template file 'deploy-local-rekor.tpl' not found in this directory "; exit 1 
+fi
+
+# Test if we have the argocd cli command:
+if ! command -v argocd &> /dev/null; then
+  echo "The argocd CLI is not installed. Please install and try again."
+  echo "See https://argo-cd.readthedocs.io/en/stable/getting_started/#2-download-argo-cd-cli for details"
+  exit 1
+fi
+
+# Test if we are logged into our cluster as an account with a 'cluster-admin' role.
+if [ "$(oc auth can-i '*' '*' --all-namespaces 2> /dev/null)" != "yes" ]; then
+  echo
+  echo "[ERROR] User '$(oc whoami)' does not have the required 'cluster-admin' role." 1>&2
+  echo "Log into the cluster with a user with the required privileges (e.g. kubeadmin) and retry."
+  exit 1
+fi
+
+# Get our domain from the ingresses config of the cluster
+domain=$( kubectl get ingresses.config.openshift.io cluster -o json|jq -r ".spec.domain" )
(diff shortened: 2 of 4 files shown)


simonbaird:
There's an older version of this PR at https://github.com/hacbs-contract/infra-deployments/pull/13 .

hack/chains/README.md
```diff
@@ -28,6 +28,11 @@ Wait a while until you see mostly healthy/synced at [Argo CD](https://openshift-
 matters for the basic pipeline functionality needed to do builds and
 demonstrate chains.)
 
+### (Optional) Deploy local cluster rekor server
+
+    cd hack/chains
+    ./deploy-local-rekor.sh
+
```
simonbaird, line 35: It would be nice to mention how to configure chains to use that rekor instance here. I guess it's this:

    ./config rekor-local

hack/chains/deploy_local_rekor.sh
```diff
@@ -0,0 +1,92 @@
...
+"""
+
+# Try accessing the log via the URL
+# curl https://rekor-server.$domain/api/v1/log
+while ! curl --fail --insecure --output /dev/null --silent "https://rekor-server.$domain/api/v1/log"; do
+  echo -n .
+  sleep 3
+done
+
+echo "Done"
```
simonbaird, line 92: Maybe this for some user friendly convenience:

    echo "https://rekor-server.$domain"

simonbaird:
Not actually related to this PR, but I got this:
```
./config.sh rekor-local
Error: open .items[.spec].spec.rules[.host].host: no such file or directory
```

simonbaird:
We also need this otherwise chains chokes on the url:

```diff
-    $0 "transparency.url: \"$REKOR_SERVER\"" $2
+    $0 "transparency.url: \"https://$REKOR_SERVER\"" $2
```

simonbaird:
Sorry, wrong button.

simonbaird:
I can see chains is trying to post the log entries, but failing with this error:

```
2022-03-18T20:45:44.437Z error [POST /api/v1/log/entries][422] createLogEntry default  &{Code:422 Message:invalid kind value: "hashedrekord"}
2022-03-18T20:45:44.453Z error Returned an error
```

But actually I don't think that's a blocker for this PR. We can file a new bug to track figuring that one out.

simonbaird:
One nitpick: All the other files are doing the kebab-case rather than the snake_case. If you want to stick with that I don't mind, but consider switching to make it tidier.

simonbaird (changes requested):
I think it's better to include the two fixups for rekor-local here in this PR. Otherwise lgtm. The argocd went green pretty fast, and it's definitely up and running. Nice patch!

hack/chains/deploy_local_rekor.sh
simonbaird: I realized now that `./config.sh rekor-local` shows it to you, I guess it's not too important.

hack/chains/README.md
simonbaird: ..or maybe the deploy script could just assume you want that and do that for you.

hack/chains/deploy_local_rekor.tpl
```diff
@@ -0,0 +1,31 @@
...
+spec:
+  destination:
+    # Define name or server, not both
+    name: ''
+    namespace: rekor-server
+    server: 'https://kubernetes.default.svc'
+  source:
+    path: helm-charts/rekor
+    repoURL: 'https://github.com/sigstore/sigstore-helm-operator'
+    targetRevision: HEAD
```
lcarva, line 14: Should we pin to a particular version for better reproducibility? I guess the downside is that we won't pick up updates automatically which might be desirable in a dev environment.

hack/chains/README.md
```diff
@@ -28,6 +28,11 @@ Wait a while until you see mostly healthy/synced at [Argo CD](https://openshift-
 matters for the basic pipeline functionality needed to do builds and
 demonstrate chains.)
 
+### (Optional) Deploy local cluster rekor server
+
+    cd hack/chains
+    ./deploy-local-rekor.sh
```
lcarva, line 34: This file doesn't exist. The one added has `-` instead of `_`.

simonbaird:
LGTM.

simonbaird:
I think it's good to merge if prow goes green.

hack/chains/deploy-local-rekor.sh
```diff
@@ -0,0 +1,92 @@
+#!/bin/bash
+#
+# A wrapper for deploying rekor into a local cluster.
+# Uses the deploy_local_rekor.tpl file, which has a placeholder for
+# {domain} which is replaced by getting the domain from the openshift
+# cluster. 
+
+# Make sure we have our template file
+if [[ ! -f $PWD/deploy_local_rekor.tpl ]];then
```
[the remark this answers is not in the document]
robnester-rh: Fixed.

simonbaird:
LGTM

PR #195, opened 2022-03-21
Add script for testing building via AppStudio

Michkov:
Script for secret creation and building via application and components
by AppStudio.

Replacement of m2-builds.sh after merge of https://github.com/redhat-appstudio/application-service/pull/74

### DIFF
--- a/hack/build/build-via-appstudio.sh
+++ b/hack/build/build-via-appstudio.sh
@@ -1,6 +1,7 @@
 #!/bin/bash
 SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )" 
 
+COMPONENT=$1
 
 if [ -z "$MY_QUAY_USER" ]; then
   echo "Missing MY_QUAY_USER variable"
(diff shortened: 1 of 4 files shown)


hack/build/build-via-appstudio.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
+SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )" 
+
+
+if [ -z "$MY_QUAY_USER" ]; then
+  echo "Missing MY_QUAY_USER variable"
+  exit 1
+fi
+
+if ! grep -q quay.io ~/.docker/config.json; then
```
scoheb, line 10: This did not work for me on Fedora 35.

I needed to use `${XDG_RUNTIME_DIR}/containers/auth.json`

Michkov: Also having Fedora 35 and it's under `~/.docker/config.json` which is fallback, primary is that `${XDG_RUNTIME_DIR}/containers/auth.json`, updated the script with the file detection

scoheb:
I have an issue with my environment such that application creation fails...

> error: timed out waiting for the condition on applications/test-application

maybe this script should abort if that happens and does not attempt to create the components?

jduimovich (approved):
LGTM

PR #197, opened 2022-03-22
Demo script for running in-cluster cosign tasks

zregvart:
When the Kaniko chains demo results in pushing to the internal OpenShift
registry with this the `cosign verify*` is run in-cluster using Tekton
tasks defined in redhat-appstudio/build-definitions#37.

Ref. https://issues.redhat.com/browse/HACBS-106

### DIFF
--- a/hack/chains/kaniko-cosign-verify.sh
+++ b/hack/chains/kaniko-cosign-verify.sh
@@ -12,11 +12,18 @@ IMAGE_DIGEST=$( kubectl get $TASKRUN_NAME -o jsonpath='{.status.taskResults[?(@.
 IMAGE_URL=$( kubectl get $TASKRUN_NAME -o jsonpath='{.status.taskResults[?(@.name == "IMAGE_URL")].value}' )
 IMAGE_REGISTRY=$( echo $IMAGE_URL | cut -d/ -f1 )
 #IMAGE_REGISTRY=$( oc registry info )
+INTERNAL_REGISTRY=$([[ $IMAGE_REGISTRY == 'image-registry.openshift-image-registry.svc:5000' ]] && echo true || echo false)
+# From https://github.com/redhat-appstudio/build-definitions/ two images are created ${GIT_SHA}-1 and ${GIT_SHA}-2
+# the ${GIT_SHA}-1 contains the tasks we need here, there is a race condition here, the image might not have been
+# built for the HEAD ref used here, in that case wait few minutes for it to be built
+APPSTUDIO_TASKS_IMGREF=quay.io/redhat-appstudio/appstudio-tasks:$(git ls-remote --heads https://github.com/redhat-appstudio/build-definitions.git refs/heads/main|cut -f 1)-1
 
-title "Make sure we're logged in to the registry"
-# Make sure we have a docker credential since cosign will need it
-# (Todo: Probably shouldn't assume kubeadmin user here)
-oc whoami -t | docker login -u kubeadmin --password-stdin $IMAGE_REGISTRY
+if ! $INTERNAL_REGISTRY; then
+  title "Make sure we're logged in to the registry"
+  # Make sure we have a docker credential since cosign will need it
+  # (Todo: Probably shouldn't assume kubeadmin user here)
+  oc whoami -t | docker login -u kubeadmin --password-stdin $IMAGE_REGISTRY
+fi
 
 title "Inspect $TASKRUN_NAME annotations"
 # Just want to show the chains related fields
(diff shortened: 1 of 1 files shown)


hack/chains/kaniko-cosign-verify.sh
```diff
@@ -29,19 +31,67 @@ title "Image digest from task result"
 kubectl get $TASKRUN_NAME -o jsonpath="{.status.taskResults[?(@.name == \"IMAGE_DIGEST\")].value}"
 echo
 
-title "Cosign verify the image"
-# Save the output data to a file so we can look at it later
-# (Actually we could just pipe it to jq because the text goes to stderr I think..?)
-show-then-run "cosign verify --key $SIG_KEY $IMAGE_URL --output-file /tmp/verify.out"
+if $INTERNAL_REGISTRY; then
+  title "Cosign verify the image - using Tekton task"
+  COSIGN_VERIFY_TASK_NAME="cosign-verify-$(openssl rand --hex 10)"
```
simonbaird, line 36: Nice, I was looking for a simple random string creator in create-signing-secret.

hack/chains/kaniko-cosign-verify.sh
```diff
@@ -29,19 +31,67 @@ title "Image digest from task result"
...
+  oc create -f <(echo "
+apiVersion: tekton.dev/v1beta1
+kind: TaskRun
+metadata:
+  name: $COSIGN_VERIFY_TASK_NAME
+spec:
+  taskRef:
+    name: cosign-verify
+    bundle: quay.io/zregvart_redhat/appstudio-tasks:2022-03-22-130413-1
```
simonbaird, line 45: Would we aim to use a different image once the tasks are merged and pushed? Do you want to add a comment about what the plan is?

simonbaird:
Only one suggestion about adding a comment about the image source, since it's a personal quay repo.

Nice patch!

hack/chains/kaniko-cosign-verify.sh
zregvart: Yeah, that's why this is in draft, once https://github.com/redhat-appstudio/build-definitions/pull/37 is merged, we can use https://quay.io/repository/redhat-appstudio/appstudio-tasks instead.

hack/chains/kaniko-cosign-verify.sh
```diff
@@ -29,19 +31,67 @@ title "Image digest from task result"
...
+  tkn taskrun logs --follow "$COSIGN_VERIFY_ATTN_TASK_NAME"
+  tkn taskrun describe "$COSIGN_VERIFY_ATTN_TASK_NAME"
+  tkn taskrun describe "$COSIGN_VERIFY_ATTN_TASK_NAME" -o=go-template='{{range .status.taskResults}}{{if eq .name "VERIFY_ATTESTATION_JSON"}}{{.value}}{{end}}{{end}}' > /tmp/verify-att.out
+else
+  title "Cosign verify the image's attestation"
+  show-then-run "cosign verify-attestation --key $SIG_KEY $IMAGE_URL --output-file /tmp/verify-att.out"
+  # There can be multiple attestations for some reason and hence multiple lines in
+  # this file, which makes it invalid json. For the sake of the demo we'll ignore
+  # all but the last line.
+  tail -1 /tmp/verify-att.out | yq-pretty
```
zregvart, line 93: Need to fix this, it belongs outside of the if/else

simonbaird:
I think this supersedes #191 FYI @joejstuart .

hack/chains/kaniko-cosign-verify.sh
```diff
@@ -29,15 +31,63 @@ title "Image digest from task result"
...
+metadata:
+  name: $COSIGN_VERIFY_TASK_NAME
+spec:
+  taskRef:
+    name: cosign-verify
+    bundle: quay.io/zregvart_redhat/appstudio-tasks:2022-03-22-130413-1
+  params:
+    - name: PUBLIC_KEY
+      value: |
+$(oc get secret signing-secrets -o=go-template='{{index .data "cosign.pub"|base64decode}}' | sed 's/^/        /g')
```
lcarva, line 49: Should we specify the namespace in this command to be `tekton-chains` in case we want to create the TaskRun in another namespace?

lcarva: Or actually... why not just use $SIG_KEY here?

hack/chains/kaniko-cosign-verify.sh
```diff
@@ -1,7 +1,6 @@
 #!/bin/bash
 
 source $(dirname $0)/_helpers.sh
-set -ue
```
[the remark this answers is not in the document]
zregvart: Thanks! I have no idea why I removed that

hack/chains/kaniko-cosign-verify.sh
```diff
@@ -11,11 +10,14 @@ TASKRUN_NAME=taskrun/$( trim-name $TASKRUN_NAME )
...
 
-title "Make sure we're logged in to the registry"
-# Make sure we have a docker credential since cosign will need it
-# (Todo: Probably shouldn't assume kubeadmin user here)
-oc whoami -t | docker login -u kubeadmin --password-stdin $IMAGE_REGISTRY
+if ! $INTERNAL_REGISTRY; then
+  title "Make sure we're logged in to the registry"
+  # Make sure we have a docker credential since cosign will need it
+  # (Todo: Probably shouldn't assume kubeadmin user here)
+  oc whoami -t | docker login -u kubeadmin --password-stdin $IMAGE_REGISTRY
```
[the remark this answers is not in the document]
zregvart: If the registry from the image reference is internal this command will fail since local machine can't access via service, i.e. would get:

```
Error response from daemon: Get "https://image-registry.openshift-image-registry.svc/v2/": dial tcp: lookup image-registry.openshift-image-registry.svc: no such host
```

That's why we pivot to running the task in-cluster below.

hack/chains/kaniko-cosign-verify.sh
zregvart: > Should we specify the namespace in this command to be tekton-chains in case we want to create the TaskRun in another namespace?

I'll add it just to be on the safe side if this is copied elsewhere, the `hack/chains/kaniko-demo.sh` that is (currently) sole client of this switches to `tekton-chains`, so that's why it works now.

> Or actually... why not just use $SIG_KEY here?

I don't want to give the task access to the secret as it contains both private and public keys.

hack/chains/kaniko-cosign-verify.sh
```diff
@@ -11,11 +11,18 @@ TASKRUN_NAME=taskrun/$( trim-name $TASKRUN_NAME )
 IMAGE_URL=$( oc get $TASKRUN_NAME -o json | jq -r '.status.taskResults[1].value' )
 IMAGE_REGISTRY=$( echo $IMAGE_URL | cut -d/ -f1 )
 #IMAGE_REGISTRY=$( oc registry info )
+INTERNAL_REGISTRY=$([[ $IMAGE_REGISTRY == 'image-registry.openshift-image-registry.svc:5000' ]] && echo true || echo false)
+# From https://github.com/redhat-appstudio/build-definitions/ two images are created ${GIT_SHA}-1 and ${GIT_SHA}-2
+# the ${GIT_SHA}-1 contains the tasks we need here, there is a race condition here, the image might not have been
+# built for the HEAD ref used here, in that case wait few minutes for it to be built
+APPSTUDIO_TASKS_IMGREF=quay.io/redhat-appstudio/appstudio-tasks:$(git ls-remote --heads https://github.com/redhat-appstudio/build-definitions.git refs/heads/main|cut -f 1)-1
```
simonbaird, line 18: I want to check that I understand the motivation here.

Most of the similar code in this repo seems to hard code a digest, which I guess means choosing stability in the stability/freshness tradeoff. Is that right? And are you choosing freshness over stability here intentionally? And is the reason we can't just use "latest" because there is no latest tag in this repo?

zregvart: Yeah, my motivation was that when looking at this I could not find any references to `quay.io/redhat-appstudio/appstudio-tasks` (only to `quay.io/redhat-appstudio/build-templates-bundle`), and that I like issues to be seen early. In this case we already saw that the build of the image failed (and was fixed) and in the future we could see issues with the tasks themselves.

zregvart:
@lcarva @simonbaird is this okay to merge?

hack/chains/kaniko-cosign-verify.sh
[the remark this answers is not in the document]
lcarva: That's not what I meant :)

I'm was talking about the `Todo` comment. You don't have to provide `-u kubeadmin`:
```
oc whoami -t | docker login -u unused --password-stdin $IMAGE_REGISTRY
```

lcarva (approved):

hack/chains/kaniko-cosign-verify.sh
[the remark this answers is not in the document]
zregvart: I get it now, I'll create a followup PR, this wasn't changed in this PR so best kept it as a separate followup cleanup PR.

lcarva:
Time to merge this?

PR #198, opened 2022-03-22
Support digest param in rekor-image-lookup script

simonbaird:
Not super important, but it makes the rekor-image-lookup.sh script
more versatile and hence more generally useful.

### DIFF
--- a/hack/chains/rekor-image-lookup.sh
+++ b/hack/chains/rekor-image-lookup.sh
@@ -5,8 +5,13 @@ IMAGE_URL=${1:-$DEFAULT_URL}
 DEFAULT_REKOR_SERVER=https://rekor.sigstore.dev
 REKOR_SERVER=${2:-$DEFAULT_REKOR_SERVER}
 
-# In reality we might know the digest already, but let's look it up
-IMAGE_DIGEST=$( skopeo inspect --no-tags docker://$IMAGE_URL | jq -r .Digest )
+if [[ $IMAGE_URL =~ ^(sha256:)?[0-9a-fA-F]{64}$ ]]; then
+  # Assume the param is already an image digest
+  IMAGE_DIGEST="$IMAGE_URL"
+else
+  # Assume the param is an image url and look up the digest with skopeo
+  IMAGE_DIGEST=$( skopeo inspect --no-tags docker://$IMAGE_URL | jq -r .Digest )
+fi
 
 # Use the digest to do a rekor search
 UUIDS=$( rekor-cli search --sha "$IMAGE_DIGEST" --rekor_server $REKOR_SERVER 2>/dev/null )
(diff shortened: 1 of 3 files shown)


hack/chains/rekor-image-lookup.sh
```diff
@@ -5,8 +5,13 @@ IMAGE_URL=${1:-$DEFAULT_URL}
 DEFAULT_REKOR_SERVER=https://rekor.sigstore.dev
 REKOR_SERVER=${2:-$DEFAULT_REKOR_SERVER}
 
-# In reality we might know the digest already, but let's look it up
-IMAGE_DIGEST=$( skopeo inspect --no-tags docker://$IMAGE_URL | jq -r .Digest )
+if [[ $IMAGE_URL =~ ^(sha256:)?[0-9a-f]+$ ]] && [[ ${#IMAGE_URL} > 63 ]]; then
```
lcarva, line 8: I wonder if this will work in bash: `^(sha256:)?[0-9a-f]{63}$`

hack/chains/rekor-verify-taskrun.sh
```diff
@@ -8,6 +8,7 @@ TASKRUN_NAME=${1:-$( tkn taskrun describe --last -o name )}
 TASKRUN_NAME=taskrun/$( trim-name $TASKRUN_NAME )
 
 # Let's not hard code the image url or the registry
+IMAGE_DIGEST=$( kubectl get $TASKRUN_NAME -o json | jq -r '.status.taskResults[0].value' )
```
lcarva, line 11: It might be better to not rely on the order of the results. How about something like this:
```
jq -r '.status.taskResults[] | select(.name=="IMAGE_DIGEST") | .value'
```

We could do something similar on the line below.

lcarva (approved):

hack/chains/rekor-image-lookup.sh
simonbaird: Yeah, I tried that, no dice. I'll give it another google, maybe the curlies need escaping or something...

hack/chains/rekor-verify-taskrun.sh
simonbaird: That's a good suggestion. Let me fix it.

simonbaird: https://github.com/simonbaird/infra-deployments/blob/c5b101b26e0476529fd36f140abb4f5b904a772e/hack/contract/fetch-data.sh#L42 is relevant.

hack/chains/rekor-image-lookup.sh
simonbaird: Ugh, it works fine. Not sure what I was doing wrong the other day..

zregvart (approved):
LGTM! This could conflict with #197, perhaps merge this first?

lcarva (approved):

simonbaird:
Failure says `The maximum number of VPCs has been reached` so I don't think it's related to the patch.

Let's merge.

PR #199, opened 2022-03-23
Draft: Enterprise Contract POC with OPA/Rego

simonbaird:
Early demo showing how rego might work.

### DIFF
--- a/hack/contract/policy/main.rego
+++ b/hack/contract/policy/main.rego
@@ -1,0 +1,12 @@
+
+package hacbs.contract.main
+import data.hacbs.contract.k8s_sanity
+import data.hacbs.contract.transparency_urls
+
+deny[msg] {
+  count(k8s_sanity.deny[msg]) > 0
+}
+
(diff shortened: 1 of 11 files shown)


hack/contract/policy/main.rego
```diff
@@ -10,3 +11,7 @@ deny[msg] {
 deny[msg] {
   count(transparency_urls.deny[msg]) > 0
 }
+
+deny[msg] {
```
lcarva, line 15: I wonder if there's an easy way to provide both `deny` and `allow` variables.

simonbaird:
Thinking about whether this should be merged, or whether we should start out in whatever repo the EC code is supposed to live in. Maybe it's better to start a new repo now, what do you think?

simonbaird:
This was a useful POC, but the code isn't going to live in this repo, so let's abandon this now.

See https://github.com/hacbs-contract/ec-policies and maybe some upcoming PRs in the build-definitions repo.

PR #200, opened 2022-03-23
create chains signing secret for ci runs

joejstuart:


.ci/oci-e2e-deployment.sh
```diff
@@ -87,6 +87,9 @@ function executeE2ETests() {
 
 createQuayPullSecrets
 
+# create signing secret to test tekton-chains
+/bin/bash "$WORKSPACE"/hack/chains/create-signing-secret.sh
+
```
simonbaird, line 92: Do we expect cosign to be installed already?

simonbaird: Maybe a `command -v cosign ...` on line 23 for consistency?

joejstuart: nope. Good catch. I just added it to the Dockerfile and a check this script.

flacatus:
The PR should fail because the openshift ci have a bug and takes the dockerfile: https://github.com/redhat-appstudio/infra-deployments/pull/200/files#diff-6704820b78e3aea8581dff31b66a527b85fc1e2635264f0c6961acb16529ff24 from main branch :(

.ci/oci-e2e-deployment.sh
Michkov: tekton-chains namespace does not exist yet. Should not be this rather in hack/bootstrap-cluster.sh since it should be set for all clusters? In case of putting it into bootrsap-cluster.sh I would do it only when cosign is installed (allow to continue without cosign)

Michkov: also possible to add into argo, probably as a PostSync job like `components/build/tekton-results/ingress-annotation-job.yaml`

Issue #202, opened 2022-03-24
./hack/bootstrap-cluster.sh not working, realpath not available


chrisahl:
I am trying to follow the README for [Bootstrap AppStudio](https://github.com/redhat-appstudio/infra-deployments#bootstrap-app-studio) and there is a failure due to realpath command not existing.  There should be:
- a check to verify this tool is available
- some documentation listing this prereq, similar to yq and jq. 

Mac Output:
```
./hack/bootstrap-cluster.sh upstream
./hack/bootstrap-cluster.sh: line 5: realpath: command not found

Installing the OpenShift GitOps operator subscription:
error: the path "/openshift-gitops/subscription-openshift-gitops.yaml" does not exist

Waiting for default project (and namespace) to exist: ................................................^C
```

PR #203, opened 2022-03-24
Fix script to work on Mac without realpath tool

chrisahl:
Signed-off-by: Chris Ahl <cahl@redhat.com>

For issue #202

### DIFF
--- a/hack/bootstrap-cluster.sh
+++ b/hack/bootstrap-cluster.sh
@@ -2,7 +2,13 @@
 
 MODE=$1
 
-ROOT="$(realpath -mq ${BASH_SOURCE[0]}/../..)"
+
+if which realpath > /dev/null; then
+  ROOT="$(realpath -mq ${BASH_SOURCE[0]}/../..)"
+else
+  ROOT="$( cd "$(dirname "$0"/..)" >/dev/null 2>&1 ; pwd -P )"
+fi
+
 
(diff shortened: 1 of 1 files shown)


hack/bootstrap-cluster.sh
```diff
@@ -2,7 +2,13 @@
 
 MODE=$1
 
-ROOT="$(realpath -mq ${BASH_SOURCE[0]}/../..)"
+
+if which realpath > /dev/null; then
+  ROOT="$(realpath -mq ${BASH_SOURCE[0]}/../..)"
+else
+  ROOT="$( cd "$(dirname "$0"/..)" >/dev/null 2>&1 ; pwd -P )"
```
Michkov, line 9: `ROOT="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"/..` - does this work for you? If so, then please use it, it's used in other scripts. (No need to have that `if`)

chrisahl: Yes, that works for me.  Updated

chrisahl: Note however that the script never completes for me.  Have to ctrl-c.  

```
Installing the OpenShift GitOps operator subscription:
subscription.operators.coreos.com/openshift-gitops-operator created

Waiting for default project (and namespace) to exist: .....................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................^C
```

PR #204, opened 2022-03-25
sync chains singing secret with argocd

joejstuart:
include chains signing secret


components/build/tekton-chains/chains-certs.yaml
```diff
@@ -113,3 +113,37 @@ spec:
       terminationGracePeriodSeconds: 30
       serviceAccount: chains-secrets-admin
       serviceAccountName: chains-secrets-admin
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: tekton-chains-signing-secret
```
Michkov, line 120: being deployed into default namespace:
```
Error creating: pods "tekton-chains-signing-secret-" is forbidden: error looking up service account default/chains-secrets-admin: serviceaccount "chains-secrets-admin" not found
```

joejstuart: I tested this with `kubectl apply` on an existing cluster so it passed. thanks for catching this.

components/build/tekton-chains/chains-certs.yaml
```diff
@@ -113,3 +113,38 @@ spec:
...
+  namespace: tekton-chains
+  annotations:
+    argocd.argoproj.io/hook: Sync
+    argocd.argoproj.io/hook-delete-policy: HookSucceeded
```
Michkov, line 124: Please remove this annotation, it's deleting the job, it's better to use default delete policy which deletes job before running new one. So it's better for debugging.

Also I noticed that when I remove the secret and run secret then it job fails for first time on:
```
I0325 13:02:26.662958 8 request.go:665] Waited for 1.160947575s due to client-side throttling, not priority and fairness, request: GET:https://172.30.0.1:443/apis/apps/v1?timeout=32s
I0325 13:02:36.861548 8 request.go:665] Waited for 11.35708957s due to client-side throttling, not priority and fairness, request: GET:https://172.30.0.1:443/apis/console.openshift.io/v1alpha1?timeout=32s
Successfully created secret signing-secrets in namespace tekton-chains
Error: open cosign.pub: permission denied
main.go:46: error during command execution: open cosign.pub: permission denied
```
But then it's restarted and fixed.

Michkov: It will be issue of the current working directory, probably should be fixed by `cd /tmp/` bofore running cosign command (current working dir is read-only, so it failed on saving cosign.pub, which is fine but better to not end with fail).

components/build/tekton-chains/chains-certs.yaml
```diff
@@ -113,3 +113,38 @@ spec:
...
+        - |
+          # Once the key-pair has been set it's marked as immutable so it can't be updated.
+          # Try to handle that nicely. The object is expected to always exist so check the data.
+          SIG_KEY_DATA=$(kubectl get secret signing-secrets -n tekton-chains -o jsonpath='{.data}')
+          [[ -n $SIG_KEY_DATA ]] && echo "Signing secret exists." && exit
+
+          # To make this run conveniently without user input let's create a random password
+          RANDOM_PASS=$( head -c 12 /dev/urandom | base64 )
+
+          # Generates the key pair secret directly in the cluster. Also creates public key file cosign.pub
```
simonbaird, line 144: Maybe no need to mention the public key file in this comment, since we aren't going to save it.

simonbaird:
I was thinking it would be better to put the secret hook in its own separate file, e.g. components/build/tekton-chains/chains-signing-secret.yaml rather than lumping it in with the certs. They are pretty different things.

Or, if there are good benefits of having them in a single file, how about naming it more generically components/build/tekton-chains/chains-certs-and-signing-secret.yaml or omponents/build/tekton-chains/chains-post-sync-hooks.yaml

joejstuart:
> I was thinking it would be better to put the secret hook in its own separate file, e.g. components/build/tekton-chains/chains-signing-secret.yaml rather than lumping it in with the certs. They are pretty different things.

The job I created depends on the service account and role/bindings, so I added it to the same file.

> Or, if there are good benefits of having them in a single file, how about naming it more generically components/build/tekton-chains/chains-certs-and-signing-secret.yaml or components/build/tekton-chains/chains-post-sync-hooks.yaml , omponents/build/tekton-chains/chains-post-sync-config.yaml

 How about chains-secrets-config.yaml since each resource is created to modify or create a secret?

Michkov (approved):

simonbaird:
LGTM, let's merge.

PR #209, opened 2022-03-25
Remove instructions for chains signing secret

simonbaird:
The key pair secret will now be created automatically using a Gitops
hook, so no need to explain how to create one manually.

### DIFF
--- a/hack/chains/create-signing-secret.sh
+++ b/hack/chains/create-signing-secret.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 ###
-### NB: There is hook that runs this after Gitops sync so you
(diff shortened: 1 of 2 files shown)


hack/chains/create-signing-secret.sh
```diff
@@ -1,4 +1,9 @@
 #!/bin/bash
+###
+### NB: There is hook that runs this after Gitops sync so you
```
joejstuart, line 3:
```suggestion
### NB: There is a hook that runs this after Gitops sync so you
```

joejstuart:
lgtm

simonbaird:
It's a readme and a comment update, so risk is approaching zero, and the e2e doesn't matter. Let's merge.

PR #211, opened 2022-03-28
Update chains readme

caugello:
Adds the tested platforms for cluster-bot.

### DIFF
--- a/hack/chains/README.md
+++ b/hack/chains/README.md
@@ -19,7 +19,8 @@ For CRC:
 
 For cluster-bot (in slack):
 
-    launch 4.9.15 # or launch latest ?
+    launch 4.9.15 $PLATFORM
+    # Use one of the following tested platforms: ovirt, openstack, azure, aws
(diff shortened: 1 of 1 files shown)


hack/chains/README.md
```diff
@@ -19,8 +19,9 @@ For CRC:
 
 For cluster-bot (in slack):
 
-    launch 4.9.15 # or launch latest ?
+    launch 4.9.15 $PLATFORM # or launch latest ?
     # ...and follow the instructions to authenticate once it's ready
+    # Use one of the following tested platforms: ovirt, openstack, azure, aws
```
simonbaird, line 24: 1. Did you want to mention which one is the most likely to work with the chains demos?
2. Every time I tried launch latest it would fail, so I'm not sure if it's useful. Did you get that to work ever? If not, then maybe we should remove it

simonbaird (approved):

hack/chains/README.md
caugello: Yes and removed the launch command.

PR #227, opened 2022-03-31
Move build definitions CI to Pipelines as Code

brunoapimentel:
The CI for the build component is being moved to Pipelines as Code in its own repository: https://github.com/redhat-appstudio/build-definitions/pull/41.


components/build/.tekton/pipelines-as-code.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: pipelinesascode.tekton.dev/v1alpha1
+kind: Repository
+metadata:
+  name: build-definitions-pac
+  namespace: build-tenmplates
```
Michkov, line 5: typo

Michkov: and the namespace should not be here, it's set in kustomization.yaml anyway.

PR #232, opened 2022-04-04
Avoid ArgoCD SharedResourceWarning on build app

Michkov:
ArgoCD is setting `app.kubernetes.io/instance` to argoCD app - https://argo-cd.readthedocs.io/en/stable/faq/#why-is-my-app-out-of-sync-even-after-syncing.
Pipelines-as-code has it set directly in yaml to `default`. That's why
we can see warning in ArgoCD -> `CustomResourceDefinition/repositories.pipelinesascode.tekton.dev is part of applications build and default`

### DIFF
--- a/components/build/pipelines-as-code/kustomization.yaml
+++ b/components/build/pipelines-as-code/kustomization.yaml
@@ -5,3 +5,5 @@ resources:
 
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
+commonLabels:
+  app.kubernetes.io/instance: "build"


chmouel:
interesting, tekton sets this as well, what is the best practice? does it need to not be set (in pipelines-as-code) ?

Michkov:
@chmouel other components are not setting it, so I was not sure if it's not really used in PaC, if not then it's probably should not be set or maybe override in kustomization.yaml is fine as well. It's [a common kubernetes label](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels).

chmouel:
yes i followed that common labels from kubernetes  and pipeline does set it (upstream) https://github.com/tektoncd/pipeline/blob/30f4051a70e67373d042471c0cd5f546bc5a93b3/config/300-pipeline.yaml#L20 

anyhoo seems fine to just do this from kustomization ...

PR #256, opened 2022-04-11
Mention gitops sync workaround in chains readme

simonbaird:
Doing this helps the Gitops sync complete, otherwise you get
endless failure:

  one or more objects failed to apply, reason: namespaces
  "application-service" not found. Retrying attempt #432 at 7:25PM.

I believe it's because the pvc-cleaner objects try to use the
application-service namespace which isn't created if you don't
install the has appliction.


Michkov:
Should be fixed by #241 (but not sure how long it will take to finish it)

simonbaird:
I mentioned the #241 fix in the readme.

simonbaird:
It's a readme change, so low risk. Let's merge.

PR #260, opened 2022-04-12
Update PipelinesAsCode to 0.6.0

Michkov:


chmouel:
not sure how it works but i guess it is only used for new deployment, 0.5 to 0.6 may need some manual tasks for upgrade

chmouel (approved):

PR #264, opened 2022-04-13
fix: Fix the SPI OAuth service deployment

metlos:
This adds two things:

* Run the OAuth service in debug mode for more detailed logging.
* Modify the deployment to authenticate with devsandbox proxy.

Hopefully, this modification will enable the service to work with Devsandbox proxy authorization.

### DIFF
--- a/components/spi/oauth-service-deployment-patch.json
+++ b/components/spi/oauth-service-deployment-patch.json
@@ -5,7 +5,7 @@
     "value": [
       {
         "name": "API_SERVER",
-        "value": "https://api-toolchain-host-operator.apps.appstudio-stage.x99m.p1.openshiftapps.com:443"
+        "value": ""
       },
       {
         "name": "DEVMODE",
--- a/components/spi/oauth-service-deployment-patch.json
+++ b/components/spi/oauth-service-deployment-patch.json
@@ -5,7 +5,7 @@
     "value": [
       {
         "name": "API_SERVER",
-        "value": ""
+        "value": "https://api-toolchain-host-operator.apps.appstudio-stage.x99m.p1.openshiftapps.com:443"
       },
       {
         "name": "DEVMODE",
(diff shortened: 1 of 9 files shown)


skabashnyuk (approved):

hack/development-mode.sh
```diff
@@ -15,6 +15,10 @@ $ROOT/hack/util-update-app-of-apps.sh $MY_GIT_REPO_URL development $MY_GIT_BRANC
 # this needs to be pushed to your fork to be seen by argocd
 $ROOT/hack/util-set-development-repos.sh $MY_GIT_REPO_URL development $MY_GIT_BRANCH
 
+# set the API server which SPI uses to authenticate users to empty string (by default) so that multi-cluster
```
scoheb, line 18: does this need to be added to `preview.sh`?

components/spi/oauth-service-deployment-patch.json
```diff
@@ -0,0 +1,12 @@
+[
+  {
+    "op": "add",
+    "path": "/spec/template/spec/containers/0/env",
+    "value": [
+      {
+        "name": "API_SERVER",
+        "value": "https://api-toolchain-host-operator.apps.appstudio-stage.x99m.p1.openshiftapps.com:443"
```
Michkov, line 8: what will happen when value will be empty? Will it use default kube API?

sparkoo: yes

Michkov (approved):

sparkoo (approved):
I don't like hardcoded apiserver value, but we need this PR to debug the issue on current deployment.

hack/development-mode.sh
metlos: Good point, it's been added.

skabashnyuk (approved):

PR #265, opened 2022-04-14
Enable github groups sync

Michkov:
Sync teams from github org redhat-appstudio as groups in Openshift.

Post merge action:

- [ ] Create secret for connection with github app - `oc create secret generic github-redhat-appstudio --from-literal=appId=190760 --from-file=privateKey=appstudio-group-sync-operator.2022-04-13.private-key.pem`


psturc:
LGTM, although I was't able to verify this against my cluster. I assume we can ignore the appstudio-e2e-deployment job for now since it doesn't deploy the `authentication` component, which was updated in this PR?

psturc (approved):

PR #266, opened 2022-04-14
Add Pipelines as Code for build-service

mbharatk:


components/build/build-service/.tekton/kustomization.yaml
```diff
@@ -0,0 +1,15 @@
+resources:
+- serviceaccount.yaml
+- pipelines-as-code.yaml
+- pvc.yaml
+
+namespace: build-service-system
```
mmorhun, line 6: I think the namespace should be `build-service`.
The default `build-service-system` is generated by operator-sdk, but in our components we don't usually use the `system` suffix: `application-service` namespace is an example.

components/build/build-service/namespace.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: build-service-system
```
mmorhun, line 4: Same as above

components/build/build-service/.tekton/pipelines-as-code.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: pipelinesascode.tekton.dev/v1alpha1
+kind: Repository
+metadata:
+  name: build-service-pac
+  namespace: build-service-system
```
mmorhun, line 5: We missed it last time. Should be consistent.

mbharatk: Done

Michkov (approved):

PR #285, opened 2022-04-22
Add overlay for e2e.

rhopp:
This PR adds new deployment mode - `e2e`. This mode is exactly the same as `preview`, but it also installs RHSSO with preconfigured realm to be used by toolchain operators instead of sso.redhat.com.

This PR also adds installation of toolchain operators in CI scripts & their configuration (to use the RHSSO installed by gitops on the same cluster).

Possible future work would be to run the tests using users from the keycloak realm (ideally through toolchain proxy) (currently we are running everything with kubeadmin).

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ See `hack/quicklab/README.md`
 
 ### Bootstrap App Studio
 Steps:
-1) Run `./hack/bootstrap-cluster.sh [$MODE]` which will bootstrap Argo CD (using OpenShift GitOps) and setup the Argo CD `Application` Custom Resources (CRs) for each component. This command will output the Argo CD Web UI route when it's finished. For upstream mode keep the $MODE empty or "upstream". For development mode and preview mode set `$MODE` to `development` or `preview`, the modes are described in section 'Development modes for your own clusters'.
+1) Run `./hack/bootstrap-cluster.sh [$MODE]` which will bootstrap Argo CD (using OpenShift GitOps) and setup the Argo CD `Application` Custom Resources (CRs) for each component. This command will output the Argo CD Web UI route when it's finished. For upstream mode keep the $MODE empty or "upstream". For development mode and preview mode set `$MODE` to `development`, `preview` or `e2e`, the modes are described in section 'Development modes for your own clusters'.
 2) Open the Argo CD Web UI to see the status of your deployments. You can use the route from the previous step and login using your OpenShift credentials (using the 'Login with OpenShift' button), or login to the OpenShift Console and navigate to Argo CD using the OpenShift Gitops menu in the Applications pulldown.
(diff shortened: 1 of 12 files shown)


hack/preview.sh
```diff
@@ -2,6 +2,8 @@
 
 ROOT="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"/..
 
+OVERLAY=$1
```
Michkov, line 5: Use development by default -> `OVERLAY=${1-development}`

hack/preview.sh
```diff
@@ -36,10 +38,16 @@ git checkout -b $PREVIEW_BRANCH
 # this needs to be pushed to your fork to be seen by argocd
 $ROOT/hack/util-set-development-repos.sh $MY_GIT_REPO_URL development $PREVIEW_BRANCH
```
Michkov, line 39: Update development to $OVERLAY

hack/preview.sh
```diff
@@ -36,10 +38,16 @@ git checkout -b $PREVIEW_BRANCH
...
 
+
 # set the API server which SPI uses to authenticate users to empty string (by default) so that multi-cluster
 # setup is not needed
 $ROOT/hack/util-set-spi-api-server.sh "$SPI_API_SERVER"
 
+if [ -n "$OVERLAY" ]; then
+    echo "Configuring E2E overlay"
+    $ROOT/hack/util-set-development-repos.sh $MY_GIT_REPO_URL $OVERLAY $PREVIEW_BRANCH
```
Michkov, line 48: not needed

README.md
```diff
@@ -144,6 +144,7 @@ The script also supports branches automatically. If you work in a checked out br
 There are two workflows for develompent provided:
 1) Development mode - work in the feature branch, apply changes related to your fork, revert the changes when the work is done
 2) Preview mode - work in a feature branch, apply script which creates new preview branch and create additional commit with for customization
+3) e2e mode - Same as `preview` mode, but also installs RHSSO with preconfigured realm to be used with toolchain operators.
```
Michkov, line 147: not sure about name of the mode, e2e is overused, in this case there is just sso 'addon', so maybe `sso mode`?

rhopp: My wishful thinking for the future with this mode is to be able to deploy HAC as well. Then this mode would contain whole appstudio experience (HAC+SSO+Appstudio) on single (ephemeral) cluster. Not sure how possible is that though.

hack/preview.sh
rhopp: hehe... That's the main reason I converted this PR to draft... After sending it I noticed this part I didn't clean up properly :-D

rhopp:
This PR is blocked by https://issues.redhat.com/browse/CRT-1511

Michkov:

PR #294, opened 2022-04-25
Use upstream patched version of chains

lcarva:
Use a nightly-build image because the current latest version, 0.8.0,
contains several bugs that make Chains unusable for our use case. Once
version 0.9.0 is published, change this reference to that.

### DIFF
--- a/components/build/tekton-chains/chains-controller-deployment.yaml
+++ b/components/build/tekton-chains/chains-controller-deployment.yaml
@@ -8,7 +8,12 @@ spec:
     spec:
       containers:
       - name: tekton-chains-controller
-        image: ghcr.io/hacbs-contract/chains/controller:cd2d8c2c9fea3902248dd811913e7ea4fa73d639
+        # Use a nightly-build image because the current latest version, 0.8.0, contains
+        # several bugs that make Chains unusable for our use case. Once version 0.9.0
+        # is published, change this reference to that. Contract team fork may be useful
+        # in future for testing features not yet upstream:
+        # image: ghcr.io/hacbs-contract/chains/controller@sha256...
+        image: gcr.io/tekton-nightly/github.com/tektoncd/chains/cmd/controller@sha256:9b4bef2e9d617dddfc0243b78db2ce8b0f176460bc408bd28e5b31cba79bdece
         volumeMounts:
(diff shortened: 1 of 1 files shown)


components/build/tekton-chains/chains-controller-deployment.yaml
```diff
@@ -8,7 +8,10 @@ spec:
     spec:
       containers:
       - name: tekton-chains-controller
-        image: ghcr.io/hacbs-contract/chains/controller:cd2d8c2c9fea3902248dd811913e7ea4fa73d639
+        # Use a nightly-build image because the current latest version, 0.8.0, contains
+        # several bugs that make Chains unusable for our use case. Once version 0.9.0
+        # is published, change this reference to that.
+        image: gcr.io/tekton-nightly/github.com/tektoncd/chains/cmd/controller@sha256:9b4bef2e9d617dddfc0243b78db2ce8b0f176460bc408bd28e5b31cba79bdece
```
lcarva, line 14: I initially wanted this to use the `:latest` tag so we would catch breaking upstream issues right away. But I've talked myself out of it because it would be a terrible development experience. Also, the `:latest` tag causes ambiguity on which image will actually be used.

We can always update this reference when there's a need to (a new bug fix, a new release, etc).

components/build/tekton-chains/chains-controller-deployment.yaml
```diff
@@ -8,7 +8,10 @@ spec:
     spec:
       containers:
       - name: tekton-chains-controller
-        image: ghcr.io/hacbs-contract/chains/controller:cd2d8c2c9fea3902248dd811913e7ea4fa73d639
```
simonbaird, line 11: Could maybe leave it with a comment, e.g.:

      # Contract team fork. May be useful in future for testing pipeline run attestation:
      #image: ghcr.io/hacbs-contract/chains/controller@sha256...

simonbaird (approved):

components/build/tekton-chains/chains-controller-deployment.yaml
lcarva: Yes! done.

simonbaird:
Can't see any reason to delay - let's merge.

PR #293, opened 2022-04-25
Add demo of enforcing EC in a pipeline

lcarva:
Signed-off-by: Luiz Carvalho <lucarval@redhat.com>

PR #295, opened 2022-04-25
Bash script to copy the chains public sig key

simonbaird:
It copies it to a secret in your current project/namespace so it can
be used conveniently without requiring access to the secret in
tekton-chains which includes the protected secret key.

Inspired by:
https://github.com/redhat-appstudio/build-definitions/pull/56
https://github.com/redhat-appstudio/infra-deployments/pull/293

### DIFF
--- a/hack/chains/release-pipeline-with-ec-demo.sh
+++ b/hack/chains/release-pipeline-with-ec-demo.sh
@@ -105,7 +105,7 @@ spec:
 
 " | oc apply -f - > /dev/null
 
-oc get pipeline simple-release -o yaml | yq e '.spec'
+oc get pipeline simple-release -o yaml | yq e '.spec' -
 
 
 title "Verify Push Secret"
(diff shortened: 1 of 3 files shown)


lcarva (approved):

simonbaird:
I'll take the previous approve from @lcarva and merge this. Let's do a rebase since it's on top anyhow afaict.

simonbaird:
Also there's one more typo fix i want to get merged quickly.

hack/chains/release-pipeline-with-ec-demo.sh
```diff
@@ -106,7 +105,7 @@ spec:
 
 " | oc apply -f - > /dev/null
 
-oc get pipeline simple-release -o yaml | yq e '.spec'
+oc get pipeline simple-release -o yaml | yq e '.spec' -
```
simonbaird, line 108: I realized after merge that maybe I need to upgrade yq..

PR #305, opened 2022-04-26
allow sbose78 to do HACBS releases in shbose

sbose78:


sbose78:
failure is unrelated, taking this in.

PR #310, opened 2022-04-27
End to end demo script for enterprise contract

zregvart:
This runs the build and the sample release pipelines. The main goal of the demo is to showcase how the test stream results from the build pipeline are picked up by the Enterprise Contract verify task run in a release pipeline.

Ref. https://issues.redhat.com/browse/HACBS-357

Also includes:

[Pipeline parameter for verify-enterprise-contract](https://github.com/redhat-appstudio/infra-deployments/commit/eba205ea9c03e9334e58e0e32380722f39ad50d1)

We need to provide the `PIPELINERUN_NAME` parameter to the `verify-enterprise-contract` task. This allows it being specified via environment variable `BUILD_PIPELINE_RUN`, and in case it is not specified the script auto-guesses a PipelineRun that has a TaskRun with a result named `HACBS_TEST_OUTPUT`.

### DIFF
--- a/hack/chains/end-to-end-demo.sh
+++ b/hack/chains/end-to-end-demo.sh
@@ -1,0 +1,140 @@
+#!/bin/bash
+#
+# A demo script that runs a build pipeline including HACBS specific tests and a sample release pipeline with HACBS Enterprise Contract test
+
+APP_GIT_REPOSITORY=https://github.com/jduimovich/single-nodejs-app
+APP_BUILDER=nodejs-builder
+APP_IMAGE_NAME="${APP_GIT_REPOSITORY/*\/}"
+APP_IMAGE_REF="quay.io/${MY_QUAY_USER}/${APP_IMAGE_NAME}"
+
+# Check for $MY_QUAY_USER
+if [[ -z ${MY_QUAY_USER+x} ]]; then
+  echo "🛑 Make sure to set the \$MY_QUAY_USER environment variable, it needs to be set to the
+   Quay.io organization where the resulting images will be pushed"
+  exit 1
+fi
+
+set -euo pipefail
+
+echo "ℹ️ Please wait..."
+
+DOCKER_LOGIN=$(docker login quay.io </dev/null 2>&1 || true)
+if [[ "${DOCKER_LOGIN}" != *"Login Succeeded"* ]]; then
+  echo "ℹ️ Please enter the quay.io credentials for ${MY_QUAY_USER}, if you're using a robot account
+and token make sure that it's allowed to push to ${APP_IMAGE_NAME} repository.
+"
+  docker login quay.io --username "${MY_QUAY_USER}"
+fi
+
+# Making the quay.io repository public, this is a requirement right now for the Test stream tasks to succeed
+if [[ "$(curl -s "https://quay.io/api/v1/repository?namespace=${MY_QUAY_USER}&public=true" | jq ".repositories[] | select(.name==\"${APP_IMAGE_NAME}\") | has(\"name\")")" != 'true' ]]; then
+  echo "🛑 Make sure that the quay.io repository '${APP_IMAGE_NAME}' exists and its visibility
+is set to public, create an empty repository here:
+
+👉 https://quay.io/new/?name=${APP_IMAGE_NAME}
+
+if the repository already exists change the visibility here:
+
+👉 https://quay.io/repository/${MY_QUAY_USER}/${APP_IMAGE_NAME}?tab=settings"
+  exit 1
+fi
+
+# We're going to assume running with kubeadmin
+if [[ "$(kubectl auth can-i '*' '*' < /dev/null 2>&1)" != 'yes' ]]; then
+  echo "🛑 Make sure that you're logged in as kubeadmin, we make assumptions here"
+  exit 1
+fi
+
+# Checking the status of the build component as a way to make sure that the cluster is okay to run the demo
+if [[ "$(kubectl get -n openshift-gitops applications.argoproj.io build -o jsonpath='{.status.sync.status}')" != 'Synced' ]]; then
+  echo "🛑 The build application is not synced. Make sure that the cluster is setup by running
+hack/bootstrap-cluster.sh first see:
+
+👉 https://github.com/redhat-appstudio/infra-deployments#bootstrapping-a-cluster"
+  exit 1
+fi
+
+echo "📜 The outline of the demo
+  1. We'll start a build pipeline using build-definitions/hack/test-build.sh
+     to build a Node.js app. This will include running the sanity checks from
+     the Test stream and gathering attestations and signatures from the
+     Contract team.
+  2. Next, a sample release pipeline will be run that will check the results
+     from the build pipeline and copy the image to the production environment.
+
+We'll show the commands being run, and point to data or visualisations.
+"
+
+HACK_CHAINS_DIR="$(dirname "$0")"
+
+BUILD_DEFINITIONS_DIR="${HACK_CHAINS_DIR}/../../../build-definitions"
+if [[ ! -d "${BUILD_DEFINITIONS_DIR}" ]]; then
+  echo "ℹ️ Did not find redhat-appstudio/build-definitions clone in ${BUILD_DEFINITIONS_DIR}, using upstream version"
+  BUILD_DEFINITIONS_DIR="$(mktemp -d build-definitions.XXXXXXXXXX)"
+  trap "rm -rf ${BUILD_DEFINITIONS_DIR}" EXIT
+  (cd "${BUILD_DEFINITIONS_DIR}" && curl -sLO https://github.com/redhat-appstudio/build-definitions/archive/refs/heads/main.zip && unzip main.zip && mv build-definitions-main/* . && rm main.zip)
+fi
+
+{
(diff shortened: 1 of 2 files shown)


hack/chains/end-to-end-demo.sh
```diff
@@ -0,0 +1,161 @@
...
+  exit 1
+fi
+
+set -euo pipefail
+
+(
+  while :
+  do
+    for i in {0..11}; do
+      printf "\r\xf0\x9f\x95\x$(printf %x $((144 + i))) Folding space..."
```
simonbaird, line 23: hmmm

simonbaird:
I wanna try it.

simonbaird:
Looks amazing.

simonbaird:
Notes and comments in no particular order:
* Clocky sometimes doesn't clean up its whitespace and I see a big indent. Maybe depends on when it is killed.  e.g.
```
$ ./end-to-end-demo.sh                                                                                                                      
                                                            📜 The outline of the demo                                                                                                                             
  1. We'll start a build pipeline using ...
```
* Maybe `sleep 1` is okay for clocky, for the times when you want to set -x and debug.
* `skopeo delete` requires admin permission in quay.io (afaict). So we should mention that in the comments. Or.. perhaps that's a big enough annoyance to reconsider that check. Our target audience would probably be comfortable figuring that out for themselves. Maybe just `echo "This demo assumes that $repo exists, is configured to be publicly accessible, and your current docker credentials have read/write permission there!"` is good enough. You could even pause while they read and ack that message. 
* The "argocd is synced" check is maybe a bit strict. I needed to comment it out since I had some unsynced stuff due to github org credentials or whatever. I would suggest making that a warning not a failure. Could also be a pause to acknowledge understanding, e.g. "Looks like your cluster isn't fully synced. Continue anyway?"

zregvart:
> * The "argocd is synced" check is maybe a bit strict. I needed to comment it out since I had some unsynced stuff due to github org credentials or whatever. I would suggest making that a warning not a failure. Could also be a pause to acknowledge understanding, e.g. "Looks like your cluster isn't fully synced. Continue anyway?"

Did you notice it checks that the `build` component in Argo is synced, not the whole cluster. I wanted that because we need the chains bit to work and it's a subtle and easy to gloss over failure if it's not deployed.

simonbaird:
http://pastebin.test.redhat.com/1047960 FYI , works for me.

simonbaird:
> > * The "argocd is synced" check is maybe a bit strict. I needed to comment it out since I had some unsynced stuff due to github org credentials or whatever. I would suggest making that a warning not a failure. Could also be a pause to acknowledge understanding, e.g. "Looks like your cluster isn't fully synced. Continue anyway?"
> 
> Did you notice it checks that the `build` component in Argo is synced, not the whole cluster. I wanted that because we need the chains bit to work and it's a subtle and easy to gloss over failure if it's not deployed.

I didn't actually. Maybe I was just too impatient. In that case, disregard my comment.

hack/chains/end-to-end-demo.sh
```diff
@@ -0,0 +1,138 @@
...
+#
+# A demo script that runs a build pipeline including HACBS specific tests and a sample release pipeline with HACBS Enterprise Contract test
+
+APP_GIT_REPOSITORY=https://github.com/jduimovich/single-nodejs-app
+APP_BUILDER=nodejs-builder
+APP_IMAGE_NAME="${APP_GIT_REPOSITORY/*\/}"
+APP_IMAGE_REF="quay.io/${MY_QUAY_USER}/${APP_IMAGE_NAME}"
+
+# Check for $MY_QUAY_USER
+if [[ -z ${MY_QUAY_USER+x} ]]; then
```
simonbaird, line 11: I don't understand the +x here.

hack/chains/end-to-end-demo.sh
```diff
@@ -0,0 +1,138 @@
...
+if [[ "${DOCKER_LOGIN}" != *"Login Succeeded"* ]]; then
+  echo "ℹ️  Please enter the quay.io credentials for ${MY_QUAY_USER}, if you're using a robot account
+and token make sure that it's allowed to push to ${APP_IMAGE_NAME} repository.
+"
+  docker login quay.io --username "${MY_QUAY_USER}"
+fi
+
+# Making the quay.io repository public, this is a requirement right now for the Test stream tasks to succeed
+if [[ "$(curl -s "https://quay.io/api/v1/repository?namespace=${MY_QUAY_USER}&public=true" | jq ".repositories[] | select(.name==\"${APP_IMAGE_NAME}\") | has(\"name\")")" != 'true' ]]; then
+  echo "🛑 Make sure that the quay.io repository '${APP_IMAGE_NAME}' exist and it's visibility
```
simonbaird, line 31: typo: "exists"
typo: "its"

hack/chains/end-to-end-demo.sh
```diff
@@ -0,0 +1,138 @@
...
+
+# We're going to assume running with kubeadmin
+if [[ "$(kubectl auth can-i '*' '*' < /dev/null 2>&1)" != 'yes' ]]; then
+  echo "🛑 Make sure that you're logged in as kubeadmin, we make assumptions here"
+  exit 1
+fi
+
+# Checking the status of the build component as a way to make sure that the cluster is okay to run the demo
+if [[ "$(kubectl get -n openshift-gitops applications.argoproj.io build -o jsonpath='{.status.sync.status}')" != 'Synced' ]]; then
+  echo "🛑 Make sure that the cluster is setup by running hack/bootstrap-cluster.sh first see
```
simonbaird, line 50: Maybe a more literal message here is useful, e.g. "The build application is not synced. Make sure ..."

hack/chains/end-to-end-demo.sh
```diff
@@ -0,0 +1,138 @@
...
+  echo "📂 Creating a new OpenShift project 'demo'
+"
+  oc new-project demo
+}
+echo "Here is the OpenShift console with the project:
+
+👉 $(oc whoami --show-console)/k8s/cluster/projects/demo
+"
+
+echo "♾️  Setting up pipelines
```
simonbaird, line 88: Is the two spaces vs one space intentional here, (and in a few other places)?

hack/chains/end-to-end-demo.sh
```diff
@@ -0,0 +1,138 @@
...
+kubectl create secret docker-registry release-demo --from-file=.dockerconfigjson="${HOME}/.docker/config.json" --dry-run=client -o yaml | kubectl apply -f -
+oc secrets link pipeline release-demo --for=pull,mount
+
+# Enterprise Contract policy configuration
+kubectl create configmap ec-policy --from-file=policy.json=<(echo '{"non_blocking_checks":["not_useful","test:conftest-clair"]}'|jq .) 2>/dev/null || echo "ℹ️  Using existing Enterprise Contract policy configuration in ec-policy ConfigMap"
+
+"${HACK_CHAINS_DIR}/copy-public-sig-key.sh"
+TASK_BUNDLE=quay.io/redhat-appstudio/appstudio-tasks:$(git ls-remote --heads https://github.com/redhat-appstudio/build-definitions.git refs/heads/main|cut -f 1)-2
+export TASK_BUNDLE
+# For simplicity tag with `production`
```
simonbaird, line 137: Comment belongs above line 121 I think.

simonbaird:
Looks good. Small nitpicks. I'll try running it again.

hack/chains/end-to-end-demo.sh
zregvart: I didn't either, see https://stackoverflow.com/a/13864829/225448

hack/chains/end-to-end-demo.sh
```diff
@@ -0,0 +1,138 @@
...
+}
+echo "Here is the OpenShift console with the project:
+
+👉 $(oc whoami --show-console)/k8s/cluster/projects/demo
+"
+
+echo "♾️  Setting up pipelines
+"
+kubectl apply -k "${BUILD_DEFINITIONS_DIR}/hack/test-build"
+kubectl apply -k "${BUILD_DEFINITIONS_DIR}/pipelines/hacbs"
```
simonbaird, line 91: I thought all this sort of stuff would come from bundles, but maybe I'm confused.

hack/chains/end-to-end-demo.sh
zregvart: Could be my shell/terminal/font but unless I do that there is no space between the emoji and text. I can see two spaces in http://pastebin.test.redhat.com/1047960, so I'll revert to using one

hack/chains/end-to-end-demo.sh
zregvart: Yup stuff got moved around and I forgot to move the comment with it

hack/chains/end-to-end-demo.sh
zregvart: I think I concluded that it was necessary, let me check without

zregvart: Yup, `build-definitions/hack/test-build.sh` doesn't work without that:

```
Error: Pipeline name nodejs-builder does not exist in namespace demo
```

simonbaird: I wonder if that is a deliberate choice, or they just didn't get around to setting them up with tkn bundle yet...

simonbaird:
Still working for me.

PR #313, opened 2022-04-27
Add hack/build/cleanup-images-and-pipelineruns.sh for dev purpose.

AndrienkoAleksandr:
Signed-off-by: Oleksandr Andriienko <oandriie@redhat.com>

### DIFF
--- a/hack/build/cleanup-images-and-pipelineruns.sh
+++ b/hack/build/cleanup-images-and-pipelineruns.sh
@@ -1,0 +1,40 @@
+#!/bin/bash
+# PipelineRun builds images for application.
+# This script removes images from Openshift internal image registry
+# to have ability re-triger PipelineRuns without skip build image tasks.  
+
+usage () {
+	echo "Usage:   $0 [--only-images ONLY_IMAGES --n NAMESPACE]"
+	echo "Example: ./cleanup-images-and-pipelineruns.sh --only-images true --n test"
+}
+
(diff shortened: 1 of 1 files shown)


hack/build/cleanup-images-and-pipelineruns.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
+# PipelineRun builds images for application.
+# This script removes images from Openshift internal image registry
+# to have ability re-triger PipelineRuns without skip build image tasks.  
+
+usage () {
+	echo "Usage:   $0 [--only-images ONLY_IMAGES --n NAMESPACE]"
+	echo "Example: ./cleanup-images-and-pipelineruns.sh --only-images true --n test"
```
Michkov, line 8: do we need script for running `oc delete is,pipelinerun --all` ?

hack/build/cleanup-images-and-pipelineruns.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
+# PipelineRun builds images for application.
+# This script removes images from Openshift internal image registry
+# to have ability re-triger PipelineRuns without skip build image tasks.
```
Michkov, line 4: Pipelines are having `rebuild` parameter when it's set to 'true' then the build is run always -> https://github.com/redhat-appstudio/build-definitions/blob/main/pipelines/base/template-build/template-build.yaml#L30

PR #317, opened 2022-04-28
Few enhancements to release-pipeline-with-ec-demo

lcarva:


simonbaird (approved):

hack/chains/release-pipeline-with-ec-demo.sh
```diff
@@ -114,6 +119,21 @@ spec:
...
+---
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: ec-policy
+data:
+  # HINT: Add "test" to the list to ignore tests
+  policy.json: |
+      {"non_blocking_checks":["not_useful"]}
+' | oc apply -f - > /dev/null
```
lcarva, line 134: I'm going to change this so it only creates the configmap if it's missing to avoid tramping over previously configured policies.

lcarva: Done

simonbaird (approved):

PR #344, opened 2022-05-09
demo: remove unnecessary chains secret

lcarva:
To push signatures and attestations to an OCI registry, Chains will use
whichever secret was used to push the corresponding image. It is not
necessary to create the same secret in the tekton-chains namespace.
Further proof of this is the fact that the demo script was linking the
secret to the pipeline ServiceAccount and the tekton-chains-controller
uses the tekton-chains-controller ServiceAccount.


zregvart:
I think the comment is off, but this is required by the `hack/chains/release-pipeline-with-ec-demo.sh` script:

https://github.com/redhat-appstudio/infra-deployments/blob/a1f782695bfe27a59c6a48adf81836aa2966338c/hack/chains/release-pipeline-with-ec-demo.sh#L8-L11

zregvart:
Though, incidentally, that `pipeline` ServiceAccount appears with a delay so first time the `end-to-end-demo.sh` is invoked that bit might fail. We might need to wait for it to be created by the Tekton operator before we proceed.

PR #347, opened 2022-05-10
Wait for the pipeline service account in e2e demo

zregvart:
If the pipeline service account has not yet been created the `hach/chains/end-to-end-demo.sh` script will fail if it tries to link the `release-demo` Docker pull secret to it. This adds loop to wait until it has been created by the Tekton operator before proceeding with the demo.

### DIFF
--- a/hack/chains/end-to-end-demo.sh
+++ b/hack/chains/end-to-end-demo.sh
@@ -87,6 +87,10 @@ echo "Here is the OpenShift console with the project:
 👉 $(oc whoami --show-console)/k8s/cluster/projects/demo
 "
 
+# wait for the pipeline service account to be created by the Tekton operator
+echo '⏳ Waiting for the pipeline service account to be created'
+while ! kubectl get serviceaccount pipeline > /dev/null 2>&1; do sleep 1; done
+
 echo "♾️ Setting up pipelines
 "
 kubectl apply -k "${BUILD_DEFINITIONS_DIR}/hack/test-build"


hack/chains/end-to-end-demo.sh
```diff
@@ -87,6 +87,9 @@ echo "Here is the OpenShift console with the project:
 👉 $(oc whoami --show-console)/k8s/cluster/projects/demo
 "
 
+# wait for the pipeline service account to be created by the Tekton operator
+while ! kubectl get serviceaccount pipeline > /dev/null 2>&1; do sleep 1; done
```
lcarva, line 91: We need to link the secret with the pipeline service account *before* the build pipeline starts! https://github.com/redhat-appstudio/build-definitions/pull/91 seems to be working for me. Can you give it a shot?

zregvart: There is a timing issue (perhaps I'm the only one seeing this) between the Tekton operator creating the `pipeline` service account after a project/namespace has been created and the script trying to link the `release-demo` pull secret to the `pipeline` service account. For me this fails if the build pipeline fails quickly (for me, because of https://github.com/redhat-appstudio/build-definitions/pull/90)

```
ℹ️ Please wait...

...snip...

🏃 Running a build pipeline

💲 build-definitions/hack/test-build.sh https://github.com/jduimovich/single-nodejs-app nodejs-builder

Building quay.io/zregvart_redhat/single-nodejs-app:2e92018
PipelineRun started: nodejs-builder-run-nrkmd

In order to track the PipelineRun progress run:
tkn pipelinerun logs nodejs-builder-run-nrkmd -f -n demo

ℹ️ Build pipeline finished

Make note that some of the steps could have been skipped over because of the
pipeline caching.

Here is the pipeline visualisation:

👉 https://console-openshift-console.apps-crc.testing/k8s/ns/demo/tekton.dev~v1beta1~PipelineRun/nodejs-builder-run-nrkmd

🏃 Running a sample release pipeline (to demo Enterprise Contract task)

💲 infra-deployments/hack/chains/release-pipeline-with-ec-demo.sh quay.io/zregvart_redhat/single-nodejs-app:2e92018 quay.io/zregvart_redhat/single-nodejs-app:production

secret/release-demo created
Error from server (NotFound): serviceaccounts "pipeline" not found
```

i.e. it fails on `oc secret link` because of the missing `pipeline` service account here:

https://github.com/redhat-appstudio/infra-deployments/blob/a6f7b9b2d4415090f858026e544621faa7c9b168/hack/chains/end-to-end-demo.sh#L132

I presume the same issue (https://github.com/redhat-appstudio/build-definitions/pull/90) would manifest with https://github.com/redhat-appstudio/build-definitions/pull/91 as well.

lcarva: Aha! So this is yet a different issue. I probably didn't see this problem because I have a cluster with 6 nodes (1/2 worker, 1/2 master) which is significantly faster than CRC.

Can we just add a short log message before entering the wait loop so we know what's happening in case something is busted and never creates the service account?

lcarva (approved):

PR #361, opened 2022-05-16
Add Michal Vala admin rights on his namespace

skabashnyuk:
Add Michal Vala admin rights on his namespace to be able to properly prepare for the SPI demo on staging.


sparkoo (approved):
looks right. thanks

PR #396, opened 2022-05-25
view RoleBinding for enterprise-contract-service

zregvart:
Adds view RoleBinding for the enterprise-contract-service namespace and initially adds the Enterprise Contract group subject.


zregvart:
Duplicate of #395

PR #399, opened 2022-05-26
Show what happened to build application when failed

Michkov:


Michkov:
@psturc @flacatus would be possible to gather `applications.argoproj.io` objects to prow logs? That would be probably more elegant than this.

psturc:
I think it's not possible to configure prow to automatically collect specific resources, but we could update the script to store the logs/objects in prow artifacts dir, which you can then inspect after a job is finished.
example:
`kubectl get applications.argoproj.io build -n ${APPLICATION_NAMESPACE} -o yaml > "${ARTIFACTS_DIR}/application-build.yaml"`

psturc:
We could also have [something like osde2e is using](https://github.com/openshift/osde2e/blob/aebbacb34028ffbcd1a299f5937458a3b6d4f4c1/pkg/e2e/e2e.go#L548). I saw this mentioned in one of our internal docs. There should be a jira task created for it.

PR #450, opened 2022-06-10
Setup HACBS test team's RBACs

dirgim:
Add the RBACs for the Test team and the integration service


Michkov:
Merged since the CI does not touch these file

PR #455, opened 2022-06-15
Remove yq installation from container tests

flacatus:
We don't need e2e container at this level and removing installation of yq. Yq installation will be added to e2e scripts to install it with GITHUB TOKEN


rhopp:
Given we are blocked on the api rate limits and there is already fix in https://github.com/openshift/release/pull/29499 I'm force-pushing this PR to unblock us.

PR #463, opened 2022-06-20
Set tekton config instead flags configmap

Michkov:
- disable pipelinesAsCode delivered with Openshift Pipelines
- enable affinity-assistant to run multiple tasks in parallel
- set prunning in the config, not by post sync job


zregvart (approved):
I've previewed this locally and forced a ArgoCD to synchronize with prune and now I can see that `openshift-pipelines/feature-flags` contains `enable-tekton-oci-bundles: "true"` and it no longer flip-flops between the default of `false` and `true`. Another point that confirms that this solves the flip-flop issue of the `feature-flags` ConfigMap is that it's `resourceVersion` is no longer increasing.

PR #471, opened 2022-06-23
fix: Remove old task param from ec demo

lcarva:


lcarva:
@redhat-appstudio/enterprise-contract ptal

robnester-rh (approved):
LGTM

PR #482, opened 2022-07-05
Gives skabashn admin role to see what's going on in Soud's namespace

skabashnyuk:
-  Needed to unblock Soud until we are working on the fix on https://issues.redhat.com/browse/SVPI-161


rottencandy (approved):
Confirming that I want @skabashnyuk to have admin access to my namespace.

PR #484, opened 2022-07-05
search for tasks in preview script

gabemontero:
/hold 

just for debug / research


gabemontero:
hmm ... tekton is not yet installed at the end of preview.sh:

```
GGM print all tasks at end of preview
error: the server doesn't have a resource type "tasks"

```

PR #486, opened 2022-07-08
add tenant "RAMESSESII2"

ramessesii2:
For implementation of Network Isolation, I need to be able to find out the Network Plugin used by AppStudio Staging cluster.


ramessesii2:
@sbose78 please take a look.

sbose78:
Do you really need something this broad?

sbose78 (approved):

components/authentication/tenants/sabhardw.yaml
```diff
@@ -0,0 +1,24 @@
...
+    resources:
+      - networks
+---
+kind: ClusterRoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: cluster-network-read-sabhardw
+subjects:
+  - kind: User
+    name: RAMESSESII2
```
sbose78, line 20: I think this should be in lowercase?

ramessesii2: Sorry, again this is just because of misunderstanding of me being a member of PipelineService cluster where I'm "RAMESSESII2"--> changed my Github handle to all lower just recently.

ramessesii2:
Hi @sbose78, if it looks good to you, please merge it or I can quickly add more changes if you need them. 😸

ramessesii2:
> Have you logged into appstudio in the past, if yes, you should add yourself to the appstudio namespace as part of this.

Hey @sbose78, no I haven't logged into appstudio before and would need access to at least one of the namespace where I can do the query for which I an trying to create an admin role for myself in the "sabhardw" ns. Thanks

Michkov (approved):

PR #497, opened 2022-07-15
setting up managed workspace for shebert for demo

scoheb:


components/authentication/tenants/shebert.yaml
```diff
@@ -10,3 +10,80 @@ roleRef:
...
+    name: scoheb
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: release-service-releasestrategy-editor-role
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: gitopsdeployment-manager
```
sbose78, line 43: ran a search and found we already have something similar - do you want to reconcile these into one?

https://github.com/redhat-appstudio/infra-deployments/blob/9da0f0b67256c53171994a66807cbcce7769119b/components/authentication/tenants/shbose.yaml#L43

scoheb: sure I can reconcile. Any suggestion as to where this can be extracted to?

sbose78: Tricky :) How about updating the one that already exists ?

sbose78 (approved):

PR #509, opened 2022-07-20
fix: ci: create shared secret

psturc:
### What

Some tests no longer use the hardcoded `appstudio-e2e-test` namespace anymore (for creating apps, components, pipelines etc.).

This has caused that these tests started failing, because those namespaces don't contain the `redhat-appstudio-registry-pull-secret` secret used to authenticate when pushing the built image to quay.io.

To fix this, we need to create a shared secret that will be consumed by the pipelinerun created in any namespace on the cluster.

(The creation of the shared secret was added quite some time ago to [e2e-tests repository's CI scripts](https://github.com/redhat-appstudio/e2e-tests/blob/a0c27237bfee12e208b46c7ff88d3f5b7a3efbe5/scripts/install-appstudio-e2e-mode.sh#L64), but I haven't realized it needs to be updated here as well)


johnmcollier (approved):
Approving to unblock HAS from getting updated in https://github.com/redhat-appstudio/infra-deployments/pull/491

/lgtm
/approve

PR #518, opened 2022-07-25
Preview optimalization - sync all applications and wait for them

Michkov:
Make preview deployment stable

- Require all applications to be Healthy and Synced
  - removed quality-dashboard since it requires manual changes
  - integrated SPI initialization
  - add wait for the components
- Set chains/rekor directly in the preview commit, avoid post deploy configuration
- Update CI to use new preview.sh -> no need to wait for components in the script, handled by preview.sh
- Blame components which are not synced


hack/preview.sh
```diff
@@ -83,6 +73,10 @@ if [ -n "$DOCKER_IO_AUTH" ]; then
     rm $AUTH
 fi
 
+rekor_server="rekor.$domain"
+sed -i "s/rekor-server.enterprise-contract-service.svc/$rekor_server/" $ROOT/argo-cd-apps/base/enterprise-contract.yaml
+yq -i e ".data |= .\"transparency.url\"=\"https://$rekor_server\"" $ROOT/components/build/tekton-chains/chains-config.yaml
```
zregvart, line 78: IIRC we don't set this currently, so we're using the public Rekor instance at rekor.sigstore.dev by default and we need to switch over manually by running `hack/chains/config.sh local-rekor` with Argo synchronization off for the Build application. We discussed removing Rekor altogether, so I'm not certain we need this.

cc @redhat-appstudio/enterprise-contract

Michkov: ah, you are setting it -> https://github.com/redhat-appstudio/infra-deployments/blob/8b04b44311fa3594af5c18f8ab3362853f022bdf/hack/preview.sh#L140 but the change is reverted by argocd...

So basically these three lines can be removed. Waiting on reaction of rest of your team.

lcarva: If you run `./hack/bootstrap.sh preview` today, it does set the transparency.url correctly and the in-cluster instance of Rekor is used. So it looks like the change is not reverted by argocd. I _think_ argocd ignores fields in a configmap that are not defined in git. Since [chains-config.yaml](https://github.com/redhat-appstudio/infra-deployments/blob/main/components/build/tekton-chains/chains-config.yaml) doesn't set the transparency.url, argocd ignores its value altogether.

The staging instance of appstudio does not have the transparency.url set because it was not bootstrapped with the preview command.

hack/preview.sh
```diff
@@ -32,24 +32,13 @@ if git rev-parse --verify $PREVIEW_BRANCH; then
...
-# Set the domain for our rekor deployment.
-# We add the modified rekor.yaml file and this will get pushed to our preview branch.
-# This shouldn't ever be updated in $MY_GIT_BRANCH.
-# If you know a better way to make this magic happen, contact rnester@redhat.com
-domain=$( kubectl get ingresses.config.openshift.io cluster --template={{.spec.domain}} )
-echo
-echo "Setting rekor server domain to: $domain"
-echo
-sed -i "s/rekor-server.enterprise-contract-service.svc/rekor.$domain/" $ROOT/argo-cd-apps/base/enterprise-contract.yaml
-git add $ROOT/argo-cd-apps/base/enterprise-contract.yaml && git commit -m "Set domain for rekor"
```
lcarva, line 44: My understanding is that if we don't create a commit with this dynamic value then argocd will revert the change.

hack/preview.sh
robnester-rh: When that was written in `preview.sh`, the only way it seemed to get the changes of the rekor url to persist were to make the changes in the preview branch and then commit them. This would then allow argo-cd to pick them up and implement them. Upon going back to upstream, the commits in the preview branch would be disregarded and the `transparency.url` field would be removed from the chains-config` secret in the cluster.

hack/preview.sh
Michkov: The commit is later in this file, so we are creating one commit with all changes by preview.

hack/preview.sh
Michkov: Ok, so what is the result, should we point to in-cluster rekor for development and testing environments (using preview.sh)? I suppose that this is not blocking this PR since the behavior is not changed.

rhopp:
I've gone throught the PR and I don't see anything suspicious... But I still leave the final `lgtm` to people who understand rekor and the stuff around better than me ;-)

hack/preview.sh
zregvart: Let's do what you have here in the PR. We want to use internal Rekor when running in preview mode.

psturc:
lgtm

I think that as a follow-up it'd make sense to update openshift-ci related scripts in application-service, build-service etc. to remove the code that runs the gitops deployment validation and SPI setup. Should we have a jira ticket for that? (if I count it correctly it will be 4 additional pull requests :)

hack/preview.sh
```diff
@@ -125,30 +119,52 @@ git checkout $MY_GIT_BRANCH
...
+while [ "$(oc get applications.argoproj.io all-components-staging -n openshift-gitops -o jsonpath='{.status.health.status} {.status.sync.status}')" != "Healthy Synced" ]; do
+  sleep 5
+done
+
+APPS=$(kubectl get apps -n openshift-gitops -o name)
+
+if echo $APPS | grep -q spi; then
+  if [ "`oc get applications.argoproj.io spi -n openshift-gitops -o jsonpath='{.status.health.status} {.status.sync.status}'`" != "Healthy Synced" ]; then
+    echo Initializing SPI
+    curl https://raw.githubusercontent.com/redhat-appstudio/e2e-tests/main/scripts/spi-e2e-setup.sh | bash -s
```
psturc, line 131: @Michkov I just realized that in case we'd need to update this script in e2e-tests repo, we should be able to test the change somehow.

Openshift CI exports an env var called `$PULL_PULL_SHA` which contains a SHA of a latest PR commit, which we could pass to `preview.sh` script (and probably rename the env var to something more appropriate, like `$E2E_TESTS_COMMIT_SHA`). Thanks to that we could test the change introduced in the e2e-tests PR.

Would it make sense to update this line accordingly?
```bash
curl https://raw.githubusercontent.com/redhat-appstudio/e2e-tests/${E2E_TESTS_COMMIT_SHA:-main}/scripts/spi-e2e-setup.sh | bash -s
```

Michkov: updated

PR #533, opened 2022-07-27
Allow John to prune old HAS OLM resources

johnmcollier:
There's some old OLM resources leftover from when we briefly deployed HAS as an OLM bundle. They're no longer needed, and seem to be orphaned from ArgoCD, so I'm adding a role to my stage user to allow myself to clean them up.


elsony (approved):
/lgtm

PR #534, opened 2022-07-27
Allow Chains to use the SharedSecret resource

lcarva:


components/build/tekton-chains/tekton-chains-controller-shared-secret.yaml
```diff
@@ -0,0 +1,28 @@
+---
```
lcarva, line 1: This should only be applied if the SharedSecret named `redhat-appstudio-user-workload` actually exists. My understanding is that this only happens in the staging environment.

This PR is in draft because I need to either:
1. Figure out how to apply an overlay to the tekton-chains Deployment only in the staging environments; or
2. Always create the `redhat-appstudio-user-workload` SharedSecret for preview and development modes.

lcarva: Ok, it looks like `SharedSecret` is already created in components/build/shared-resources/redhat-appstudio-user-workload-sharedsecret.yaml

So let's take approach 2 and this PR is good as is. Moving it out of draft.

gabemontero (commented):
just a minor comment and question (though if my assumptions are incorrect perhaps a req for change)

README.md
```diff
@@ -88,11 +88,6 @@ SPI Vault instance has to be manually initialized. There is a script to help wit
 2) Clone SPI operator repo `git clone https://github.com/redhat-appstudio/service-provider-integration-operator && cd service-provider-integration-operator`
 3) run `vault-init.sh` script from repo root directory `./hack/vault-init.sh`
 
-
-#### Post-bootstrap Tekton Chains Configuration
```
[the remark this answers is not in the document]
lcarva: Yes, good idea. Just added one.

components/build/tekton-chains/tekton-chains-controller-shared-secret.yaml
```diff
@@ -0,0 +1,28 @@
...
+        - mountPath: /secret/default-push-secret/config.json
+          subPath: .dockerconfigjson
+          name: default-push-secret
+      volumes:
+      - name: default-push-secret
+        csi:
+          driver: csi.sharedresource.openshift.io
+          readOnly: true
+          volumeAttributes:
+            sharedSecret: redhat-appstudio-user-workload
```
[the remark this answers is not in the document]
lcarva: It comes from the [release.yml](https://github.com/lcarva/infra-deployments-1/blob/256695d472ff2ec6b87bb0e3b6f9c922c5c1a883/components/build/tekton-chains/kustomization.yaml#L10)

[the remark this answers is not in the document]
gabemontero: ahhh ... the beautiful misdirection / mystification of `kustomize` :-) .... thanks for the answer

gabemontero:
/lgtm

let's let @Michkov chime in when he is back on the clock tomorrow and if he is good he can apply the approve label

lcarva:
Flagging this as WIP (I can't figure how to move it back to a draft).

The e2e tests caught a valid issue. If the underlying Secret does not exist, the path is a directory instead of a file. This causes the `k8schain` library to crash.

components/build/tekton-chains/tekton-chains-controller-shared-secret.yaml
lcarva: It seems like if the `SharedSecret` is used when the underlying `Secret` does not exist, the mounted path is always a directory. This causes the `k8schain` library used by Chains to error - meaning no image signature nor attestation. 

I submitted a [fix](https://github.com/google/go-containerregistry/pull/1420 ) for the library, but I have no idea how long that will take to get merged and released. I can create a fork but would really prefer not to do that...

@Michkov, can we always create the `build-templates/redhat-appstudio-user-workload` secret? I think an "empty" one should be enough.

lcarva: I think I have a workaround. Waiting for e2e test results.

components/build/tekton-chains/tekton-chains-controller-shared-secret.yaml
```diff
@@ -0,0 +1,58 @@
...
+        - name: default-push-secret
+          mountPath: /secret/default-push-secret
+        - name: docker-config
+          mountPath: /secret/docker-config
+        command:
+        - ln
+        - -s
+        - -f
+        - /secret/default-push-secret/.dockerconfigjson
+        - /secret/docker-config/config.json
```
gabemontero, line 39: so do you know what `ls -la /secret/docker-config` looks like in the underlying Pod after the secret referenced by the `redhat-appstudio-user-workload` shared secret is created and the CSI driver attempts to store its contents ?

fwiw this sort of approach has never been tried before with shared resources

if I had cycles I'd try this out on the side but I don't know when in the short term I'll be able to

README.md
```diff
@@ -88,10 +88,19 @@ SPI Vault instance has to be manually initialized. There is a script to help wit
 2) Clone SPI operator repo `git clone https://github.com/redhat-appstudio/service-provider-integration-operator && cd service-provider-integration-operator`
 3) run `vault-init.sh` script from repo root directory `./hack/vault-init.sh`
 
+### Optional: Use SharedSecret with Tekton Chains
 
-#### Post-bootstrap Tekton Chains Configuration
+During the build pipeline, it is possible to use the `redhat-appstudio-user-workload`
```
lcarva, line 93: This is no longer valid. Live/hot loading works now with the initContainer approach.

components/build/tekton-chains/tekton-chains-controller-shared-secret.yaml
gabemontero: I believe @lcarva told me in the hacbs weekly today that it worked (though I'm still mildly curious what the ls -la look like in the running pod if you `oc rsh` into it) ... which I think is also what he is conveying with https://github.com/redhat-appstudio/infra-deployments/pull/534#discussion_r932301954

lcarva: It's an empty directory if the `Secret` does not exist. As soon as it is created, it gets populated properly.

The initContainer workaround was suggested in the kubernetes issue linked in the comment.

README.md
lcarva: Updated.

simonbaird (approved):

PR #541, opened 2022-07-28
Disable chains transparency logging by default

simonbaird:
We like rekor and will probably use it in the future, but for the
HACBS MVP/POC it will be disabled. The reasoning is that we can save
some time and effort by avoiding trying to figure out the best way
to provision rekor instances for each tenant.
    
Note: We may also want to remove the rekor instance from
`argo-cd-apps/base/enterprise-contract.yaml` since it won't be used,
and review the rekor related parts of `hack/preview.sh`. This patch
ignores all of that and just does a minimal change to the default
chains configuration.

JIRA: [HACBS-888](https://issues.redhat.com//browse/HACBS-888)
See also: [HACBS-766](https://issues.redhat.com//browse/HACBS-766)


lcarva:
Change looks good. But I suspect e2e tests will fail. 

[This](https://github.com/redhat-appstudio/e2e-tests/blob/26d1ba9b2582541c7986001a96a439bb1c3b791d/pkg/utils/tekton/controller.go#L494-L508) might have to return an empty string if `transparency.enabled` is unset or set to `false`.

zregvart:
We also need to remove this from the `preview.sh` (e2e tests run with `preview.sh`):
https://github.com/redhat-appstudio/infra-deployments/blob/ad75019c63ba1ac33ff1c4b063fcf6ff2d298521/hack/preview.sh#L76-L78

lcarva:
> We also need to remove this from the `preview.sh` (e2e tests run with `preview.sh`):
> 
> https://github.com/redhat-appstudio/infra-deployments/blob/ad75019c63ba1ac33ff1c4b063fcf6ff2d298521/hack/preview.sh#L76-L78

@zregvart, I think it's ok to leave that alone. By having the `transparency.url` set, it makes it easier for us to enable the rekor integration ad-hoc.

zregvart:
Perhaps we also need to make sure e2e tests don't fail without Rekor. Right now it seems we're getting 400 from Rekor, perhaps a version mismatch between CLI and the Rekor version?

Michkov:

PR #553, opened 2022-08-01
update SPI

sparkoo:
- updates SPI to 0.7.0
 - own kustomize for Vault, reflecting changes in spi vault deployment
 - new SPI configuration of Vault


components/spi/kustomization.yaml
```diff
@@ -1,9 +1,11 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
```
Michkov, line 2: this is already in the file

hack/preview.sh
```diff
@@ -39,6 +39,8 @@ $ROOT/hack/util-set-development-repos.sh $MY_GIT_REPO_URL development $PREVIEW_B
 # set the API server which SPI uses to authenticate users to empty string (by default) so that multi-cluster
 # setup is not needed
 yq -i e ".0.value.0.value=\"$SPI_API_SERVER\"" $ROOT/components/spi/oauth-service-deployment-patch.json
```
Michkov, line 41:
```suggestion
yq -i e ".0.value.value=\"$SPI_API_SERVER\"" $ROOT/components/spi/oauth-service-deployment-patch.json
```
You changed format of oauth-service-deployment-patch.json

Michkov (changes requested):
`StatefulSet.apps "spi-vault" is invalid: spec: Forbidden: updates to statefulset spec for fields other than 'replicas', 'template', 'updateStrategy', 'persistentVolumeClaimRetentionPolicy' and 'minReadySeconds' are forbidden` so I had to manually remove the StatefulSet.

Now spi-controller-manager  and spi-oauth-service failing on:
```
{"level":"error","ts":"2022-08-04T08:36:59.002Z","logger":"setup","caller":"runtime/proc.go:255","msg":"failed to initialize the token storage","error":"error while authenticating: unable to log in to auth method: unable to log in with Kubernetes auth: Put \"https://spi-vault-spi-system.apps.hacbs-build-01.hosted.upshift.rdu2.redhat.com/v1/auth/kubernetes/login\": x509: certificate signed by unknown authority","stacktrace":"runtime.main\n\t/usr/local/go/src/runtime/proc.go:255"}
```

components/spi/kustomization.yaml
sparkoo: fixed

Michkov (approved):

PR #564, opened 2022-08-04
Enable the JVM cache in CI namespace

stuartwdouglas:
The CI for JVM build service is running in the jvm-build-service
namespace, we need to enable JVM support for this namespace to activate
the cache.


Michkov:
Tests passed only reporting failed.

PR #579, opened 2022-08-08
remove cache instance in jvm-build-service namespace

gabemontero:
As part of the next step following the merging of https://github.com/redhat-appstudio/jvm-build-service/pull/105 and in particular what is noted in https://github.com/redhat-appstudio/jvm-build-service/blob/650846d2011d745b82c72cd59cf1e2398daeddfa/deploy/cache/base/deployment.yaml#L1-L3 we are removing the reference to the yaml that defines the jvm-build-service cache in the `jvm-build-service` namespace 

@Michkov @stuartwdouglas @phillip-kruger @brunoapimentel @psturc FYI

### DIFF
--- a/components/build/jvm-build-service/kustomization.yaml
+++ b/components/build/jvm-build-service/kustomization.yaml
@@ -1,7 +1,6 @@
 resources:
 - allow-argocd-to-manage.yaml
 - https://github.com/redhat-appstudio/jvm-build-service/deploy/crds/base?ref=345ea0982fc5f57ef55c3e7e689e460e70cd5313
-- https://github.com/redhat-appstudio/jvm-build-service/deploy/cache/base?ref=345ea0982fc5f57ef55c3e7e689e460e70cd5313
 - https://github.com/redhat-appstudio/jvm-build-service/deploy/operator/base?ref=345ea0982fc5f57ef55c3e7e689e460e70cd5313
 - .tekton/
 - namespace.yaml
@@ -12,12 +11,6 @@ resources:
 namespace: jvm-build-service
 
 images:
-- name: hacbs-jvm-cache
-  newName: quay.io/redhat-appstudio/hacbs-jvm-cache
-  newTag: 345ea0982fc5f57ef55c3e7e689e460e70cd5313
-- name: quay.io/QUAY_USERNAME/hacbs-jvm-cache
-  newName: quay.io/redhat-appstudio/hacbs-jvm-cache
-  newTag: 345ea0982fc5f57ef55c3e7e689e460e70cd5313
 - name: hacbs-jvm-operator
   newName: quay.io/redhat-appstudio/hacbs-jvm-controller
   newTag: 345ea0982fc5f57ef55c3e7e689e460e70cd5313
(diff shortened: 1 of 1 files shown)


gabemontero:
/hold

ideally we do not merge this until https://github.com/redhat-appstudio/e2e-tests/pull/168 has merged and we verify jvm-build-service is OK in the e2e run

gabemontero:
Not sure of all the particulars yet, but I'm pretty sure my change here is incomplete.  The app studio bootstrap failed, and I still see the jvm-build-workspace-artifact-cache pods getting created in the jvm-build-service namespace, with the same exception noted in this PRs description.

Investigating / trying to sort out.

gabemontero:
heads up @Michkov @stuartwdouglas @phillip-kruger @brunoapimentel @psturc I edited the description to remove the bit about the artifact cache in the jvm-build-service namespace (and only in that namespace) getting that amazon sdk stack trace that @Michkov and @psturc discovered and reported in https://coreos.slack.com/archives/C03KN2B47GU/p1659958982619609?thread_ts=1659958940.328259&cid=C03KN2B47GU

The artifact caches in the test namespaces do not exhibit that exception.  

And as noted in that slack thread, and in other active PRs, this stack trace appears to have no bearing on artifactbuilds/dependencybuilds.

The setup of the artfact cache in the jvm-build-service namespace is occurring because of https://github.com/redhat-appstudio/infra-deployments/pull/564 i.e. @stuartwdouglas made the explicit choice to add it.  It did not appear to be a short term addition based on the PR's description.  I'll try to confirm that with him when he is next available.

gabemontero:
the new jvm e2e test failed

now, with pipeline run pruning, it is hard to tell if that got in the way, or if this change caused the problem, or if the test is simply a victim of RPC delays

crafting a main branch PR to cross reference

gabemontero:
same error - though I suppose it is conceivable the error in this case is related to this change (i.e. it is perhaps premature somehow)

stuartwdouglas:
I added this because there are some Java builds on CI that use the HACBS build definitions, and thus require the use of the cache. The real issue here is that startup fails with rebuilds set to false, which is fixed in https://github.com/redhat-appstudio/jvm-build-service/pull/170

gabemontero:
the situation discussed in https://coreos.slack.com/archives/C03KN2B47GU/p1659994799379339 with the latest theory from @stuartwdouglas at https://coreos.slack.com/archives/C03KN2B47GU/p1660017969344259?thread_ts=1659994799.379339&cid=C03KN2B47GU is also a factor, perhaps even a bigger one I think, though certainly we need both items addressed

gabemontero:
OK all green e2e

@stuartwdouglas - just to make sure when you are back on the clock

1) are you saying in https://github.com/redhat-appstudio/infra-deployments/pull/579#issuecomment-1208969956 that the cache service this PR is no longer including in gitops mgmt needed for hacbs java tests previously defined in CI (and not the tests for the jvm-build-service we have had to add then remove multiple times)?  And hence this PR should be close and never done, deferred to another time after other changes have landed?

2) Or is the "this" you are referring to with "I added this" is in fact https://github.com/redhat-appstudio/infra-deployments/blob/9d81c822d6fefe327db208cf611fd7c5a6c41c28/components/build/jvm-build-service/system-config.yaml#L14-L21 which leads to the creation of the artifact cache in the jvm-build-servivce, and per the comment in https://github.com/redhat-appstudio/jvm-build-service/blob/650846d2011d745b82c72cd59cf1e2398daeddfa/deploy/cache/base/deployment.yaml#L1-L3 we should go ahead and make this change now?

thanks

gabemontero:
I did just realize, this PR might need https://github.com/redhat-appstudio/build-definitions/pull/179 to merge first 

that sound right @stuartwdouglas ?

stuartwdouglas:
We definitely should merge that other PR first. I have rebased it so it should be ready to go.

gabemontero:
https://github.com/redhat-appstudio/build-definitions/pull/179 has merged 

/hold cancel

/assign @Michkov 
/assign @stuartwdouglas

gabemontero:
green e2e's @Michkov @stuartwdouglas no need to rebase at the time of this comment's posting

gabemontero:
going to run the e2e again, just in case the bundle was not updated after https://github.com/redhat-appstudio/build-definitions/pull/179 merged but before this referenced it (the bundle updates looking at quay.io seems like it was minutes after 179 merged, but let's make sure)

/test appstudio-e2e-deployment

gabemontero:
ok still green

Michkov (approved):

PR #613, opened 2022-08-15
Feature toggle Rekor

zregvart:
This allows for customizing Tekton Chains transparency configuration and
correspondingly the Rekor instance. Three modes are now available:
 * off      - transparency off
 * local    - transparency on, with an in-cluster Rekor instance
 * sigstore - transparency on, with Sigstore public Rekor instance

These are set via ApplicationSet list generator and the custom "rekor"
parameter set there, see argo-cd-apps/base/enterprise-contract.yaml.

It is possible to change this parameter in a gitops manner and the
resulting state should reflect the value of the parameter, e.g. changing
from "local" to "off" should remove the "rekor" ArgoCD Application and
configure the "chains-config" ConfigMap accordingly.

The in-cluster deployed instance of Rekor no longer has an Ingress to
it's service, so it is no longer available outside of the cluster.

Ref. https://issues.redhat.com/browse/HACBS-888


components/enterprise-contract/rekor-local/kustomization.yaml
```diff
@@ -0,0 +1,11 @@
...
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+resources:
+  - ../base/
+  - rekor.yaml
+
+patches:
+  - tekton-chains-config.yaml
+#  - remove-rekor-ingress.yaml
```
simonbaird, line 11: Is this line intentionally commented out?

zregvart: Doh, forgot to uncomment that one

zregvart: Did it unnecessarily complicated before, now it should be okay

robnester-rh (approved):
LGTM

argo-cd-apps/overlays/development/repo-overlay.yaml
```diff
@@ -98,12 +98,13 @@ spec:
     targetRevision: main
 ---
 apiVersion: argoproj.io/v1alpha1
-kind: Application
+kind: ApplicationSet
 metadata:
   name: enterprise-contract
-  finalizers:
-    - resources-finalizer.argocd.argoproj.io
+  namespace: openshift-gitops
```
lcarva, line 104: Why is this needed? (It is not needed for Applications)

components/enterprise-contract/base/tekton-chains/kustomization.yaml
```diff
@@ -25,5 +27,6 @@ patchesStrategicMerge:
 # Mount the shared secret
 - tekton-chains-controller-shared-secret.yaml
 
-apiVersion: kustomize.config.k8s.io/v1beta1
-kind: Kustomization
+patches:
+# Remove the chains-config
+- remove-chains-config.yaml
```
lcarva, line 32: Why do we do this?

components/build/kustomization.yaml
```diff
@@ -3,7 +3,6 @@ resources:
 - tekton-results/
 - build-templates/
 - pipelines-as-code/
-- tekton-chains/
```
lcarva, line 6: I think it makes sense to move this from `build` to `enterprise-contract`. I'm just worried what will happen during an upgrade. Will we have two Applications trying to manage the same resources?

argo-cd-apps/overlays/development/repo-overlay.yaml
zregvart: Removed

components/enterprise-contract/base/tekton-chains/kustomization.yaml
zregvart: Because we don't ArgoCD to load the chains-config ConfigMap from the Helm charts and from the configuration in the `rekor-*/tekton-chains-config.yaml`. Having two sources for the same object makes ArgoCD flip-flop between them

components/build/kustomization.yaml
zregvart: There should be only one Application, well -- two: `enterprise-contract` and (in some cases) `rekor`. ArgoCD should make sure that the definitions that for Tekton Chains that were loaded from the `build` application are replaced with the configuration now loaded from `enterprise-contract` application.

Michkov:

PR #633, opened 2022-08-21
PLNSRVCE-431: Change ArgoCD components to use HAS components for PaC integration

brunoapimentel:
For dogfooding purposes, this PR proposes a switch to a HAS Application/Component model for each of the ArgoCD components that live in the infra-deployments repository.

This means that instead of manually defining Pipelines-as-code Repositories in YAML files, their creation should be handled by the build-service as part of the reconciling process of the HAS Component CRs.

Some issues arise from this change, and will be addressed as comments in this PR.


argo-cd-apps/base/appstudio.yaml
```diff
@@ -1,32 +1,34 @@
 apiVersion: argoproj.io/v1alpha1
 kind: Application
 metadata:
-  name: e2e-tests
-
+  name: appstudio
```
brunoapimentel, line 4: All HAS Components should live under the same namespace as the main HAS Application. The proposal here is to have them all defined in the same folder and in a separate ArgoCD component. Another approach would be to keep the component.yaml in their respective ArgoCD component folders, but to force them to be created in the same namespace.

components/appstudio/application-service.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: appstudio.redhat.com/v1alpha1
+kind: Component
+metadata:
+  name: application-service-component
+  annotations:
+    pipelinesascode: "1"
+    skipprcreation: "1"
```
brunoapimentel, line 7: This annotation is a proposal to have the build-service to skip the creation of the PR that sets up the basic PaC resources in the related repository.

On the tests made, however, no PRs were created since the Github App did not have access to the redhat-appstudio org repositories. No errors were found in the build-service logs or the Component's status.

This might indicate that this proposed change to the build-service is not necessary.

brunoapimentel:
In order to test the pipeline execution, I changed one of the HAS Components to point to a private fork instead of the usual redhat-appstudio repository and gave the Github App access to it.

The pipeline ran as expected, but the PR to define the initial PaC resources was also created.

brunoapimentel:
We'll be closing this PR since changing all the components at the same time is very risky. A new PR changing only the build-service to this new model will be made in the near future.

PR #650, opened 2022-08-25
PLNSRVCE-624: set hacbs tekton bundle key

Michkov:
Set global hacbs tekton bundle so it can be used directly in hacbs
workflow.


psturc (approved):
lgtm, verified on public OCP 4.11 cluster

PR #653, opened 2022-08-25
add quotes around $CRCBINARY

amfred:
Recent versions of CRC on Mac often install it into a path with a space character.  We need to surround $CRCBINARY with quotes.

Tested on macos with:
CRC version: 2.7.1+8b30b973
OpenShift version: 4.11.0
Podman version: 4.1.1


amfred:
The appstudio-e2e-deployment test isn't relevant to this change, because it doesn't use the CRC scripts.

jduimovich (approved):
LGTM

PR #661, opened 2022-08-29
Allow access tekton-pipelines to shared secret

Michkov:
- allow to set custom registry secret
    - allow Avinal to create secret in the namespace


avinal:
Do we need to add a service account in tekton-results too? As in the jvm-build service.

avinal (approved):
Looks good to me 👍🏼

PR #668, opened 2022-08-30
Grant rbean tekton-triggers-admin

ralphbean:
I need rights to delete a validatingwebhookconfiguration to let argo
recreate it, to prove or disprove a theory in:
https://coreos.slack.com/archives/C031USXS2FJ/p1661870515286729


ralphbean:
On reflection, argo is not recreating the validatingwebhookconfiguration that I want to delete. Some other controller is.

PR #684, opened 2022-09-01
fix: increase timeout for e2e

psturc:
### Why
Currently there's a 1h (default) timeout for running e2e. Since the number of tests is growing and takes more time to finish, we are sometimes hitting the limit. If that happens, tests get interrupted and a CI check is marked as failed.

This is a temporary solution until we enable parallel e2e test run for infra-deployments (which was already enabled for e2e-tests repo)


psturc:
```
Ran 87 of 92 Specs in 4424.652 seconds
SUCCESS! -- 87 Passed | 0 Failed | 0 Pending | 5 Skipped
```

PR #697, opened 2022-09-02
Change tekton-results images to HACBS built images

avinal:
## Changes

- Tekton Results is now being built in HACBS, so using the latest images
from the HACBS build for both watcher and api
- Refer: openshift-pipelines/tektoncd-results#8

Signed-off-by: Avinal Kumar <avinal@redhat.com>


avinal:
My understanding is that just one update script can take care of all the image tags since api and watcher images are built simultaneously.

PR #721, opened 2022-09-09
Fix spi domain resolution

lcarva:
Not all cluster have `apps.` as part of the domain.


lcarva:
Looks like the file `hack/util-set-spi-vault-host.sh` has been removed and there wasn't much interest in this fix, so I'll go ahead an close this out :shrug:

PR #723, opened 2022-09-13
Add pact broker.

Katka92:
Adding Pact Broker for HAC-dev contract tests as it was agreed on the Architectures call.


psrna:
@sbose78 @elsony can you please review and approve? My understanding is that pre-kcp is frozen for critical fixes only for the reason of having a stable appstudio demo deployment. However this change is not touching any of the appstudio services and I believe that thus it qualifies for an exception. It was also agreed on the arch call that pact broker can be deployed on the staging cluster. Thank you!

argo-cd-apps/overlays/development/repo-overlay.yaml
```diff
@@ -107,3 +107,14 @@ spec:
   source:
     repoURL: https://github.com/redhat-appstudio/infra-deployments.git
     targetRevision: pre-kcp
+---
+apiVersion: argoproj.io/v1alpha1
+kind: Application
+metadata:
+  name: enterprise-contract
```
zregvart, line 114: Did you mean to name this `enterprise-contract`?

Katka92: Thank you, fixed.

MatousJobanek:
@psrna out of curiosity. What is the reason behind adding the contract tests for pre-kcp when the AppStudio components will be updated only in exceptional cases (as you have mentioned in your comment)?

Or maybe I misunderstood your intention - do you want to install it in appstudio-stage cluster and use it only as a place for storing the contracts for future kcp-based work? If so, then it's probably fine, I'm just not sure about the plans for maintaining this cluster. IIRC, as soon as we have the main branch working on kcp, then we wanted to destroy this cluster.

psrna:
> 

@MatousJobanek yes, the intent was to deploy it on appstudio-stage cluster (hence the pre-kcp branch) as it was advised during the arch call. The pack broker deployment should be used only for storing/managing the contracts and not for the contract tests themselves.

My understanding is that the cluster will not be destroyed immediately after the kcp migration, but will be used as a DEMO cluster. But either cluster would work for us. The old appstudio-stage cluster or the new one, kcp based. If you prefer to deploy to the new one, let us know. Thanks.

MatousJobanek:
> The pack broker deployment should be used only for storing/managing the contracts and not for the contract tests themselves.

right, this is what I meant by the second part of my comment. Store the contracts in appstudio-stage cluster, but use the contracts for tests against main.

> My understanding is that the cluster will not be destroyed immediately after the kcp migration, but will be used as a DEMO cluster. 

There won't be any DEMO cluster. As a demo environment, we will most likely use kcp-stable (or any other CPS instance created for this purpose, but I don't think that there is any need for it).
In this context, sync with AppStudio QE - what are their plans for testing the promotion from kcp-unstable to kcp-stable, if they have any.
When we have AppStudio running on CPS, then there is no reason for maintaining this cluster, so we would love to destroy it as soon as possible. If there are any other plans for the cluster, I'm not aware of them.

In the Arch call, it was mentioned that the pact broker needs to be installed along with QE dashboard (no clue why) which is currently in appstudio-stage cluster - you are right. I don't have any information on what is the plan for migrating these components.

TBH, I don't have a strong opinion on where the pact broker should be installed:
* appstdudio-stage cluster
* appstudio-kcp cluster
* a kcp workspace in CPS
* any other place ...

As I see it, at the end of the day we should dogfood our stuff as much as possible, so ideally it should be deployed as another AppStudio application/component.

So it's up to you - you will manage it and take care of the migration if needed.
I just wanted to mention it here so you are aware of the fact that the appstudio-stage cluster can be destroyed in a few months.

Katka92:
@sbose78 @elsony Sorry, I didn't mean to remove your request for review, that was by accident. :grimacing:

rhopp:
I'm not really sure... but I guess this won't work. If something, ArgoCD could be able to create Template, but I doubt it will create objects out of that template.

Katka92:
I've tried to separate CRs into files, removing the template file. I tried that running `bootstrap-cluster.sh`. The manual setup of secrets is required in addition, but it should work fine.

rhopp (approved):

PR #727, opened 2022-09-14
PLNSRVCE-658: <do not merge yet> readd kcp'ed jvm-build-service but leverages a bunch of simplifications over pre-kcp integration

gabemontero:
so while I cannot test yet with the known "pipelineruns missing from 'api-resources'" situation for us, I did some preliminary modeling of how jvm-build-service would add itself back in

In crafting the changes, and comparing with the pre-kcp branch, it motivated me to craft https://github.com/redhat-appstudio/jvm-build-service/pull/258 to simplify how we reference the 5 jvm-build-service images, where I believe I was able to remove the need for image ref specification in the system configmap and deployment env vars for the product flow, dev flow, and openshift ci flow.

That resulted in removing some of the kustomization operations in the pre-kcp branch.

I see staging as 
- get https://github.com/redhat-appstudio/jvm-build-service/pull/258 merged after sign off from @stuartwdouglas 
- track the various efforts with @Michkov  and the infra team so that we can get pipeline runs available and synch'ing in some form of apptudio kcp
- test https://github.com/redhat-appstudio/jvm-build-service/pull/236, get it working, then review/iterate with team / get it merged
- then come back to this PR

That said, if there is some obvious feedback on this PR reviewers want to provide while this is still in draft state, please do.

### DIFF
--- a/components/jvm-build-service/kustomization.yaml
+++ b/components/jvm-build-service/kustomization.yaml
@@ -1,0 +1,22 @@
+resources:
+- https://github.com/redhat-appstudio/jvm-build-service/deploy/kcp?ref=bc1ecfaf82f1edaef5b4d80e53d093e371048fb7
+
+images:
+- name: hacbs-jvm-operator
+  newName: quay.io/redhat-appstudio/hacbs-jvm-controller
+  newTag: e927aa1b0dcd479ce0ad4f4a62a87b3b1b7dc423
+
+namespace: jvm-build-service
+
+patches:
+  - target:
+      kind: Deployment
+      name: hacbs-jvm-operator
+    patch: |-
+      # reduce CPU request to be able to run on testing instance
+      - op: replace
+        path: /spec/template/spec/containers/0/resources/requests/cpu
+        value: 10m
+
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
--- a/argo-cd-apps/base/jvm-build-service.yaml
+++ b/argo-cd-apps/base/jvm-build-service.yaml
@@ -1,0 +1,42 @@
+apiVersion: argoproj.io/v1alpha1
+kind: ApplicationSet
+metadata:
+  name: jvm-build-service
+spec:
+  generators:
+  - list:
+      elements:
+      - kcp-name: kcp-stable
+      - kcp-name: kcp-unstable
+  template:
+    metadata:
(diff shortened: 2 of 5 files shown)


components/jvm-build-service/kustomization.yaml
```diff
@@ -0,0 +1,22 @@
+resources:
+- https://github.com/redhat-appstudio/jvm-build-service/deploy/kcp?ref=bc1ecfaf82f1edaef5b4d80e53d093e371048fb7
```
gabemontero, line 2: no longer need to manage the system-config / operator-images items that were present in the pre-kcp branch

components/jvm-build-service/kustomization.yaml
```diff
@@ -0,0 +1,22 @@
...
+
+images:
+- name: hacbs-jvm-operator
+  newName: quay.io/redhat-appstudio/hacbs-jvm-controller
+  newTag: e927aa1b0dcd479ce0ad4f4a62a87b3b1b7dc423
```
gabemontero, line 7: the operator will now be able to determine the image ref for the other jvm build service images built with this commit ^^ and use them

components/jvm-build-service/kustomization.yaml
Michkov, line 2: Cannot be evaluated:
```
# kustomize build https://github.com/redhat-appstudio/jvm-build-service/deploy/kcp?ref=bc1ecfaf82f1edaef5b4d80e53d093e371048fb7
Error: accumulating resources: accumulation err='accumulating resources from '../kcp-manager': '/tmp/kustomize-788111474/deploy/kcp-manager' must resolve to a file': recursed accumulation of path '/tmp/kustomize-788111474/deploy/kcp-manager': add operation does not apply: doc is missing path: "/spec/template/spec/containers/2/args/-": missing value

```

gabemontero: yeah expected - the commit level is out of date, and the PR has not merged yet

for now, I'm just depicting future intent with this change.

I should have  conveyed that in a comment.

this will be corrected once the jvm-build-service PR has merged, which I'll do once I've been able to successfully bring up on kcp locally; that will most likely coincide with my switching this out of draft status.

gabemontero:
closing in favor of https://github.com/redhat-appstudio/infra-deployments/pull/758

PR #728, opened 2022-09-15
Prototype of adding shared resources for preview mode

Michkov:
- Changed structure to include overlays
- preview.sh is updating apiexport-patch.yaml with identityHash
- added shared-resources and self binding for development environment


components/has/overlays/dev/apiexport-patch.yaml
```diff
@@ -0,0 +1,16 @@
...
+  - group: ""
+    resource: secrets
+  - group: ""
+    resource: configmaps
+  - identityHash: gitopsrvc-appstudio-shared
+    group: "appstudio.redhat.com"
+    resource: "environments"
+  - identityHash: gitopsrvc-appstudio-shared
+    group: "appstudio.redhat.com"
+    resource: "applicationsnapshotenvironmentbindings"
```
MatousJobanek, line 16: I would rather keep the permissionClaims together with the rest of the APIExport in the API repository and patch only the actual identityHash here. 
I know that it would mean that we need to rely on the position in the list of permission claims to be able to set the correct identity hash, but it's still better than replicating the whole permissionClaims section here.

MatousJobanek (commented):
I know that it's still a draft, just leaving a comment here that we will most likely move the whole HAS & GitOps-appstudio API to a separate repository. 
https://coreos.slack.com/archives/C02CTEB3MMF/p1663258906060969?thread_ts=1663084426.737819&cid=C02CTEB3MMF

components/has/overlays/dev/apiexport-patch.yaml
Michkov: For debugging purposes it's easier to have it in one file. Validation if API exporter points to right identity hash could be complicated. Also this format gives me way to apply this file directly to create placeholder IdentityHash when apiexport does not exists yet but we need it's identityHash

MatousJobanek: > For debugging purposes it's easier to have it in one file. Validation if API exporter points to right identity hash could be complicated. 

I fully understand that, but this would mean that we need to keep the permissionClaims in sync with the original APIExport, right? And have a validation that checks if there is any update in the original permissionClaims which wasn't reflected here. This can be tricky as well.

> Also this format gives me way to apply this file directly to create placeholder IdentityHash when apiexport does not exists yet but we need it's identityHash

But wouldn't it be easier to just create a set of empty APIExports without any APIResourceSchema or permissionClaims? You technically don't need that, right?

johnmcollier: Wondering too if we should wait on the discussions around using a shared APIExport for HAS + gitops-appstudio-shared? If we do move to that

Michkov: We need to have permissionClaims so that we can identify to which APIExport it should point via IdentityHash.

Michkov:
Marking as ready for now, with these changes when running preview.sh it will deploy application-service successfully and it's possible to create applications and components.

PR #729, opened 2022-09-15
add support for HACBS service provider workspace

MatousJobanek:
* refactoring of `hack/configure-kcp.sh` to use functions so we can configure appstudio & hacbs worskpace via one function
* changed the templating of `argo-cd-apps/base/authorization-in-kcp.yaml` to use matrix so it's applied for both SP workspaces
* slightly changed the structure of `components/authorization/kcp` by adding:
    * base - which is applied for both workspaces
    * appstudio - only for appstudio specific authz
    * hacbs - only for HACBS specific authz

with these changes, we will be able to create ApplicationSet using `redhat-hacbs-workspace-{{kcp-name}}` as the destination name for all HACBS components

!!! DO NOT MERGE YET !!! waiting until the `redhat-hacbs` service provider workspace is created in CPS

cc @ralphbean


hack/configure-kcp.sh
```diff
@@ -21,48 +21,44 @@ function extra_params() {
...
+    if grep -q "insecure-skip-tls-verify: true" ${KCP_KUBECONFIG}; then
+      sed -i 's/certificate-authority-data: .*/insecure-skip-tls-verify: true/' /tmp/${SYNC_TARGET}-syncer.yaml
+    fi
+    kubectl apply -f /tmp/${SYNC_TARGET}-syncer.yaml --kubeconfig ${CLUSTER_KUBECONFIG}
+  fi
+  
+  BIND_SCOPE="system:authenticated"
+  if [[ ${KCP_INSTANCE_NAME} == "kcp-stable" ]] || [[ ${KCP_INSTANCE_NAME} == "kcp-unstable" ]] && [[ ${ROOT_WORKSPACE} == "root" ]]
+  then
+    BIND_SCOPE="rh-sso:16270929"
```
scoheb, line 57: can we leave a comment here as to what this refers to?

scoheb: actually, I now see this was already here...nonetheless if you know, a comment would be nice

ralphbean:
Wanted to verify preview and bootstrap changes in kcp-stable against my user workspace, but it is having some OOM issues today: https://coreos.slack.com/archives/C035ZBD755E/p1663262918662849?thread_ts=1663262799.902799&cid=C035ZBD755E

ralphbean:
OK, with kcp-stable working again, I can confirm bootstrap in preview mode against kcp-stable works as well as expected.

ralphbean:
@MatousJobanek I think this needs an additional change in `hack/destroy.sh` to add removal of the redhat-hacbs workspace.

hack/configure-kcp.sh
MatousJobanek: added comment in 966caac
I hope that it clarifies the purpose of it

MatousJobanek:
@ralphbean Good catch - I totally forgot to update that as well. 
Done in [d358b12](https://github.com/redhat-appstudio/infra-deployments/pull/729/commits/d358b1263d9d4b4c9175d4b154250b16982cd2fe)
Thanks

ralphbean (approved):

ralphbean:
Awaiting https://coreos.slack.com/archives/C035ZBD755E/p1662729348114149

Michkov (approved):

ralphbean:
OK, I think https://issues.redhat.com/browse/KCP-228 is done now and this can be merged.

PR #732, opened 2022-09-16
sync replace after configuring kcp

MatousJobanek:
Figured out that hard refresh is not enough for loading the newly added cluster definition and that the Applications has to be sync-replaced. :man_shrugging:

### DIFF
--- a/hack/configure-kcp.sh
+++ b/hack/configure-kcp.sh
@@ -173,3 +173,7 @@ for APP in $(kubectl get apps -n openshift-gitops -o name --kubeconfig ${CLUSTER
   kubectl patch ${APP} -n openshift-gitops --type merge -p='{"metadata": {"annotations":{"argocd.argoproj.io/refresh": "hard"}}}' --kubeconfig ${CLUSTER_KUBECONFIG}
 done
 
+echo "Triggering replace sync of all Applications:"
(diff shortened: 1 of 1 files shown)


hack/configure-kcp.sh
```diff
@@ -173,3 +173,7 @@ for APP in $(kubectl get apps -n openshift-gitops -o name --kubeconfig ${CLUSTER
   kubectl patch ${APP} -n openshift-gitops --type merge -p='{"metadata": {"annotations":{"argocd.argoproj.io/refresh": "hard"}}}' --kubeconfig ${CLUSTER_KUBECONFIG}
 done
 
+echo "Triggering replace sync of all Applications:"
+for APP in $(kubectl get apps -n openshift-gitops -o name --kubeconfig ${CLUSTER_KUBECONFIG}); do
+  kubectl patch ${APP} -n openshift-gitops --type merge -p='{"operation": {"sync":{"syncOptions": ["Replace=true"]}}}' --kubeconfig ${CLUSTER_KUBECONFIG}
```
Michkov, line 178: does it help? I can see mostly issue that the cluster is created twice with the same name, so it's not related to application itself.

MatousJobanek: what do you mean by

>  the cluster is created twice with the same name

I guess that such an error is a result of having two secrets using the same cluster name, right? If so, then this is something to fix/improve in the scripts.

Michkov (approved):

PR #736, opened 2022-09-19
Addressing leftovers after the merge of "SPI on KCP/CPS"

skabashnyuk:
## Goal
Addressing leftovers after the merge of https://github.com/redhat-appstudio/infra-deployments/pull/715

### DIFF
--- a/argo-cd-apps/base/spi.yaml
+++ b/argo-cd-apps/base/spi.yaml
@@ -1,3 +1,4 @@
+---
 apiVersion: argoproj.io/v1alpha1
 kind: ApplicationSet
 metadata:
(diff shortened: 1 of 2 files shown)


hack/preview.sh
```diff
@@ -39,12 +39,30 @@ if [ -n "$MY_GITHUB_ORG" ]; then
     $ROOT/hack/util-set-github-org $MY_GITHUB_ORG
 fi
 
+
+echo
+echo -n "Waiting for 'spi-system' namespace to exist: "
+while ! kubectl get namespace spi-system --kubeconfig ${KCP_KUBECONFIG} &> /dev/null ; do
```
Michkov, line 45: argoCD is not applied yet, so the spi-system namespace will not be created. The current code is creating the namespace and route, which is then updated by argoCD

argo-cd-apps/base/spi.yaml
```diff
@@ -41,7 +41,7 @@ spec:
             duration: 15s # the amount to back off. Default unit is seconds, but could also be a duration (e.g. "2m", "1h")
 ---
 apiVersion: argoproj.io/v1alpha1
-kind: ApplicationSet
+kind: Application
 metadata:
   name: spi-vault
 spec:
```
Michkov, line 47: generators and template should be removed

Michkov: should be similar to https://github.com/redhat-appstudio/infra-deployments/blob/pre-kcp/argo-cd-apps/base/spi.yaml

also https://github.com/redhat-appstudio/infra-deployments/blob/main/argo-cd-apps/overlays/development/repo-overlay.yaml#L34-L50 should be updated

argo-cd-apps/base/spi.yaml
```diff
@@ -41,42 +41,34 @@ spec:
...
-      name: spi-vault
-    spec:
-      project: default
+  project: default
 
-      source:
-        path: components/spi/vault
-        repoURL: https://github.com/redhat-appstudio/infra-deployments.git
-        targetRevision: main
+    source:
```
Michkov, line 50: `source`(and other elements on the level) must be same level as `project`

Michkov (approved):

MatousJobanek (approved):

PR #740, opened 2022-09-20
Integration of application-api

Michkov:
- Changed structure to include overlays
- added application-api

When APIBinding created in different workspace the application-service starts sucessfully and creates github repositories

### DIFF
--- a/components/application-api/base/kustomization.yaml
+++ b/components/application-api/base/kustomization.yaml
@@ -1,0 +1,4 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- https://github.com/redhat-appstudio/application-api/config/kcp?ref=f978dd513b9d0cfab6a588d142ba8198704201e7
--- a/components/application-service/base/manager_resources_patch.yaml
+++ b/components/application-service/base/manager_resources_patch.yaml
@@ -18,4 +18,7 @@ spec:
         env:
           - name: IMAGE_REPOSITORY
             value: ""
-
+        args:
+          - '--leader-elect'
+          - '--api-export-name'
+          - application-api
--- a/components/application-service/overlays/dev/kustomization.yaml
+++ b/components/application-service/overlays/dev/kustomization.yaml
@@ -1,0 +1,10 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- ../../base/
+
+configMapGenerator:
+- literals:
+  - GITHUB_ORG=""
+  name: github-config
+  behavior: replace
--- a/argo-cd-apps/base/application-api.yaml
+++ b/argo-cd-apps/base/application-api.yaml
@@ -1,7 +1,7 @@
 apiVersion: argoproj.io/v1alpha1
 kind: ApplicationSet
 metadata:
-  name: application-service
+  name: application-api
 spec:
   generators:
   - list:
@@ -10,17 +10,17 @@ spec:
       - kcp-name: kcp-unstable
   template:
     metadata:
-      name: '{{kcp-name}}-application-service'
+      name: '{{kcp-name}}-application-api'
     spec:
       project: default
 
       source:
-        path: components/application-service/overlays/{{kcp-name}}
+        path: components/application-api/base
         repoURL: https://github.com/redhat-appstudio/infra-deployments.git
         targetRevision: main
 
       destination:
-        namespace: application-service
+        namespace: application-api
         name: redhat-appstudio-workspace-{{kcp-name}}
 
       syncPolicy:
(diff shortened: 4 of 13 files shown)


components/application-api/base/kustomization.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- https://github.com/johnmcollier/application-api/config/kcp?ref=addAPIs
```
Michkov, line 4: needs to be updated after https://github.com/redhat-appstudio/application-api/pull/1 is merged

Michkov: Updated

johnmcollier (approved):
Looks good from my end 👍

components/application-service/base/manager_resources_patch.yaml
```diff
@@ -18,4 +18,7 @@ spec:
         env:
           - name: IMAGE_REPOSITORY
             value: ""
-
+        args:
+          - '--leader-elect'
+          - '--api-export-name'
+          - application-api
```
johnmcollier, line 24: So these changes will get HAS "running", but in order to support private Git flows in AppStudio, HAS still needs the SPI APIs made available to it. 

I'm wondering what our options are here:
1) Include a permission claim for the SPI APIExport identity hash
    - This does mean that _any_ clients needing to use the application-api APIExport, would need to set the SPI identity hash
    - Wouldn't that defeat the purpose of having application-api and SPI sepratate then?

2) Include a dummy APIExport for has under `config/application-service` that includes permission claims for both application-api and spi.
    - I suppose it could live under `config/kcp` in the HAS repo too. I'm flexible either way.
    
@Michkov @MatousJobanek what do you think?

Michkov: I would not block this PR, let's progress step by step.

I am also for option 2, so that we can patch the APIExport here in infra-deployment for kcp-stable and kcp-unstable (providing identity hashes) and create a script which would generate hashes for dev environment.

components/application-service/overlays/dev/kustomization.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- ../../base/
+
+configMapGenerator:
+- literals:
+  - GITHUB_ORG=""
+  name: github-config
+  behavior: replace
```
MatousJobanek, line 10: originally, this was used in `base` - which means for all environments (CPS & dev). The same is also used in pre-kcp version. 
TBH, I don't know the HAS implementation details, so only mentioning this so we don't miss anything.

MatousJobanek (approved):

components/application-service/base/manager_resources_patch.yaml
MatousJobanek: I was thinking about this - there are a few options. Let me list them here keeping the first two from your comment.

1. Include a permission claim for the SPI API in the application-api APIExport
   * yes, this is the most straightforward way of doing it, but it would mean that we need to combine permission claims from both HAS an GitOps, which is probably nothing bad but could be a bit messy
   * if someone needs to create the APIExport for it's own testing of a 3rd party component that depends on it, then it would need to ideally drop the permission claims from the APIExport.
   * 1b. - an alternative way is having overlays that automatically adds or removes the permission claims based on when and how it's supposed to be used. With this, we could have one APIExport with all needed permission claims in it (from HAS & GitOps), but still providing the way of reusing it by other teams
 
2. Include a dummy APIExport for HAS under config/application-service that includes permission claims for both application-api and spi.
    * This is my favorite option too. We would have separate APIExports for HAS & GitOps claiming from the other APIExports, but not providing any API.
    * This will be also similar to what the Build APIExport will look like (not providing any API, but claiming from application-api and pipeline-service)
    * it's just a bit more APIExports to manage (by setting the right identity hashes), but we will need to do it anyway, so it's probably not a big deal.

3. Keep permission claims only in infra-deployments repository
    * instead of having the permission claims directly in the APIExports in the repository, we could keep it in infra-deployments and add it together with the identity hashes. 
    * TBH, I would prefer to keep the permission claims part directly in the corresponding repositories (inside the APIExports) and set only the identity hashes in infra-deployments. However, this will be tricky to do as well since we will need to rely on the position of the permission claim while doing the patch of the identity hash.
    * see https://github.com/redhat-appstudio/infra-deployments/pull/728#discussion_r972722124

But I agree with @Michkov - let's not block this PR and solve it in the next one

components/application-service/overlays/dev/kustomization.yaml
Michkov: I moved this to dev since preview is changing GITHUB_ORG. And preview will be deploying only dev overlay. So I rather changed only in dev. I am not sure right now if the kcp-stable and kcp-unstable will be using same GITHUB_ORG if not then it can be easily patched in their overlay folders.

PR #743, opened 2022-09-20
SPI on KCP/CPS

skabashnyuk:
- Do not include shared-configuration-file secret as part of kustomize output. 
- Give skabashniuk admin rights for kcp-5ttoh8uv3y7i namespace. That is a temporary solution to figure out why secrets created from kcp/spi-system namespace do not appear in kcp-5ttoh8uv3y7i  on the workload cluster.

Secret exist in spi-system
Знімок екрана 2022-09-20 о 18 14 33
can't be located in  kcp-5ttoh8uv3y7i 
Знімок екрана 2022-09-20 о 18 14 17


components/authorization/cluster/kcp-5ttoh8uv3y7i-admin.yaml
```diff
@@ -0,0 +1,12 @@
+kind: RoleBinding
```
Michkov, line 1: the file name could be more generic, eg. kcp-spi-namespace-admin.yaml, since you will also include there rolebinding for unstable kcp namespace.

Michkov (approved):

MatousJobanek (approved):

PR #745, opened 2022-09-21
Add script for install Pipeline Service

Michkov:
Script for installing Pipeline Service for testing purposes.

### DIFF
--- a/hack/install-pipeline-service.sh
+++ b/hack/install-pipeline-service.sh
@@ -7,7 +7,7 @@ MODE=preview parse_flags $@
 
 
 PIPELINE_SERVICE_DIR=$(mktemp -d)
-git clone -b record-server-ca-during-registration --depth 1 https://github.com/lcarva/pipeline-service/ $PIPELINE_SERVICE_DIR
+git clone -b record-server-ca-during-registration --depth 1 https://github.com/openshift-pipelines/pipeline-service/ $PIPELINE_SERVICE_DIR
 export WORK_DIR="${PIPELINE_SERVICE_DIR}/gitops/sre/"
 export WORKSPACE_DIR=$WORK_DIR
 
--- a/hack/install-pipeline-service.sh
+++ b/hack/install-pipeline-service.sh
@@ -1,0 +1,45 @@
+#!/bin/bash -e
+
+ROOT="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"/..
+
+source ${ROOT}/hack/flags.sh "The install-pipeline-service.sh installs Pipeline Service for development and testing on non-production clusters / kcp instances."
+MODE=${MODE:-preview} parse_flags $@
+
+
+PIPELINE_SERVICE_DIR=$(mktemp -d)
+git clone -b record-server-ca-during-registration --depth 1 https://github.com/lcarva/pipeline-service/ $PIPELINE_SERVICE_DIR
+export WORK_DIR="${PIPELINE_SERVICE_DIR}/gitops/sre/"
+export WORKSPACE_DIR=$WORK_DIR
+
+if [ "$ROOT_WORKSPACE" == "~" ]; then
+  ROOT_WORKSPACE=$(KUBECONFIG=${KCP_KUBECONFIG} kubectl ws '~' | cut -f2 -d'"')
+fi
+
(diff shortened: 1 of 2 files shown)


hack/install-pipeline-service.sh
```diff
@@ -0,0 +1,45 @@
...
+  ROOT_WORKSPACE=$(KUBECONFIG=${KCP_KUBECONFIG} kubectl ws '~' | cut -f2 -d'"')
+fi
+
+KUBECONFIG=$KCP_KUBECONFIG $PIPELINE_SERVICE_DIR/images/access-setup/content/bin/setup_kcp.sh --kcp-workspace pipeline-service --kcp-org $ROOT_WORKSPACE
+
+KUBECONFIG=$CLUSTER_KUBECONFIG $PIPELINE_SERVICE_DIR/images/access-setup/content/bin/setup_compute.sh
+
+$PIPELINE_SERVICE_DIR/images/cluster-setup/bin/install.sh
+
+$PIPELINE_SERVICE_DIR/images/kcp-registrar/register.sh --kcp-org $ROOT_WORKSPACE --kcp-workspace pipeline-service --kcp-sync-tag v0.8.2
```
bnallapeta, line 24: Need to use KUBECONFIG=$CLUSTER_KUBECONFIG without which it will point to whichever kubeconfig the user currently has exported.

Michkov: the script itself handles this, there is always `KUBECONFIG` variable set before calling `kubectl`. The script is written to work on multiple clusters, it reads kubeconfigs from credentials folder

bnallapeta: Okay. But how is it different for Line 20 where you're explicitly using KUBECONFIG=$CLUSTER_KUBECONFIG?

Michkov: setup_compute.sh uses `kubectl` command directly and expects that KUBECONFIG is pointing to cluster.

register.sh does not expect that user passes KUBECONFIG since it is taken from credential files.

bnallapeta: Got it. Thanks for clarifying this.

hack/install-pipeline-service.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/bash -e
+
+ROOT="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"/..
+
+source ${ROOT}/hack/flags.sh "The install-pipeline-service.sh installs Pipeline Service for development and testing on non-production clusters / kcp instances."
+MODE=${MODE:-preview} parse_flags $@
```
lcarva, line 6: Should it always force `preview`? e.g. `MODE=preview parse_flags $@`

hack/install-pipeline-service.sh
```diff
@@ -0,0 +1,45 @@
...
+
+
+PIPELINE_SERVICE_DIR=$(mktemp -d)
+git clone -b record-server-ca-during-registration --depth 1 https://github.com/lcarva/pipeline-service/ $PIPELINE_SERVICE_DIR
+export WORK_DIR="${PIPELINE_SERVICE_DIR}/gitops/sre/"
+export WORKSPACE_DIR=$WORK_DIR
+
+if [ "$ROOT_WORKSPACE" == "~" ]; then
```
lcarva, line 14: Should there me something early in this script that slurps in the values from `hack/preview.env` ?

hack/install-pipeline-service.sh
Michkov: updated

hack/install-pipeline-service.sh
Michkov: done by `parse_flags`

lcarva (approved):

hack/install-pipeline-service.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/bash -e
+
+ROOT="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"/..
+
+source ${ROOT}/hack/flags.sh "The install-pipeline-service.sh installs Pipeline Service for development and testing on non-production clusters / kcp instances."
+MODE=preview parse_flags $@
+
+
+PIPELINE_SERVICE_DIR=$(mktemp -d)
+git clone -b record-server-ca-during-registration --depth 1 https://github.com/lcarva/pipeline-service/ $PIPELINE_SERVICE_DIR
```
lcarva, line 10: FYI, this is in main now: https://github.com/openshift-pipelines/pipeline-service/pull/245

lcarva (approved):

PR #748, opened 2022-09-21
Add Release Service members

scoheb:
- initial Release Service team members for onboarding


Michkov:
@MatousJobanek Do we have any rules for adding people? I am not in the list so I do not feel that I can decide to add new users.

MatousJobanek (approved):
I don't think that we have any rules for this. 
And please add yourself as well ;-)

PR #755, opened 2022-09-22
SPI  on CPS/Stable

skabashnyuk:
- Reconfigure  VAULTHOST to correct vault route on workload cluster
 - `BASEURL` to OAuth service on CPS/Stable

Context

Знімок екрана 2022-09-22 о 11 41 24
Знімок екрана 2022-09-22 о 11 41 16

### DIFF
--- a/components/authorization/cluster/kcp-spi-namespace-admin.yaml
+++ b/components/authorization/cluster/kcp-spi-namespace-admin.yaml
@@ -10,16 +10,3 @@ roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
   name: admin
----
-kind: RoleBinding
-apiVersion: rbac.authorization.k8s.io/v1
-metadata:
-  name: kcp-1ry9c9jzxyzk-admin
-  namespace: kcp-1ry9c9jzxyzk
-subjects:
-  - kind: User
-    name: skabashnyuk
-roleRef:
-  apiGroup: rbac.authorization.k8s.io
-  kind: ClusterRole
-  name: admin


components/spi/config-patch.json
```diff
@@ -2,11 +2,11 @@
...
     "op": "replace",
     "path": "/data/VAULTHOST",
-    "value": "https://spi-vault-spi-vault.apps.appstudio-kcp.g2pi.p1.openshiftapps.com"
+    "value": "https://vault-spi-vault.apps.appstudio-kcp.g2pi.p1.openshiftapps.com"
   },
   {
     "op": "replace",
     "path": "/data/BASEURL",
-    "value": "https://spi-oauth-kcp-2l71pp1tc2f1.apps.appstudio-kcp.g2pi.p1.openshiftapps.com"
+    "value": "https://spi-oauth-kcp-5ttoh8uv3y7i.apps.appstudio-kcp.g2pi.p1.openshiftapps.com"
```
MatousJobanek, line 10: but this would mean that it would use one endpoint for both kcp instances, is this what you want?

components/authorization/cluster/kcp-spi-namespace-admin.yaml
```diff
@@ -10,3 +10,16 @@ roleRef:
...
+metadata:
+  name: kcp-1ry9c9jzxyzk-admin
+  namespace: kcp-1ry9c9jzxyzk
+subjects:
+  - kind: User
+    name: skabashnyuk
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: admin
```
MatousJobanek, line 25: just out of curiosity - why do you need that? All modifications in this namespace should happen via the syncer.

MatousJobanek: I fully understand that it's a huge black box, but this is exactly the reason why I would love to not edit anything in these namespaces (even accidentally). We just don't know what we could break there.

MatousJobanek (approved):

PR #760, opened 2022-09-22
Add APIExport & APIBinding for HAS

johnmcollier:
- Adds a simple APIExport for HAS that claims the APIs from `application-api` and `spi`
   - Uses placeholder values for the identityhashes 
   - Also adds a stub APIBinding for HAS
- Updates the KCP manifest location for HAS from `config/kcp` to `config/kcp-manager`
   - Prevents a collision with the old APIExport & APIResourceSchemas for the time being.
- Adds a stub APIBinding for `application-api`
   - Required so that the HAS APIExport can successfully use its APIs


Michkov:
Solved by #763

PR #765, opened 2022-09-26
automatically accept all permission claims in APIBindings

MatousJobanek:


hack/create-user-workspace.sh
```diff
@@ -39,5 +39,18 @@ kubectl ws create ${USER_APPSTUDIO_WORKSPACE}  --ignore-existing --type root:uni
...
 
+echo
+echo "Patching APIBindings to accept all permission claims:"
+API_BINDINGS=$(kubectl get apibindings.apis.kcp.dev -l provided-by=infra-deployments -o name)
+for API_BINDING in ${API_BINDINGS}
+do
+  EXPORT_PERMISSION_CLAIMS=$(kubectl get ${API_BINDING} -o jsonpath='{.status.exportPermissionClaims}')
+  for EXPORT_CLAIM in $(echo ${EXPORT_PERMISSION_CLAIMS} | jq -c '.[]')
+  do
+    ACCEPTED_CLAIMS=${ACCEPTED_CLAIMS}$(echo ${EXPORT_CLAIM} | jq '. += {"state": "Accepted"}' | jq -c)","
```
Michkov, line 50: ACCEPTED_CLAIMS is global variable so it's used by all API_BINIDINGS in the cycle, it should be emptied at the beginning of first for cycle

hack/create-user-workspace.sh
```diff
@@ -39,5 +39,18 @@ kubectl ws create ${USER_APPSTUDIO_WORKSPACE}  --ignore-existing --type root:uni
 kubectl kustomize ${ROOT}/apibindings/${SERVICE_NAME}/ | sed "s|\${APPSTUDIO_SP_WORKSPACE}|${APPSTUDIO_SP_WORKSPACE}|g;s|\${HACBS_SP_WORKSPACE}|${HACBS_SP_WORKSPACE}|g" | \
   kubectl apply -f -
 
+echo
+echo "Patching APIBindings to accept all permission claims:"
+API_BINDINGS=$(kubectl get apibindings.apis.kcp.dev -l provided-by=infra-deployments -o name)
+for API_BINDING in ${API_BINDINGS}
+do
+  EXPORT_PERMISSION_CLAIMS=$(kubectl get ${API_BINDING} -o jsonpath='{.status.exportPermissionClaims}')
```
Michkov, line 47:
```suggestion
  EXPORT_PERMISSION_CLAIMS=$(kubectl get ${API_BINDING} -o jsonpath='{.status.exportPermissionClaims}')
  ACCEPTED_CLAIMS=
```

Michkov (changes requested):

hack/create-user-workspace.sh
MatousJobanek: good catch,thanks!

hack/create-user-workspace.sh
```diff
@@ -39,5 +39,18 @@ kubectl ws create ${USER_APPSTUDIO_WORKSPACE}  --ignore-existing --type root:uni
...
+  for EXPORT_CLAIM in $(echo ${EXPORT_PERMISSION_CLAIMS} | jq -c '.[]')
```
Michkov, line 49: must be `for EXPORT_CLAIM in $(echo "${EXPORT_PERMISSION_CLAIMS}" | jq -c '.[]')`, otherwise `echo ${EXPORT_PERMISSION_CLAIMS}` returns only 'y', not sure why

MatousJobanek: interesting, I don't see such behavior.
But anyway, in this case using the double quote makes complete sense - fixed in: 1c3a6f5

Michkov (approved):

PR #772, opened 2022-09-29
Preview - add pipeline service integration

Michkov:
Force users to install pipeline-service or set environment variables to connect to existing instance.


mmorhun (approved):
LGTM

hack/create-user-workspace.sh
```diff
@@ -16,6 +16,7 @@ SERVICE_NAME=${1}
 ROOT_WORKSPACE=${ROOT_WORKSPACE:-"root"}
 APPSTUDIO_WORKSPACE=${APPSTUDIO_WORKSPACE:-"redhat-appstudio"}
 HACBS_WORKSPACE=${HACBS_WORKSPACE:-"redhat-hacbs"}
+PIPELINE_SERVICE_WORKSPACE=${PIPELINE_SERVICE_WORKSPACE:-"pipeline-service"}
```
MatousJobanek, line 19: this is not the right workspace name for kcp-stable:
```suggestion
PIPELINE_SERVICE_WORKSPACE=${PIPELINE_SERVICE_WORKSPACE:-"redhat-pipeline-service-compute"}
```

hack/preview-template.env
```diff
@@ -24,6 +24,15 @@ export MY_GITHUB_TOKEN=
 
 # Optional
 
+## Pipeline Service pointers
+### Pipeline Service workspace, default $ROOT_WORKSPACE:pipeline-service
+### On kcp-stable - root:redhat-pipeline-service-compute
+export PIPELINE_SERVICE_SP_WORKSPACE=
```
MatousJobanek, line 30: ok, now I see that it was mentioned here, but I would love to use the default values that correspond to the production ones.
BTW, don't forget that create-user-workspace.sh doesn't load values from this file. you should be able to use it without preview mode - in CPS using the APIExports from `root`

PR #773, opened 2022-09-29
Convert ROOT_WORKSPACE='~' to full path in flags.sh

Michkov:
- Access workspaces directly, not needed to go via ROOT_WORKSPACE
- Allow to define HOME_WORKSPACE

### DIFF
--- a/hack/create-user-workspace.sh
+++ b/hack/create-user-workspace.sh
@@ -14,6 +14,7 @@ fi
 SERVICE_NAME=${1}
 
 ROOT_WORKSPACE=${ROOT_WORKSPACE:-"root"}
+HOME_WORKSPACE=${HOME_WORKSPACE:-"~"}
 APPSTUDIO_WORKSPACE=${APPSTUDIO_WORKSPACE:-"redhat-appstudio"}
 HACBS_WORKSPACE=${HACBS_WORKSPACE:-"redhat-hacbs"}
 PIPELINE_SERVICE_WORKSPACE=${PIPELINE_SERVICE_WORKSPACE:-"pipeline-service"}
@@ -23,12 +24,12 @@ then
   export KUBECONFIG=${KCP_KUBECONFIG}
 fi
 
-echo "Accessing the home workspace:"
-kubectl ws '~'
-
 if [ "${ROOT_WORKSPACE}" == "~" ]; then
   ROOT_WORKSPACE=$(kubectl ws . --short)
 fi
+
+echo "Accessing the home workspace:"
+kubectl ws $HOME_WORKSPACE
 
 APPSTUDIO_SP_WORKSPACE=${APPSTUDIO_SP_WORKSPACE:-${ROOT_WORKSPACE}:${APPSTUDIO_WORKSPACE}}
 HACBS_SP_WORKSPACE=${HACBS_SP_WORKSPACE:-${ROOT_WORKSPACE}:${HACBS_WORKSPACE}}
--- a/hack/create-user-workspace.sh
+++ b/hack/create-user-workspace.sh
@@ -25,7 +25,7 @@ then
 fi
 
 if [ "${ROOT_WORKSPACE}" == "~" ]; then
-  ROOT_WORKSPACE=$(kubectl ws . --short)
+  ROOT_WORKSPACE=$(kubectl ws '~' --short)
 fi
 
 echo "Accessing the home workspace:"
--- a/hack/flags.sh
+++ b/hack/flags.sh
@@ -81,6 +81,11 @@ parse_flags() {
   fi
   export ROOT_WORKSPACE=${ROOT_WORKSPACE:-"root"}
 
+  # Convert home ROOT_WORKSPACE to full path
+  if [ "$ROOT_WORKSPACE" == "~" ]; then
+    ROOT_WORKSPACE=$(KUBECONFIG=${KCP_KUBECONFIG} kubectl ws '~' --short)
+  fi
+
   # Check config files and version compatibility
   if kubectl version -o yaml --kubeconfig ${CLUSTER_KUBECONFIG} | yq '.serverVersion.gitVersion' | grep -q kcp; then
     echo CLUSTER_KUBECONFIG=${CLUSTER_KUBECONFIG} points to KCP not to cluster.
--- a/hack/configure-kcp.sh
+++ b/hack/configure-kcp.sh
@@ -104,13 +104,7 @@ EOF
 configure_service_provider_workspace() {
   echo "Creating and accessing '${SP_WORKSPACE_NAME}':"
   KUBECONFIG=${KCP_KUBECONFIG} kubectl ws ${ROOT_WORKSPACE}
-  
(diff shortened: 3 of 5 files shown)


hack/flags.sh
```diff
@@ -81,6 +81,11 @@ parse_flags() {
   fi
   export ROOT_WORKSPACE=${ROOT_WORKSPACE:-"root"}
 
+  # Convert home ROOT_WORKSPACE to full path
+  if [ "$ROOT_WORKSPACE" == "~" ]; then
+    ROOT_WORKSPACE=$(KUBECONFIG=${KCP_KUBECONFIG} kubectl ws '~' --short)
```
MatousJobanek, line 86: we need to export it, right?
```suggestion
    export ROOT_WORKSPACE=$(KUBECONFIG=${KCP_KUBECONFIG} kubectl ws '~' --short)
```

hack/create-user-workspace.sh
```diff
@@ -22,12 +22,8 @@ then
...
 fi
 
-echo "Accessing the home workspace:"
-kubectl ws '~'
-
-if [ "${ROOT_WORKSPACE}" == "~" ]; then
-  ROOT_WORKSPACE=$(kubectl ws . --short)
-fi
+echo "Accessing the root workspace:"
+kubectl ws $ROOT_WORKSPACE
```
MatousJobanek, line 26: we don't use `flags.sh` (nor preview.sh) in this script (yet). 
you can run create-user-workspace.sh and it should automatically create appstudio workspace in your home workspace and bind everything from `root:`

hack/create-user-workspace.sh
```diff
@@ -22,12 +22,8 @@ then
   export KUBECONFIG=${KCP_KUBECONFIG}
 fi
 
-echo "Accessing the home workspace:"
-kubectl ws '~'
```
MatousJobanek, line 26: We should create the appstudio/hacbs user workspace in a home workspace only. No matter what `ROOT_WORKSPACE` is set to. `ROOT_WORKSPACE` defines the place where service provider workspaces are created

Michkov: in that case we can maybe have `SP_ROOT_WORKSPACE` and `HOME_WORKSPACE`

MatousJobanek: I would say that we should create these workspaces always in a home workspace - this is also the expected behavior/location from the HAC point of view. If we need something else in the future, then we can improve it later.
As for `SP_ROOT_WORKSPACE` - yeah, we can rename it, why not.

Michkov: Expecting that e2e-tests will need this soon, since they will be using one account

flacatus: Agree, I already detect a issue in my  local installation:

```log
application.argoproj.io/all-components patched
application.argoproj.io/dev-application-api patched
application.argoproj.io/dev-application-service patched
application.argoproj.io/dev-authorization-appstudio patched
application.argoproj.io/dev-authorization-hacbs patched
application.argoproj.io/dev-enterprise-contract patched
application.argoproj.io/dev-spi patched
application.argoproj.io/spi-vault patched
Accessing the home workspace:
Current workspace is "root:users:uz:hm:rh-sso-redhat-appstudio-qe".
Creating & accessing AppStudio workspace 'appstudio':
Workspace "appstudio" (type root:universal) already exists.
Workspace "appstudio" (type root:universal) is ready to use.
Current workspace is "root:users:uz:hm:rh-sso-redhat-appstudio-qe:appstudio".
apibinding.apis.kcp.dev/pipeline-service unchanged
Error from server (Forbidden): error when creating "STDIN": apibindings.apis.kcp.dev "application-api" is forbidden: spec.reference.workspace.path: not pointing to an ancestor or child of an ancestor of "root:users:uz:hm:rh-sso-redhat-appstudio-qe:appstudio"
Error from server (Forbidden): error when creating "STDIN": apibindings.apis.kcp.dev "application-service" is forbidden: spec.reference.workspace.path: not pointing to an ancestor or child of an ancestor of "root:users:uz:hm:rh-sso-redhat-appstudio-qe:appstudio"
Error from server (Forbidden): error when creating "STDIN": apibindings.apis.kcp.dev "spi" is forbidden: spec.reference.workspace.path: not pointing to an ancestor or child of an ancestor of "root:users:uz:hm:rh-sso-redhat-appstudio-qe:appstudio"
Accessing the home workspace:
Current workspace is "root:users:uz:hm:rh-sso-redhat-appstudio-qe".
Creating & accessing AppStudio workspace 'hacbs':
Workspace "hacbs" (type root:universal) already exists.
Workspace "hacbs" (type root:universal) is ready to use.
Current workspace is "root:users:uz:hm:rh-sso-redhat-appstudio-qe:hacbs".
apibinding.apis.kcp.dev/pipeline-service unchanged
Error from server (Forbidden): error when creating "STDIN": apibindings.apis.kcp.dev "application-api" is forbidden: spec.reference.workspace.path: not pointing to an ancestor or child of an ancestor of "root:users:uz:hm:rh-sso-redhat-appstudio-qe:hacbs"
Error from server (Forbidden): error when creating "STDIN": apibindings.apis.kcp.dev "application-service" is forbidden: spec.reference.workspace.path: not pointing to an ancestor or child of an ancestor of "root:users:uz:hm:rh-sso-redhat-appstudio-qe:hacbs"
Error from server (Forbidden): error when creating "STDIN": apibindings.apis.kcp.dev "spi" is forbidden: spec.reference.workspace.path: not pointing to an ancestor or child of an ancestor of "root:users:uz:hm:rh-sso-redhat-appstudio-qe:hacbs"
Current workspace is "root:users:uz:hm:rh-sso-redhat-appstudio-qe:ci-kvzst".

```

hack/flags.sh
Michkov: not needed, it's already exported on line 82

hack/create-user-workspace.sh
Michkov: the script is called from preview.sh, so at least for that usecase it will be working, but could break current usecase. Created #778 for now which should be safe.

Michkov: Code updated to use '~' by default but allow to modify using HOME_WORKSPACE

hack/create-user-workspace.sh
Michkov: Keeping default '~' but allowing to override by `HOME_WORKSPACE`, in case that it will not be supported in future we can just remove it.

hack/create-user-workspace.sh
```diff
@@ -23,13 +24,13 @@ then
...
-
 if [ "${ROOT_WORKSPACE}" == "~" ]; then
   ROOT_WORKSPACE=$(kubectl ws . --short)
```
MatousJobanek, line 28: then we need to change this to:
```suggestion
  ROOT_WORKSPACE=$(kubectl ws '~' --short)
```

hack/create-user-workspace.sh
MatousJobanek: ok, fair enough :+1:

MatousJobanek (approved):

PR #774, opened 2022-09-29
preview - Update of identityhash code

Michkov:
Based on comments in #763

### DIFF
--- a/hack/preview.sh
+++ b/hack/preview.sh
@@ -95,16 +95,19 @@ echo "SPI configured"
 # evaluate APIExports pointers
 evaluate_apiexports() {
   APIEXPORT_FILES=$1
+  # Get existing APIExports in service provider namespace
   NEW_IDENTITY_HASHES=$(oc get apiexports.apis.kcp.dev -o jsonpath='{range .items[*]}{@.metadata.name}:{@.status.identityHash}{"\n"}{end}' --kubeconfig ${KCP_KUBECONFIG})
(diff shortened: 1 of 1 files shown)


amfred (approved):
lgtm

MatousJobanek:
How about the flow of entering the workspaces: https://github.com/redhat-appstudio/infra-deployments/pull/763#discussion_r982244296 shouldn't this be fixed as well?

plus which name of the application-service APIExport should we use? https://github.com/redhat-appstudio/infra-deployments/pull/763#discussion_r982274493 - is this the same as for https://github.com/redhat-appstudio/infra-deployments/pull/763#discussion_r983191555?

components/application-service/base/kustomization.yaml
```diff
@@ -13,3 +13,12 @@ namespace: application-service
...
   - ./manager_resources_patch.yaml
+
+# APIExport managed by infra-deployments for now
+patchesStrategicMerge:
+- |-
+  apiVersion: apis.kcp.dev/v1alpha1
+  kind: APIExport
+  metadata:
+    name: has
+  $patch: delete
```
MatousJobanek, line 24: Just asking, but why couldn't we delete the `has` APIExport from the application-service repo (for example as part of https://github.com/redhat-appstudio/application-service/pull/192)
I thought that our goal is to have one APIExport that would be used as the only source of truth. Or am I missing something?

Michkov: Removed the removing part from this PR, handled in #790  for has

PR #778, opened 2022-09-29
create-user-workspace: allow to define HOME_WORKSPACE

Michkov:
Allow to change user workspace by setting HOME_WORSKPACE env variable.


Michkov:
included in #773

PR #783, opened 2022-09-30
Script for resolving identityHashes

Michkov:
Resolve identity hashes from template.


hack/resolve-identityhashes.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+
+ROOT="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"/..
+
+if [[ -z ${2} ]]
+then
+  echo "Resolve identityhashes placeholders "
+  echo "     resolve-identityhashes.sh <kcp-stable|kcp-unstable> file-to-resolve"
```
MatousJobanek, line 8: I saw your PR in the application-service, and it's nice that it's included in the pipeline, but it makes any local changes (for dev/testing purposes) in the infra-deployments very hard.

My thought was that there would be a script executable directly from the infra-deployments for example:
```
generate-apiexport-overlays.sh ./components/application-service
```
that would take the path to the application folder and:
1. get the APIExport from the application-service directory by calling `kubectl kustomize <provided-path>/base` and extract the actual APIExport content by some fancy `sed` command
2. store the copy of the APIExport in `/dev` overlay
3. replace placeholders with kcp-stable identity hashes and store the modified APIExport in `/kcp-stable` overlay
4. replace placeholders with kcp-unstable identity hashes and store the modified APIExport in `/kcp-unstable` overlay

which means all the steps at once. That would make sure that everything is in sync even if it's called "locally". The same script could be used also in PR checks. And you could also run the same in the pipeline.

WDYT?

Michkov: Yop, that could work, I wanted to avoid using external tools (kubectl/kustomize) since they are not part of the image for the CI (https://github.com/redhat-appstudio/build-definitions/blob/main/tasks/update-infra-deployments.yaml#L17), but that could be updated.

Michkov: @MatousJobanek  Script rewritten to your proposal, https://github.com/redhat-appstudio/build-definitions/pull/212 created.

Michkov: Usage example -> https://github.com/redhat-appstudio/build-service/pull/65 (still need some updates here in infra-deployments after these changes are merged)

Michkov: Added generated resources by running script on https://github.com/redhat-appstudio/build-service/pull/65

hack/generate-apiexport-overlays.sh
```diff
@@ -0,0 +1,44 @@
...
+  echo "component dir must contain 'base' dir"
+fi
+
+TEMP=$(mktemp -d)
+kubectl kustomize $BASE_DIR | csplit -sf $TEMP/resource - /---/ '{*}'
+
+APIEXPORT_FILES=$(grep -lr 'kind: APIExport' $TEMP)
+if [ -n $APIEXPORT_FILES ]; then
+  cat $APIEXPORT_FILES > $COMPONENT_DIR/overlays/dev/apiexport.yaml
+fi
```
MatousJobanek, line 27: If the repository contains more than one APIExport, then this could fail. 
```suggestion
mapfile -t APIEXPORT_FILES < <(grep -lr 'kind: APIExport' $TEMP)
if [ ${#APIEXPORT_FILES[@]} != 1 ]; then
  echo "ERROR: the repository contains ${#APIEXPORT_FILES[@]} APIExports. One APIExport per component is supported!";
  exit 1
fi
cat ${APIEXPORT_FILES[0]} > $COMPONENT_DIR/overlays/dev/apiexport.yaml
```

MatousJobanek (approved):
This is cool :+1: 
Now, we need to:
* align all APIExports in the component repositories and regenerate the overlays here.
* modify the enterprise-contract HACBS component so it is supported by the script as well 
* Add a GH actions workflow that checks that everything is aligned

which can be done in separate PRs, of course :-), but there is one thing to add to this PR: update the REAME, please

README.md
```diff
@@ -12,31 +12,41 @@ You may use the `application-service` component as an example for how to add you
...
+1) Add an APIBinding:
    - For an AppStudio component to [apibindings/appstudio](apibindings/appstudio)
    - For an HACBS component to [apibindings/hacbs](apibindings/hacbs)
-8) Open a PR for all the above.
+1) Generate APIExports for overlays if component APIExports requires identityHashes:
+   - Execute `hack/generate-apiexport-overlays.sh components/(component-name)`
+   - The script generates `apiexport.yaml` files in `overlays`
+   - `overlays/dev/apiexport.yaml` contains APIExports from the component repository
+   - `overlays/kcp-stable/apiexport.yaml` replaces `identityHash: placeholder` by value in file `identity/kcp-stable/placeholder`
+   - `overlays/kcp-unstable/apiexport.yaml` replaces `identityHash: placeholder` by value in file `identity/kcp-stable/placeholder`
```
johnbieren, line 42: typo - should say by value in file with path to the unstable placeholder instead of stable

Michkov: updated, thanks

PR #784, opened 2022-09-30
fix: macos compatibility: sed

psturc:


hack/configure-kcp.sh
```diff
@@ -46,7 +46,7 @@ configure_compute_workspace() {
     echo "Creating SyncTarget..."
     KUBECONFIG=${KCP_KUBECONFIG} kubectl kcp workload sync ${SYNC_TARGET} --syncer-image ghcr.io/kcp-dev/kcp/syncer:main --resources=services,routes.route.openshift.io -o /tmp/${SYNC_TARGET}-syncer.yaml
     if grep -q "insecure-skip-tls-verify: true" ${KCP_KUBECONFIG}; then
-      sed -i 's/certificate-authority-data: .*/insecure-skip-tls-verify: true/' /tmp/${SYNC_TARGET}-syncer.yaml
+      sed -i.bak 's/certificate-authority-data: .*/insecure-skip-tls-verify: true/' /tmp/${SYNC_TARGET}-syncer.yaml && rm /tmp/${SYNC_TARGET}-syncer.yaml.bak
```
psturc, line 49: https://stackoverflow.com/a/22084103

hack/preview.sh
```diff
@@ -115,12 +115,12 @@ evaluate_apiexports() {
...
         echo Unknown IDENTITY_HASH for $REQUEST
       else
-        sed -i "s/\b$REQUEST\b/$IDENTITY_HASH/g" $APIEXPORTFILE
+        sed -i.bak "s/\b$REQUEST\b/$IDENTITY_HASH/g" $APIEXPORTFILE && rm "$APIEXPORTFILE.bak"
       fi
     done
   done
 }
-APIEXPORTS=$(find -name '*apiexport*.yaml' | grep overlays/dev)
+APIEXPORTS=$(find . -name '*apiexport*.yaml' | grep overlays/dev)
```
psturc, line 123: https://stackoverflow.com/a/17548983

MatousJobanek (approved):
The usage of `sed -i` as well as `sed -i.bak` might be tricky - IIRC some platforms might not support any of the versions at all. 
In some of my scripts I just forward the result of sed to a temp file and then copy it back - something like this:
```
sed ... /path/to/file > /tmp/result-of-sed && mv /tmp/result-of-sed /path/to/file
```

PR #785, opened 2022-09-30
Fix variable quoting in preview.sh for APIExports

johnbieren:


johnbieren:
@dirgim fyi
If you try deploying without this fix, the apiexports are all added to the hacbs workspace, creating duplicates (and issues because of that)

Michkov (approved):

PR #792, opened 2022-10-04
Fix variable quoting in preview.sh for APIExports

johnbieren:
The last PR wasn't rebased and reintroduced this bug, breaking hacbs in dev envs


johnbieren:
@Michkov

PR #794, opened 2022-10-05
preview fix: filter hacbs by folder components/hacbs

### LINKED ISSUE #793: preview.sh breaks if `hacbs` is in your absolute path
`ROOT="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"/..` gets the full path of your directory. If the directory is something like `/home/user/git_stuff/hacbs/infra-deployments`, then
https://github.com/redhat-appstudio/infra-deployments/blob/main/hack/preview.sh#L127..L131
will break because it finds `hacbs` in every APIExports' path

Michkov:
Do not look just for hacbs string but for components/hacbs since it's looking to absolute path.

Fixes #793


dirgim (approved):
I checked this fix locally and it lgtm

PR #803, opened 2022-10-06
manually get jvm build service cache mem fix into staging cluster

gabemontero:
manual alternative to https://github.com/redhat-appstudio/infra-deployments/pull/791

we need to get to 243a385d392cf79db71556c86a951dbb415213 to pick up cache mem leak fixes from @stuartwdouglas so that the PaC builds running on the staging cluster for jvm-build-service work more reliably, and we can more reliably get quay.io/redhat-appstudio image updates for changes after 243a385d392cf79db71556c86a951dbb415213

Per https://quay.io/repository/redhat-appstudio/hacbs-jvm-cache?tab=tags and https://quay.io/repository/redhat-appstudio/hacbs-jvm-build-request-processor?tab=tags there are tags for this level

Also FYI per https://github.com/redhat-appstudio/jvm-build-service/blob/243a385d392cf79db71556c86a951dbb4152139f/deploy/kcp/apiexport_jvmbuildservice.yaml the kcp api resource schema there matches what we currently have in infra-deployments for example https://github.com/redhat-appstudio/infra-deployments/blob/971fd91fb8dd510c3b7f3ebfca92a37c688b2477/components/hacbs/jvm-build-service/overlays/dev/apiexport.yaml

i.e. v202209270158

Lastly, the level we are moving from, 7d203d042f5ce4b49751fb9552cb71cab10ef757, does not include the mem leak fixes that culminated in https://github.com/redhat-appstudio/jvm-build-service/commit/ac09a788d20577ddb15b19728308317345b6565d

/assign @Michkov

### DIFF
--- a/components/hacbs/jvm-build-service/base/kustomization.yaml
+++ b/components/hacbs/jvm-build-service/base/kustomization.yaml
@@ -1,5 +1,5 @@
 resources:
-- https://github.com/redhat-appstudio/jvm-build-service/deploy/kcp?ref=7d203d042f5ce4b49751fb9552cb71cab10ef757
+- https://github.com/redhat-appstudio/jvm-build-service/deploy/kcp?ref=243a385d392cf79db71556c86a951dbb415213
 
 patchesStrategicMerge:
 - self-binding.yaml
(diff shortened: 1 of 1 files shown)


stuartwdouglas:
There has been a very large number of changes since the last pre-kcp update. I think this should work, but there is a chance that this will also need an update to build-definitions (e.g. there is no sidecar any more, but we still have the legacy endpoint used by the sidecar so this should still work, but it is not actively tested).

gabemontero:
forgot that staging cluster is still pre-kcp, to this PR is targeted to wrong branch

and @Michkov made a good point - will the corresponding build-definitions work with our later levels (probably not)

he is looking into moving the push secret so we can use the appstudio/base bundle

PR #808, opened 2022-10-08
Add initial support for running GitOps Service on KCP

jgwest:
Initial PR in support of [GITOPSRVCE-226](https://issues.redhat.com/browse/GITOPSRVCE-226)

This PR:
- Adopts the latest set of K8s resources from managed-gitops repo
- Adds the necessary component metadata for this new component
- Adds database secret generation to preview.sh
- Adds StatefulSet to the syncer, with a WARNING comment.

Known issues:
- Until we have virtual workspace support, the controllers will not be happy. This is part of another story
- Likewise, the Argo CD in gitops-service-argocd namespace isn't yet happy. This will be worked on as part of another story.


components/gitops/overlays/kcp-stable/apiexport.yaml
```diff
@@ -0,0 +1,36 @@
...
+kind: APIExport
+metadata:
+  name: gitops-appstudio-shared
+spec:
+  latestResourceSchemas:
+  - v202209151420.applicationpromotionruns.appstudio.redhat.com
+  - v202209151420.applicationsnapshotenvironmentbindings.appstudio.redhat.com
+  - v202209151420.applicationsnapshots.appstudio.redhat.com
+  - v202209151420.applications.appstudio.redhat.com
+  - v202209151420.environments.appstudio.redhat.com
```
Michkov, line 29: these schemas are provided by application-api

hack/preview.sh
```diff
@@ -86,6 +86,17 @@ oc --kubeconfig ${KCP_KUBECONFIG}  create -n spi-system secret generic shared-co
 rm $TMP_FILE
 echo "SPI configured"
 
+echo "start GitOps Service config"
+if ! kubectl get namespace gitops &>/dev/null; then
```
Michkov, line 90: must contain `--kubeconfig ${KCP_KUBECONFIG}`

hack/preview.sh
```diff
@@ -86,6 +86,17 @@ oc --kubeconfig ${KCP_KUBECONFIG}  create -n spi-system secret generic shared-co
...
+  KUBECONFIG=${KCP_KUBECONFIG} kubectl create namespace gitops
+fi
+if ! kubectl get secret -n gitops gitops-postgresql-staging &>/dev/null; then
```
Michkov, line 93: must contain `--kubeconfig ${KCP_KUBECONFIG}`

hack/configure-kcp.sh
```diff
@@ -44,7 +44,14 @@ configure_compute_workspace() {
...
   if [[ -z "$(kubectl get synctargets.workload.kcp.dev ${SYNC_TARGET} --kubeconfig ${KCP_KUBECONFIG} 2>/dev/null)" ]]; then
     echo "Creating SyncTarget..."
-    KUBECONFIG=${KCP_KUBECONFIG} kubectl kcp workload sync ${SYNC_TARGET} --syncer-image ghcr.io/kcp-dev/kcp/syncer:v0.9.0 --resources=services,routes.route.openshift.io -o /tmp/${SYNC_TARGET}-syncer.yaml
+    # CAUTION: '--resource=statefulsets.apps' is temporarily added as a synchronized resource to the syncer, until the GitOps 
+    # Service team moves Argo CD out of KCP. The KCP team does not support StatefulSets as of this writing (Oct 2022).
+    # 
+    # Until KCP adds support, StatefulSets should not be used by components except in the short term and in limited contexts,
+    # and with full understanding of the consequences. - @jgwest
+
+    KUBECONFIG=${KCP_KUBECONFIG} kubectl kcp workload sync ${SYNC_TARGET} --syncer-image ghcr.io/kcp-dev/kcp/syncer:v0.9.0 --resources=services,routes.route.openshift.io,statefulsets.apps -o /tmp/${SYNC_TARGET}-syncer.yaml
```
MatousJobanek, line 53: I thought that StatefulSets are not supported by kcp. :man_shrugging:

jgwest: This is the story for moving Argo CD off KCP: https://issues.redhat.com/browse/GITOPSRVCE-237

Once complete,  we will be able to remove statefulsets from the syncer.

MatousJobanek: hmm... I thought that the main goal was to run the ArgoCD instance(s) in a kcp workspace, not directly in an OSD cluster. So I'm surprised that you mentioned that you want to move it off kcp...
cc @alexeykazakov 

As for StatefulSet - when you did the experiment with running ArgoCD in a kcp workspace, you didn't mention anything related to unsupported StatefulSets, so this is new to me as well. 
Have you talked to kcp folks regarding this particular use-case of StatefulSet? If it's OK to go with that?
I'm slightly afraid of using something that is not supported by kcp, because we know that there can be some unexpected behavior even with the supported types - not sure what will happen with our syncer/workspace/wokload-cluster when using the unsupported ones.

MatousJobanek (commented):
btw, it looks like you didn't use the hack/generate-apiexport-overlays.sh script. 
you need to fix the APIExport on your side and then use the script to generate them here.

hack/configure-kcp.sh
jgwest: Discussed here: https://coreos.slack.com/archives/C02CTEB3MMF/p1662580655163479

jgwest: > I'm slightly afraid of using something that is not supported by kcp, 

We are as well, GITOPSRCE-237 will allow us to avoid this, by running Argo CD on workload cluster, rather than via KCP.

alexeykazakov: @MatousJobanek, yes, while stateful sets work well enough for ArgoCD atm they may break any moment. However because the work has already been done it was decided to proceed and add the already tested and working manifests to the infra-deployment to deploy ArgoCD via KCP. That will allow us to unblock development which requires a running ArgoCD. Meanwhile the team will work on https://issues.redhat.com/browse/GITOPSRVCE-237.

jgwest:
PR updated based on review comments: PR been updated to now use injection of identityHashes, refactored GitOps Service into two separate Argo CD Applications (the APIExport script did not like it when 2 APIExports in a single file, when one consumed the other), and general refactoring of resources.

MatousJobanek (approved):

PR #811, opened 2022-10-10
Template API Version update

bamachrn:
APIVersion for the template resource needs update as it is moved from default group


amfred (approved):
Tested this config on the staging cluster.

MatousJobanek (changes requested):
Could you please clarify why are we working on the pre-kcp version? This branch is frozen and only exceptional super-critical fixes should be merged there.
Is there any request that requires these changes?

PR #818, opened 2022-10-11
[WIP]Added prometheus and grafana templates to argoCD

bamachrn:
This PR
* Installs prometheus operator using operator hub
* configure prometheus secrets, configmaps, deployments using template 
* Configuring Servicemonitor
* Installs grafana operator using operator hub
* configure grafana crd
* Configuring grafana data sources


components/monitoring/base/grafanadatasource_prometheus.yaml
```diff
@@ -0,0 +1,55 @@
+apiVersion: integreatly.org/v1alpha1
+kind: GrafanaDataSource
+metadata:
+  name: prometheus-customer-grafanadatasource
+spec:
+  name: DevSandbox Monitor
```
amfred, line 6: We won't need the DevSandbox bits here.

components/monitoring/base/prometheus_servicemonitors.yaml
```diff
@@ -0,0 +1,163 @@
...
+      port: https
+    namespaceSelector:
+      matchNames:
+      - toolchain-host-operator
+    selector:
+      matchLabels:
+        control-plane: "controller-manager"
+        operators.coreos.com/toolchain-host-operator.toolchain-host-operator: ""
+
+#Sandbox Member Operator
```
amfred, line 31: We won't need this.

components/monitoring/base/grafanadashboard_sandbox_usage.yaml
```diff
@@ -0,0 +1,648 @@
+apiVersion: integreatly.org/v1alpha1
+kind: GrafanaDashboard
+metadata:
+  name: sandbox-usage-dashboard
```
amfred, line 4: We won't need this, but if you want a test dashboard, maybe the ones that Sergeii recommended would work.  If that's too much trouble for one PR, we can just start with a dashboard-less Grafana and add more dashboards in another PR.

amfred (changes requested):

components/monitoring/base/install_prometheus.yaml
```diff
@@ -0,0 +1,28 @@
...
+  spec:
+    channel: beta
+    installPlanApproval: Automatic
+    name: prometheus
+    source: community-operators
+    sourceNamespace: openshift-marketplace
+    startingCSV: prometheusoperator.0.56.3
+parameters:
+- name: MONITORING_NAMESPACE
+  value: workload-monitoring
```
xcoulon, line 28: I thought the namespace would be `openshift-user-workload-monitoring`?

argo-cd-apps/base/monitoring-workload-in-cluster.yaml
```diff
@@ -0,0 +1,34 @@
...
+spec:
+  project: workload-monitoring
+
+  source:
+    path: components/monitoring/base
+    repoURL: https://github.com/redhat-appstudio/infra-deployments.git
+    targetRevision: main
+
+  destination:
+    namespace: workload-monitoring
```
MatousJobanek, line 15: are you sure about the namespace name? why not `openshift-user-workload-monitoring`?

components/monitoring/base/install_prometheus.yaml
```diff
@@ -0,0 +1,28 @@
+apiVersion: template.openshift.io/v1
+kind: Template
+metadata:
+  name: install-prometheus-operator
+objects:
```
MatousJobanek, line 5: let's not use openshift template here

xcoulon: yes, we don't need Openshift templates here. If anything needs to be changed, it should be handled by Kustomize.

components/monitoring/base/install_prometheus.yaml
bamachrn: I did not want to put openshift here. because there are a lot of default openshift namespaces there which are mostly from the openshifts own components. but yes it is fine to use `openshift-user-workload-monitoring`

amfred: Argh, this note from the documentation about `openshift-user-workload-monitoring` (https://docs.openshift.com/container-platform/4.6/monitoring/enabling-monitoring-for-user-defined-projects.html) concerns me:

`You must have access to the cluster as a user with the cluster-admin role to enable monitoring for user-defined projects in OpenShift Container Platform. Cluster administrators can then optionally grant users permission to configure the components that are responsible for monitoring user-defined projects.`

And this is even more concerning:
https://docs.openshift.com/container-platform/4.6/monitoring/managing-metrics.html#managing-metrics 

`Only cluster administrators have access to the third-party UIs provided with OpenShift Container Platform Monitoring.`
`Developers can only use the Developer perspective and not the Administrator perspective. As a developer, you can only query metrics for one project at a time. Developers cannot access the third-party UIs provided with OpenShift Container Platform monitoring that are for core platform components. Instead, use the Metrics UI for your user-defined project.`

It sounds like it might be easier to grant our developers the access they need if we have our own Prometheus instance in our own namespace.  But what do you all think?

bamachrn: so what is the bes suggestion to use. I have added this to kustomization so that we can change in one place

xcoulon: > You must have access to the cluster as a user with the cluster-admin role to enable monitoring for user-defined projects in OpenShift Container Platform. Cluster administrators can then optionally grant users permission to configure the components that are responsible for monitoring user-defined projects.

My understanding is that cluster-admin privileges are needed when we want to create the `cluster-monitoring-config` ConfigMap in the `openshift-monitoring`, which will trigger the deployment of a second Prometheus instance in the `openshift-user-workload-monitoring` namespace. We may not need such privileges if we want to deploy Prometheus ourselves, but if using the `openshift-user-workload-monitoring` namespace causes troubles, then I guess that another (non `openshift-*`) namespace would be fine, too.

xcoulon: that being said, would it make sense to use something like `appstudio-`, `redhat-` or a combination of both as a prefix for such a namespace, if we decide not to use `openshift-user-workload-monitoring`? 
It's easier to identify and we could even have a rule (implemented by a webhook for example) to prevent users from creating namespaces with such a prefix.

argo-cd-apps/base/monitoring-workload-in-cluster.yaml
```diff
@@ -0,0 +1,34 @@
+apiVersion: argoproj.io/v1alpha1
+kind: Application
+metadata:
+  name: monitoring-workload-in-cluster
+
+spec:
+  project: workload-monitoring
```
Michkov, line 7: project must be `default`

components/monitoring/base/kustomization.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - install_prometheus.yaml
+  - prometheus-view.yaml
+
```
Michkov, line 6:
```
error validating data: ValidationError(OperatorGroup.spec.targetNamespaces): invalid type for com.coreos.operators.v1.OperatorGroup.spec.targetNamespaces: got "string", expected "array"
```
`one or more objects failed to apply (dry run)`
```suggestion
commonAnnotations:
  argocd.argoproj.io/sync-options: SkipDryRunOnMissingResource=true
```

components/monitoring/base/kustomization.yaml
```diff
@@ -0,0 +1,22 @@
...
+namespace: "openshift-user-workload-monitoring"
+patches:
+  - target:
```
Michkov, line 9: why to patch local files? Why not to modify them directly?

components/monitoring/base/install_prometheus.yaml
```diff
@@ -0,0 +1,23 @@
...
+apiVersion: operators.coreos.com/v1alpha1
+kind: Subscription
+metadata:
+  name: prometheus
+  namespace: openshift-user-workload-monitoring
+spec:
+  channel: beta
+  installPlanApproval: Automatic
+  name: prometheus
+  source: community-operators
```
Michkov, line 21: Don't we want to have supported environment? Community operator from beta channel does not seem to be right to me.

components/monitoring/base/kustomization.yaml
Michkov: I had to remove these patches to be able to run this.

components/monitoring/base/kustomization.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - install_prometheus.yaml
```
Michkov, line 4: Installation failed with -> `Failed, InterOperatorGroupOwnerConflict, intersecting operatorgroups provide the same apis`

argo-cd-apps/base/monitoring-workload-in-cluster.yaml
```diff
@@ -0,0 +1,34 @@
+apiVersion: argoproj.io/v1alpha1
```
Michkov, line 1: Create also record in argo-cd-apps/overlays/development/repo-overlay.yaml:
```
kind: Application
metadata:
  name: monitoring-workload-in-cluster
spec:
  source:
    targetRevision: main
```

Michkov (changes requested):

argo-cd-apps/base/monitoring-workload-in-cluster.yaml
Michkov: Currently only default project is used. I am ok with creation of new projects, but that must be created probably in bootstrap script, not sure if can be created by ArgoCD using `default` project.

components/monitoring/base/install_prometheus.yaml
MatousJobanek: Let's not use the `redhat-` prefix - namespaces with this prefix have the same limitation as with `openshift-`. IIRC, you cannot even create such a namespace with dedicated-admin privileges.
So, if you want to deploy it to another namespace, then let's use the `appstudio-` prefix for that.

> It's easier to identify and we could even have a rule (implemented by a webhook for example) to prevent users from creating namespaces with such a prefix.

that's not needed - all namespaces in the cluster are created either via infra-deployments repo or via kcp syncer. There is no risk of facing any namespace name conflicts.

Anyway, I'll leave the decision about the location up to @xcoulon :wink:

xcoulon: ah right, I forgot about the `redhat-` prefix which is also reserved!

bamachrn: lets use `appstudio-workload-monitoring`

components/monitoring/base/kustomization.yaml
bamachrn: @Michkov I am curious, how are you testing this? 
 `$kubectl kustomize .` generates the yamls properly here. any ideas on how should I test them?

components/monitoring/base/install_prometheus.yaml
xcoulon: ok, let's use that for now. Anyways, it should be easily changed with Kustomize

components/monitoring/base/kustomization.yaml
bamachrn: I wanted to patch them through kustomization.yaml to make sure, in future if we plan to change the namespace we dont need change it every where. We are planning to add a lot of configs here for single change updating all the files does not make much sense to me.

components/monitoring/base/kustomization.yaml
Michkov: having dev cluster, using preview mode to deploy on it. In this case it should be possible to reproduce by creation of argoCD app.

components/monitoring/base/install_prometheus.yaml
bamachrn: Prometheus is available in beta channel from comunity operators only in OSD. I dont see much option there. And I believe we can trust the operators published in openshift operatorhub :thinking:

argo-cd-apps/base/monitoring-workload-in-cluster.yaml
bamachrn: done

argo-cd-apps/base/monitoring-workload-in-cluster.yaml
bamachrn: done

components/monitoring/base/kustomization.yaml
Michkov: cannot see anymore

components/monitoring/base/install_prometheus.yaml
```diff
@@ -0,0 +1,23 @@
...
+metadata:
+  name: prometheus
+  namespace: appstudio-workload-monitoring
+spec:
+  channel: beta
+  installPlanApproval: Automatic
+  name: prometheus
+  source: community-operators
+  sourceNamespace: openshift-marketplace
+  startingCSV: prometheusoperator.0.56.3
```
Michkov, line 23: do not use startingCSV that will always install this one and will be updating to newest one. Without this line it will install newest one directly.

bamachrn: make sense. done

Michkov (approved):

components/monitoring/base/kustomization.yaml
MatousJobanek: if we wanted to change the namespace name in the future, then it would make sense to change it in the local files as well - having something different in the patches and in the local files doesn't make any sense and it just brings confusion. 
Since the files are located in the same folder, in the same repository, and without having additional overlays, then using patches doesn't have any reasoning.

MatousJobanek (changes requested):
@bamachrn could you please drop the patches from the kustomize? https://github.com/redhat-appstudio/infra-deployments/pull/818/files#r996785896

argo-cd-apps/overlays/development/repo-overlay.yaml
```diff
@@ -179,4 +179,13 @@ spec:
...
         targetRevision: main
+---
+apiVersion: argoproj.io/v1alpha1
+kind: Application
+metadata:
+  name: monitoring-workload-in-cluster
+spec:
+  source:
+    repoURL: https://github.com/redhat-appstudio/infra-deployments.git
+    targetRevision: main
```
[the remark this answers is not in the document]
amfred: Actually, no.  We only want this for stage and prod.

components/monitoring/base/install_prometheus.yaml
amfred: I agree with Bama here - it's the only option I found.

components/monitoring/base/kustomization.yaml
bamachrn: ok, agree with the confusion part. but we need to update in so many places. It may be one time and never happen also. removing it.

argo-cd-apps/overlays/development/repo-overlay.yaml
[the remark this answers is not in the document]
bamachrn: This is why I did not add it initially, but thought we may need it in dev. Yes it makess sense to not add it to local setup.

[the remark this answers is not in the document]
Michkov: Ok, in that case it must be added to https://github.com/redhat-appstudio/infra-deployments/blob/main/argo-cd-apps/overlays/development/delete-applications.yaml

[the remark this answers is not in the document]
bamachrn: ah! yes

components/monitoring/base/install_prometheus.yaml
amfred: Available Prometheus monitors as of now:
image
The AWS one sends it to their Prometheus service ($$), and the others don't actually deploy Prometheus.

amfred (approved):
/lgtm

PR #820, opened 2022-10-11
Add ApplicationSnapshots into build-service APIExport

mmorhun:
Signed-off-by: Mykola Morhun <mmorhun@redhat.com>

### DIFF
--- a/components/build-service/overlays/dev/apiexport.yaml
+++ b/components/build-service/overlays/dev/apiexport.yaml
@@ -25,3 +25,6 @@ spec:
   - group: appstudio.redhat.com
     identityHash: application-api
     resource: applications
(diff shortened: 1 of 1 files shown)


components/build-service/overlays/dev/apiexport.yaml
```diff
@@ -25,3 +25,6 @@ spec:
   - group: appstudio.redhat.com
     identityHash: application-api
     resource: applications
+  - group: appstudio.redhat.com
+    identityHash: application-api
+    resource: applicationsnapshots
```
Michkov, line 30: should be added here -> https://github.com/redhat-appstudio/build-service/blob/main/config/kcp/build-service-apiexport.yaml then it will create PR here

PR #825, opened 2022-10-12
Connect Syncer to Local  CKCP via internal service

jduimovich:
This PR contains the following changes.

A new option CKCP_SYNCER_USE_INTERNAL_SERVICE  in preview mode will allow the syncer to connect to ckcp via the local kubernetes service instead of the public route. 
This is required when the local kube (CRC) has an IP of 127.0.0.1 as the syncer will try to connect to the public API endpoint and cannot reach it.

There is a new utility that warns if new vars are found in preview-template and not in preview.env so than when updates are made end users can easily know they may need to resolve their their preview.env files .

### DIFF
--- a/components/ckcp/certs.yaml
+++ b/components/ckcp/certs.yaml
@@ -74,5 +74,6 @@ spec:
   dnsNames:
     - localhost
     - $HOSTNAME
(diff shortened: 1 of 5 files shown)


dirgim (approved):
/lgtm

PR #827, opened 2022-10-12
PLNSRVCE-762: add IMAGE_TAG,IMAGE_REPO env vars to jvm bld svc

gabemontero:
@Michkov I was able to verify this with https://github.com/redhat-appstudio/jvm-build-service/pull/313 

With these, I was able to get jvm build service "demo'able" on KCP per my recent messages in the team's slack channel and [PLNSRVCE-762](https://issues.redhat.com//browse/PLNSRVCE-762)

That said, I'm guessing this needs changes directed by you, and I'll add some disclaimers / questions as in line comments below to confirm/deny that

@stuartwdouglas FYI

### DIFF
--- a/components/hacbs/jvm-build-service/base/kustomization.yaml
+++ b/components/hacbs/jvm-build-service/base/kustomization.yaml
@@ -4,6 +4,17 @@ resources:
 patchesStrategicMerge:
 - self-binding.yaml
 
+patches:
+  - patch: |-
+      - op: add
+        path: /spec/template/spec/containers/0/env/-
+        value:
+          name: IMAGE_TAG
+          value: b911059a17e1efdbc2ddb496604e673d0c491c43
+    target:
+      kind: Deployment
+      name: hacbs-jvm-operator
+
 images:
 - name: hacbs-jvm-operator
   newName: quay.io/redhat-appstudio/hacbs-jvm-controller
--- a/components/hacbs/jvm-build-service/base/kustomization.yaml
+++ b/components/hacbs/jvm-build-service/base/kustomization.yaml
@@ -1,5 +1,5 @@
 resources:
-- https://github.com/redhat-appstudio/jvm-build-service/deploy/kcp?ref=7d3e4b05319ef1584abbab1b9fa36c7072b685c8
+- https://github.com/redhat-appstudio/jvm-build-service/deploy/kcp?ref=6dfe2aa04bd48604d962fba24f542668428ae719
 
 patchesStrategicMerge:
 - self-binding.yaml
@@ -10,7 +10,7 @@ patches:
         path: /spec/template/spec/containers/0/env/-
         value:
           name: IMAGE_TAG
-          value: b911059a17e1efdbc2ddb496604e673d0c491c43
+          value: 6dfe2aa04bd48604d962fba24f542668428ae719
     target:
       kind: Deployment
       name: hacbs-jvm-operator
@@ -18,7 +18,7 @@ patches:
 images:
 - name: hacbs-jvm-operator
   newName: quay.io/redhat-appstudio/hacbs-jvm-controller
-  newTag: 7d3e4b05319ef1584abbab1b9fa36c7072b685c8
+  newTag: 6dfe2aa04bd48604d962fba24f542668428ae719
 
 namespace: jvm-build-service
 
--- a/components/hacbs/jvm-build-service/overlays/dev/kustomization.yaml
+++ b/components/hacbs/jvm-build-service/overlays/dev/kustomization.yaml
@@ -5,3 +5,14 @@ resources:
 
 patchesStrategicMerge:
   - apiexport.yaml
+
+patches:
+  - patch: |-
+      - op: add
+        path: /spec/template/spec/containers/0/env/-
+        value:
+          name: IMAGE_REPO
+          value: redhat-appstudio
+    target:
+      kind: Deployment
+      name: hacbs-jvm-operator


components/hacbs/jvm-build-service/base/kustomization.yaml
```diff
@@ -4,6 +4,26 @@ resources:
 patchesStrategicMerge:
 - self-binding.yaml
 
+patches:
+  - patch: |-
+      - op: add
+        path: /spec/template/spec/containers/0/env/-
+        value:
+          name: IMAGE_TAG
```
gabemontero, line 12: I intentionally did not institute the pre KCP flexibility we had with per image overrides in preview.env

Rationale: 
- even for the jvm bld svc dev flow, we only need to override the repo and tag to handle all the images 
- I'm still holding out hope I can switch back to the controller being able to construct the image refs for the cache and req processor (we no longer have sidecars or dep analyser usage) by the controller examining the image ref for its pod

components/hacbs/jvm-build-service/base/kustomization.yaml
```diff
@@ -4,6 +4,26 @@ resources:
...
+          value: b911059a17e1efdbc2ddb496604e673d0c491c43
```
gabemontero, line 13: If I ever caught how the appstudio staging bot knows what fields in this file it needs to update when jvm build service repo PRs merge, I've lost it.

wherever that intelligence is, does it need additional modification to update this field as well?

thanks

components/hacbs/jvm-build-service/base/kustomization.yaml
gabemontero: depending on how ^^ pans out, my thinking is we can always add those preview.env related override later on

components/hacbs/jvm-build-service/base/kustomization.yaml
Michkov: https://github.com/redhat-appstudio/jvm-build-service/blob/main/.tekton/controller-push.yaml#L23-L26 , unfortunately it can use only basic linux commands, so sed is useful but it's not using context -> hard to apply when you have multiple lines.

components/hacbs/jvm-build-service/base/kustomization.yaml
```diff
@@ -4,6 +4,26 @@ resources:
...
+    target:
+      kind: Deployment
+      name: hacbs-jvm-operator
+  - patch: |-
+      - op: add
+        path: /spec/template/spec/containers/0/env/-
+        value:
+          name: IMAGE_REPO
+          value: redhat-appstudio
```
Michkov, line 22: what is this IMAGE_REPO?

components/hacbs/jvm-build-service/base/kustomization.yaml
gabemontero: gotcha / bummer

fingers crossed we can munge some bash that works

thanks for the pointer

components/hacbs/jvm-build-service/base/kustomization.yaml
gabemontero: yeah this one is not "required" but rather a "nice to have" wrt to the development flow that @stuartwdouglas created for us

in that development flow, `redhat-appstudio` as the repository for the image ref is changed to the developer's personal repository on `quay.io` (i.e. `gabemontero` for me).

so certainly the appstudio staging bot will not need to change this as jvm-build-service PRs merged.

but in my infra-deployment branch I'm using for development I've changed the value of this env var to `gabemontero`

components/hacbs/jvm-build-service/base/kustomization.yaml
Michkov: it is available

components/hacbs/jvm-build-service/base/kustomization.yaml
Michkov: so this patch would be needed only in dev overlay and should be modified by preview.sh?

gabemontero: ok I've pushed that update @Michkov PTAL

Michkov: So could you rebase to newest version and remove changes in apiexport?

gabemontero: ok I've rebased this PR to latest infra-deployments main as of this comment
and I've updated the jvm-build-service commit sha to the last valid one from 7:52 AM US Eastern

that said, per the slack thread https://coreos.slack.com/archives/C03KN2B47GU/p1666189974098079 I pinged you on @Michkov something unfortunately seems wrong with my update https://github.com/gabemontero/jvm-build-service/blob/71d073569161dd3c1e4492f525cac748f84fc037/.tekton/controller-push.yaml and the controller on push pipelinerun is not getting generated on the staging cluster.  As a result, there is no controller quay image for commit 040c0513e26a05461e3be76db4835542dfa962de

I'm trying to reproduce somehow while I wait to hear from you.

But in the meantime we need to 

/hold

this PR until that problem is fixed.

gabemontero: oh, on remove the apiexport changes ... not sure I follow.  I've changed what jvm-build-service permission claims need to be, so the apiexport changes here reflect that.

gabemontero: false alarm on the controller push job ....PR 332 from jvm-build-service lead to https://console-openshift-console.apps.appstudio-stage.x99m.p1.openshiftapps.com/k8s/ns/jvm-build-service/tekton.dev~v1beta1~PipelineRun/jvm-build-controller-on-push-qzhhk succeeding 

/hold cancel

gabemontero: also https://github.com/redhat-appstudio/infra-deployments/pull/842 is updated correctly

gabemontero: and re apiexports I think I see now that https://github.com/redhat-appstudio/infra-deployments/pull/842/files is making the same edits as in this PR, since I made those updates in jvm-build-service

gabemontero: apiexport changes reverted PTAL

Michkov (approved):

PR #840, opened 2022-10-14
add Syncer manifests

MatousJobanek:
* adds the syncer manifests used in production
* they are not deployed in preview mode
* they don't contain the kubeconfig secrets


argo-cd-apps/base/syncer.yaml
```diff
@@ -0,0 +1,43 @@
+apiVersion: argoproj.io/v1alpha1
+kind: ApplicationSet
+metadata:
+  name: syncer
```
Michkov, line 4: syncer is very generic, proposing to change to `kcp-syncer` (everywhere, filenames, components/)

components/syncer/kcp-unstable/appstudio-internal-syncer.yaml
```diff
@@ -0,0 +1,221 @@
...
+        args:
+        - dns
+        - start
+        env:
+        - name: NAMESPACE
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.namespace
+        image: ghcr.io/kcp-dev/kcp/syncer:main
+        imagePullPolicy: IfNotPresent
```
Michkov, line 197: should be Always or we should bind to specific tag.

MatousJobanek: good catch - forgot about the second deployment 3c341b2

PR #848, opened 2022-10-19
Improve scripts to end when subcommand fails

Michkov:
- exit script when subcommand fails
- improve text output

### DIFF
--- a/hack/flags.sh
+++ b/hack/flags.sh
@@ -59,6 +59,10 @@ parse_flags() {
 
   if [ "${MODE}" == "preview" ]
   then
+    if [ "${PARSE_FLAGS_RUN}" == "1" ]; then
+      # parse_flags was already called so we do not need to proceed again
+      return
+    fi
     if [[ -n ${KCP_KUBECONFIG_FLAG}${CLUSTER_KUBECONFIG_FLAG} ]]
     then
       echo "ERROR: You cannot use the parameter --kcp-kubeconfig nor --cluster-kubeconfig in preview mode - use only the './hack/preview.env' file" >&2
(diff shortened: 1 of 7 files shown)


hack/flags.sh
```diff
@@ -59,6 +59,9 @@ parse_flags() {
 
   if [ "${MODE}" == "preview" ]
   then
+    if [ "${PARSE_FLAGS_RUN}" == "1" ]; then
+      return
```
MatousJobanek, line 63: could you add a comment to explain why it returns - it doesn't have to be obvious for the first glance.

hack/install-ckcp.sh
```diff
@@ -18,19 +18,27 @@ if [ "$1" == "--destroy" ]; then
   oc delete project spi-vault
 fi
 
+set -e
+
```
MatousJobanek, line 22: we could move it to the very top of the script

hack/util-patch-spi-config.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/bash -e
```
MatousJobanek, line 1: let's keep the scripts consistent - either use this one or the `set -e` one

hack/install-ckcp.sh
Michkov: moved and disabled for destroy part

hack/util-patch-spi-config.sh
Michkov: updated

hack/util-set-github-org
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/bash -e
```
MatousJobanek, line 1:
```suggestion
#!/bin/bash

set -e
```

MatousJobanek (approved):

PR #852, opened 2022-10-19
Quicklab: fix directory removal

Michkov:
- fix typo in param
- change destination dir to be only one dir otherwise it's not deleted with the PVC deletion


lcarva:
> * change destination dir to be only one dir otherwise it's not deleted with the PVC deletion

Oh wow! I didn't realize that's why it wasn't working. Thanks!

lcarva (approved):

PR #858, opened 2022-10-20
Add kcp stable/unstable identity hashes for enterprise-contract

johnbieren:


johnbieren:
@Michkov the filename should be the apiexport name, right? Not the directory name in `components`?

Michkov:
@johnbieren yes, let's name it by apiexport name. Currently it's not needed since no other APIExport is using enterprisecontract APIExport. But it's ok to add it

johnbieren:
> @johnbieren yes, let's name it by apiexport name. Currently it's not needed since no other APIExport is using enterprisecontract APIExport. But it's ok to add it

Okay, this should be correct then https://github.com/hacbs-contract/enterprise-contract-controller/blob/main/config/kcp/apiexport_enterprisecontract.yaml#L4

No one currently uses their APIExport, but one of my colleagues from release-service is making a change to release-service that has us using their APIs, so not having the hashes will block him (hence my pr)

Michkov (approved):

PR #849, opened 2022-10-19
Add prometheus CR to deploy prometheus pod and rules

bamachrn:
* In this PR
   * Prometheus CR is deployed with github oauth and openshift oauth
   * Prometheus Rule for scrape selection is added
   
* This PR assumes the `prometheus-proxy-config` is already available in the cluster with required value

PR #859, opened 2022-10-20
feat: deploy Grafana using operator

xcoulon:


components/monitoring/base/grafana-app.yaml
```diff
@@ -0,0 +1,148 @@
...
+    auth:
+      disable_login_form: False
+      disable_signout_menu: True
+    auth.basic:
+      enabled: False
+    auth.proxy:
+      enabled: True
+      enable_login_token: True
+      header_property: username
+      header_name: X-Forwarded-User
```
bamachrn, line 58: How about allowing the users to have edit access to create their own dashboard?

components/monitoring/base/grafana-app.yaml
```diff
@@ -0,0 +1,148 @@
...
+      - name: OAUTH2_PROXY_CLIENT_SECRET
+        valueFrom:
+          secretKeyRef:
+            key: client-secret
+            name: grafana-oauth2-proxy-config
+      - name: OAUTH2_PROXY_COOKIE_SECRET
+        valueFrom:
+          secretKeyRef:
+            key: cookie-secret
+            name: grafana-oauth2-proxy-config
```
bamachrn, line 94: Let's make sure we are adding this secret to README as well

bamachrn:
@xcoulon  this needs clean up. seems like changes from multiple places came together

components/monitoring/base/configure-prometheus.yaml
```diff
@@ -0,0 +1,223 @@
+---
```
xcoulon, line 1: this file will be removed from this PR once https://github.com/redhat-appstudio/infra-deployments/pull/849 is merged (but I need it for now to test the connection to Prometheus)

xcoulon:
> @xcoulon this needs clean up. seems like changes from multiple places came together

yes, I don't know what happened here, but it should be better now that I pushed https://github.com/redhat-appstudio/infra-deployments/commit/7a80a788e8479a721c6aab9ef52f096c2957586c 🤷‍♂️

components/monitoring/base/grafana-app.yaml
xcoulon: As already discussed on slack (IIRC), I don't think we need something like this for now.

components/monitoring/base/grafana-app.yaml
xcoulon: this should be covered by https://github.com/redhat-appstudio/infra-deployments/pull/883

amfred (approved):

MatousJobanek (approved):

PR #861, opened 2022-10-20
chore: adds required release-service CRD permissionclaims

theflockers:
- adds EnterpriseContractPolicy CRD permissionclaims to release-service kcp overlays.

### DIFF
--- a/components/hacbs/release-service/overlays/dev/apiexport.yaml
+++ b/components/hacbs/release-service/overlays/dev/apiexport.yaml
@@ -26,6 +26,9 @@ spec:
   - group: appstudio.redhat.com
     identityHash: application-api
     resource: applicationsnapshotenvironmentbindings
+  - group: appstudio.redhat.com
+    identityHash: enterprisecontract
+    resource: enterprisecontractpolicies
   - group: tekton.dev
     identityHash: pipeline-service
     resource: pipelineruns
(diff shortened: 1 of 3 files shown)


theflockers:
> I believe this will be done by our tekton push job in our repo once your draft PR merges

I see. Just changed that one to _ready to review_. I will change this to draft and close it if required.

parasense (approved):
I do not see any glaring issue here. I am curious how one would generate those hashes, and why they are different for kcp-{stable,unstable} ? Do we know what hashing algorithm is being used? Looks like ~64'ish characters, so maybe SHA-512 ? 

Regardless, LGTM :+1:

johnbieren:
> I do not see any glaring issue here. I am curious how one would generate those hashes, and why they are different for kcp-{stable,unstable} ? Do we know what hashing algorithm is being used? Looks like ~64'ish characters, so maybe SHA-512 ?
> 
> Regardless, LGTM +1

I am not sure of the hashing algorithm but basically when you create an APIExport, it gets an identityHash similar to a uuid... tmbk it is not programatically created. That is why it is different for kcp-stable and kcp-unstable, as they are two different instances so two different APIExports that each got their own hash (or uuid as I think of it)

MatousJobanek:
As for the question for generating the actual identity and the identity hash out of it:
https://github.com/kcp-dev/kcp/blob/c5c5e3c20306700274b74a93e922edeaee177d57/pkg/reconciler/apis/apiexport/crypto.go

theflockers:
Already done automatically by the push job. closing it.

PR #865, opened 2022-10-21
introduce appstudio & hacbs CWTs

MatousJobanek:
* introduce appstudio & hacbs CWTs
* there are some parts put into a comment - it's a preparation for environment sub-workspaces. 
* they are empty for now - creation of the APIBindings is still done via the script. Will be replaced as soon as this is done https://issues.redhat.com/browse/ASC-156


davidmogar (approved):
LGTM.

PR #870, opened 2022-10-25
Fix permissions

Katka92:
Add RoleBinding with view ClusterRole. Add a file with those permissions to kustomization.yaml.

### DIFF
--- a/components/authentication/pact-broker.yaml
+++ b/components/authentication/pact-broker.yaml
@@ -23,6 +23,9 @@ subjects:
   - kind: User
     apiGroup: rbac.authorization.k8s.io
     name: Katka92
+  - kind: Group 
+    apiGroup: rbac.authorization.k8s.io
+    name: QE team
 roleRef:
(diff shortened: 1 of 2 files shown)


components/authentication/pact-broker.yaml
```diff
@@ -27,3 +27,17 @@ roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
   name: pact-broker-maintainer
+---
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: pact-broker-maintainers
```
MatousJobanek, line 34: it would conflict with the other ClusterRoleBinding
```suggestion
  name: pact-broker-maintainers-view
```

MatousJobanek (approved):

components/authentication/pact-broker.yaml
Katka92: Thank you

rhopp (approved):

PR #871, opened 2022-10-25
downscale the unstable syncer to zero

MatousJobanek:
there seems to be a problem with running two syncers in the same cluster at the same time - let's downscale the unstable syncer to zero for now


Michkov (approved):

gabemontero:
see https://coreos.slack.com/archives/C032EJ007C0/p1666622368784409 for tl;dr

PR #882, opened 2022-11-02
[WIP] Add prometheus servicemonitors for scrapping services

bamachrn:
* This PR adds servicemonitors to prometheus
* Service selection is based on the Labels not on namespaces

### DIFF
--- a/components/monitoring/base/prometheus-servicemonitors.yaml
+++ b/components/monitoring/base/prometheus-servicemonitors.yaml
@@ -1,0 +1,46 @@
+---
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  namespace: appstudio-workload-monitoring
+  name: prometheus
+  labels:
+    prometheus: appstudio-kcp-workload
+spec:
+  endpoints:
+  - bearerTokenSecret:
+      key: ""
+    interval: 15s
+    port: oauth2-proxy
+    scheme: HTTPS
+    tlsConfig:
+      caFile: /etc/prometheus/tls/tls.crt # volume mounted in the `prometheus` container of the prometheus pods
+      serverName: "prometheus-oauth2.appstudio-workload-monitoring.svc"
+  namespaceSelector:
+    matchNames:
+    - appstudio-workload-monitoring
+  selector:
+    matchLabels:
+      app.kubernetes.io/instance: prometheus
+---
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  namespace: appstudio-workload-monitoring
+  name: spi-operator
+  labels:
+    prometheus: appstudio-kcp-workload
+spec:
+  endpoints:
+    - bearerTokenSecret:
+        name: prometheus-k8s-token-x6qjc
+        key: token
+      scheme: https
+      tlsConfig:
+        insecureSkipVerify: true
+      port: https
+  namespaceSelector: {}
+  selector:
+    matchLabels:
+      control-plane: "controller-manager"
+      app.kubernetes.io/part-of: "service-provider-integration-operator"


components/monitoring/base/prometheus-servicemonitors.yaml
```diff
@@ -0,0 +1,46 @@
...
+kind: ServiceMonitor
+metadata:
+  namespace: appstudio-workload-monitoring
+  name: spi-operator
+  labels:
+    prometheus: appstudio-kcp-workload
+spec:
+  endpoints:
+    - bearerTokenSecret:
+        name: prometheus-k8s-token-x6qjc
```
xcoulon, line 36: where does this secret come from, and could we have a shorter name instead? Eg: `prometheus-k8s-token`

components/monitoring/base/prometheus-servicemonitors.yaml
```diff
@@ -0,0 +1,46 @@
...
+kind: ServiceMonitor
+metadata:
+  namespace: appstudio-workload-monitoring
+  name: prometheus
+  labels:
+    prometheus: appstudio-kcp-workload
+spec:
+  endpoints:
+  - bearerTokenSecret:
+      key: ""
```
xcoulon, line 12: why is it empty?

components/monitoring/base/prometheus-servicemonitors.yaml
```diff
@@ -0,0 +1,46 @@
...
+spec:
+  endpoints:
+    - bearerTokenSecret:
+        name: prometheus-k8s-token-x6qjc
+        key: token
+      scheme: https
+      tlsConfig:
+        insecureSkipVerify: true
+      port: https
+  namespaceSelector: {}
```
xcoulon, line 42: could we label the `appstudio` namespaces to avoid scanning _all_ namespaces?

bamachrn: we want to scan all the namespaces here. Because we dont know about the generated namespace name by ArgoCD.

components/monitoring/base/prometheus-servicemonitors.yaml
bamachrn: This is the token used by the `prometheus-k8s` SA, I am not sure how to modify this secret name. But we should be able to decide this name on run time, I have a command for that. I am trying to figure out, how to do it using Argo.

components/monitoring/base/prometheus-servicemonitors.yaml
bamachrn: this is local to prometheus installation, we are accessing this through service and cert file directly.

components/monitoring/base/prometheus-servicemonitors.yaml
xcoulon: I know we can't control the name of the generated namespaces, so could we add labels on these namespaces?

components/monitoring/base/prometheus-servicemonitors.yaml
xcoulon: I thought oauth2-proxy needed a token in the requests? 🤔

components/monitoring/base/prometheus-servicemonitors.yaml
xcoulon: I thought we agreed that for now, all secrets could be generated by a script, which is why I wonder why you have such a name here

bamachrn: this is not manually geerated. This is generated by prometheus operator/CR. so we dont need anything for generating this.

components/monitoring/base/prometheus-servicemonitors.yaml
bamachrn: :thinking: this works in local, and the pre-kcp stage. I will re-verify this.

components/monitoring/base/prometheus-servicemonitors.yaml
bamachrn: thats a great idea @MatousJobanek any thoughts on this please?

amfred: I love the idea of adding an appstudio label to all of "our" namespaces, but I don't know how to do it.

MatousJobanek: as of now, (in kcp environment) there is no way to add some labels to `kcp-*` namespaces. 
Instaed of using ServiceMonitors, we should use PodMonitors because we can control labels at Pods

xcoulon: > Instead of using ServiceMonitors, we should use PodMonitors because we can control labels at Pods

how does it defer from labels on services?

bamachrn: let me look into how prometheus supports the podmonitor. then also the issue remains same we need to grep through all the namespaces in the cluster.

components/monitoring/base/prometheus-servicemonitors.yaml
```diff
@@ -0,0 +1,125 @@
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  namespace: appstudio-workload-monitoring
```
MatousJobanek, line 4: As discussed in a slack thread - ServiceMonitors should be ideally stored in a component namespace:
https://coreos.slack.com/archives/CHK0J6HT6/p1670234963843059?thread_ts=1669991703.870609&cid=CHK0J6HT6

PR #883, opened 2022-11-02
add script to create secrets for Prometheus and Grafana

xcoulon:
the `./hack/setup-monitoring.sh` script can be used to:
- create the `prometheus-oauth2-proxy` and `grafana-oauth2-proxy` secrets in the `appstudio-workload-monitoring` namespace. These secrets are used by the OAuth2 proxies in Prometheus and Grafana to restrict access to the UIs to the members of the [Red Hat Appstudio SRE organization](https://github.com/organizations/redhat-appstudio-sre)
- create secrets for Grafana datasources, so it can connect to remote Prometheus instances using a token associated with the `grafana` service account created in the `appstudio-workload-monitoring` namespace where Prometheus is deployed.


bamachrn:
Lets add a README update as well. For how to run the script and  that we need to have these env variables set in local machine before running the script.

bamachrn:
we can re-use https://github.com/redhat-appstudio/infra-deployments/blob/pre-kcp/hack/monitoring-template.env

xcoulon:
> Lets add a README update as well. For how to run the script and that we need to have these env variables set in local machine before running the script.

done in 678487d

xcoulon:
> we can re-use https://github.com/redhat-appstudio/infra-deployments/blob/pre-kcp/hack/monitoring-template.env

what would be the benefit? To be honest, I already have this in my local `.zshrc`, so I would not need such a template

bamachrn:
Agree but we are not doing this for us only.  It would be easy to not miss anything while running the script.

amfred:
I also think we'll want an environment template file but this will work and unblock other PRs so let's get it in.

bamachrn:
looks good to me /lgtm

README.md
```diff
@@ -209,27 +209,24 @@ Authorization in `root:redhat-appstudio` and `root:redhat-hacbs` workspaces in C
 
 For access to the OpenShift staging cluster, the user must be added to the `stage` team in the `redhat-appstudio-sre` Github organization.
 
-## Enabling Monitoring for workload cluster
+## Monitoring for workload clusters
 
-Prometheus Operator is deployed through ArgoCD. This is configure to run in `appstudio-workload-monitoring` namespace. 
-This prometheus is accessed through exported route, which is secured by github oauth. Github oauth secrets are configured in the cluster through secret
+Both Prometheus and Grafana UIs are protected by an OAuth2 proxy which delegates the authentication to GitHub. 
+Users must belong to the [Red Hat Appstudio SRE organization](https://github.com/redhat-appstudio-sre) team configured in the OAuth2 proxy to be allowed to access these UIs.
```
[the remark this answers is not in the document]
xcoulon: AFAIK, yes, as long as you have the values (or you have access to the settings to generate a new secret).

README.md
```diff
@@ -209,27 +209,24 @@ Authorization in `root:redhat-appstudio` and `root:redhat-hacbs` workspaces in C
...
-```
+Make sure you have the following environment variables:
+- `PROMETHEUS_GITHUB_CLIENT_ID`
+- `PROMETHEUS_GITHUB_CLIENT_SECRET`
+- `PROMETHEUS_GITHUB_COOKIE_SECRET`
+- `GRAFANA_GITHUB_CLIENT_ID`
+- `GRAFANA_GITHUB_CLIENT_SECRET`
+- `GRAFANA_GITHUB_COOKIE_SECRET`
+
+The `PROMETHEUS_GITHUB_CLIENT_ID`/`PROMETHEUS_GITHUB_CLIENT_SECRET` and `GRAFANA_GITHUB_CLIENT_ID`/`GRAFANA_GITHUB_CLIENT_SECRET` value pairs must match an existing OAuth application on GitHub - see [OAuth apps](https://github.com/organizations/redhat-appstudio-sre/settings/applications) in the [Red Hat Appstudio SRE organization](https://github.com/organizations/redhat-appstudio-sre). The `PROMETHEUS_GITHUB_COOKIE_SECRET` and `GRAFANA_GITHUB_COOKIE_SECRET` can be generated using the [followiing instructions](https://oauth2-proxy.github.io/oauth2-proxy/docs/configuration/overview#generating-a-cookie-secret).
```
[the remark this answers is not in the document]
xcoulon: thanks!

README.md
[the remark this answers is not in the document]
xcoulon: @Michkov I stand corrected: you'll need to configure your OAuth2 proxy with the client id and secret, and GitHub will redirect to the callback URL configured in the settings, so basically a single cluster at a time that can use a given client id/secret pair.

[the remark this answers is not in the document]
Michkov: Just thinking if it should not be described in generic way and then subsection about configuration of staging instance. (but this is not blocking merge of this PR from my side)

README.md
```diff
@@ -209,27 +209,64 @@ Authorization in `root:redhat-appstudio` and `root:redhat-hacbs` workspaces in C
...
 ```
+$ PROMETHEUS_URL=`oc get route/prometheus-k8s -n openshift-monitoring -o json | jq -r '.status.ingress[0].host'`
+
+$ PROMETHEUS_URL=`oc get route/prometheus -n appstudio-workload-monitoring -o json | jq -r '.status.ingress[0].host'`
+```
+
+`GRAFANA_TOKEN` is obtained by requesting a token for the `grafana-oauth` service account in the **Prometheus cluster**:
+```
+$ GRAFANA_TOKEN=`oc create token grafana-oauth -n appstudio-workload-monitoring`
+```
```
amfred, line 265: Do we need to ask users to enter the $PROMETHEUS_URL and $GRAFANA_TOKEN or can the script calculate them automatically?

README.md
```diff
@@ -209,27 +209,64 @@ Authorization in `root:redhat-appstudio` and `root:redhat-hacbs` workspaces in C
...
+
+Grafana datasources contain the connection settings to the Prometheus instances. These datasources are stored in secrets in the `appstudio-workload-monitoring` of the **Grafana cluster**.
+
+The Prometheus endpoints used by Grafana are protected by an OAuth proxy running as a sidecar container and which checks that the incoming requests contain a valid token. A token is valid if it belongs to a service account of the **Prometheus cluster** and which has the RBAC permission to "get namespaces". This can be obtained with the `cluster-monitoring-view` cluster role.
+
+```
+$ ./hack/setup-monitoring.sh grafana-datasource-secret $DATASOURCE_NAME $PROMETHEUS_URL $GRAFANA_TOKEN
+```
+
+`DATASOURCE_NAME` is the name of the secret itself, as well as the base name of the YAML file it contains. Grafana supports datasources from multiple secrets, so their names must be unique.
```
amfred, line 252: I'm not sure what to provide for $DATASOURCE_NAME from this doc. What would be an example of a datasource name?  Does it have to correspond to some resource or namespace that already exists, or can it be an arbitrary string?

README.md
```diff
@@ -209,27 +209,64 @@ Authorization in `root:redhat-appstudio` and `root:redhat-hacbs` workspaces in C
...
+
+$ ./hack/setup-monitoring.sh oauth2-secret grafana-oauth2-proxy $GRAFANA_GITHUB_CLIENT_ID $GRAFANA_GITHUB_CLIENT_SECRET $GRAFANA_GITHUB_COOKIE_SECRET
+```
+
+The `PROMETHEUS_GITHUB_CLIENT_ID`/`PROMETHEUS_GITHUB_CLIENT_SECRET` and `GRAFANA_GITHUB_CLIENT_ID`/`GRAFANA_GITHUB_CLIENT_SECRET` value pairs must match an existing "OAuth application" on GitHub - see [OAuth apps](https://github.com/organizations/redhat-appstudio-sre/settings/applications) in the [Red Hat Appstudio SRE organization](https://github.com/organizations/redhat-appstudio-sre). The `PROMETHEUS_GITHUB_COOKIE_SECRET` and `GRAFANA_GITHUB_COOKIE_SECRET` can be generated using the [following instructions](https://oauth2-proxy.github.io/oauth2-proxy/docs/configuration/overview#generating-a-cookie-secret).
+
+
+#### Grafana Datasources
+
+Grafana datasources contain the connection settings to the Prometheus instances. These datasources are stored in secrets in the `appstudio-workload-monitoring` of the **Grafana cluster**.
```
amfred, line 244:
```suggestion
Grafana datasources contain the connection settings to the Prometheus instances. These datasources are stored in secrets in the `appstudio-workload-monitoring` namespace of the **Grafana cluster**.
```

README.md
```diff
@@ -209,27 +209,64 @@ Authorization in `root:redhat-appstudio` and `root:redhat-hacbs` workspaces in C
...
-Prometheus Operator is deployed through ArgoCD. This is configure to run in `appstudio-workload-monitoring` namespace. 
-This prometheus is accessed through exported route, which is secured by github oauth. Github oauth secrets are configured in the cluster through secret
+Note:
 
-`prometheus-proxy-config`
+This section refers to "Grafana cluster" and "Prometheus cluster" as the clusters on which Grafana and Prometheus are deployed, respectively. In a multi-cluster topology, there will be a single cluster on which Grafana is deployed, whereas Prometheus will be deployed on all clusters where collecting metrics is needed.
+
+### Setup
+
+First, create the `appstudio-workload-monitoring` namespace if it does not exist yet:
```
amfred, line 220:
```suggestion
First, create the `appstudio-workload-monitoring` namespace on each Prometheus or Grafana cluster, if it does not exist yet:
```

README.md
```diff
@@ -209,27 +209,64 @@ Authorization in `root:redhat-appstudio` and `root:redhat-hacbs` workspaces in C
...
+Both Prometheus and Grafana UIs are protected by an OAuth2 proxy running as a sidecar container and which delegates the authentication to GitHub. 
+Users must belong to the [Red Hat Appstudio SRE organization](https://github.com/redhat-appstudio-sre) team configured in the OAuth2 proxy to be allowed to access the Web UIs.
+
+Create the secrets with the following commands:
+
+```
+$ ./hack/setup-monitoring.sh oauth2-secret prometheus-oauth2-proxy $PROMETHEUS_GITHUB_CLIENT_ID $PROMETHEUS_GITHUB_CLIENT_SECRET $PROMETHEUS_GITHUB_COOKIE_SECRET
+
+$ ./hack/setup-monitoring.sh oauth2-secret grafana-oauth2-proxy $GRAFANA_GITHUB_CLIENT_ID $GRAFANA_GITHUB_CLIENT_SECRET $GRAFANA_GITHUB_COOKIE_SECRET
+```
```
amfred, line 237: I think you should specify here that the promethus-oauth2-proxy secret needs to be created for each Prometheus cluster.  And do you need a different "OAuth application" for each cluster or is it the same one for all clusters?

And then specify that you only need to create the grafana-oauth2-proxy secret once, on the Grafana cluster.

amfred (changes requested):

README.md
```diff
@@ -209,27 +209,64 @@ Authorization in `root:redhat-appstudio` and `root:redhat-hacbs` workspaces in C
...
-`prometheus-proxy-config`
+This section refers to "Grafana cluster" and "Prometheus cluster" as the clusters on which Grafana and Prometheus are deployed, respectively. In a multi-cluster topology, there will be a single cluster on which Grafana is deployed, whereas Prometheus will be deployed on all clusters where collecting metrics is needed.
+
+### Setup
+
+First, create the `appstudio-workload-monitoring` namespace if it does not exist yet:
+
+```
+$ oc create namespace appstudio-workload-monitoring
+```
```
amfred, line 224: Do you want them to run this command, or
```
$ ./hack/setup-monitoring.sh  namespace
```

README.md
```diff
@@ -209,27 +209,64 @@ Authorization in `root:redhat-appstudio` and `root:redhat-hacbs` workspaces in C
...
+
+$ PROMETHEUS_URL=`oc get route/prometheus-oauth -n appstudio-workload-monitoring -o json | jq -r '.status.ingress[0].host'`
+```
+
+`GRAFANA_TOKEN` is obtained by requesting a token for the `grafana-oauth` service account in the **Prometheus cluster**:
+```
+$ GRAFANA_TOKEN=`oc create token grafana-oauth -n appstudio-workload-monitoring`
+```
+
+The `grafana-oauth` service account is created by `.components/monitoring/base/configure-prometheus.yaml` along with a binding to the `cluster-monitoring-view` cluster role.
```
amfred, line 267: How do we get Argo CD to apply this configuration?

README.md
xcoulon: In a multi-cluster setup, each pair of `$PROMETHEUS_URL` and `$GRAFANA_TOKEN` values will be retrieved from the workload cluster were Prometheus is deployed, but used in the "control plane" cluster where the Grafana datasource secret is going to be created. So for now I believe it's simpler to ask the user to do it manually.

README.md
xcoulon: clarified in 3f1a2da

README.md
xcoulon: I initially had a function for this in the script, but then I thought it was maybe not needed, for such a simple command. However, I can restore it if it makes things easier. Also, IIRC @bamachrn said that the namespace would be created by ArgoCD? (I haven't verified that yet, though)

README.md
xcoulon: good point, let me clarify that! And yes, it's a different OAuth Application for each cluster, indeed

xcoulon: done in 13337c0

README.md
xcoulon: I'm not sure I understand your question: this `grafana-oauth` service account and its associated cluster role binding are defined in `components/monitoring/base/configure-prometheus.yaml`, so it going to be managed by ArgoCD, isn't it?

README.md
bamachrn: this means we need to update the env variables before running for each prometheus cluster. could we put this somewhere?

components/monitoring/base/configure-prometheus.yaml
```diff
@@ -118,9 +133,17 @@ spec:
     - --upstream=http://localhost:9090
     - --tls-cert=/etc/tls/private/tls.crt
     - --tls-key=/etc/tls/private/tls.key
+    - --request-logging=true
     - --cookie-secret=SECRET
     # see https://github.com/openshift/oauth-proxy/#delegate-authentication-and-authorization-to-openshift-for-infrastructure
     - --openshift-delegate-urls={"/":{"group":"","resource":"namespaces","verb":"get"}}
+    securityContext:
```
bamachrn, line 140: we are already adding container security context. I believe we dont need pod security context again. or we can add pod securitycontext and remove the container security context as it is same for both the containers.

xcoulon: apparently the security context defined in the Prometheus CR's `.spec` does not apply to the extra containers. When running `oc rollout restart statefulset/prometheus-prometheus`, I had multiple warning which were fixed after adding these security contexts.

README.md
xcoulon: would you have any pointers to this, please?

README.md
xcoulon: > could we put this somewhere?

did you mean documenting it? If so, please see in 98c9d0f

README.md
```diff
@@ -110,7 +110,7 @@ which will:
 #### Workspaces:
 If `-rw | --root-workspace` parameter is not specified, then by default, all workspaces are automatically created under the `root` workspace.
 There are two workspaces created per kcp instance:
-* `redhat-appstudio-internal-compute` - This is the workspace where the SyncTarget for the OpenShift workload cluster is configured. If the root workspace is different from `root`, then the name of the workspace is set to `compute` to work around [this issue](https://github.com/kcp-dev/kcp/issues/1843). (The name of the workspace can be overridden by setting the `COMPUTE_WORKSPACE` variable)
+* `redhat-appstudio-internal-compute` - This is the workspace where the SyncTarget for the OpenShift Prometheus cluster is configured. If the root workspace is different from `root`, then the name of the workspace is set to `compute` to work around [this issue](https://github.com/kcp-dev/kcp/issues/1843). (The name of the workspace can be overridden by setting the `COMPUTE_WORKSPACE` variable)
```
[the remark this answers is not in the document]
xcoulon: reverted in dd05d11

README.md
amfred: It's actually automatically creating the namespaces when installing the operators (using the CreateNamespace=true directive), rather than explicitly creating the namespaces.  Here's an example: https://github.com/redhat-appstudio/infra-deployments/blob/4f7edc301a7d01d7710cb75aff7b2c41a1c203f1/argo-cd-apps/base/has.yaml#L28

README.md
```diff
@@ -190,6 +222,71 @@ Access to namespaces is managed by [components/authentication](components/authen
...
+The Prometheus endpoints used by Grafana are protected by an OAuth proxy running as a sidecar container and which checks that the incoming requests contain a valid token. A token is valid if it belongs to a service account of the **Prometheus cluster** and which has the RBAC permission to "get namespaces". This can be obtained with the `cluster-monitoring-view` cluster role.
+
+```
+$ ./hack/setup-monitoring.sh grafana-datasource-secret $DATASOURCE_NAME $PROMETHEUS_URL $GRAFANA_TOKEN
+```
+
+`DATASOURCE_NAME` is the name of the secret itself, as well as the base name of the YAML file it contains. Grafana supports datasources from multiple secrets, so their names must be unique.
+`DATASOURCE_NAME` is an arbitrary name, for example `cluster-1-prometheus-openshift-ds` for Prometheus running in the `openshift-monitoring` namespace of Cluster-1.
+
+`PROMETHEUS_URL` is obtained from the route created for Prometheus in the `openshift-monitoring` and `appstudio-workload-monitoring` namespaces:
```
amfred, line 273:
```suggestion
`PROMETHEUS_URL` is obtained from the route created for Prometheus in the `openshift-monitoring` and `appstudio-workload-monitoring` namespaces in the **Prometheus cluster**:
```

README.md
amfred: Perfect, thanks.

README.md
amfred: OK that makes sense.  I would recommend restructuring this section in the order that people would run the commands, to make it easier to follow.

> #### Grafana Datasources
> 
> Grafana datasources contain the connection settings to the Prometheus instances. These datasources are stored in secrets in the `appstudio-workload-monitoring` namespace of the **Grafana cluster**.
> 
> The Prometheus endpoints used by Grafana are protected by an OAuth proxy running as a sidecar container and which checks that the incoming requests contain a valid token. A token is valid if it belongs to a service account of the **Prometheus cluster** and which has the RBAC permission to "get namespaces". This can be obtained with the `cluster-monitoring-view` cluster role.
>
> In a multi-cluster setup, each pair of $PROMETHEUS_URL and $GRAFANA_TOKEN values will be retrieved from the workload cluster were Prometheus is deployed, but used in the "control plane" cluster where the Grafana datasource secret is going to be created.  Run the following commands to obtain the tokens from each workload cluster where Prometheus is deployed:

(and then the commands to get those $PROMETHEUS_URL and $GRAFANA_TOKEN values)

> Using the values you just obtained from the **Prometheus cluster**, run the following command on the **Grafana cluster**:
> 

(and then the grafana-datasource-secret command and explanation about the arbitrary DATASOURCE_NAME that you recently added)

> Repeat this procedure for each workload cluster where Prometheus is deployed.

README.md
```diff
@@ -190,6 +222,71 @@ Access to namespaces is managed by [components/authentication](components/authen
 
 Users can be added to organizations by Michal Kovarik <mkovarik@redhat.com> and by Shoubhik Bose <shbose@redhat.com>.
 
+## Monitoring for Prometheus clusters
+
+Note:
+
+This section refers to "Grafana cluster" and "Prometheus cluster" as the clusters on which Grafana and Prometheus are deployed, respectively. In a multi-cluster topology, there will be a single cluster on which Grafana is deployed, whereas Prometheus will be deployed on all clusters where metrics needed to be collected.
```
amfred, line 229:
```suggestion
This section uses **Grafana cluster** and **Prometheus cluster** to refer to the clusters on which Grafana and Prometheus are deployed, respectively. In a multi-cluster topology, there will be a single cluster on which Grafana is deployed, whereas Prometheus will be deployed on all clusters where metrics needed to be collected.
```

README.md
amfred: You're right that it's a simple command, but if we can automate creating the namespace, even better.  I just don't have sample code handy for that.  I wouldn't put a ton of effort into automating this one line, since we're already doing this semi-manually.

README.md
amfred: These changes look good to me, thanks.

README.md
amfred: Because we're jumping between clusters, I added the **Prometheus cluster** here to make it easier even though you could infer it.

There are two commands below to get the PROMETHUS_URL, but only one place to use that URL.  Do these two commands return the same value (in which case we can just provide one way to get it)?  Or do I need to use both values somehow?

README.md
```diff
@@ -190,6 +222,71 @@ Access to namespaces is managed by [components/authentication](components/authen
...
+
+$ PROMETHEUS_URL=`oc get route/prometheus-oauth -n appstudio-workload-monitoring -o json | jq -r '.status.ingress[0].host'`
+```
+
+`GRAFANA_TOKEN` is obtained by requesting a token for the `grafana-oauth` service account in the **Prometheus cluster**:
+```
+$ GRAFANA_TOKEN=`oc create token grafana-oauth -n appstudio-workload-monitoring`
+```
+
+The `grafana-oauth` service account is created by `components/monitoring/base/configure-prometheus.yaml` along with a binding to the `cluster-monitoring-view` cluster role.
```
amfred, line 286: Please explain how configure-prometheus.yaml gets applied here. Do I need to run another command to apply this using Argo? Or is this already applied to all clusters if you run the bootstrap or preview scripts?

amfred (changes requested):
Sorry for the second round of edit requests.  I'm trying to follow along.  I got farther this time but still ran into some issues.

README.md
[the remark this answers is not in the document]
amfred: I think this is explained more clearly now.

README.md
amfred: I haven't gotten far enough to see if it works.  Since it's in the base configuration, I think this prometheus configuration will be applied to every cluster where you run the bootstrap script.  Let's verify that and then I would add that statement here to avoid confusion.  Otherwise people might think they need to `kubectl apply` it.

README.md
xcoulon: I've updated that section in 01b893b, could you give it another look and let me know if it's ok?

README.md
xcoulon: ok, I suggest we address that in a separate PR, once we have a `./components/monitoring/kustomize.yaml` in place (it's currently missing after the `main` branch was switch to pre-KCP)

README.md
xcoulon: ok, makes sense. Yes, it should be applied by ArgoCD on every cluster (as opposed to Grafana on a single cluster), but we should mention that for the sake of clarity. I'll add that once we have `./components/monitoring` in place.

README.md
```diff
@@ -139,7 +139,39 @@ spec:
...
+```
+which takes care of creation of the workspaces, SyncTarget, and the representation of the cluster in ArgoCD for the given kcp instance. So to fully finish the upstream configuration run the script for both kcp instances using the parameter `-kn kcp-stable` and `-kn kcp-unstable`.
+
+#### Access Argo CD Web UI
+Open the Argo CD Web UI to see the status of your deployments. You can use the route from the previous step and login using your OpenShift credentials (using the 'Login with OpenShift' button), or login to the OpenShift Console and navigate to Argo CD using the OpenShift Gitops menu in the Applications pulldown.
+   ![OpenShift Gitops menu with Cluster Argo CD menu option](documentation/images/argo-cd-login.png?raw=true "OpenShift Gitops menu")
+
+If your deployment was successful, you should see several applications running, such as "all-components", "application-service", and so on.
+
+### Optional: OpenShift Local (formerly CRC) Post-Bootstrap Configuration
```
MatousJobanek, line 174: I guess that this is a zombie :ghost: from the kcp/pre-kcp branch switch.

hack/setup-monitoring.sh
```diff
@@ -0,0 +1,77 @@
...
+# -----------------------------------------------------------------
+if declare -f "$1" > /dev/null
+then
+    # call arguments verbatim
+    "$@"
+else
+    # Show a helpful error
+    echo "'$1' is not a valid command" >&2
+    echo "available commands:"
+    echo "namespace                   Create the 'appstudio-workload-monitoring' namespace"
```
MatousJobanek, line 72: the script doesn't have this function/command, does it?

hack/setup-monitoring.sh
```diff
@@ -0,0 +1,77 @@
...
+    echo "  $0 oauth2-secret NAME CLIENT_ID CLIENT_SECRET COOKIE_SECRET"
+    exit 1
+  fi
+
+  NAME=$1
+  CLIENT_ID=$2
+  CLIENT_SECRET=$3
+  COOKIE_SECRET=$4
+
+  oc create secret generic $NAME \
```
MatousJobanek, line 20: would it make sense to use `apply` instead of `create` - in case it needs to be replaced with new values?

hack/setup-monitoring.sh
```diff
@@ -0,0 +1,77 @@
...
+  withCredentials: true
+  isDefault: false
+  jsonData:
+    httpHeaderName1: 'Authorization'
+    timeInterval: 5s
+    tlsSkipVerify: true
+  secureJsonData:
+    httpHeaderValue1: 'Bearer $TOKEN'"
+  
+  oc create secret generic $NAME -n appstudio-workload-monitoring --from-literal=$NAME.yaml="$DATA"
```
MatousJobanek, line 60: the same as for `create` vs `apply`

MatousJobanek (commented):
Looks good :+1: 
btw, would it make sense to move the monitoring README to a separate file and just reference it from the main README?

README.md
xcoulon: removed in 67ce1d0

hack/setup-monitoring.sh
xcoulon: keeping the `oc create secret` command as it suppors the `--from-literal` flags, but using `--dry-run=client -o yaml` options to generate the YAML representation and `oc apply -f -` to apply if afterwards.
see c3454ee

hack/setup-monitoring.sh
xcoulon: see c3454ee

README.md
amfred: Per our conversation in Slack, please add `kustomize build components/monitoring/base | oc apply -f - `

amfred: I see that in the latest commit, thanks.

amfred (approved):

README.md
```diff
@@ -190,6 +222,71 @@ Access to namespaces is managed by [components/authentication](components/authen
...
+
+```
+# on each Prometheus cluster
+$ ./hack/setup-monitoring.sh oauth2-secret prometheus-oauth2-proxy $PROMETHEUS_GITHUB_CLIENT_ID $PROMETHEUS_GITHUB_CLIENT_SECRET $PROMETHEUS_GITHUB_COOKIE_SECRET
+
+# on the Grafana cluster
+$ ./hack/setup-monitoring.sh oauth2-secret grafana-oauth2-proxy $GRAFANA_GITHUB_CLIENT_ID $GRAFANA_GITHUB_CLIENT_SECRET $GRAFANA_GITHUB_COOKIE_SECRET
+```
+
+The `PROMETHEUS_GITHUB_CLIENT_ID`/`PROMETHEUS_GITHUB_CLIENT_SECRET` and `GRAFANA_GITHUB_CLIENT_ID`/`GRAFANA_GITHUB_CLIENT_SECRET` value pairs must match an existing "OAuth Application" on GitHub - see [OAuth apps](https://github.com/organizations/redhat-appstudio-sre/settings/applications) in the [Red Hat Appstudio SRE organization](https://github.com/organizations/redhat-appstudio-sre). The `PROMETHEUS_GITHUB_COOKIE_SECRET` and `GRAFANA_GITHUB_COOKIE_SECRET` can be generated using the [following instructions](https://oauth2-proxy.github.io/oauth2-proxy/docs/configuration/overview#generating-a-cookie-secret).
```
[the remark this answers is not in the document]
amfred: Not a blocker though.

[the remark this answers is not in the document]
amfred: Not a blocker though.

PR #900, opened 2022-11-09
Add stump file for preview script to create apiexport

Michkov:
Preview script requires to have apiexport for dev so that it can be created for getting identity hash and later updated by argoCD.


MatousJobanek (approved):

Michkov:
replaced by #904

PR #901, opened 2022-11-09
WIP: Add annotations

amfred:


components/application-service/base/kustomization.yaml
```diff
@@ -12,4 +12,19 @@ images:
 namespace: application-service
 
 patches:
-  - ./manager_resources_patch.yaml
+- ./manager_resources_patch.yaml
+- patch: |-
```
MatousJobanek, line 16: I can be wrong, but you cannot combine the strategic merge patch and json patch in one. You need to put them in different sections:
```yaml
patchesStrategicMerge:
- ./manager_resources_patch.yaml

patches:
- patch: |-
```

components/application-service/base/kustomization.yaml
```diff
@@ -12,4 +12,19 @@ images:
...
+    - op: add
+      path: /metadata/labels
+      value:
+        appstudio.redhat.com/namespacetype: "controller"
```
MatousJobanek, line 20: this will replace the already existing labels because you define that you want to add the whole `labels` section with only one value - one label
```suggestion
    - op: add
      path: /metadata/labels/appstudio.redhat.com~1namespacetype
      value:
        controller
```

components/application-service/base/kustomization.yaml
```diff
@@ -12,4 +12,19 @@ images:
...
+    - op: add
+      path: /metadata/annotations
+      value:
+        k8s.ovn.org/acl-logging: |-
+      {
+        "deny": "info",
+        "allow": "info"
+      }
```
MatousJobanek, line 28: the same here, plus I don't think that this format of the annotation value is supported by kustomize
```suggestion
    - op: add
      path: /metadata/annotations/k8s.ovn.org~1acl-logging
      value:
        '{"deny": "info", "allow": "info"}'
```

PR #925, opened 2022-11-15
Disable rekor integration

lcarva:
This aligns with the current plan to not have Rekor enabled initially, but introduce it post-mvp phase.

### DIFF
--- a/components/build/tekton-chains/chains-config.yaml
+++ b/components/build/tekton-chains/chains-config.yaml
@@ -21,7 +21,6 @@ data:
   # For image signing iiuc
   artifacts.oci.storage: "oci"
 
-  # We might want to turn this off by default later so we aren't
-  # filling up rekor.sigstore.dev with garbage.
-  transparency.enabled: "true"
-  #transparency.url: "https://rekor.sigstore.dev"
+  # Rekor integration is disabled for now. It is planned to be re-introduced
+  # in the future.
+  transparency.enabled: "false"


lcarva:
@redhat-appstudio/enterprise-contract ptal

zregvart:
I guess e2e tests need an update first...

```
[INFO] Cloning tests from branch pre-kcp repository https://github.com/redhat-appstudio/e2e-tests.git
Cloning into '/go/src/github.com/redhat-appstudio/infra-deployments/tmp/e2e-tests'...
fatal: Remote branch pre-kcp not found in upstream origin
Unexpected end of command stream
[ERROR] Job failed with code 128. 
```

simonbaird (approved):

lcarva:
e2e test failures seem to be unrelated.

Michkov:
So https://github.com/redhat-appstudio/infra-deployments/blob/main/hack/preview.sh#L78-L80 should be removed as well?

lcarva:
> So https://github.com/redhat-appstudio/infra-deployments/blob/main/hack/preview.sh#L78-L80 should be removed as well?

I'd rather not as part of this PR. Rekor is something we still want, just not right now. There's still value in setting `transparency.url` because it allows us to simply toggle `transparency.enabled` to use Rekor which is helpful in a dev environment.

There's more work that needs to be done around Rekor. This PR is the bare minimum needed right now.

lcarva:
/lgtm

(re-adding label after rebase)

PR #926, opened 2022-11-16
Minor improvements to hack/build/build-via-appstudio.sh

tkdchen:


hack/build/build-via-appstudio.sh
```diff
@@ -51,4 +51,5 @@ if [ -z "$COMPONENT" ]; then
 else
   create-component $COMPONENT
 fi
-echo "Run this to show running builds: tkn pr list"
+echo "Run this from physical cluster to show running builds:
+oc get pipelineruns --all-namespaces"
```
tkdchen, line 55: This might be irrelative for now, since it was made to get pipelineruns easily when kcp is enable. If `tkn pr list` is still a preferred way to list pipelinesruns, I'll revert it.

hack/build/build-via-appstudio.sh
```diff
@@ -36,7 +36,7 @@ fi
 function create-component {
   GIT_URL=$1
   NAME=$(echo $GIT_URL | grep -o '[^/]*$')
-  IMAGE=quay.io/$MY_QUAY_USER/$NAME
+  [ -z "$WITHOUT_OUTPUT_IMAGE" ] && IMAGE=quay.io/$MY_QUAY_USER/$NAME
```
Michkov, line 39: SKIP_OUTPUT_IMAGE is better var name

PR #934, opened 2022-11-16
Point to e2e test on main branch

lcarva:


lcarva:
This is probably not the right fix, or it needs more work :(

PR #935, opened 2022-11-16
Add application-api component

johnmcollier:
Adds the application-api component as it was not present on the pre-kcp branch. 

Newer versions of the HAS and gitops components do not have the CRDs in their kustomize manifests, so it's necessary to add the application-api kustomize manifests in infra-deployments. As we had a separate component for application-api on the KCP-branch, I figure it makes sense to do so as well for the non-KCP branch.

### DIFF
--- a/components/application-api/kustomization.yaml
+++ b/components/application-api/kustomization.yaml
@@ -1,0 +1,10 @@
+resources:
+- allow-argocd-to-manage.yaml
+- argocd-permissions.yaml
+- https://github.com/redhat-appstudio/application-api/config/crd?ref=77cba9006505d430248f863f81539cd8b39efaed
+
+
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+namespace: application-api
--- a/components/application-api/kustomization.yaml
+++ b/components/application-api/kustomization.yaml
@@ -6,5 +6,3 @@ resources:
 
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
-
-namespace: application-api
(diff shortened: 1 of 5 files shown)


johnmcollier:
```
 /bin/bash: ./scripts/install-appstudio-kcp.sh: No such file or directory 
 ```
 
 😄

File: components/application-api/kustomization.yaml [hunk 1]
```diff
@@ -0,0 +1,10 @@
+resources:
+- allow-argocd-to-manage.yaml
+- argocd-permissions.yaml
+- https://github.com/redhat-appstudio/application-api/config/crd?ref=77cba9006505d430248f863f81539cd8b39efaed
+
+
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+namespace: application-api
```
MatousJobanek, line 10: just out of curiosity - is there anything namespace-scoped?

MatousJobanek (commented):
Please update also the `repo-overlay.yaml` file in `developement` directory

components/application-api/kustomization.yaml
johnmcollier: I added it to be consistent, but since there's nothing namespace scoped it can be removed

MatousJobanek (approved):

components/application-api/kustomization.yaml
MatousJobanek: it was a question just out of curiosity - you didn't have to remove it :wink: :smile: 
but either way is fine :+1:

flacatus:
To help e2e get fixed we need this merged without tests

PR #939, opened 2022-11-18
Added prometheus configs to ArgoCD

bamachrn:
This PR adds all the changes for promethes deployment through ArgoCD


amfred (approved):

bamachrn:
should we be running e2e-kcp-stable for CI? I believe we need to update it to without KCP mode. @MatousJobanek @amfred

PR #940, opened 2022-11-18
fix: add missing files in application-api application

rhopp:
Signed-off-by: Radim Hopp <rhopp@redhat.com>

### DIFF
--- a/components/application-api/allow-argocd-to-manage.yaml
+++ b/components/application-api/allow-argocd-to-manage.yaml
@@ -1,0 +1,14 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: grant-argocd
+  namespace: application-api
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: admin
+subjects:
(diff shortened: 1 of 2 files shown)


johnmcollier (approved):

flacatus (approved):

rhopp:
I'm force-merging this one to unblock us from this deadlock. This should unblock (or at least keep us moving forward in) https://github.com/openshift/release/pull/34064

MatousJobanek:
@jgwest @Michkov  do we really need these additional (Cluster)Role(Bindings) defined here? I thought that ArgoCD should have cluster-admin permissions. :man_shrugging: 

In addition, I see that the same set of RBAC is defined per component - having these ClusterRole(Binding)s per component doesn't make any sense because:
* it's cluster-scoped
* they are not limited to the particular `customresourcedefinitions` resource names.

But still - why ArgoCD needs them at all?

Michkov:
@sbose78 could you answer @MatousJobanek question?

Michkov:
@sbose78 Answer:
```
This was needed to ensure Argo CD has permissions to CRUD in the namespace.
... after we remove cluster-admin privlieges from the ArgoCD
```

PR #948, opened 2022-11-22
Reorganization of SPI deployment

skabashnyuk:
- Decouple SPI Vault deployment from spi controller and Oauth service.
 - Used simular to `kcp` branch configuration layout.

### DIFF
--- a/hack/preview.sh
+++ b/hack/preview.sh
@@ -96,7 +96,7 @@ if [ -n "$DOCKER_IO_AUTH" ]; then
 fi
 
 rekor_server="rekor.$domain"
-sed -i.bak "s/rekor-server.enterprise-contract-service.svc/$rekor_server/" $ROOT/argo-cd-apps/base/enterprise-contract.yaml && rm $ROOT/argo-cd-apps/base/enterprise-contract.yaml.bak
+sed -i "s/rekor-server.enterprise-contract-service.svc/$rekor_server/" $ROOT/argo-cd-apps/base/enterprise-contract.yaml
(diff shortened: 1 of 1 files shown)


argo-cd-apps/overlays/development/repo-overlay.yaml
```diff
@@ -74,6 +74,7 @@ spec:
   source:
     repoURL: https://github.com/redhat-appstudio/infra-deployments.git
     targetRevision: main
+    path: components/spi/overlays/development
```
Michkov, line 77: not needed here, since it's patch so only differences should be here - targeRevios and repoURL are here as placeholders which are replaced by scripts which are called by preview mode.

argo-cd-apps/base/vault.yaml
```diff
@@ -0,0 +1,33 @@
+apiVersion: argoproj.io/v1alpha1
+kind: Application
+metadata:
+  name: vault
```
Michkov, line 4: Please unify naming to `spi-vault`

argo-cd-apps/base/vault.yaml
```diff
@@ -0,0 +1,33 @@
...
+spec:
+  project: default
+
+  source:
+    path: components/vault
+    repoURL: https://github.com/redhat-appstudio/infra-deployments.git
+    targetRevision: main
+
+  destination:
+    namespace: spi-system
```
Michkov, line 15: spi-vault - then you do not need to set `namespace: spi-vault` in `components/vault/kustomization.yaml` (which I would rather have `components/spi-vault`)

Michkov (approved):

flacatus (approved):

PR #956, opened 2022-11-23
drop unnecessary admin permissions

MatousJobanek:
drops most of the unnecessary admin permissions + does a few more changes:
* renames `pipeline-maintainer` to `component-maintainer` and uses as the main "component maintainer" ClusterRole
* grants view permission (at the cluster-level) to all authenticated users
* removes all tenant-related manifests - which means it deletes the whole `components/authentication/tenants/` directory
* drops the `-ci` part of the name from the filenames and RoleBindings - they supposed to represent maintainer RoleBinding in general (not only related to a CI)
* adds the ability to exec a pod to component maintainers (instead fo having separate Role & RoleBinding in `has-debug.yaml` file)
* drops some additional extra permissions like the `prune-has.yaml` one as this should be managed by a controller automatically cc @johnmcollier


johnmcollier (approved):
LGTM from my end.

The HAS pruner role is no longer needed, and has been replaced by similar functionality in HAS.

However for the recent memory issues with HAS, still being able to GET/LIST any HAS/Application-api resources (Application, Component, CDQ, Environment, Snapshot, and SnapshotEnvironmentBinding) at the cluster level would be beneficial for debugging currently. I can't remember if the cluster-wide view permissions includes CRDs, if not, would it be possible to keep a cluster role around for that?

components/authentication/component-maintainer.yaml
```diff
@@ -33,3 +33,9 @@ rules:
     verbs:
       - get
       - list
+  - apiGroups:
+    - ''
+    resources:
+    - pods/exec
```
johnmcollier, line 39: Just thought about this while working with pprof, `pods/portforward` would be good here too 🙂

Michkov (approved):

components/authentication/component-maintainer.yaml
```diff
@@ -1,7 +1,7 @@
 kind: ClusterRole
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
-  name: pipeline-maintainer
+  name: component-maintainer
```
Michkov, line 4: pipeline-maintainer is still used in authentication/e2e-tests.yaml and authentication/group-sync/group-sync-ci.yaml

components/authentication/everyone-can-view.yaml
```diff
@@ -1,11 +1,11 @@
...
 metadata:
-  name: view-quality-dashboard
-  namespace: quality-dashboard
+  name: everyone-view
 subjects:
-  - kind: Group
-    name: QE team
+- kind: Group
+  apiGroup: rbac.authorization.k8s.io
+  name: 'system:authenticated'
```
Michkov, line 8: authentication/group-sync/view-group-sync.yaml can be removed

components/authentication/component-maintainer.yaml
MatousJobanek: here you go d89a550

components/authentication/component-maintainer.yaml
MatousJobanek: I missed that plus one more occurrence 5d1f710

MatousJobanek:
> However for the recent memory issues with HAS, still being able to GET/LIST any HAS/Application-api resources (Application, Component, CDQ, Environment, Snapshot, and SnapshotEnvironmentBinding) at the cluster level would be beneficial for debugging currently.

good point @johnmcollier - I totally forgot that the controllers (and CRDs) are not installed via OLM and thus not included in the aggregated `view` ClusterRole. This should take care of it: f9d2a88

PR #958, opened 2022-11-23
Enable pprof in HAS on staging

johnmcollier:
Adds a patch to set the `ENABLE_PPROF` environment variable to true in HAS on staging


kim-tsao (approved):

elsony (approved):
/lgtm

PR #961, opened 2022-11-24
Delete PaC secret before creating a new one

tkdchen:
When rerun setup-for-integration.sh on a provisioned OpenShift cluster, it fails to create secret in build-service namespace due to the existing one. This patch deletes the existing secret in advance before creation.

Meanwhile, the code of deletion, creation and message report are grouped together for build-service namespace specifically.


Michkov:
Fix for this problem is included in #966

PR #968, opened 2022-11-24
One namespace for component's Pipeline-as-Code CI

Michkov:
Unify .tekton directories into tekton-ci component.

Reasons:
- Shared-resources should not be used anymore, which would require to
  create/manage secret in each namespace. With one namespaces it would
  be managed only there
- SRE engineers are interested in the service, so they will see only the
  service in the service namespace - not pods of CI pipelines
- Quota can be set on service namespace, CI pipelineRuns will not take
  resources from it
- In case of multiple environments we can select one when the CI will be
  installed, as part of the commit the CI is disabled for developemnt
  environment


components/tekton-ci/repository.yaml
```diff
@@ -0,0 +1,98 @@
...
+  name: service-provider-integration
+spec:
+  url: "https://github.com/redhat-appstudio/service-provider-integration-operator"
+---
+apiVersion: pipelinesascode.tekton.dev/v1alpha1
+kind: Repository
+metadata:
+  name: e2e-tests
+spec:
+  url: "https://github.com/redhat-appstudio/e2e-tests"
```
bnallapeta, line 21: This is included by mistake? What would e2e folder contain after this change? Is it needed? (maybe in e2e repository?)
If not, need to remove this block as well as the e2e folder under components.

scoheb:
what are the issues you are trying to solve with this or what are the benefits?

bnallapeta (commented):
need to update e2e-tests to tekton-ci [here](https://github.com/Michkov/infra-deployments/blob/merged-tekton-ci/hack/destroy-cluster.sh#L44).

components/tekton-ci/repository.yaml
Michkov: folder in components removed, but this code is needed, to be able to run pipelines defined in https://github.com/redhat-appstudio/e2e-tests/tree/main/.tekton

Michkov:
> need to update e2e-tests to tekton-ci [here](https://github.com/Michkov/infra-deployments/blob/merged-tekton-ci/hack/destroy-cluster.sh#L44).

Updated

Michkov:
> what are the issues you are trying to solve with this or what are the benefits?

@scoheb Description of commit and PR updated with reasons.

gabemontero (approved):

PR #971, opened 2022-11-25
Allow component-maintainer role to list secrets

johnmcollier:
Will be handy when creating/deleting secrets via the OpenShift UI.

... plus I think the memory issues we're encountering with HAS is indirectly related to the sheer number of secrets on the cluster, and being able to list cluster-wide would help 😄


MatousJobanek (approved):

MatousJobanek:
>... plus I think the memory issues we're encountering with HAS is indirectly related to the sheer number of secrets on the cluster, and being able to list cluster-wide would help smile

but this won't help you with listing secrets cluster-wide  - the role is granted only in component namespaces.

johnmcollier:
Right, forgot this wasn’t bound cluster wide 🤦 , this can be closed

PR #983, opened 2022-11-29
Necessary permission for the component maintainers to trigger deployment restart

skabashnyuk:
In SPI we have a couple of secrets that are needed to be configured outside of ArgoCD. 
After that is done it is necessary to restart deployment to apply these configuration changes.

```
kubectl rollout restart  deployment/spi-oauth-service  -n spi-system
kubectl rollout restart  deployment/spi-controller-manager  -n spi-system
```


components/authentication/component-maintainer.yaml
```diff
@@ -43,3 +43,11 @@ rules:
     - pods/portforward
     verbs:
     - create
+  - apiGroups:
+      - 'apps'
+      - 'extensions'
```
Michkov, line 48: why extensions? deployments are in apps

PR #990, opened 2022-12-01
Allow testing with repo url that ends with .git

tkdchen:

I have been using this snippet for my testing for PLNSRVCE-693. Hopefully, it is useful for others as well.

### DIFF
--- a/hack/build/build-via-appstudio.sh
+++ b/hack/build/build-via-appstudio.sh
@@ -35,7 +35,8 @@ fi
 
 function create-component {
   GIT_URL=$1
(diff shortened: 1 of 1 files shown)


gbenhaim:
shouldn't the component the uses the repo url know how to handle a url that ends with `.git`?

Michkov:
Agree with @gbenhaim this is about hiding, I prefer to not do this.

Michkov:
@tkdchen so now I can see what this is doing, it's just fixes component name, it's not touching repository. So please reopen.

PR #1005, opened 2022-12-05
Remove monitoring from hack in favour of setup through ArgoCD

bamachrn:
We are now setting up monitoring using ArgoCD, so removing the manual setup from hack


Michkov (approved):

amfred:
Skipping the E2E tests because this isn't going to affect them and we're having a merge race.

amfred (approved):

PR #1008, opened 2022-12-05
RHDP-420 install keycloak&toolchain in dev env

rhopp:
This PR should enable us to (in CI and/or in development mode):
- install toolchain operators & configure them to use with RHSSO deployed by this PR
- install RHSSO and configures some default REALM to be used by toolchain operators
  - This realm has currently just one user (user1/user1) for testing purposes. We can add as many users as we want.

This will allow us to for example to run e2e-tests in more real-world-like environment (where our tests would be communicating through toolchain proxy with the cluster - instead of using `kubeadmin` for everything like novadays)

Usage: 
```
./hack/bootstrap-cluster.sh preview
```
This will normally install cluster in preview mode. The only difference is that it also deploys keycloak (but doesn't wire it anywhere - I didn't want to introduce another overlay just for keycloak)
```
./hack/bootstrap-cluster.sh preview --toolchain
```
This will do the same as previous command, but also installs toolchain operators (and these operators are configured to use keycloak of toolchain/sandbox team)
```
./hack/bootstrap-cluster.sh preview --toolchain --keycloak
```
This will do the same as previous command, but also reconfigures toolchain operators to use keycloak which is deployed on the same cluster (deployed as part of `development` overlay)


rhopp:
@amfred you might like this one.

hack/bootstrap-cluster.sh
```diff
@@ -1,6 +1,37 @@
...
+  echo "Usae: $0 MODE [-t|--toolchain] [-kc|--keycloak] [-h|--help]"
+  echo "  MODE             upstream/preview (default: upstream)"
+  echo "  -t, --toolchain  (only in preview mode) Install toolchain operators"
+  echo "  -kc, --keycloak  (only in preview mode) Configure the toolchain operator to use keycloak deployed on the cluster"
+  echo "  -h, --help       Show this help message and exit"
+  echo
+  echo "Example usage: \`$0 preview --toolchain --keycloak"
+}
+
 MODE=$1
```
Michkov, line 14: move MODE=$1 after while

hack/preview.sh
```diff
@@ -231,3 +264,9 @@ while :; do
   jq -r '.message' <<< "$STATE"
   sleep $INTERVAL
 done
+
+
+if [ -n $KEYCLOAK ] && [ -n $TOOLCHAIN ]; then
+  echo "Restarting toolchain registration service to pick up keycloak's certs."
+  oc delete deployment/registration-service -n toolchain-host-operator
```
Michkov, line 271: Probably will be necessary to wait until deployment is done before running any tests.

rhopp: Good point. I'll add wait here.

.ci/oci-e2e-deployment.sh
```diff
@@ -103,7 +103,7 @@ export KUBECONFIG_TEST="/tmp/kubeconfig"
 curl https://raw.githubusercontent.com/redhat-appstudio/e2e-tests/pre-kcp/scripts/provision-openshift-user.sh | bash -s
 export KUBECONFIG="${KUBECONFIG_TEST}"
 
-timeout --foreground 15m "$WORKSPACE"/hack/bootstrap-cluster.sh preview
+timeout --foreground 15m "$WORKSPACE"/hack/bootstrap-cluster.sh preview --keycloak --toolchain
```
psturc, line 106: note that we are using [centralized scripts now](https://github.com/openshift/release/blob/62cafb73ce055b3f29f8a06cee0d6f301821225f/ci-operator/config/redhat-appstudio/infra-deployments/redhat-appstudio-infra-deployments-main.yaml#L31), so changes in this file won't be reflected in the CI job. In fact, the whole `.ci` folder should be [deleted](https://issues.redhat.com/browse/HACBS-789).

hack/bootstrap-cluster.sh
rhopp: Sorry @Michkov, I don't understand this.
If I move that line after `while` loop, it would break the script, because of the `shift`s in that `while` will shift the positional parameters.
The way it's now it stores the `MODE` first (this is the first mandatory parameter - I could maybe figure out some validation here though) and then shifts through remaining positional parameters in `while` loop.

Michkov: right, in that case could you put it into '*)'? So that it will be possible to run `./hack/bootstrap-cluster.sh --toolchain --keycloak preview`

.ci/oci-e2e-deployment.sh
rhopp: Good point. I'll remove the changes from here and I'll create separate PRs to update our centralized scripts (this PR shouldn't affect fucntionality of our scripts).

rhopp:
@Michkov @psturc I've adressed your concerns. Please take a look.

Michkov (approved):

PR #1016, opened 2022-12-07
Add SA, Role and RoleBindings for Internal Services Controller

scoheb:
- As part of HACBS-1443, we have developed an operator that will run on a private, internal cluster provided by AppSRE.
- This operator will reach out to the HACBS cluster to watch for InteralRequest CRs and execute internal pipelines.
- This PR created the objects needed for the operator to watch the CRs

### DIFF
--- a/components/authentication/internal-services.yaml
+++ b/components/authentication/internal-services.yaml
@@ -1,0 +1,16 @@
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: internal-services
+  namespace: internal-services
+subjects:
+  - kind: Group
+    apiGroup: rbac.authorization.k8s.io
+    name: Release team
+  - kind: User
+    apiGroup: rbac.authorization.k8s.io
+    name: Michkov
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: component-maintainer
--- a/components/internal-services/internal_service_request_role.yaml
+++ b/components/internal-services/internal_service_request_role.yaml
@@ -1,0 +1,32 @@
+# permissions for end users to view applications.
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: internal-service-request-role
+rules:
+  - apiGroups:
+      - appstudio.redhat.com
+    resources:
+      - internalrequests
+    verbs:
+      - create
+      - delete
+      - get
+      - list
+      - patch
+      - update
+      - watch
+  - apiGroups:
+      - appstudio.redhat.com
+    resources:
+      - internalrequests/finalizers
+    verbs:
+      - update
+  - apiGroups:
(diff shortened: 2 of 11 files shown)


components/internal-services/internal_service_request_role.yaml
```diff
@@ -0,0 +1,34 @@
+# permissions for end users to view applications.
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: internal-service-request-role
+  labels:
+    rbac.authorization.k8s.io/aggregate-to-view: 'true'
```
Michkov, line 7: I would not aggregate to view since there are verbs - delete, create, update, patch.

Michkov (changes requested):

components/internal-services/kustomization.yaml
```diff
@@ -0,0 +1,9 @@
+resources:
```
Michkov, line 1: please also include README.md (at leaste minimal) in components/internal-services.

components/internal-services/internal_service_request_role.yaml
scoheb: removed

components/internal-services/kustomization.yaml
scoheb: added

scoheb:
@Michkov please take another look.

scoheb:
test seems unrelated

Michkov (approved):

components/internal-services/kustomization.yaml
```diff
@@ -0,0 +1,9 @@
...
+- internal_service_request_role.yaml
+- internal_service_request_role_binding.yaml
+- internal_service_request_service_account.yaml
```
Michkov, line 4: `components/internal-services/internal_service_service_account_token.yaml` should be included as well

components/internal-services/internal_service_service_account_token.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: internal-service-request-service-account-secret
```
Michkov, line 4: Somebody should be able to read the token - please create record in `components/authentication` (I know that you are almighty admin but still :stuck_out_tongue_winking_eye:)

components/internal-services/kustomization.yaml
scoheb: added

components/internal-services/internal_service_service_account_token.yaml
scoheb: added

components/authentication/internal-services.yaml
```diff
@@ -0,0 +1,16 @@
+kind: RoleBinding
```
Michkov, line 1: not in kustomization.yaml

Michkov: fix pushed

PR #1044, opened 2022-12-14
Split Prometheus and Grafana components

xcoulon:
Prometheus should be deployed on each workload cluster, whereas Grafana
should be deployed on the "control-plane" cluster only, so it makes sense
to split the configuration and the READMEs


components/monitoring/prometheus/base/install-prometheus.yaml
```diff
@@ -20,3 +20,4 @@ spec:
   name: prometheus
   source: community-operators
   sourceNamespace: openshift-marketplace
+  startingCSV: prometheusoperator.0.56.3
```
Michkov, line 23: should not be there, on new installation it will install this version and will be doing updates, so it's better to not having it here -> so newest is just installed.

amfred: The fix is to just remove the line, @Michkov ?

Michkov: yes

components/monitoring/prometheus/base/kustomization.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- install-prometheus.yaml
+- prometheus-view.yaml
+- configure-prometheus.yaml
+- prometheus-servicemonitors.yaml
+
+namespace: "appstudio-workload-monitoring"
```
bamachrn, line 9: May be we should remove this namespace line. as this by default changes all the namespaces to monitoring namespace for all the components in the base directory. We may need to exclude the servicemonitors from this.

amfred (changes requested):

components/monitoring/prometheus/base/prometheus-servicemonitors.yaml
```diff
@@ -8,7 +8,7 @@ metadata:
 spec:
   endpoints:
   - bearerTokenSecret:
-      name: "prometheus-k8s-token-xhrjb"
+      name: "prometheus-k8s-token-md9md"
       key: token
```
bamachrn, line 12: Let change it back to `prometheus-k8s-token-xhrjb` as that is needed for the current deployment in stage

components/monitoring/prometheus/base/prometheus-servicemonitors.yaml
```diff
@@ -34,7 +34,7 @@ metadata:
 spec:
   endpoints:
   - bearerTokenSecret:
-      name: "prometheus-k8s-token-xhrjb"
```
bamachrn, line 37: same as above

components/monitoring/prometheus/base/prometheus-servicemonitors.yaml
```diff
@@ -60,7 +60,7 @@ metadata:
...
+      name: "prometheus-k8s-token-md9md"
       key: token
```
bamachrn, line 64: same as above

components/monitoring/prometheus/base/prometheus-servicemonitors.yaml
```diff
@@ -88,7 +88,7 @@ spec:
     port: https
     scheme: https
     bearerTokenSecret:
-      name: "prometheus-k8s-token-xhrjb"
+      name: "prometheus-k8s-token-md9md"
       key: token
```
bamachrn, line 92: same as above

bamachrn (changes requested):
Please update the secret name for prometheus sa

components/monitoring/prometheus/base/kustomization.yaml
xcoulon: I thought ServiceMonitors will be provided by operators in their own namespace, so we would not need the `prometheus-servicemonitors.yaml` at all?

components/monitoring/prometheus/base/install-prometheus.yaml
xcoulon: removed in e278564

components/monitoring/grafana/README.md
```diff
@@ -63,12 +57,21 @@ $ PROMETHEUS_URL=`oc get route/prometheus-oauth -n appstudio-workload-monitoring
...
-- The same token can be used in datasources secrets related to the Prometheus instances deployed in the `openshift-monitoring` and `appstudio-workload-monitoring` namespaces.
 
 Using the values obtained from the **Prometheus cluster**, run the following command on the **Grafana cluster**:
 
 ```
 $ ./hack/setup-monitoring.sh grafana-datasource-secret $DATASOURCE_NAME $PROMETHEUS_URL $GRAFANA_OAUTH_TOKEN
+```
+
+Notes: 
+- The `grafana-oauth` service account is created by `components/monitoring/base/prometheus/configure-prometheus.yaml` along with a binding to the `cluster-monitoring-view` cluster role.
```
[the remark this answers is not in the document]
xcoulon: I don't understand your question: when installing/configuring Prometheus, a Service Account named `grafana-oauth` will be created and used here to generate a token. Maybe the sentence in the README above is confusing?

components/monitoring/prometheus/base/prometheus-servicemonitors.yaml
xcoulon: my apologies, I had already noticed that unneeded change and reverted it but I suspect I forgot to push it earlier 🤷‍♂️

bamachrn: this PR is aleady doing that.

components/monitoring/prometheus/base/kustomization.yaml
bamachrn: we need that for scrapping prometheus it-self. But as we are already putting namespaces in the yamls. we can remove it anyway

xcoulon: as discussed on Slack, we'll keep this transformer for now.

components/monitoring/grafana/README.md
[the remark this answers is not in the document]
amfred: I wasn't understanding why a file called configure-prometheus.yaml was creating things for Grafana, but looking at it again now, it makes sense where it is.  
Nevermind, this looks good to me!

amfred (approved):

PR #1055, opened 2022-12-15
RBAC for composing Releases

scoheb:
- Enable the abilty for the Release team to compose Releases

### DIFF
--- a/components/authentication/kustomization.yaml
+++ b/components/authentication/kustomization.yaml
@@ -16,6 +16,7 @@ resources:
 - tekton-ci.yaml
 - internal-services.yaml
 - pipeline-service-sre.yaml
+- release-composer.yaml
 
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
--- a/components/authentication/release-composer.yaml
+++ b/components/authentication/release-composer.yaml
@@ -1,0 +1,51 @@
+kind: ClusterRoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: release-composer-releasestrategy
+subjects:
+  - kind: Group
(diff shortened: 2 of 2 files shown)


scoheb:
@johnbieren @Michkov please review

components/authentication/managed-workspaces.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
```
johnbieren, line 1: This file should be added in `kustomization.yaml` too, no?

scoheb: yep! good catch

scoheb: fixed

johnbieren (approved):

Michkov:
@ralphbean @MatousJobanek @sbose78 Do you think that managed-namespaces should be handled by infra-deployments?

@scoheb Do you have any ticket regarding this work which could help me to understand?

MatousJobanek:
There is an ongoing discussion in a slack thread. 
https://coreos.slack.com/archives/C02K1781K70/p1671111375323959
TBH, first of all, it needs to be clarified:
* what we are trying to do here
* why in this way
* what is missing in the StoneSoup functionality that we cannot follow the standard way of provisioning workspaces
* if there is anything missing, why do we have to hack the process and our cluster to provide functionality which is not supported yet

sbose78 (approved):
Valid questions by Matous, we should get to them without necessarily blocking this PR. Context would be great :)

Based on the slack thread, it doesn't look like Scott has any other way of accomplishing this, and if this helps us move the needle, so be it.

scoheb:
No longer needed. I will provide a summary shortly.

PR #1079, opened 2022-12-21
Feature toggle Rekor

zregvart:
This allows for customizing Tekton Chains transparency configuration and
correspondingly the Rekor instance. Three modes are now available:
 * off      - transparency off
 * local    - transparency on, with an in-cluster Rekor instance
 * sigstore - transparency on, with Sigstore public Rekor instance

These are set via ApplicationSet list generator and the custom "rekor"
parameter set there, see argo-cd-apps/base/enterprise-contract.yaml.

It is possible to change this parameter in a gitops manner and the
resulting state should reflect the value of the parameter, e.g. changing
from "local" to "off" should remove the "rekor" ArgoCD Application and
configure the "chains-config" ConfigMap accordingly.

The in-cluster deployed instance of Rekor no longer has an Ingress to
it's service, so it is no longer available outside of the cluster.

The `runAsUser` by default is set to 65533, so the annotations on the
`rekor` namespace are set to allow that. Still when the `rekor-server`
Pod is starting the following is logged:

```
spec:2022/12/20 19:56:39 warning: could not resolve current working directory: stat .: permission denied
```

The creation of`treeID` is forced (`createtree.forced=true`), found no
other way for Rekor not to fail with:

```
api/error.go:72	error processing request	{"requestID": "rekor-server-7fc8f96d54-gkbp6/YAfKPyHOLp-000029", "handler": "tlog.GetLogInfo", "statusCode": 500, "clientMessage": "unexpected error communicating with transparency log", "error": "grpc error: rpc error: code = NotFound desc = tree 1234567890 not found"}
```

This will cause data loss if the `rekor-createtree` Job is re-run,
creating a new `treeID`, and the `rekor-server` Pod is restarted to
reconfigure the new `treeID` from the `rekor-config` ConfigMap.

We can't manage `tekton-chans/chains-config` ConfigMap from the
`pipeline-service` as we need to modify it based on the state of the
Rekor mode from `enterprise-contract` Application, so the
`chains-config` ConfigMap was removed from the `pipeline-service` and
was added to `enterprise-contract` where it can be managed/patched.

### DIFF
--- a/argo-cd-apps/base/enterprise-contract.yaml
+++ b/argo-cd-apps/base/enterprise-contract.yaml
@@ -1,77 +1,52 @@
-apiVersion: argoproj.io/v1alpha1
-kind: Application
-metadata:
-  name: rekor
-
-spec:
-  project: default
-  source:
-    path: helm-charts/rekor
-    repoURL: 'https://github.com/sigstore/sigstore-helm-operator'
-    targetRevision: "143b8d71fc861d7c7ac93a15d0fbd33654b75908"
-
-    helm:
-      valueFiles:
-        - values.yaml
-      values: |-
-        server:
-          strategy:
-            type: Recreate
-          ingress:
-            hostname: rekor-server.enterprise-contract-service.svc
-            annotations:
-              route.openshift.io/termination: "edge"
-        mysql:
-          auth:
-            password: "password"
-
-  destination:
-    namespace: enterprise-contract-service
-    server: https://kubernetes.default.svc
-
-  syncPolicy:
-    # Comment this out if you want to manually trigger deployments (using the
-    # Argo CD Web UI or Argo CD CLI), rather than automatically deploying on
-    # every new Git commit to your directory.
-    automated:
-      prune: true
-      selfHeal: true
-
-    syncOptions:
-      - CreateNamespace=true
-
-    retry:
-      limit: 50 # number of failed sync attempt retries; unlimited number of attempts if less than 0
-      backoff:
-        duration: 15s # the amount to back off. Default unit is seconds, but could also be a duration (e.g. "2m", "1h")
 ---
 apiVersion: argoproj.io/v1alpha1
-kind: Application
+kind: ApplicationSet
 metadata:
   name: enterprise-contract
-
+  namespace: openshift-gitops
 spec:
-  project: default
-
-  source:
-    path: components/enterprise-contract
-    repoURL: https://github.com/redhat-appstudio/infra-deployments.git
-    targetRevision: main
-
-  destination:
-    namespace: enterprise-contract-service
-    server: https://kubernetes.default.svc
-
+  generators:
+    - list:
+        elements:
+          # Determines which Rekor (Tekton Chains transparency log) variant will
+          # be used, can be:
+          #  * off      - transparency off
+          #  * local    - transparency on, with an in-cluster Rekor instance
+          #  * sigstore - transparency on, with Sigstore public Rekor instance
+          - rekor: local
   syncPolicy:
+    # Allows us to delete the "rekor" Application if "rekor" above is not "local"
+    preserveResourcesOnDeletion: false
+  template:
+    metadata:
+      name: enterprise-contract
 
-    # Comment this out if you want to manually trigger deployments (using the
-    # Argo CD Web UI or Argo CD CLI), rather than automatically deploying on
-    # every new Git commit to your directory.
-    automated:
-      prune: true
-      selfHeal: true
+    spec:
+      project: default
+      source:
+        # Uses the overlay specified by "rekor" generator property above, this
+        # will setup in-cluster Rekor (if set to "local"), and change the
+        # chains-config ConfigMap based on that setting
+        path: components/enterprise-contract/rekor-{{rekor}}
+        repoURL: https://github.com/redhat-appstudio/infra-deployments.git
+        targetRevision: main
+      destination:
+        namespace: openshift-gitops
+        server: https://kubernetes.default.svc
 
-    retry:
-      limit: 50 # number of failed sync attempt retries; unlimited number of attempts if less than 0
-      backoff:
-        duration: 15s # the amount to back off. Default unit is seconds, but could also be a duration (e.g. "2m", "1h")
+      syncPolicy:
+
+        # Comment this out if you want to manually trigger deployments (using the
+        # Argo CD Web UI or Argo CD CLI), rather than automatically deploying on
+        # every new Git commit to your directory.
+        automated:
+          prune: true
+          selfHeal: true
+
+        syncOptions:
+          - CreateNamespace=true
+
+        retry:
+          limit: 50 # number of failed sync attempt retries; unlimited number of attempts if less than 0
+          backoff:
+            duration: 15s # the amount to back off. Default unit is seconds, but could also be a duration (e.g. "2m", "1h")
--- a/argo-cd-apps/base/enterprise-contract.yaml
+++ b/argo-cd-apps/base/enterprise-contract.yaml
@@ -13,7 +13,7 @@ spec:
           #  * off      - transparency off
           #  * local    - transparency on, with an in-cluster Rekor instance
           #  * sigstore - transparency on, with Sigstore public Rekor instance
-          - rekor: local
+          - rekor: off
   syncPolicy:
     # Allows us to delete the "rekor" Application if "rekor" above is not "local"
     preserveResourcesOnDeletion: false
--- a/components/enterprise-contract/base/chains-config.yaml
+++ b/components/enterprise-contract/base/chains-config.yaml
@@ -3,6 +3,9 @@ kind: ConfigMap
 metadata:
   name: chains-config
   namespace: tekton-chains
+  labels:
+    app.kubernetes.io/instance: default
+    app.kubernetes.io/part-of: tekton-chains
 
 data:
   # See https://tekton.dev/docs/chains/config/
--- a/components/enterprise-contract/rekor-local/rekor-namespace.yaml
+++ b/components/enterprise-contract/rekor-local/rekor-namespace.yaml
@@ -1,0 +1,9 @@
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: rekor
+  annotations:
+    # UID used by rekor deployment
+    openshift.io/sa.scc.uid-range: 65530-65535
+    openshift.io/sa.scc.mcs: s0:c26,c5
--- a/components/enterprise-contract/rekor-local/rekor.yaml
+++ b/components/enterprise-contract/rekor-local/rekor.yaml
@@ -1,0 +1,54 @@
+---
+apiVersion: argoproj.io/v1alpha1
+kind: Application
+metadata:
+  name: rekor
+  namespace: openshift-gitops
+
+spec:
+  project: default
+  source:
+    path: charts/rekor
+    repoURL: https://github.com/sigstore/helm-charts/
+    targetRevision: 46f53f756fa59760c251bf6bde98cac241096453
+
+    helm:
+      valueFiles:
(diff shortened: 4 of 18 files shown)


argo-cd-apps/base/enterprise-contract.yaml
```diff
@@ -1,77 +1,52 @@
...
+    metadata:
+      name: enterprise-contract
+
+    spec:
+      project: default
+      source:
+        # Uses the overlay specified by "rekor" generator property above, this
+        # will setup in-cluster Rekor (if set to "local"), and change the
+        # chains-config ConfigMap based on that setting
+        path: components/enterprise-contract/rekor-{{rekor}}
```
lcarva, line 30: I'm trying to understand this (so bear with me!).

Are you using an ApplicationSet just as a means to select a different path? And if so, to use a different mode of Rekor, we'd have to change the value in the generators list above? Wouldn't it be easier to just use a regular Application and document that to use a different mode, you must change the path to a different one?

argo-cd-apps/base/enterprise-contract.yaml
```diff
@@ -1,77 +1,52 @@
...
-
+  generators:
+    - list:
+        elements:
+          # Determines which Rekor (Tekton Chains transparency log) variant will
+          # be used, can be:
+          #  * off      - transparency off
+          #  * local    - transparency on, with an in-cluster Rekor instance
+          #  * sigstore - transparency on, with Sigstore public Rekor instance
+          - rekor: off
```
lcarva, line 16: What happens if I do this?
```yaml
  generators:
    - list:
        elements:
          - rekor: off
          - rekor: local
          - rekor: sigstore
```

components/enterprise-contract/base/chains-config.yaml
```diff
@@ -3,6 +3,9 @@ kind: ConfigMap
 metadata:
```
lcarva, line 3: Are we sure we want to move the Chains configuration from the pipeline-service to the enterprise-contract? It seems a little odd, and confusing, to have the Deployment and the configuration split like this.

components/enterprise-contract/rekor-local/rekor-namespace.yaml
```diff
@@ -0,0 +1,8 @@
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: rekor
+  annotations:
+    openshift.io/sa.scc.uid-range: 65530-65535
```
lcarva, line 7: Let's document why these annotations are needed. 

Also, is this a supported/documented approach, or is this something that is going to bite us later?

components/enterprise-contract/rekor-local/rekor.yaml
```diff
@@ -0,0 +1,52 @@
...
+    path: charts/rekor
+    repoURL: https://github.com/sigstore/helm-charts/
+    targetRevision: 46f53f756fa59760c251bf6bde98cac241096453
+
+    helm:
+      valueFiles:
+        - values.yaml
+      values: |-
+        createtree:
+          force: true
```
lcarva, line 20: It might be wroth leaving a comment here explaining why this is needed.

hack/chains/config.sh
```diff
@@ -84,8 +84,8 @@ case "$1" in
     ;;
 
   rekor-local )
```
lcarva, line 86: Aren't the `rekor-*` targets irrelevant now?

hack/preview.sh
```diff
@@ -71,10 +71,13 @@ fi
...
 # patch argoCD applications to point to your fork
-oc kustomize $ROOT/argo-cd-apps/overlays/development | yq e "select(.spec.source.repoURL==\"https://github.com/redhat-appstudio/infra-deployments.git\")
-  | del(.spec.destination, .spec.syncPolicy, .spec.project, .spec.source.path, .metadata.namespace)
-  | .spec.source.repoURL=\"$MY_GIT_REPO_URL\"
-  | .spec.source.targetRevision=\"$PREVIEW_BRANCH\"
+oc kustomize $ROOT/argo-cd-apps/overlays/development | yq e "select(.spec.source.repoURL==\"https://github.com/redhat-appstudio/infra-deployments.git\" or .spec.template.spec.source.repoURL==\"https://github.com/redhat-appstudio/infra-deployments.git\")
+  | del(.metadata.namespace, .spec.destination, .spec.syncPolicy, .spec.project, .spec.source.path, .spec.template.spec.destination, .spec.template.spec.syncPolicy, .spec.template.spec.project, .spec.template.spec.source.path)
+  | (.spec | select(has(\"source\")) | .source.repoURL)=\"$MY_GIT_REPO_URL\"
+  | (.spec | select(has(\"source\")) | .source.targetRevision)=\"$PREVIEW_BRANCH\"
+  | (.spec | select(has(\"template\")) | .template.spec.source.repoURL)=\"$MY_GIT_REPO_URL\"
```
lcarva, line 78: Is this change needed to take into account the ApplicationSet structure?

lcarva (commented):
This is quite a complex approach.

What's the workflow here for switching between the different modes of Rekor?

argo-cd-apps/base/enterprise-contract.yaml
zregvart: Regular Applications don’t garbage collect, so switching from “rekor-local” to “off” would keep the Rekor running

argo-cd-apps/base/enterprise-contract.yaml
zregvart: Don’t, ArgoCD would infinity loop over the chains-config ConfigMap

components/enterprise-contract/base/chains-config.yaml
zregvart: Can’t manage it otherwise, would lead to having two applications compete over it. I tried `replacements`, that works only within the same kustomization. The idea there was to have a ConfigMap in `enterprise-contract-Service` and lookup values from `tekton-chains`, that does not seem possible.

components/enterprise-contract/rekor-local/rekor-namespace.yaml
zregvart: The only way I can make it work with Rekor helm charts, by default each namespace gets a range of UIDs, no idea what they are or how to pass to the helm chart.

hack/preview.sh
zregvart: Yeah, I tried to do it using `with` but lucked out

argo-cd-apps/base/enterprise-contract.yaml
lcarva: I think that's a pretty good indicator that this is not the right tool for the job...

All we need is a single Application to deploy Rekor locally. The other modes are not deploying Rekor. They're just manipulating the Chains configuration. I also think it should be perfectly valid to have Rekor running, but have Chains not using it, especially during development and debugging phases. I really disagree with coupling the Rekor deployment to the Chains configuration management. The Rekor service shouldn't care if anyone is using it, that's a Chains concern.

We should have a specific Chains config and specific mode for Rekor in staging. For development/preview mode we could have a default, and leverage an env var in preview.env to switch the Rekor mode.

I still don't see the benefits of using an ApplicationSet. In fact, it seems like it makes it easier for a developer to shoot themselves in the foot.

simonbaird:
I looked over it, but I don't think I can contribute a lot of useful feedback.

That said, it does seem quite complicated, and I am also not sure it's wise to pull chains config back from pipeline-service ownership.

simonbaird:
Re the "feature toggle" concept, maybe we don't need to have it quite so high-tech, i.e. fully toggleable on a running system with single gitops config change might be overkill. E.g. if removing Rekor from a running system took some (well understood, well documented) manual steps, I think that's probably okay.

Can we build something less good here and maybe remove some of the complexity?

argo-cd-apps/base/enterprise-contract.yaml
zregvart: I don't think there is any value in having Rekor running without Tekton Chains using it. After all if this is something that one wishes to have, what's stopping them from deploying Rekor manually? _Validity_ of this is more of a personal preference, for Application Studio there is no other use for Rekor other than to store entries created by Tekton Chains. And that use case is, as far as I understand, still under debate. If there were any other use cases for Rekor, they would surely be voiced by now.

The benefit of using ApplicationSet is that it allows switching between these three different modes without having anything leftover or inconsistent upon making the switch. There is no other way that I know of doing this with ArgoCD.

If the debate is around having three modes, usefulness of them or technical implementation I would welcome this, but this to me seems rather pointless to debate.

zregvart:
With regards to feature toggle: this PR introduces it, what downsides of using it are there? One that is being voiced is complexity, I fail to see any. The only alternative that I could think of would be to have commented out section in `kustomization.yaml` to allow deploying Rekor. I don't think that is any better than what is done here.

With regards to managing `chains-config` from `enterprise-contract-service`: Rekor and Tekton Chains are inevitably linked together by the architecture of App Studio, there is no use for Rekor other than to store attestations produced by Tekton Chains. This allows us to deploy three different kinds of clusters by changing one value in the configuration, allowing for flexibility of business requirements. This is **exactly** what we require with the uncertainty we have around Rekor usage.

I fail to see how any other outcome would not be a half-solution to the problem.

Roming22:
> I looked over it, but I don't think I can contribute a lot of useful feedback.
> 
> That said, it does seem quite complicated, and I am also not sure it's wise to pull chains config back from pipeline-service ownership.

I agree with the second part of this statement. I believe a PR should be opened against [pipeline-service](https://github.com/openshift-pipelines/pipeline-service) to implement the easy configuration of the transparency mode and deployment of rekor. A simple PR should then be submitted to this repo to add/edit the required `kustomization.yaml` to select the desired behavior.

argo-cd-apps/base/enterprise-contract.yaml
lcarva: My point is that Rekor is a standalone component. At its core, Rekor couldn't care less if Chains is using it or not. On the other hand, Chains should be aware of Rekor if it intends to use it. This is the natural relationship between these two components. I think we should avoid murking that relationship.

My other point is that by using ApplicationSets in this case, it's very easy for someone to do something that puts their cluster in an unstable position, as you mentioned earlier. I think we should avoid knowingly adding such things.

Also, I disagree that toggling off Rekor should get rid of the deployment because this is yet another way to mess things up. What if someone decides that we should turn off Rekor for a little bit, for whatever reason? It would be a bad experience if that also deleted data. I think once Rekor is deployed, any sort of removal should be manual.

zregvart: > My point is that Rekor is a standalone component. At its core, Rekor couldn't care less if Chains is using it or not. On the other hand, Chains should be aware of Rekor if it intends to use it. This is the natural relationship between these two components. I think we should avoid murking that relationship.

In theory it is, in practice it is not. Most definitely by everything we are doing here Rekor is tied to Chains. There is no Rekor without Chains.

> My other point is that by using ApplicationSets in this case, it's very easy for someone to do something that puts their cluster in an unstable position, as you mentioned earlier. I think we should avoid knowingly adding such things.

I don't follow this argument, it is very easy to make other mistakes that put cluster in unstable state. We should not be bound by what can be done by mistake but what we enable as functionality.

> Also, I disagree that toggling off Rekor should get rid of the deployment because this is yet another way to mess things up. What if someone decides that we should turn off Rekor for a little bit, for whatever reason? It would be a bad experience if that also deleted data. I think once Rekor is deployed, any sort of removal should be manual.

This argument doesn't make sense to me. It is not intended for this to be used to "turn off Rekor for a little bit". The intent for this is to deploy and configure the use of Rekor when the cluster is provisioned. By this line of thinking what if someone decides to delete all storage of a cluster _for a little bit for whatever reason_.

PR #1092, opened 2022-12-26
PLNSRVCE-934: remove hacbs param from pipeline selector config

tkdchen:


components/build-service/build-pipeline-selector.yaml
```diff
@@ -9,25 +9,25 @@ spec:
         name: docker-build
         bundle: quay.io/redhat-appstudio-tekton-catalog/pipeline-docker-build:3809d7784aa17aed2426c8f440008e83f62d64c6
       pipelineParams:
-        - name: hacbs
-          value: "true"
+        - name: skip-checks
+          value: "false"
```
Michkov, line 13: skip-checks is handled directly by build-service so we should not have it defined in the selector.

PR #1101, opened 2023-01-03
PLNSRVCE-885: build-selector - add FBC pipeline

Michkov:
Select fbc-builder pipeline when language in devfile is set to fbc.


Michkov:
based on #1099

PR #1112, opened 2023-01-05
ignore diff in imagePullSecrets in tekton-ci

Michkov:
With new imagePullSecrets the argocd marked pipeline service account as OutOfSync. This change is ignoring secrets in list which starts 'pipeline-dockercfg'


argo-cd-apps/base/tekton-ci.yaml
```diff
@@ -14,6 +14,13 @@ spec:
     namespace: tekton-ci
     server: https://kubernetes.default.svc
 
+
+  ignoreDifferences:
+  - group: ""
+    kind: "ServiceAccount"
```
lcarva, line 20: Should we set `name` here as well since this is only applicable to the `pipeline` service account anyways?

lcarva (approved):
/lgtm
/approve

simonbaird (approved):

PR #1113, opened 2023-01-05
Updated README to make it explicit and grafana oauth token non expiring

bamachrn:
This PR 
* updates the grafana oauth token to make it non-expiring till tha grafana sa is explicitly deleted.
* updated manual commads to make it suitable for current stage setup


xcoulon (approved):
Awesome! I like the idea of binding the token to an existing secret, so we don't have to deal with expiration ;)

amfred (approved):
/lgtm

PR #1144, opened 2023-01-11
Provision SPI's grafana dashboard

skabashnyuk:
- Service provider's dashboard 
 - SPI components state  dashboard
 - Dedicate to SPI's SLO dashboard

We decided to use individual configmaps per dashboard to make sure we are not hitting the limit 1Mb per configmap.


bamachrn (commented):
we are just creating the structure here right? we are not putting the dashboard definitions.

bamachrn (approved):
this makes sense

PR #1180, opened 2023-01-20
Added token expiration duration to readme

bamachrn:
This PR updates the token duration for grafana datasource authentications


xcoulon (approved):

amfred:
Setting a long expiration time on a token is a useful hack, but I don't want to recommend it.  We really need to figure out how to automatically rotate it correctly.

Omeramsc:
> Setting a long expiration time on a token is a useful hack, but I don't want to recommend it. We really need to figure out how to automatically rotate it correctly.

Can [this](https://kubernetes.io/docs/reference/access-authn-authz/service-accounts-admin/#create-token) be a possible solution?  or using an operator/sidecar container to renew the token

xcoulon:
> > Setting a long expiration time on a token is a useful hack, but I don't want to recommend it. We really need to figure out how to automatically rotate it correctly.
> 
> Can [this](https://kubernetes.io/docs/reference/access-authn-authz/service-accounts-admin/#create-token) be a possible solution? or using an operator/sidecar container to renew the token

this looks interesting, indeed!

PR #1201, opened 2023-01-23
Fix tekton-chains resource limits and set a new signing-secret

Mo3m3n:


ramessesii2 (approved):

Michkov (approved):

scoheb:
Changing the tekton chains secret will cause other services to fail without rotating the pub secret. Can you please announce this on slack?

Michkov:
@scoheb yop, planned. Basically old key was lost with https://github.com/redhat-appstudio/infra-deployments/pull/982

Mo3m3n:
old keys were lost before https://github.com/redhat-appstudio/infra-deployments/pull/982, however redeploying the controller made the issue visible and there was no backup of the old keys. New keys have now a backup copy.

PR #1248, opened 2023-01-30
fix: build-via-appstudio.sh - link secret

Michkov:
Registry-auth support was removed in [PLNSRVCE-910](https://issues.redhat.com//browse/PLNSRVCE-910) and it's replaced by secrets linked to the serviceaccount.


psturc (approved):
verified

I don't think we need to wait for CI to pass, since the updated script is not related to e2e tests

PR #1270, opened 2023-02-02
Improve support for overriding SPI images

scoheb:
- Add OAUTH image and tag support


scoheb:
FYI @redhat-appstudio/spi-team

PR #1273, opened 2023-02-03
grafana: move component code to subfolders

Michkov:
To avoid git conflicts moving grafana configurations of services to sub-folders.

Action required -> if you have update set in .tekton folder then update file path.


zregvart (approved):

Michkov:
@skabashnyuk @scoheb Please update your .tekton push.yaml with new path.

scoheb:
Created redhat-appstudio/release-service#167

PR #1289, opened 2023-02-06
Update pipeline-service

Roming22:


Mo3m3n:
@Roming22 can you add to the commit message why the update and what to expect from it ?
This makes it easier to debug and maintain these changes

Roming22:
> @Roming22 can you add to the commit message why the update and what to expect from it ? This makes it easier to debug and maintain these changes

Done

Michkov:
I think that this PR is not needed, multi-cluster branch is synced with branch so when it's merged to main then it's also synced to multi-cluster. CC @gbenhaim

PR #1304, opened 2023-02-09
Added ServiceMonitor for pipeline-service

bnallapeta:
- This PR goes along with https://github.com/openshift-pipelines/pipeline-service/pull/473 opened on Pipeline Service.
- Adds the required Service Monitor and cluster role binding in order to scrape metrics from the custom pipeline service metrics exporter.

### DIFF
--- a/components/monitoring/prometheus/base/servicemonitors/pipeline-service.yaml
+++ b/components/monitoring/prometheus/base/servicemonitors/pipeline-service.yaml
@@ -1,0 +1,33 @@
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  name: pipeline-service
+  namespace: appstudio-workload-monitoring
+  labels:
+    app: pipeline-metrics-exporter
+    prometheus: appstudio-workload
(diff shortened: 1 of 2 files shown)


components/monitoring/prometheus/base/servicemonitors/pipeline-service.yaml
```diff
@@ -0,0 +1,33 @@
...
+  labels:
+    app: pipeline-metrics-exporter
+    prometheus: appstudio-workload
+spec:
+  selector:
+    matchLabels:
+      app: pipeline-metrics-exporter
+  endpoints:
+    - path: /metrics
+      port: metrics
```
bamachrn, line 15: this port matches the service exporter port right?

bnallapeta: Yes, [here](https://github.com/openshift-pipelines/pipeline-service/blob/main/operator/gitops/argocd/pipeline-service/metrics-exporter/service.yaml#L11) is the link for it

PR #1288, opened 2023-02-06
Update pipeline-service

Roming22:
* Expose the `public-key` secret in the `tekton-chains` namespace to all users of the clusters to allow users to decode tekton-chains signed payloads.

Signed-off-by: Romain Arnaud <rarnaud@redhat.com>

PR #1321, opened 2023-02-13
Update pipeline service to have the custom exporter changes

bnallapeta:
This will also include changes to expose the public-key secret in the tekton-chains namespace to all users of the clusters to allow users to decode tekton-chains signed payloads.


Michkov:
https://github.com/redhat-appstudio/infra-deployments/pull/1288 should be closed since this is replacement?

bnallapeta:
> #1288 should be closed since this is replacement?

Just closed #1288.

PR #1333, opened 2023-02-14
Ensure openshift-cluster-csi-drivers exists

jduimovich:
This pull request includes two changes and apply to preview mode.

1. For some configurations (new version of CRC), the  `openshift-cluster-csi-drivers` namespace is missing, due to the Storage capability not being enabled.  This pull request enables the storage capability which creates the namespace.  The namespace is used by the pipelines service as it installs a csi driver.  
2. On CRC/WSL there are issues reading from `/dev/stdin`  for the patch-file due to windows/wsl interactions so this change updates the install script to workaround the issue via an intermediate temp file for the patch-file parameter.


hack/preview.sh
```diff
@@ -31,6 +41,7 @@ spec:
         authClientLibraryURL: $RHSSO_URL/auth/js/keycloak.js
         authClientPublicKeysURL: $RHSSO_URL/auth/realms/redhat-external/protocol/openid-connect/certs
 EOF
+  oc patch ToolchainConfig/config -n toolchain-host-operator --type=merge --patch-file=$AUTH
```
Michkov, line 44: remove $AUTH file

Michkov (approved):

Michkov:
Shared secrets which were using CSI driver were removed in #1836, closing this PR since it should not be needed anymore.

PR #1338, opened 2023-02-14
move to 4.12 version of sharead resources, including official images; add reminders for moves to OCP 4.13, 4.14

gabemontero:
I could not find any jira trackers for this yet (not surprising, 4.13 does not GA until April) but I wanted to register some placeholders somewhere to possibly help remind us when the time comes

### DIFF
--- a/components/shared-resources/globally-shared-secrets-clusterrole.yaml
+++ b/components/shared-resources/globally-shared-secrets-clusterrole.yaml
@@ -13,6 +13,7 @@ rules:
       - redhat-appstudio-user-workload
     verbs:
       - use
+# the rule to use the csi-scc can be removed when we move to 4.13 OCP
   - verbs:
       - use
     apiGroups:
--- a/components/shared-resources/kustomization.yaml
+++ b/components/shared-resources/kustomization.yaml
@@ -1,19 +1,25 @@
 resources:
 # when stonesoup rebases on OCP 4.14, shared resources will be part of the GA install; as such, if we are still using shared resources,
 # all these raw.githubusercontent refs to install shared resources can go away
-- https://raw.githubusercontent.com/openshift/csi-driver-shared-resource-operator/c60295f4e59c0d3f17fd1c5b920296ec58231db1/assets/config_configmap.yaml
-- https://raw.githubusercontent.com/openshift/csi-driver-shared-resource-operator/c60295f4e59c0d3f17fd1c5b920296ec58231db1/assets/csidriver.yaml
-- https://raw.githubusercontent.com/openshift/csi-driver-shared-resource-operator/c60295f4e59c0d3f17fd1c5b920296ec58231db1/assets/metrics_service.yaml
-- https://raw.githubusercontent.com/openshift/csi-driver-shared-resource-operator/c60295f4e59c0d3f17fd1c5b920296ec58231db1/assets/node_sa.yaml
-- https://raw.githubusercontent.com/openshift/csi-driver-shared-resource-operator/c60295f4e59c0d3f17fd1c5b920296ec58231db1/assets/node.yaml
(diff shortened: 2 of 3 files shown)


gabemontero:
per the arch call today, I'll open some jira's for failures seen here, but I want to circle back on the exact process desired before I do.

also, I wasnt to see if certain failures are flakes or perm fails.

For ^^ 

1) the one test case failure 

```
 Summarizing 1 Failure:
  [FAIL] [e2e-demos-suite] create an nodejs application with dockerfile and devfile [It] waits public component nodejs-dockerfile pipeline to be finished [e2e-demo]
  /tmp/tmp.pizHlaARPT/tests/e2e-demos/e2e-demo.go:191
```

2) from the ginkgo output, the pipeline run that failed was 

```
devfile-sample-bqas-tlc4l
```

3) could not find any pipeline run logs, but the pipeline run json had this log snippet

```
"message": "The step \"clone\" in TaskRun \"devfile-sample-bqas-tlc4l-clone-repository\" failed to pull the image \"\". The pod errored with the message: \"Back-off pulling image \"registry.redhat.io/openshift-pipelines/pipelines-git-init-rhel8:v1.8.2-8@sha256:a538c423e7a11aae6ae582a411fdb090936458075f99af4ce5add038bb6983e8\".\"",
```

gabemontero:
so the nodejs failure wrt the clone step missing its image is flake ^^

adambkaplan:
@gabemontero when we upgrade to 4.13, is there a way that we can disable the shared resource CSI driver initially?

gabemontero:
> @gabemontero when we upgrade to 4.13, is there a way that we can disable the shared resource CSI driver initially?

no

gabemontero:
/hold

some hypershift dependencies have most likely resulted in leaving shared resources in tech preview for 4.13

I'll report back when I know more

gabemontero:
it is official - while there is an accepted "it runs on hypershift" version of shared resources, in correspondence with storage and hypershift teams, we have deemed that GA level integration between shared resources and hypershift will not get to GA level by 4.13 feature freeze.

So **IF** we continue to use shared resources in appstudio/stonesoup, we will still need to install the driver like we are today

One improvement that will be in 4.13 (via new storage team admission plugins) is that shared resources is a trusted CSI driver, and as such, we no longer need the custom SCC aggregated.  To pick that up, we'll need to bump the csi-driver-shared-resource-operator commit level to `6bf074077d1d1df13caa1b146d733519633b74b4` and then we can delete the related SCC and clusterroles around that.

Pending when a jira for 4.13 migration lands on or queues, I'll update this PR to at least capture all this in the comments here.  I'll post again once I've rebased and done all that.

@sbose78 @adambkaplan @Michkov FYI

gabemontero:
hmm ... driver is fine, and the webhook was launched, though its log is unreadable

and ultimately gitops complained about shared resources being unstable 

trying to sort through things in the must gather, plus trying to test locally (though I did not bring up a hypershift cluser)

gabemontero:
yeah shared resources comes up fine in my non-hypershift local cluster

gabemontero:
so yep e2e's pass with shared resources at 4.12 without the webhook.

I was also able bring up a 4.12 hypershift cluster and reproduce the situation with the webhook.  Bottom line, we won't use the webhook until we get shared resources via OCP GA install, and we no longer install shared resources ourselves.

the driver enforces the same restrictions, just and pod provisioning type vs. obj creation time.

pushing some clean up, then will unhold, request tagging for merge

adambkaplan (approved):
/lgtm

gabemontero:
an image pull flake during cloning  in a build-service python test:

```
"message": "The step \"clone\" in TaskRun \"test-component-lbiu-599k9-clone-repository\" failed to pull the image \"\". The pod errored with the message: \"Back-off pulling image \"registry.redhat.io/openshift-pipelines/pipelines-git-init-rhel8:v1.8.2-8@sha256:a538c423e7a11aae6ae582a411fdb090936458075f99af4ce5add038bb6983e8\".\"",
```

/retest

PR #1364, opened 2023-02-16
WIP - Update tekton-results database secret name

avinal:
## Changes

**NOTE: this is a breaking change. Needs openshift-pipelines/pipeline-service#504**

The latest changes to the tekton-results manifests in pipeline-service will use `tekton-results-postgres` as the secret name. 

## Action Items

The data fields provided to this secret must follow the naming convention as provided here: https://github.com/bitnami/containers/blob/main/bitnami/postgresql/15/debian-11/rootfs/opt/bitnami/scripts/postgresql-env.sh.

### Required Changes to `tekton-results-postgres` secret's data fields

Refer: https://github.com/openshift-pipelines/tektoncd-results/blob/downstream/config/api.yaml#L38-L53

```
db.name -> DB_NAME
db.host -> DB_HOST
db.password -> POSTGRES_PASSWORD
db.user -> POSTGRES_USER
```

### DIFF
--- a/components/pipeline-service/staging/base/external-secrets/tekton-results/tekton-results-database.yaml
+++ b/components/pipeline-service/staging/base/external-secrets/tekton-results/tekton-results-database.yaml
@@ -1,7 +1,7 @@
 apiVersion: external-secrets.io/v1beta1
 kind: ExternalSecret
 metadata:
-  name: tekton-results-database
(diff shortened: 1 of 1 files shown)


avinal:
Local deployment is only required for development and patches can be used to achieve the desired manifests. Closing because no longer required.

PR #1368, opened 2023-02-17
docs: fixed wrong url for development config

tnevrlka:
I tried to click on the development configuration link on [preview mode for your clusters section](https://redhat-appstudio.github.io/infra-deployments/docs/development/deployment.html#preview-mode-for-your-clusters) which pointed me to a [non-existing link](https://redhat-appstudio.github.io/infra-deployments/argo-cd-apps/overlays/development/). 

I have corrected the url.

### DIFF
--- a/docs/development/deployment.md
+++ b/docs/development/deployment.md
@@ -30,7 +30,7 @@ To enable development for a team or individual to test changes on your own clust
 
 There are a set of scripts that help with this, and minimize the changes needed in your forks.
 
-The [development configuration](../../argo-cd-apps/overlays/development) includes a kustomize overlay that can redirect the default components individual repositories to your fork.
+The [development configuration](https://github.com/redhat-appstudio/infra-deployments/tree/main/argo-cd-apps/overlays/development) includes a kustomize overlay that can redirect the default components individual repositories to your fork.
 The script also supports branches automatically. If you work in a checked out branch, each of the components in the overlays will mapped to that branch by setting `targetRevision:`.
 
 Preview mode works in a feature branch, apply script which creates new preview branch and create additional commit with customization.


docs/development/deployment.md
```diff
@@ -30,7 +30,7 @@ To enable development for a team or individual to test changes on your own clust
 
 There are a set of scripts that help with this, and minimize the changes needed in your forks.
 
-The [development configuration](../../argo-cd-apps/overlays/development) includes a kustomize overlay that can redirect the default components individual repositories to your fork.
+The [development configuration](https://github.com/redhat-appstudio/infra-deployments/tree/main/argo-cd-apps/overlays/development) includes a kustomize overlay that can redirect the default components individual repositories to your fork.
```
Michkov, line 33: it's better to have it relative, so it works properly also in forks.

tnevrlka: The problem is with github.io website which does not show directories afaik. It is done similarly to my way in https://github.com/redhat-appstudio/infra-deployments/blob/07d3ebacf22f17e5f96f4cabac32d95c000ec82c/docs/development/spi.md?plain=1#L7.

Michkov: your example is pointing to different git repository

Michkov: ok, I see it now

Michkov (approved):

docs/development/deployment.md
tnevrlka: fyi: there are different examples that point to this repo https://github.com/redhat-appstudio/infra-deployments/blob/07d3ebacf22f17e5f96f4cabac32d95c000ec82c/docs/deployment/extending-the-service.md?plain=1#L60

PR #1369, opened 2023-02-17
Fix source path for quality-dashboard

gbenhaim:
Fix source path for quality-dashboard


gbenhaim:
@flacatus with this fix quality-gate should work.

PR #1384, opened 2023-02-20
[WIP]Add logging with fluentd to cloudwatch

bamachrn:
This PR
* installs cluster-logging operator to the `openshift-logging` namespace
* configures 
     * fluentd log collectors deployed in each of the nodes
     * cluster log forwarder to forward log from fluentd to cluoudwatch with region set to `us-east-1`
     * aws authentication is set using external secret
 
* this PR `does not enable` the high resource comsuming elasticsearch and kibana by default.
* We are also going from fluentbit to fluentd for logforwarding due to fluentbit yamls are not ready for production yet 
https://docs.fluentbit.io/manual/administration/configuring-fluent-bit/yaml


gbenhaim:
@bamachrn you would also need to add changes similar to what I added in https://github.com/redhat-appstudio/infra-deployments/pull/1385/files

components/monitoring/logging/base/configure-logging.yaml
```diff
@@ -0,0 +1,34 @@
+---
+apiVersion: "logging.openshift.io/v1"
+kind: "ClusterLogging"
+metadata:
+  name: "instance"
+  namespace: "openshift-logging"
```
gbenhaim, line 6: please add `argocd.argoproj.io/sync-options: SkipDryRunOnMissingResource=true` annotation

https://argo-cd.readthedocs.io/en/stable/user-guide/sync-options/#skip-dry-run-for-new-custom-resources-types

components/monitoring/logging/base/configure-logging.yaml
```diff
@@ -0,0 +1,34 @@
...
+spec:
+  collection:
+    logs:
+      fluentd: {}
+      type: fluentd
+---
+apiVersion: logging.openshift.io/v1
+kind: ClusterLogForwarder
+metadata:
+  name: instance
```
gbenhaim, line 16: please add `argocd.argoproj.io/sync-options: SkipDryRunOnMissingResource=true` annotation

https://argo-cd.readthedocs.io/en/stable/user-guide/sync-options/#skip-dry-run-for-new-custom-resources-types

components/monitoring/logging/base/configure-logging.yaml
```diff
@@ -0,0 +1,34 @@
...
+apiVersion: "logging.openshift.io/v1"
+kind: "ClusterLogging"
+metadata:
+  name: "instance"
+  namespace: "openshift-logging"
+spec:
+  collection:
+    logs:
+      fluentd: {}
+      type: fluentd
```
gbenhaim, line 11: @amfred so it's ok that we are not using fluent bit?

components/monitoring/logging/base/configure-logging.yaml
gbenhaim, line 6: please include the namespace in the kustomization instead.

components/monitoring/logging/base/configure-logging.yaml
```diff
@@ -0,0 +1,34 @@
+---
+apiVersion: "logging.openshift.io/v1"
+kind: "ClusterLogging"
+metadata:
+  name: "instance"
```
gbenhaim, line 5: can we have a descriptive name?

components/monitoring/logging/base/configure-logging.yaml
gbenhaim, line 16: can we have a descriptive name?

components/monitoring/logging/base/configure-logging.yaml
```diff
@@ -0,0 +1,34 @@
...
+  namespace: "openshift-logging"
+spec:
+  collection:
+    logs:
+      fluentd: {}
+      type: fluentd
+---
+apiVersion: logging.openshift.io/v1
+kind: ClusterLogForwarder
```
gbenhaim, line 14: I think that this should be on its own file in a dedicated directory (base/cluster-log-forwarder/cloudwatch.yaml), and include it in the resources of the staging kustomization.
This will allow us to deploy the logging operator in dev environments without configuring forwarding.

components/monitoring/logging/base/install-logging-operator.yaml
```diff
@@ -0,0 +1,58 @@
...
+kind: Subscription
+metadata:
+  name: "elasticsearch-operator"
+  namespace: "openshift-operators-redhat" 
+spec:
+  channel: "stable-5.5" 
+  installPlanApproval: "Automatic" 
+  source: "redhat-operators" 
+  sourceNamespace: "openshift-marketplace"
+  name: "elasticsearch-operator"
```
gbenhaim, line 37: why do we need elastic search? we forward the logs to cloudwatch https://docs.openshift.com/container-platform/4.12/logging/config/cluster-logging-collector.html#cluster-logging-removing-unused-components-if-no-elasticsearch_cluster-logging-collector

components/monitoring/logging/base/configure-logging.yaml
```diff
@@ -0,0 +1,34 @@
...
+metadata:
+  name: instance
+  namespace: openshift-logging
+spec:
+  outputs:
+    - cloudwatch:
+        groupBy: logType
```
gbenhaim, line 21: I think it would be better to use `namespaceName` so we will also group logs by their origin namespace.

gbenhaim (changes requested):

components/monitoring/logging/base/configure-logging.yaml
bamachrn: we can not change this name. this is fixed by the clusterlogging operator. https://docs.openshift.com/container-platform/4.10/logging/cluster-logging-deploying.html

components/monitoring/logging/base/configure-logging.yaml
bamachrn: moving to another file makes sense. but we are using logging operator only for logforwarding here. we are not using the elasticsearch and kibana.

components/monitoring/logging/base/configure-logging.yaml
bamachrn: yeah. but I believe we are more interested on the application logs. As audit and infra logs would be available from sre @amfred how do you think on this?

components/monitoring/logging/base/configure-logging.yaml
bamachrn: we can not change this name. this is fixed by the clusterlogging operator. https://docs.openshift.com/container-platform/4.12/logging/cluster-logging-deploying.html

components/monitoring/logging/base/configure-logging.yaml
gbenhaim: I see that you are also forwarding audit an infra logs. What do you mean that it will be available from sre?

components/monitoring/logging/base/configure-logging.yaml
amfred: Yes, fluent bit wasn't working and it's not well supported yet.

components/monitoring/logging/base/install-logging-operator.yaml
amfred: We don't need elasticsearch.

bamachrn: we need this for deploying the fluentd collectors. we are not running the elasticsearch pods. But fluentd is getting deployed as elasticsearch log colletctor.

bamachrn: we are using the above documentation link. thats why we are deploying only fluentd in [configure-logging.yaml](https://github.com/redhat-appstudio/infra-deployments/pull/1384/files#diff-7ba5c34961c5175a02d801bc5b8d296c62cb25ff83cbb06c76e6493505ae696e)

components/monitoring/logging/base/configure-logging.yaml
bamachrn: yes, My plan is remove the audit and infra logforwarding once we know we can access the audit logs from sre. may be lets make it to groupby `namespaceName` to make it similar to the host cluster log forwarding setup by sre team.

components/monitoring/logging/base/install-logging-operator.yaml
bamachrn: let me see if we can get collectors working without the elasticsearch operator.

components/monitoring/logging/base/configure-logging.yaml
gbenhaim: SRE doesn't have any connection to our member clusters, so I don't think we can count on them for infra and audit logs.

PR #1385, opened 2023-02-20
Support overlays in the Prometheus app

gbenhaim:
No that we have overlays for the Prometheus component, we need to update the matching argo app to support them.


bamachrn (approved):

bamachrn:
@gbenhaim please rebase this on top of main.

PR #1391, opened 2023-02-22
Add scope for a client.

Katka92:
As a new change in chrome is introduced, now it is expected that keycloak has a scope `nameandterms` which is not by default present in a keycloak. Adding it to the default scopes of a `cloud-services` client in keycloak.

PR introducing breaking change: https://github.com/RedHatInsights/insights-chrome/pull/2284/files
Temporary fix before this PR is merged: https://github.com/RedHatInsights/insights-chrome/pull/2289

### DIFF
--- a/components/dev-sso/keycloak-realm.yaml
+++ b/components/dev-sso/keycloak-realm.yaml
@@ -21,6 +21,11 @@ spec:
         webOrigins:
           - '*'
         directAccessGrantsEnabled: true
+        defaultClientScopes: 
+          - nameandterms
+    clientScopes:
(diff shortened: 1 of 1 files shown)


flacatus:
@Katka92 seems like is broking our e2e tests this changes. What exactly is doing your changes?

flacatus:
`"the server has asked for the client to provide credentials"`

When im using the users kubeconfig generated with the new realm I see: "the server has asked for the client to provide credentials"

flacatus:
Not sure exactly where is the root of this problem...

PR #1395, opened 2023-02-22
Update group-sync configurations

gbenhaim:
- Remove the sync of redhat-appstudio-sre, it's not needed.
- Update the secret patch in the production overlay.

### DIFF
--- a/components/authentication/production/github-redhat-appstudio-secret-patch.yaml
+++ b/components/authentication/production/github-redhat-appstudio-secret-patch.yaml
@@ -1,0 +1,4 @@
+---
+- op: add
+  path: /spec/dataFrom/0/extract/key
+  value: production/group-sync/github-redhat-appstudio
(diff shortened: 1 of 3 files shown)


components/authentication/base/group-sync/external-secrets/kustomization.yaml
```diff
@@ -1,5 +1,4 @@
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
-  - github-redhat-appstudio-sre.yaml
```
Michkov, line 4: group 'stage' is synced from this -> components/authentication/base/everyone-can-view.yaml so it should be removed or replaced.

gbenhaim: thanks, I'll remove it. I think it's to wide.

gbenhaim: ok, lets keep the redhat-appstudio-sre sync for now since it's useful for debugging.

Michkov (approved):

PR #1404, opened 2023-02-23
Remove unused shared secrets

Michkov:
Only used shared secret is redhat-appstudio-user-workload for ability to push to defalt user-workload quay repository.

Other secrets were used by internal services for pushing to quay and CI automation, that was solved by moving them to single namespace tekton-ci and setting external secrets for them.


psturc (approved):

components/shared-resources/globally-shared-secrets-clusterrole.yaml
```diff
@@ -11,9 +11,6 @@ rules:
       - sharedsecrets
     resourceNames:
       - redhat-appstudio-user-workload
-      - snyk-shared-secret
```
sbose78, line 14: remind me, what's the state of the snyk usage?

Michkov: currently pipelines are having parameter which expect secret name - https://github.com/redhat-appstudio/build-definitions/blob/main/pipelines/template-build/template-build.yaml#L49 
It's set for our pipelineruns in tekton-ci namespace. So used by our services and optional for customers (must set it).

Michkov: HACBS-1778

PR #1432, opened 2023-03-02
Remove duplicate crb

gbenhaim:
ClusterRoleBinding/prometheus-pipeline-service-exporter-reader is part of applications argocd/monitoring-workload-prometheus-stone-prd-rh01 and pipeline-service-stone-prd-rh01.


bamachrn (changes requested):
this is not exactly duplicate of the other declaration https://github.com/openshift-pipelines/pipeline-service/blob/main/operator/gitops/argocd/pipeline-service/metrics-exporter/clusterrolebinding.yaml. may be we should remove the `prometheus` name from the other declaration. 

Then we dont need this PR. if we remove it from here, prometheus will not be able to access the metrics from pipeline-services

PR #1437, opened 2023-03-02
chore(STONEO11Y-28): o11y component deployment

Omeramsc:
Adds the o11y component files to deploy its PrometheusRules whenever there are new commits on the o11y repository.


argo-cd-apps/base/member/infra-deployments/o11y/011y.yaml
```diff
@@ -0,0 +1,39 @@
...
+  generators:
+    - merge:
+        mergeKeys:
+          - nameNormalized
+        generators:
+          - clusters:
+              values:
+                sourceRoot: components/o11y
+                environment: staging
+                clusterDir: base
```
gbenhaim, line 15: @Omeramsc it won't work like this. You need to add the list generator as well. See https://github.com/redhat-appstudio/infra-deployments/blob/main/argo-cd-apps/base/member/infra-deployments/authentication/authentication.yaml#L16

argo-cd-apps/overlays/development/kustomization.yaml
```diff
@@ -56,3 +56,8 @@ patches:
       kind: ApplicationSet
       version: v1alpha1
       name: image-controller
+  - path: development-overlay-patch.yaml
```
gbenhaim, line 59: Prom isn't deployed currently in dev. So we need to omit the o11y app from it. https://github.com/redhat-appstudio/infra-deployments/blob/main/argo-cd-apps/overlays/development/delete-applications.yaml

components/o11y/production/base/kustomization.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
```
gbenhaim, line 1: you can move base to be in `components/o11y/`
since we don't have cluster-specific overlays, we don't need to have `base` under the environment overlays.

gbenhaim (changes requested):

components/o11y/staging/base/kustomization.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+commonLabels:
```
gbenhaim, line 3: same comment as above

gbenhaim (changes requested):

argo-cd-apps/base/member/infra-deployments/o11y/011y.yaml
Omeramsc: ack, fixed. is that sufficient?

argo-cd-apps/overlays/development/kustomization.yaml
Omeramsc: do you mean it should be on delete *instead* of of this kustomization, or in addition to it? (I modified to the former)

components/o11y/production/base/kustomization.yaml
Omeramsc: done

components/o11y/staging/base/kustomization.yaml
Omeramsc: done

argo-cd-apps/overlays/production/kustomization.yaml
```diff
@@ -73,3 +73,8 @@ patches:
       kind: ApplicationSet
       version: v1alpha1
       name: image-controller
+  - path: development-overlay-patch.yaml
```
gbenhaim, line 76: should be `production-overlay-patch.yaml`

components/o11y/base/kustomization.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
```
gbenhaim, line 2: I missed it before, but you would also have to add the production and stage overlays, since the ApplicationSet will look for them.

`components/o11y/staging/kustomization.yaml`
`components/o11y/production/kustomization.yaml`

both include `../base` in their resources list.

argo-cd-apps/overlays/development/kustomization.yaml
```diff
@@ -55,4 +55,4 @@ patches:
     target:
       kind: ApplicationSet
       version: v1alpha1
-      name: image-controller
+      name: image-controller
```
gbenhaim, line 58: missing a new line at the end of the file.

argo-cd-apps/overlays/development/kustomization.yaml
gbenhaim: the change that you've made is ok.

argo-cd-apps/overlays/development/kustomization.yaml
Omeramsc: fixed

components/o11y/base/kustomization.yaml
Omeramsc: fixed

argo-cd-apps/overlays/production/kustomization.yaml
Omeramsc: fixed

yftacherzog:
@Omeramsc I think the git message mentions the wrong Jira

Omeramsc:
> @Omeramsc I think the git message mentions the wrong Jira

Was I drunk? ...thanks, will be fixed. :)

gbenhaim (approved):

Omeramsc:
@gbenhaim  @zregvart  Any idea why is it failing? The only change since it passed was a very basic rebase...

zregvart:
@Omeramsc I've asked on [Slack](https://redhat-internal.slack.com/archives/C02FANRBZQD/p1678439401468699), what we can do is to rebase and see if it passes again.

From what I can see a 2min timeout was reached in the test that was expecting a PipelineRun to be created.

PR #1440, opened 2023-03-02
Create missing namespace when bootstrapping an upstream test cluster

Roming22:
Current version fails to install an upstream version of the cluster because the namespace in which the application is created does not exists.


Roming22:
Note that upstream deployment will still fail because ArgoCD does not seem to be looking for applications in that namespace.

PR #1448, opened 2023-03-06
Reverting updates to managed-gitops service components to last known good version

anandf:
Reverting PR https://github.com/redhat-appstudio/infra-deployments/pull/1349 to revert managed-gitops service components in Production environment to last known good version.
**Note**: The changes are being reverted only in the Production environment so that the development and staging environment can be used for debugging the issue and finding the root cause.

### DIFF
--- a/components/gitops/development/kustomization.yaml
+++ b/components/gitops/development/kustomization.yaml
@@ -2,10 +2,10 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 
 resources:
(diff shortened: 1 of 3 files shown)


Michkov:
Reason, ticket? 

Expecting that we should keep all environments in sync, debugging should be done on separate instance.

jgwest:
Merging even though there is still a single failing test: prod is already env is already failing, and this PR should address that, and the failure is in an unrelated component, and is already a known failure, as identified in #app-studio-ci-reports.

PR #1470, opened 2023-03-08
Add required permissions in pipeline-service testing NS

Mo3m3n:


Mo3m3n:
I have thought about it and I wanted to see what the ClusterRole "admin" can do to double check but I don't have permissions to list its capabilities.
Anyway I have updated the PR per your suggestion which makes sense to me

sbose78:
> try these out in a cluster-bot cluster :)

sbose78 (approved):

PR #1489, opened 2023-03-10
enable fmuntean to manage spacerequest in own namespace

mfrancisc:
Need ability to test and debug SpaceRequest CRs in my own namespace


amfred (approved):
Seems reasonable.

xcoulon (approved):

PR #1544, opened 2023-03-21
Set application name for PaC

Roming22:
Update the name developers see when Pipelines as Code sends Checks API results back.

Production: "Red Hat Trusted App Pipeline"
Staging: "RHTAP Staging"

### DIFF
--- a/components/pipeline-service/production/base/pac-app-name.yaml
+++ b/components/pipeline-service/production/base/pac-app-name.yaml
@@ -1,0 +1,4 @@
+---
+- op: replace
+  path: /data/application-name
+  value: Red Hat Trusted App Pipeline
--- a/components/pipeline-service/staging/base/pac-app-name.yaml
+++ b/components/pipeline-service/staging/base/pac-app-name.yaml
@@ -1,0 +1,4 @@
+---
+- op: replace
+  path: /data/application-name
+  value: RHTAP Staging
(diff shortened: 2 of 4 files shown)


Roming22:
>  why?
The application name is what is displayed in GitHub when feedback on the pipelineruns is sent. Currently the default value of `Pipeline As Code CI` is used.
This PR changes that to set the PaC name to match the GitHub App name for each environment.

components/pipeline-service/production/base/pac-app-name.yaml
```diff
@@ -0,0 +1,4 @@
+---
+- op: replace
+  path: /data/application-name
+  value: Red Hat Trusted Application Pipeline
```
adambkaplan, line 4: I think we should use the same name as the GitHub app. For projects with large contributor bases, many of the folks who submit PRs will only interact with RHTAP through the GitHub app/integrations:

```suggestion
  value: Red Hat Trusted App Pipeline
```

components/pipeline-service/staging/base/pac-app-name.yaml
```diff
@@ -0,0 +1,4 @@
+---
+- op: replace
+  path: /data/application-name
+  value: RHTAP Staging
```
adambkaplan, line 4: Adding context from our discussions with the Naming/Branding team. Use of the acronym is acceptable here because users in staging will either be Red Hat employees, or select users who we are working with directly.

For production we cannot assume that anyone knows what RHTAP means until we define the acronym explicitly.

components/pipeline-service/production/base/pac-app-name.yaml
Roming22: Done

PR #1547, opened 2023-03-21
[STONE-774] Use redhat-appstudio-appdata-staging on staging

johnmcollier:
Fixes https://issues.redhat.com/browse/STONE-774

Updates the production and staging overlays to specify the correct GitHub org in HAS for GitOps repository. Production will use https://github.com/redhat-appstudio-appdata and staging will use https://github.com/redhat-appstudio-appdata-staging


maysunfaisal (approved):

components/has/staging/kustomization.yaml
```diff
@@ -3,3 +3,6 @@ kind: Kustomization
 resources:
   - ../base
   - ../base/external-secrets
+
+patches:
+  - ./gitops-org-patch.yaml
```
kim-tsao, line 8: Do we need to have the token patch and target name as it's configure in the production kustomization.yaml file?

johnmcollier: No, the default external secret resource exists under the base folder, and points to staging.

PR #1556, opened 2023-03-22
Enable maintainer to manually approve..

sbose78:
operator upgrades

Typically, if the operator upgrade strategy is set to 'manual', when a new version is available for approval - 
* User would need to find the relevant installPlan
* Edit the InstallPlan and set "approved" to `true`.
```
spec:
    approval: Manual
    approved: false # flip to true
```

OpenShift Consoles makes this less error prone but users/maintainers still need permissions on the InstallPlan resource to be able to do this.


components/authentication/base/pipeline-service-sre.yaml
```diff
@@ -3,6 +3,14 @@ apiVersion: rbac.authorization.k8s.io/v1
...
   name: pipeline-service-sre
 rules:
+- apiGroups:
+  - operators.coreos.com
+  resources:
+  - installplans
+  verbs:
+  - get
+  - list
+  - patch
```
adambkaplan, line 13: No update?

sbose78:
Adam, Ann,
Reviews addresses.

amfred (approved):

PR #1562, opened 2023-03-23
[STONEBUGS-89] Set GITHUB_ORG for HAS via Config Map

johnmcollier:
Fixes https://issues.redhat.com/browse/STONEBUGS-89

Instead of patching the `GITHUB_ORG` environment variable directly, I've updated the staging and production overlays to set them via Config Map.


components/has/production/kustomization.yaml
```diff
@@ -3,8 +3,14 @@ kind: Kustomization
 resources:
   - ../base
   - ../base/external-secrets
+
+configMapGenerator:
+- literals:
+  - GITHUB_ORG="redhat-appstudio-appdata"
```
kim-tsao, line 9: Just confirming the name of this env var.  Is it `GITHUB_ORG` or `GITOPS_ORG` which was in the deleted gitops-org-patch.yaml

adambkaplan: this is either a bug or a bug fix - good catch @kim-tsao

kim-tsao (approved):

PR #1605, opened 2023-03-30
contract: Use new collection format in default ECP

simonbaird:
The collections key in the EC policy configuration is now deprecated. The preferred way now is to use include with the @collection syntax, so let's use that in the default ECP CRs.

Ref: https://issues.redhat.com/browse/HACBS-2044


zregvart (approved):

lcarva (approved):
/lgtm

zregvart (approved):

PR #1613, opened 2023-04-03
Promote SPI staging to production

skabashnyuk:
${SUBJ}


sparkoo:
staging kustomize base is not on this version yet because of https://github.com/redhat-appstudio/service-provider-integration-operator/pull/553

sparkoo (approved):

PR #1649, opened 2023-04-12
Disable CSV Copy for all the operators from OLM

bamachrn:
This PR
* Disables copying CSVs from OLM for all the stonesoup clusters


amfred:
This looks like what we want to do (see the thread on #wg-stonestone-platform) but let's get Gal's review.

amfred:
Also waiting for a passing test

bamachrn:
its taking more than 2hours

Michkov:
```
'one or more objects failed to apply, reason: ApplicationSet.argoproj.io "disable-csvcopy" is invalid: spec.generators[0].merge.generators[0].clusters.values: Invalid value: "array": spec.generators[0].merge.generators[0].clusters.values in body must be of type object: "array" (retried 5 times).'
```

amfred (approved):
Let's see if this really applies the change to all namespaces.

PR #1657, opened 2023-04-13
Remove quay provider from spi shared config in preview mode

jrichter1:
both the 'github' and 'quay' parts were getting the values from the same environment variables, so setting them essentially generated a config with 2 identical providers (which is invalid)

removing the quay part since nobody seems to be using it anyway


rsoaresd:
From what I know, regarding SPI, only oauth is being used. But I will
confirm with the QE team.

PR #1668, opened 2023-04-14
fix: Remove unknown field retention_in_days field from ClusterLogForwarder

jannfis:
The field `rentention_in_days` from the log forwarder's output gets removed once the resource is applied to the cluster, leading to a reconciliation loop in Argo CD.

I suspect it's either the default, or invalid.


amfred:
We're going to solve this issue with a Lambda function instead.

amfred (approved):

components/monitoring/logging/base/configure-logging/configure-logforwarder.yaml
```diff
@@ -10,7 +10,6 @@ spec:
     - cloudwatch:
         groupBy: namespaceName
         region: us-east-1
-        retention_in_days: 180
```
sbose78, line 13: Agreed 👍 
https://github.com/openshift/cluster-logging-operator/blob/ea2d07afa39bb63edcf74208cbc49fa3eeaa72a0/apis/logging/v1/cluster_log_forwarder_types.go#L161

sbose78 (approved):

PR #1676, opened 2023-04-17
RHTAP-825: fix grafana deployment

Michkov:
Allow to first install subscription and then CRDs used by it.


psturc (approved):
/lgtm

verified

PR #1699, opened 2023-04-20
Expose GH webhook secret to sprayproxy

enarha:
Sprayproxy will use the secret to validate the webhook request.


enarha:
This is marked as WIP because External Secrets Operator is still not available on the host clusters.

gbenhaim (changes requested):
The external-secrets-operator is now available on the hosts clusters.
You would need to also add a secret for staging.
You need need to add the `sprayproxy` to the allow secret of the cluster secret store - https://github.com/redhat-appstudio/infra-deployments/blob/main/components/cluster-secret-store/base/appsre-stonesoup-vault-secret-store.yaml#L29

gbenhaim (approved):

PR #1718, opened 2023-04-24
Removing tekton-chains public key

Roming22:
The key is now available as a shared resources to anyone logged on the cluster. See: https://github.com/openshift-pipelines/pipeline-service/blob/main/operator/gitops/argocd/pipeline-service/tekton-chains/public-key.yaml


adambkaplan (approved):
/lgtm

PR #1722, opened 2023-04-25
Add cert-manager operator

jsztuka:
This component is mainly used by mutating webhook that is going to be used on staging env.
Reference to planned webhook:
https://github.com/redhat-appstudio/integration-service/pull/152

Let me please know if the component placement is correct and if it needs to be mentioned somepleace else to get automatically deployed.

This change could be applied once the Summit is over.

### DIFF
--- a/components/cert-manager/base/cert-manager.yaml
+++ b/components/cert-manager/base/cert-manager.yaml
@@ -1,0 +1,12 @@
+---
+apiVersion: operators.coreos.com/v1alpha1
+kind: Subscription
+metadata:
+  name: cert-manager
+  namespace: openshift-operators
+spec:
+  channel: stable
+  name: cert-manager
+  source: community-operators
+  sourceNamespace: openshift-marketplace
+  installPlanApproval: Automatic
--- a/components/cert-manager/base/kustomization.yaml
+++ b/components/cert-manager/base/kustomization.yaml
@@ -1,0 +1,6 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - cert-manager.yaml
+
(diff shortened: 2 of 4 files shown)


johnmcollier:
We've used [OpenShift's CA injection](https://github.com/openshift/openshift-docs/blob/main/modules/certificate-injection-using-operators.adoc) in the past for webhooks, have you considered using it here in lieu of Cert Manger?

https://github.com/redhat-appstudio/application-service/pull/52

jsztuka:
> We've used [OpenShift's CA injection](https://github.com/openshift/openshift-docs/blob/main/modules/certificate-injection-using-operators.adoc) in the past for webhooks, have you considered using it here in lieu of Cert Manger?
> 
> [redhat-appstudio/application-service#52](https://github.com/redhat-appstudio/application-service/pull/52)

Thats exactly what I was trying to achieve, but had no luck with my case. Webhook stopped mutating after such change.

jsztuka:
I tried to modify my codebase according to the one you mentioned above and I'm getting following results:
```
Error from server: error when creating "integrationtestscenario-pass.yaml": conversion webhook for appstudio.redhat.com/v1alpha1, Kind=IntegrationTestScenario failed: Post "https://integration-service-webhook-service.integration-service.svc:443/convert?timeout=30s": x509: certificate signed by unknown authority
```
is there something else requested from our side to get this thing working? Assigning/creating certs for this case?

jsztuka:
Tried to figure out the way to run mutating webhook without cert-manager but was not successfull getting rid of error mentioned above. Looks like enabling cert-manager subscription for certain amount of time will be the way to go.

jsztuka:
Get the changes working(without cert-manager), this could be set closed.

PR #1743, opened 2023-04-27
[DEVHAS-259] Fix replicas patch for HAS in production

johnmcollier:
Kustomize didn't like mixing patches where `path` was specified, and also not specified. So explicitly specifying it for all patches


kim-tsao (approved):
/lgtm

PR #1745, opened 2023-04-28
Fixing mismatched file names in kustomize reference

sbose78:
Tested with
```
cd components/tekton-ci
kustomize build production
```

### DIFF
--- a/components/tekton-ci/production/kustomization.yaml
+++ b/components/tekton-ci/production/kustomization.yaml
@@ -17,9 +17,9 @@ patches:
       kind: ExternalSecret
       group: external-secrets.io
       version: v1beta1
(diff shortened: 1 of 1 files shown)


amfred:
What about line 20 in components/tekton-ci/production/kustomization.yaml - should that also be snyk-shared-token?

components/tekton-ci/production/kustomization.yaml
```diff
@@ -17,9 +17,9 @@ patches:
       kind: ExternalSecret
       group: external-secrets.io
       version: v1beta1
-  - path: snyk-shared-secret.yaml
+  - path: snyk-shared-token.yaml
```
sbose78, line 20: @amfred This ?

PR #1773, opened 2023-05-03
Do not merge: Add overlays for remote secrets

mshaposhnik:
Add overlays for remote secrets  enable on dev environment, and disable for STG/PROD


sparkoo:
personally, I would not touch staging and prod overlays at all. Just patch the development one.

mshaposhnik:
@sparkoo so we than rely just on the defaul vaule in the operator CLI, it's sounds little dangerous. Alsothat not allows us to enable it  for the development (this is incovient as that probably be the mainstream feature for nearest future..)

sparkoo:
I'm ok to rely on default value, we're relying on it now. We've agreed to freeze main branch in SPI repo and infra-deployment is frozen as well for regular image updates, so I don't think it's really dangerous.
Only question is, if we need to enable this for developers to play with the feature. If not, I would not merge it at all now and enable the feature everywhere after freeze.

metlos:
We only need to enable other teams to play with this, so IMHO relying on default off in our main and explicitly enable here just in the dev overlay is enough. For the SPI team, we can switch it on by default in our develop branch that we're using during the code freeze so that we don't have to wait for after the freeze is lifted (thawed?)

scoheb:
I was able to manually enable the feature flag as described in this PR and I was able to validate this feature locally on my dev cluster.

PR #1808, opened 2023-05-11
Add script for setup namespace for appstudio execution

Michkov:
This is script for setting namespace which is not managed by toolchain operator. For testing purposes only.


psturc (approved):
/lgtm

/retest

PR #1817, opened 2023-05-16
GitOps: apply #1816 to staging only,  update components/gitops/staging/base/kustomization.yaml

jgwest:
Same PR as #1816 but only applied to staging.


jgwest:
Tests passed on #1817.

jgwest:
Tests passed on https://github.com/redhat-appstudio/infra-deployments/pull/1816

PR #1824, opened 2023-05-17
Update pipeline-service in development/staging

Roming22:
* Create `pipelines-scc` SecurityContextConstraint via ArgoCD.
* Disable the pipeline service account
  * Turn off creation of RBAC resources
  * Add job to create the service certificate authority for Chains, which was provided by the OpenShift Pipelines Operator RBAC feature (not documented!)
* Disable image signing test

### DIFF
--- a/components/pipeline-service/development/kustomization.yaml
+++ b/components/pipeline-service/development/kustomization.yaml
@@ -8,8 +8,8 @@ commonAnnotations:
   argocd.argoproj.io/sync-options: SkipDryRunOnMissingResource=true
 
 resources:
-  - git::https://github.com/openshift-pipelines/pipeline-service.git/developer/openshift/gitops/argocd/pipeline-service?ref=01d8e1eb09fbfe1665e73d654bdc806207a72bd4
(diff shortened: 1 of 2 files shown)


adambkaplan:
/hold

I have a few concerns:

- Image signing test is disabled, this means we are not verifying that chains is working properly end to end.
- We are using a job to re-create the ca-bundle for chains to access the internal registry (I believe this is why we need it?). This might be prone to a race condition between the operator and ArgoCD.

Roming22:
@adambkaplan 
1. The test was deactivated in pipeline-service, but my understanding is that [there is a test](https://github.com/redhat-appstudio/e2e-tests/blob/8578eca880ec8f692d0463b7f42e1b1e884a083b/tests/build/chains.go#L61) in the e2e suite.
2. Is it that the job might be run before the old cabundle has been deleted? Could it be solved by moving the job to another sync-wave, or do we have to be smarter about it?

Roming22:
@adambkaplan I've opened https://github.com/openshift-pipelines/pipeline-service/pull/645 to address #2.

gabemontero (approved):
/lgtm

PR #1852, opened 2023-05-26
feat(RHTAP-965): Add dashboard to track PipelineRuns and TaskRuns

jhutar:


avi-biton:
@jhutar The dashboards json files should be located in your team's repository, please take a look at `o11y` example:
https://github.com/redhat-appstudio/infra-deployments/tree/main/components/monitoring/grafana/base/o11y
https://github.com/redhat-appstudio/o11y/tree/main/grafana

avi-biton:
> I assume it is because of flexibility - it is easier to change that dashboard in our repo than in infra-deployments right?
Yes you are right


> Are there some requirements on the repos? I assume they just have to public.
Thats about it

> If we later update the json in our repo, how do we propagate the change to Stage (or Prod)?
You simply point it to the right commit sha in `infra-deployments` kustomize file  
See example: https://github.com/redhat-appstudio/infra-deployments/blob/main/components/monitoring/grafana/base/o11y/kustomization.yaml#L4

jhutar:
Oh, so if I would have `?ref=main` (or `?ref=HEAD`?) it would not work?

Michkov:
@jhutar @avi-biton in this case it's dashobard for PipelineRuns and TaskRuns so in theory it could be part of [pipeline-service](https://github.com/openshift-pipelines/pipeline-service) CC @Roming22 

I don't think that it would be necessary to create separate repository just for one dashboard - in that case I would add it directly to infra-deployments.

avi-biton:
> Oh, so if I would have `?ref=main` (or `?ref=HEAD`?) it would not work?

I didn't try that, but you can test it in the development cluster

avi-biton:
> @jhutar @avi-biton in this case it's dashobard for PipelineRuns and TaskRuns so in theory it could be part of [pipeline-service](https://github.com/openshift-pipelines/pipeline-service) CC @Roming22
> 
> I don't think that it would be necessary to create separate repository just for one dashboard - in that case I would add it directly to infra-deployments.

I agree, if no such repository exists, the dashboard can be added directly to `infra-deployments`, no need to create a special repository for that

jhutar:
Hello. I have created extra repo. Our team now owns 3 dashboards and we plan to continue on them. Please let me know if current status looks good.

Roming22:
@Michkov Different team will need different dashboards to access the information relevant to their work. I would not be surprised that some of the data in those dashboards are duplicated, and I personally prefer that over having to remember which graph is in which dashboard.

We can discuss this offline in an arch call if necessary.

jhutar:
Hello @avi-biton or anybody. Could you please re-review?

Michkov:
Previous e2e-test passed - https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/redhat-appstudio_infra-deployments/1852/pull-ci-redhat-appstudio-infra-deployments-main-appstudio-e2e-tests/1669280093983215616

Merging

PR #1869, opened 2023-05-29
Add Health check to cluster-log-forwarder

srivickynesh:
JIRA: [RHTAPWATCH-240](https://issues.redhat.com/browse/RHTAPWATCH-240)

Add custom health checks for cluster log forwarder resources


ifireball:
I don't think that creating a configmap like this would work. AFAIK that config map needs to be in the same namespace that Argo is running from - and when you use the GtiOps operator you can't change it manually, but instead you need to make the operator change it for you.

Can you point to a documentation that describes that this would work? At the ver least the configuration seems to be on the wrong cluster...

components/monitoring/logging/staging/configure-logging/cluster-log-forwarder-health-check.yml
```diff
@@ -0,0 +1,56 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: argocd-cm
+  namespace: openshift-gitops
```
gbenhaim, line 5: This won't work since we are deploying using an operator. https://argocd-operator.readthedocs.io/en/latest/reference/argocd/#resource-customizations
Maybe it would be better to contribute it to the upstream project https://github.com/argoproj/argo-cd/tree/master/resource_customizations

I think that this check should be generic and not check a specific log forwarder (e.g splunk/cloudwatch).

gbenhaim (changes requested):

PR #1882, opened 2023-05-30
PLNSRVCE-1307: pipeline service update dev/staging: reenable exporter with caching client

gabemontero:
So @adambkaplan @Roming22 let's decided if we want to 

- see the benefit alone of the non-caching client wrt metrics exporter stability, first in dev/staging, then prod, as this PR currently is the next step in that path
- or not bother on seeing the impact of a caching client on its own, and do not re-enable the metrics exporter until we rework the granularity/structure of the current completion and scheduled metrics

I'll mention in planning tomorrow as well if you all do not comment here beforehand.

### DIFF
--- a/components/pipeline-service/development/kustomization.yaml
+++ b/components/pipeline-service/development/kustomization.yaml
@@ -8,8 +8,8 @@ commonAnnotations:
   argocd.argoproj.io/sync-options: SkipDryRunOnMissingResource=true
 
 resources:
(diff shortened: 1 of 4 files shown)


Roming22 (approved):

gabemontero:
ocp cluster install fail

gabemontero:
OCP cluster install fail

PR #1888, opened 2023-06-01
Add redhat-appstudio/remote-secret repository to CI

skabashnyuk:
It is implementation https://github.com/redhat-appstudio/book/pull/98 in a separate repository


Michkov (approved):
I installed the github app to the repository so it should be ready

PR #1890, opened 2023-06-01
[DEVHAS-377] separate stage and prod deployments

kim-tsao:
We are refactoring our environment overlays to always deploy to stage and dev but we will not promote to prod until it's safe to do so.  Promotion to prod will be a manual effort for now.  Corresponding changes to the HAS repo will need to be made to coordinate with this change.

**Changes**:
- The base folder now contains a new subfolder to organize remaining argcod related patches
- Added a [components](https://kubectl.docs.kubernetes.io/guides/config_management/components/) folder to allow reuse of patches to avoid having to copy the same files across all environments
- Refactored the environment folders into an `overlays` parent.  Updated their respective kustomization files to include all the deployment info from the base kustomization.yaml

**Tests**

- Ran `kustomize build` before and after the changes and verified the outputs were the same across all environments

https://issues.redhat.com/browse/DEVHAS-377


johnmcollier:
> Added a [components](https://kubectl.docs.kubernetes.io/guides/config_management/components/) folder to allow reuse of patches to avoid having to copy the same files across all environments

Why can't we keep the patches in base?

components/has/overlays/staging/kustomization.yaml
```diff
@@ -0,0 +1,29 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - ../../base/argocd-roles
+  - ../../base/external-secrets
```
maysunfaisal, line 5: these resources are in both staging and prod, so I would think it would be nice to see them in base

kim-tsao: The reason why they're not in base is because `external-secrets` is not in dev

kim-tsao:
> > Added a [components](https://kubectl.docs.kubernetes.io/guides/config_management/components/) folder to allow reuse of patches to avoid having to copy the same files across all environments
> 
> Why can't we keep the patches in base?

The patches modify the deployment which no longer exist in base.  I would get an error saying the Deployment object could not be found if I kept them in base.

components/has/components/enable-profiling/kustomization.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: kustomize.config.k8s.io/v1alpha1
+kind: Component
```
johnmcollier, line 2: A bit of a nit, but `components/has/components` seems a little redundant. Could we just throw these under base instead, like https://github.com/redhat-appstudio/infra-deployments/tree/main/components/has/base/external-secrets?

components/has/overlays/development/kustomization.yaml
```diff
@@ -0,0 +1,25 @@
...
+
+
+components:
+  - ../../components/manager-resources
+  - ../../components/enable-profiling
+
+images:
+  - name: quay.io/redhat-appstudio/application-service
+    newName: quay.io/redhat-appstudio/application-service
+    newTag: 05ee33115e4cd7ebfad61d96ffe5ef36c8159dab
```
johnmcollier, line 15: This doesn't match the commit SHA up above (a32c938ee1fcafaed4c64abd3ed4864121fd363c)

kim-tsao: yeah, WIP.  Had to manually patch it but I'm waiting for another rebase

johnmcollier:
@kim-tsao What kustomize version is needed to use the `Components` type?

I get:
```
Error: json: unknown field "components"
```

When I run kustomize build. Have we verified this works with ArgoCD yet? IIRC, the kustomize it uses is back level? And we've had issues in the past getting newer features to work

components/has/components/enable-profiling/kustomization.yaml
kim-tsao: The patches in the components folder update the deployment spec which has now been moved under the separate environment overlays.  We can't apply patches from outside the directory because of this [security restriction](https://kubectl.docs.kubernetes.io/faq/kustomize/#security-file-foo-is-not-in-or-below-bar).  The alternative was to copy every component overlay into each environment directory which I wanted to avoid.

kim-tsao:
> @kim-tsao What kustomize version is needed to use the `Components` type?
> 
> I get:
> 
> ```
> Error: json: unknown field "components"
> ```
> 
> When I run kustomize build. Dunno what version I was on before, but moving up to latest resolved it.
> 
> Have we verified this works with ArgoCD yet? IIRC, the kustomize it uses is back level? And we've had issues in the past getting newer features to work. Since the development overlay doesn’t use components, it may not be seen in the CI if there’s an issue

Is there a way we can verify it works with ArgoCD?  When I was re-vamping our structure, I did notice they had a similar approach see their [k-components\https://github.com/redhat-appstudio/infra-deployments/tree/main/argo-cd-apps/k-components] folder

johnmcollier:
Ah, if they're already using Components then it should be fine. 

Yeah, without tests against the staging overlays, and since [bootstrap-cluster.sh](https://github.com/redhat-appstudio/infra-deployments/blob/main/hack/bootstrap-cluster.sh) uses the development overlays, I'm not sure how else we'd test otherwise (porting over changes?)

components/has/components/enable-profiling/kustomization.yaml
johnmcollier: Hmm. Could we call `components` something else then? They're not really components of HAS I guess? Maybe `common`?

components/has/overlays/production/kustomization.yaml
```diff
@@ -1,10 +1,20 @@
...
   - https://github.com/redhat-appstudio/application-service/config/prometheus/?ref=a32c938ee1fcafaed4c64abd3ed4864121fd363c
 
+components:
+  - ../../components/manager-resources
+  - ../../components/enable-profiling
+
+images:
+  - name: quay.io/redhat-appstudio/application-service
+    newName: quay.io/redhat-appstudio/application-service
+    newTag: 05ee33115e4cd7ebfad61d96ffe5ef36c8159dab
```
maysunfaisal, line 16: should this tag match line 6 above?

components/has/components/enable-profiling/kustomization.yaml
kim-tsao: > I guess Kustomize likes to call it components https://github.com/kubernetes-sigs/kustomize/blob/master/examples/components.md

@johnmcollier  @maysunfaisal  If you want, I can adopt the argoCD `k-components` name.  I'm guessing the convention stands for "kustomize components"?  At least this distinguishes itself from an RHTAP component.

components/has/overlays/staging/kustomization.yaml
```diff
@@ -0,0 +1,29 @@
...
+    newTag: 05ee33115e4cd7ebfad61d96ffe5ef36c8159dab
+
+configMapGenerator:
+  - literals:
+      - GITHUB_ORG="redhat-appstudio-appdata-staging"
+    name: github-config
+    behavior: replace
+
+patches:
+  - path: ./replicas_patch.yaml
```
maysunfaisal, line 26: Can this be a component patch? Since staging and prod are using the same replica patch spec which is 3, they can be included as component here?

components/has/overlays/production/kustomization.yaml
kim-tsao: yes, similar to what John commented above, I missed this when I had to rebase but now that I have to rebase again, I will make sure I fix this.  Still waiting for https://github.com/redhat-appstudio/infra-deployments/pull/1928 to merge

components/has/overlays/staging/kustomization.yaml
kim-tsao: yep, that's in the back of my mind too.  I think it's a good idea to move it to components

components/has/overlays/staging/kustomization.yaml
```diff
@@ -0,0 +1,29 @@
...
+
+components:
+  - ../../components/manager-resources
+  - ../../components/enable-profiling
+
+
+images:
+  - name: quay.io/redhat-appstudio/application-service
+    newName: quay.io/redhat-appstudio/application-service
+    newTag: 05ee33115e4cd7ebfad61d96ffe5ef36c8159dab
```
maysunfaisal, line 17: tag should be same as line 6? 🤔

hack/util-set-github-org
```diff
@@ -3,4 +3,4 @@
 ROOT="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"/..
 
 echo "Changing AppStudio Gitlab Org to \"$1\""
-yq e -i "(.configMapGenerator[].literals[] | select(. == \"*GITHUB*\")) = \"GITHUB_ORG=$1\"" $ROOT/components/has/base/kustomization.yaml
+yq e -i "(.configMapGenerator[].literals[] | select(. == \"*GITHUB*\")) = \"GITHUB_ORG=$1\"" $ROOT/components/has/overlays/development/kustomization.yaml
```
maysunfaisal, line 6: Since we dont have a base Kustomization and we consider development overlays as default like here, then these two lines also need to be updated https://github.com/redhat-appstudio/infra-deployments/blob/main/hack/preview.sh#L171-L172

components/has/components/enable-profiling/kustomization.yaml
maysunfaisal: sure we can adopt `k-components`, if someone already named it that

hack/util-set-github-org
kim-tsao: good catch! I'll try to update it in the next rev if I keep the current approach.  I'm also thinking about [John's suggestion](https://github.com/johnmcollier/infra-deployments/tree/separate_envs/components/has) to keep the base overlay for dev and stage.  I think it's a bit cleaner.

components/has/components/enable-profiling/kustomization.yaml
kim-tsao: folder has been renamed

kim-tsao:
I did not add the new configmap yet because I need to make the corresponding changes in our HAS repo first.  I'll introduce the environment variable in https://github.com/redhat-appstudio/application-service/pull/344 and push that once this PR gets merged, then I'll make another change in this repo to add the new env var values

kim-tsao:
made some additional changes since I got a little confused with the pre and post review structure.  Since the base kustomization has all the content for dev, the preview.sh and util script to set the github org should point to base and not the development overlay since that is now just a reference to base.

johnmcollier (approved):

johnmcollier (approved):
/approve
/lgtm

PR #1897, opened 2023-06-02
Pipeline Service Dashboard

adambkaplan:
Add a grafana dashboard for Pipeline Service


Roming22:
I"m surprised that there is not a staging and production definition for the dashboards. Don't we need to validate dashboards changes before promoting them?

PR #1909, opened 2023-06-06
Enable CI for pipeline-service-exporter component

enarha:
pipeline-service-exporter is already part of the pipeline-service component. Initially onboarded as a RHTAP user component we are switching to RHTAP component mode and enabling the CI for the repo.


gabemontero (approved):
/approve

PR #1917, opened 2023-06-07
Remote secret operator deployment

skabashnyuk:
Enabled for dev/stage. Disable for prod.
Context: https://issues.redhat.com/browse/SVPI-510 CI/CD of Remote-secret controller with infra-deployment


metlos (commented):
Is there way of not using images and config from your fork of SPI?

mshaposhnik (approved):
That's huge... Maybe also a few words into SPI repo of how to use the latest tag/image ?

PR #1940, opened 2023-06-12
GitOps Service team RBAC changes for staging/production (#604)

keithchong:
See https://issues.redhat.com/browse/GITOPSRVCE-604


components/gitops/staging/base/authentication/gitops-rolebindings.yaml
```diff
@@ -0,0 +1,58 @@
+---
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: gitops-namespace-all-access
+  namespace: gitops
+subjects:
+  - kind: User
```
Michkov, line 8: We are syncing groups from redhat-appstudio teams -> so maybe this could be used https://github.com/orgs/redhat-appstudio/teams/gitops-service-team ?
Example: https://github.com/redhat-appstudio/infra-deployments/blob/main/components/authentication/base/internal-services.yaml#L7-L9

jgwest (changes requested):
Hey Keith, looks great!

### 1) Audit vs original story

Feel free to double check/poke holes in my audit versus the original story, below. I think these items are missing from what is defined in the story:

All GitOps Service Team member should be able to:

- Staging:
With the current PR, GitOps Service team members can write to all GitOps namespaces, when they should only be able to read from them.
    - Once the ability to write to gitops namespaces is done, make sure that GitOps Service team members can still kill pods in GitOps namespaces.
    - Ability to read/write all API resources in your personal Namespace on staging (we might need to add individual rolebindings for each team member's namespace, here)

- Production:
    - With the current PR, GitOps Service team members can exec into pod on prod, which they shouldn't be able to do.

### 2) Unify (reduce YAML duplication) of Roles between staging and prod

It also couldn't hurt to unify some of the Roles/ClusterRoles that are duplicated between on staging and production.
e.g. have a folder like `/components/gitops/base/authentication', and in there move the Roles/ClusterRoles that are the same between staging and prod. You would then include that base folder from the existing staging/prod `kustomization.yaml`s.

For example, this is the same in both, and probably could be moved to a single shared location:
```yaml
kind: ClusterRole
metadata:
  name: gitops-component-maintainer
rules:
  - apiGroups:
      - '*'
    resources:
      - '*'
    verbs:
      - get
      - list
      - watch
  - apiGroups:
      - operators.coreos.com
    resources:
      - installplans
    verbs:
      - get
      - list
      - update
      - patch
```

(The Cluster/RoleBindings would obviously be different in staging/prod, but I assume at least the Roles are, or can be made, the same).


### 3) I like @Michkov's  idea of referencing the team (via a Group), rather than referencing individual team members.

That way, once team members are removed from the GitHub group, they are automatically removed from the role list (rather than needing to remember to update the YAML).

components/gitops/staging/base/authentication/gitops-rolebindings.yaml
jgwest: :+1: from me

jgwest (approved):
Looks great, thanks @keithchong, and thanks for the individual testing, refactoring, and for the test steps you posted!

PR #1947, opened 2023-06-13
Fix Tekton Results servicemonitors scheme

enarha:
Tekton results exports metrics over HTTP and not HTTPS. Changing that requires upstream update of code, which is not given will be accepted. As a fix here we change the scheme to HTTP.


sbose78 (approved):

adambkaplan:
@enarha I created https://issues.redhat.com/browse/PLNSRVCE-1345 to mark this as a tech debt item. We should use HTTPS to ensure metrics data is encrypted in transit.

PR #1953, opened 2023-06-14
doc(RHTAPWATCH-278): document the monitoring stack

yftacherzog:
This change documents the design for providing central monitoring for RHTAP


yftacherzog:
@gbenhaim @avi-biton @amisstea @Kousalya1998 please take a look

components/monitoring/prometheus/README.md
```diff
@@ -1,34 +1,67 @@
-## Data Plane Clusters Prometheus Instances
-We use the Openshift-provided Prometheus deployments, platform and
-user-workload-monitoring (UWM), alongside a Prometheus instance deployed by the RHOBS
+## RHTAP Central Monitoring
+The RHTAP monitoring solution is based on three Prometheus instances deployed to each
+Production and staging cluster. Each cluster writes a subset of the metrics it generates
```
avi-biton, line 3:
```suggestion
Production and Staging cluster. Each cluster writes a subset of the metrics it generates
```

components/monitoring/prometheus/README.md
```diff
@@ -1,34 +1,67 @@
...
 In Production and Staging, UWM Prometheus is enabled using OCM (since it maintains the
 Prometheus configurations).
 In Development it's enabled by deploying a configmap using ArgoCD.
 
-### Observability Operator (OBO) Prometheus
-Federates the Platform Prometheus instance.
+#### Observability Operator (OBO) Prometheus
+Federates the Platform and UWM Prometheus instances.
 
 There are limitations for both built-in Prometheus instances that do no allow us to
```
avi-biton, line 47:
```suggestion
There are limitations for both built-in Prometheus instances that do not allow us to
```

components/monitoring/prometheus/README.md
yftacherzog: done

components/monitoring/prometheus/README.md
yftacherzog: done

eisraeli (approved):
LGTM

components/monitoring/prometheus/README.md
```diff
@@ -1,34 +1,67 @@
-## Data Plane Clusters Prometheus Instances
-We use the Openshift-provided Prometheus deployments, platform and
-user-workload-monitoring (UWM), alongside a Prometheus instance deployed by the RHOBS
+## RHTAP Central Monitoring
+The RHTAP monitoring solution is based on three Prometheus instances deployed to each
+Production and Staging cluster. Each cluster writes a subset of the metrics it generates
```
ifireball, line 3: does this include the host cluser?

components/monitoring/prometheus/README.md
```diff
@@ -1,34 +1,67 @@
...
+into Observatorium (RHOBS), marking each metric with a label containing its cluster of origin.
+
+Observatorium holds two different tenant for RHTAP, one for Production and one for
+Staging. The metrics collected for each of those tenants are made available to the
+AppSRE team via the AppSRE Grafana instance.
```
ifireball, line 8: Is the role of this only to make things available for AppSRE? If not, lets frame things differently.

components/monitoring/prometheus/README.md
```diff
@@ -1,34 +1,67 @@
...
+   rhobs <--> |query|grafana(AppSRE Grafana)
+   subgraph "RHTAP Cluster"
+       subgraph CMO
+           UWM
+           Platform
+       end
+       MS
+   end
+
+```
```
ifireball, line 24: Perhaps you should explicitly add the host cluster to the chart to both show it and emphasize the idea that data is streaming from multiple clusters.

Also maybe add some blocks to indicate the kinds of client service each instance scrapes.

Is a federation connection between Prometheus instances significantly different from a scraping connection between Prometheus and a client service? If not, the arrows might need to say "scrape" rather then "federate". Also are the arrow really bidirectional? As far as I can understand that data flow goes in one direction.

ifireball (changes requested):

components/monitoring/prometheus/README.md
yftacherzog: It does.

ifireball: perhaps make the text explicit about that

components/monitoring/prometheus/README.md
yftacherzog: I'll look into adding the stuff you suggested. My main reason to stop there is the cost-effectiveness of this exercise. Namely, it takes (me) a lot of time and I'm not sure it will be worth it in terms of readers' experience.

> Is a federation connection between Prometheus instances significantly different from a scraping connection
The main difference is the API endpoint being accessed –`metrics` (usually) vs. `federate`.

> Also are the arrow really bidirectional?
Not really. The main reason for me to do them this way is that there's no native support in mermaid for doing backward arrows - only forward or bidirectional. The only way to achieve this is to do a back and forth arrow and delete one direction. Doing that, the arrow is not straight anymore, which raises the question why is the arrow different from others.

e.g. doing this:
```mermaid
flowchart BT
   foo --- bar -->foo
```
and deleting one of the directions:
```mermaid
flowchart BT
   foo --- bar -->foo
   linkStyle 0 stroke:none
```
I settled on doing it bidirectional as saying: "this side is asking for data and the other delivers it"

yftacherzog: Thinking about it again, I think the main issue here with the directions of the arrows was that in this particular case, if I did this trick, it made things appear in an incorrect order. i.e. one of the prometheus instances was placed above the MS promotheus.

components/monitoring/prometheus/README.md
yftacherzog: done

components/monitoring/prometheus/README.md
yftacherzog: done

components/monitoring/prometheus/README.md
ifireball: Its hard to get details of HTTP request/response in single arrow, I suggest you focus instead on the logical data flow.

> My main reason to stop there is the cost-effectiveness of this exercise. Namely, it takes (me) a lot of time and I'm not sure it will be worth it in terms of readers' experience.

Well a picture is worth 1K words...

yftacherzog: If a tree falls in the forest, and someone takes a picture of that, but no one bother looking at that picture – does it make a sound? :)

I'll do my best.

yftacherzog: @ifireball take another look. Let me know your thoughts.

ifireball: Looks better IMO, but could be improved in a few small ways:

- "RHTAP Member Cluster" should probably become plural
- I don't know what "CMO" is, can you spell it out?
- Can you get away with saying "User Workload Monitor" instead of UWM?
- Instead of saying just (Generic resources) can you provide a few key examples to make thing clearer? (The chart text is actually better at this because you've used "pods" as your key.

yftacherzog: @ifireball how about now?

ifireball: /approve

scoheb (approved):
approving on behalf of team. seems like this PR is blocking the tide merge queue?

PR #1970, opened 2023-06-19
RHTAPWATCH-322: Fluentd collectors are incompatible with Splunk ClusterLogForwarder configuration

eisraeli:
Fluentd collectors are incompatible with Splunk ClusterLogForwarder configuration.
    
As detailed in
https://docs.openshift.com/container-platform/4.13/logging/cluster-logging-external.html#logging-forward-splunk_cluster-logging-external
fluentd is not supported with the Splunk Log Forwarder configuration.
    
Therefore, this commit swaps fluentd with vector.

### DIFF
--- a/components/monitoring/logging/base/configure-logging/configure-log-collectors.yaml
+++ b/components/monitoring/logging/base/configure-logging/configure-log-collectors.yaml
@@ -2,11 +2,12 @@
 apiVersion: "logging.openshift.io/v1"
 kind: "ClusterLogging"
 metadata:
-  annotations:
-    argocd.argoproj.io/sync-options: SkipDryRunOnMissingResource=true
   name: "instance"
+  annotations:
+    logging.openshift.io/preview-vector-collector: enabled
+    argocd.argoproj.io/sync-options: SkipDryRunOnMissingResource=true
 spec:
   collection:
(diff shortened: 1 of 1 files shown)


components/monitoring/logging/base/configure-logging/configure-log-collectors.yaml
```diff
@@ -2,11 +2,13 @@
 apiVersion: "logging.openshift.io/v1"
 kind: "ClusterLogging"
 metadata:
+  name: "instance"
+  namespace: "openshift-logging"
```
amisstea, line 6: What determined the namespace up until now?

eisraeli: @amisstea The `kustomization.yaml` file contains `namespace: "openshift-logging"`

Haven't really decided where it's better to list the namespace.

eisraeli: Anyway, just to match what was already done in
`/infra-deployments/components/monitoring/logging`
I kept the namespace name just in the `kustomization.yaml` file.

amisstea (approved):

PR #1977, opened 2023-06-20
Add JBSConfig to activate JVM Build Service

stuartwdouglas:
Initial attempt at rebuilding the dependencies of the JVM Build Service.

Creating as a draft for now. I think this will also require defining a new role with permissions to modify the JVM Build Service objects, to allow me to rebuild failures etc. 

[STONEBLD-1221](https://issues.redhat.com//browse/STONEBLD-1221)


stuartwdouglas:
closing, we will do this via the build tenant at some point.

PR #1982, opened 2023-06-20
Remote-secret set correct location of aws secret manager credentials

skabashnyuk:
Related to https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/71761


xbaran4:
Shouldn't we create a new ExternalSecret for credentials that the RemoteSecret operator will use? Currently, both RS operator and SPI operator would use the credentials meant for RS. Or am I missing something?

mshaposhnik (approved):

PR #1983, opened 2023-06-20
feat(HACBS-2310): add clusterrole for release pipelines

johnbieren:
Release PipelineRuns need access to Custom Resources in dev and managed namespaces. This commit adds a ClusterRole to provide those permissions. We can then create RoleBindings in the namespaces we need to pull from so that the pipelineruns can access the resources.


scoheb (changes requested):
role name?

scoheb:
[FAIL] [e2e-demos-suite] DEVHAS-234: creates an application with springboot component from RHTAP samples [It] waits public component springboot-component pipeline to be finished [e2e-demo]

scoheb:
not related to the content of this PR

/retest

johnbieren:
Yeah, Kasem reported that some known issues were affecting e2e today, so I figured it was just due to that (although I admittedly didn't dive in)

PR #1989, opened 2023-06-21
Fix timeout for remote-secret-controller-in-cluster-local

Katka92:
The application was not created during the given timeout which failed the script. Increasing the timeout should solve the problem.
Related issue: [HAC-4343](https://issues.redhat.com/browse/HAC-4343)


Katka92:
@albarbaro @rsoaresd Hello, I hope I've pinged the right people :sweat_smile: Could you please look at those changes? :slightly_smiling_face: Thanks

albarbaro (approved):

PR #1996, opened 2023-06-22
Reconfigure RemoteSecret for stage to use spi's AWS secret manager

skabashnyuk:
- Terraform is not working. A new credential for AWS secret manager would not be delivered. We don't know when it will start working
- The idea is to reuse the same to spi-operator/spi-oauth service AWS secret manager, with a difference in instance id. That would basically make a name of secrets in AWS SM have different names. But will keep the same prefix.
- Prod we keep it the same for a while. I hope it will be fixed until we enable RemoteSecret on prod.


xbaran4 (approved):
Under these circumstances, I think we can move ahead with this change.

PR #2014, opened 2023-06-26
PLNSRVCE-1069: remove servicemonitor definition that is not currently being processed

gabemontero:
Per conversation with @amisstea , as this folder is not being processed (he indicated there were some errors in sibling folders), removing the exporter service monitor from infra-deployments, and we'll define in pipeline-service repo, as other components are doing.

https://github.com/openshift-pipelines/pipeline-service/pull/685 is adding to pipeline-service

I am leaving the SA and RBAC definitions for now, in case this subfolder gets reactivated, and those artifacts become useful again.

@adambkaplan @ramessesii2


gabemontero:
green appstudio-e2e-tests

amisstea (approved):

PR #2023, opened 2023-06-27
bump pipeline service in prod

gabemontero:
Pull in the https://github.com/redhat-appstudio/infra-deployments/pull/2015 changes to dev/staging into production.  Note:  the https://github.com/redhat-appstudio/infra-deployments/pull/2015/files#diff-d99d1444c5e672b40e18d7e3637b32e4722598a1e086575e084da610d3455dc5 change from the prior PR should already be available in prod.

Examination of stage looks good after about 18 hours.

Reminder - any e2e results do not stem from this change, as we are only changing prod, and the e2e's work off dev (vs. stage or prod)

### DIFF
--- a/components/pipeline-service/production/base/kustomization.yaml
+++ b/components/pipeline-service/production/base/kustomization.yaml
@@ -8,7 +8,7 @@ commonAnnotations:
   argocd.argoproj.io/sync-options: SkipDryRunOnMissingResource=true
 
 resources:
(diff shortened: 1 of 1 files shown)


gabemontero:
same unrelated flakes .... I'll see about a clean run, but as this change just bumps prod to the same working level as stage, and this change does not affect the e2e's, I may just hit the merge button by EOB.

PR #2045, opened 2023-06-29
do not run image-controller pipelines in tekton-ci ns

sbose78:
.. they will run on rhtap-build


sbose78:
Taking this in for the time-being, will revert if needed.

PR #2051, opened 2023-07-03
fix(RHTAPWATCH-356): obo conflicts when using OSD on AWS

avi-biton:
- OBO (Observability operator) is not deployed by default
- Adding --obo/-o to enable obo deployment in development environment

Examples:   
`./hack/bootstrap-cluster.sh preview --obo`
`./hack/bootstrap-cluster.sh preview -o`

### DIFF
--- a/hack/bootstrap-cluster.sh
+++ b/hack/bootstrap-cluster.sh
@@ -3,7 +3,7 @@
 ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"/..
 
 main() {
-    local mode keycloak toolchain
+    local mode keycloak toolchain obo
     while [[ $# -gt 0 ]]; do
         key=$1
         case $key in
@@ -13,6 +13,10 @@ main() {
             ;;
         --keycloak | -kc)
             keycloak="--keycloak"
+            shift
+            ;;
+        --obo | -o)
+            obo="--obo"
             shift
             ;;
         preview | upstream)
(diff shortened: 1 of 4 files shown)


hack/bootstrap-cluster.sh
```diff
@@ -52,19 +56,20 @@ main() {
...
     esac
 }
 
 print_help() {
-    echo "Usae: $0 MODE [-t|--toolchain] [-kc|--keycloak] [-h|--help]"
+    echo "Usae: $0 MODE [-t|--toolchain] [-kc|--keycloak] [-o|--obo] [-h|--help]"
     echo "  MODE             upstream/preview (default: upstream)"
     echo "  -t, --toolchain  (only in preview mode) Install toolchain operators"
     echo "  -kc, --keycloak  (only in preview mode) Configure the toolchain operator to use keycloak deployed on the cluster"
+    echo "  -o, --obo        (only in preview mode) Install and configure the Observability operator"
```
yftacherzog, line 69: We're also deploying Prometheus federation with that option enabled. I'm not sure if we can come up with a better name for the flag, but the description should also mention something about the Prometheus instance that will be deployed.

hack/preview.sh
```diff
@@ -112,6 +121,12 @@ update_patch_file () {
 update_patch_file "${ROOT}/argo-cd-apps/k-components/inject-infra-deployments-repo-details/application-patch.yaml"
 update_patch_file "${ROOT}/argo-cd-apps/k-components/inject-infra-deployments-repo-details/application-set-patch.yaml"
 
+# Add monitoringstack to $ROOT/components/monitoring/prometheus/development/kustomization.yaml to deploy Observability operator
```
yftacherzog, line 124: Not sure we need this comment.

hack/bootstrap-cluster.sh
avi-biton: I changed the description to:  
`(only in preview mode) Install Observability operator and Prometheus instance for federation`

hack/preview.sh
avi-biton: I removed the comment and updated the prompt

yftacherzog (approved):
/lgtm

yftacherzog:
@gbenhaim can you take a look?

hmariset (approved):
lgtm

PR #2054, opened 2023-07-03
chore(RHTAPWATCH-351): segment-bridge component deployment

Omeramsc:
Deploy a component for the User Journey effort as segment-bridge. This is needed in order to have a Service account with read permissions to the UserSignup resources from the host cluster. Those resources will be used in a script that'll  map between cluster usernames and SSO ID's

### DIFF
--- a/argo-cd-apps/overlays/development/delete-applications.yaml
+++ b/argo-cd-apps/overlays/development/delete-applications.yaml
@@ -38,5 +38,11 @@ $patch: delete
 apiVersion: argoproj.io/v1alpha1
 kind: ApplicationSet
 metadata:
-  name: segment-bridge
+  name: segment-bridge-member
+$patch: delete
+---
+apiVersion: argoproj.io/v1alpha1
+kind: ApplicationSet
+metadata:
+  name: segment-bridge-host
 $patch: delete
--- a/components/authentication/base/segment-bridge.yaml
+++ b/components/authentication/base/segment-bridge.yaml
@@ -1,0 +1,64 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: segment-bridge-read-access
+  namespace: toolchain-host-operator
+rules:
+- apiGroups: [""]
+  resources: ["UserSignup"]
+  verbs: ["get", "list", "watch"]
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: segment-bridge-host-sa
+  namespace: toolchain-host-operator
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: sa-read-permissions
+  namespace: toolchain-host-operator
+subjects:
+  - kind: ServiceAccount
+    name: segment-bridge-host-sa
+    namespace: toolchain-host-operator
+roleRef:
+  kind: Role
+  name: segment-bridge-read-access
+  apiGroup: rbac.authorization.k8s.io
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: segment-bridge-host
+  namespace: toolchain-host-operator
+  annotations:
+    kubernetes.io/service-account.name: segment-bridge-host-sa
+type: kubernetes.io/service-account-token
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: segment-bridge-read-host-sa-secret
+  namespace: toolchain-host-operator
+rules:
+- apiGroups: [""]
+  resources: ["secrets"]
+  resourceNames: ["segment-bridge-host"]
+  verbs: ["get"]
+--- 
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: segment-bridge-read-host-sa-secret-binding
+  namespace: toolchain-host-operator
+subjects:
+  - kind: User
+    name: Omeramsc
+  - kind: User
+    name: ifireball
+roleRef:
+  kind: Role
+  name: segment-bridge-read-host-sa-secret
+  apiGroup: rbac.authorization.k8s.io
--- a/components/segment-bridge/README.md
+++ b/components/segment-bridge/README.md
@@ -4,4 +4,3 @@ Component used to allow reading resources from the host and member clusters
 
 Serves as part of the User Journey effort,
 related information can be found at the [segment-bridge repository](https://github.com/redhat-appstudio/segment-bridge)
-
--- a/components/segment-bridge/base/serviceaccount.yaml
+++ b/components/segment-bridge/base/serviceaccount.yaml
@@ -10,7 +10,7 @@ rules:
 apiVersion: v1
 kind: ServiceAccount
 metadata:
-  name: segment-bridge-cluster-sa
+  name: segment-bridge-member-sa
   namespace: segment-bridge
 ---
 apiVersion: rbac.authorization.k8s.io/v1
@@ -19,7 +19,7 @@ metadata:
   name: sa-read-permissions-all-namespaces
 subjects:
   - kind: ServiceAccount
-    name: segment-bridge-cluster-sa
+    name: segment-bridge-member-sa
     namespace: segment-bridge
 roleRef:
   kind: ClusterRole
@@ -29,9 +29,34 @@ roleRef:
 apiVersion: v1
 kind: Secret
 metadata:
-  name: segment-bridge
+  name: segment-bridge-member
   namespace: segment-bridge
   annotations:
-    kubernetes.io/service-account.name: segment-bridge
+    kubernetes.io/service-account.name: segment-bridge-member-sa
 type: kubernetes.io/service-account-token
-
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: segment-bridge-read-member-sa-secret
+  namespace: segment-bridge
+rules:
+- apiGroups: [""]
+  resources: ["secrets"]
+  resourceNames: ["segment-bridge-member"]
+  verbs: ["get"]
+--- 
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: segment-bridge-read-member-sa-secret-binding
+  namespace: segment-bridge
+subjects:
+  - kind: User
+    name: Omeramsc
+  - kind: User
+    name: ifireball
+roleRef:
+  kind: Role
+  name: segment-bridge-read-member-sa-secret
+  apiGroup: rbac.authorization.k8s.io
--- a/argo-cd-apps/base/host/kustomization.yaml
+++ b/argo-cd-apps/base/host/kustomization.yaml
@@ -2,6 +2,8 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
   - sprayproxy
+  - segment-bridge
 components:
   - ../../k-components/deploy-to-host-cluster-merge-generator
   - ../../k-components/inject-argocd-namespace
+
--- a/argo-cd-apps/base/host/segment-bridge/kustomization.yaml
+++ b/argo-cd-apps/base/host/segment-bridge/kustomization.yaml
@@ -1,0 +1,7 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - segment-bridge-host.yaml
+components:
+  - ../../../k-components/inject-infra-deployments-repo-details
+
--- a/argo-cd-apps/base/host/segment-bridge/segment-bridge-host.yaml
+++ b/argo-cd-apps/base/host/segment-bridge/segment-bridge-host.yaml
@@ -1,7 +1,7 @@
 apiVersion: argoproj.io/v1alpha1
 kind: ApplicationSet
 metadata:
-  name: segment-bridge
+  name: segment-bridge-host
 spec:
   generators:
     - merge:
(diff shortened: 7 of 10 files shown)


argo-cd-apps/overlays/development/delete-applications.yaml
```diff
@@ -34,3 +34,10 @@ kind: ApplicationSet
 metadata:
   name: cluster-secret-store
 $patch: delete
+---
+apiVersion: argoproj.io/v1alpha1
+kind: ApplicationSet
+metadata:
+  name: segment-bridge
+$patch: delete
```
yftacherzog, line 42: Does this mean that it will not be deployed in dev? If so, do we need to maintain the dev configs under the `segment-bridge` component?

components/segment-bridge/README.md
```diff
@@ -0,0 +1,3 @@
+---
+title: SEGMENT-BRIDGE
+---
```
yftacherzog, line 3:
```suggestion
# SEGMENT-BRIDGE
```

argo-cd-apps/overlays/development/delete-applications.yaml
Omeramsc: While this is a mistake I'll fix, The current requirement is for the staging cluster, so if we'll start not including this in envs, prod should be selected as well.

yftacherzog: ok. Do you want it on all stage clusters? I thought you mentioned you want it on a specific cluster.

Omeramsc: Staging host cluster is what we need. is differentiation possible in those files?

yftacherzog: Yes. There are ways to differentiate between the different clusters, but as far as I understand, the starting point is the collection of clusters it will be deployed on. I think currently, you're deploying only to member clusters.

Have a look here: https://github.com/redhat-appstudio/infra-deployments/tree/main/argo-cd-apps/base

Your configs are under `member`. I think it means that it will only be deployed to member clusters. You can either have it under `all-clusters` and then differentiate by name like we did for the prometheus configs, or you can have it under `host` and then I think it will only go to the host cluster.

argo-cd-apps/base/member/infra-deployments/segment-bridge/segment-bridge.yaml
```diff
@@ -0,0 +1,41 @@
+apiVersion: argoproj.io/v1alpha1
+kind: ApplicationSet
+metadata:
+  name: segment-bridge
+spec:
+  generators:
```
ifireball, line 6: seems like you can simplify this to just the `clusters` generator

components/segment-bridge/README.md
```diff
@@ -0,0 +1 @@
+# SEGMENT-BRIDGE
```
ifireball, line 1: At least refer to our repo so people know what his is, you can also write a few words...

components/segment-bridge/base/role.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
```
ifireball, line 2: This gives us cluster-wide permissions, while we only need permissions for a particular namespace.

components/segment-bridge/base/rolebinding.yaml
```diff
@@ -0,0 +1,14 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: sa-read-permissions
+  namespace: segment-bridge
```
ifireball, line 6: Not seeing namespace creation YAML for this namespace and we don't need our own namspace anyway, we can just put the account in the `toolchain-host-operator` namespace.

ifireball (changes requested):

components/segment-bridge/README.md
Omeramsc: fixed

components/segment-bridge/README.md
Omeramsc: fixed

argo-cd-apps/base/member/infra-deployments/segment-bridge/segment-bridge.yaml
Omeramsc: Not too sure about this actually, most of the configs looks the same. does removing the `merge` or `lists` add anything?

components/segment-bridge/base/role.yaml
Omeramsc: fixed

components/segment-bridge/base/rolebinding.yaml
Omeramsc: fixed

argo-cd-apps/base/member/infra-deployments/segment-bridge/segment-bridge.yaml
ifireball: it would make the code simpler

argo-cd-apps/base/host/segment-bridge/sprayproxy.yaml
```diff
@@ -0,0 +1,41 @@
+apiVersion: argoproj.io/v1alpha1
```
ifireball, line 1: why is this file called `sprayproxy.yaml`?

components/segment-bridge/base/serviceaccount.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
+kind: ServiceAccount
```
ifireball, line 2: we need to do something about the token for accessing this service account. AFAIK by default we would get a random token, but we need something a bit less random so we can can configure it for our client script and not have to replace it every time the cluster gets reprovisioned

ifireball (dismissed):

components/segment-bridge/README.md
```diff
@@ -0,0 +1,6 @@
...
+
+Component used to allow reading resources from the host cluster
+
+Serve as part of the User Journey effort,
```
yftacherzog, line 5:
```suggestion
Serves as part of the User Journey effort,
```

components/segment-bridge/base/serviceaccount.yaml
Omeramsc: I'm looking up if we can somehow set this beforehand, maybe by setting a token secret? If not, will we need to run an oc command each time?

components/segment-bridge/README.md
Omeramsc: fixed

argo-cd-apps/base/host/segment-bridge/sprayproxy.yaml
Omeramsc: fixed.

Omeramsc:
Added a secret to create a long-lived token. While still depends on k8s generating the token instead of pre-setting it up, its better than what we currently have

argo-cd-apps/base/member/infra-deployments/segment-bridge/segment-bridge.yaml
Omeramsc: done

components/segment-bridge/base/serviceaccount.yaml
Omeramsc: As mentioned in previous channels, currently I couldn't find a way to pre-set the token, and I believe it is not doable. I've set a secret-access-token so the generated token will at least be long-lived.

components/segment-bridge/development/kustomization.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- ../base
```
gbenhaim, line 4: I don't think that the sandbox operator are installed by default in development. IIRC the user should specify he/she wants to install them using a flag.

gbenhaim (changes requested):

components/segment-bridge/base/service-account-token.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: segment-bridge
+  namespace: toolchain-host-operator
```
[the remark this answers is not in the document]
ifireball: yes

argo-cd-apps/base/host/segment-bridge/segment-bridge.yaml
```diff
@@ -0,0 +1,35 @@
...
+kind: ApplicationSet
+metadata:
+  name: segment-bridge
+spec:
+  generators:
+    - clusters:
+        values:
+          sourceRoot: components/segment-bridge
+          environment: staging
+          clusterDir: ""
```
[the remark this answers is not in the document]
Omeramsc: Is it or the `merge` generator essential? what is their use?

argo-cd-apps/base/member/infra-deployments/kustomization.yaml
```diff
@@ -17,5 +17,6 @@ resources:
   - internal-services
   - image-controller
   - o11y
+  - segment-bridge
```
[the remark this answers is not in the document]
Omeramsc: right, left it in by mistake. will remove

components/segment-bridge/base/role.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: segment-bridge-read-access
+rules:
+- apiGroups: [""]
```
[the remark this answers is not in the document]
Omeramsc: as in `UserSignup`, right? will do.

components/segment-bridge/development/kustomization.yaml
Omeramsc: I see there's a script to run on a dev env, but I don't see anything that needs to be configured in those files, or any other component that does that. @MatousJobanek Any idea if there's anything I need to add for it to be configured?

argo-cd-apps/base/host/segment-bridge/segment-bridge.yaml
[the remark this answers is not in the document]
Omeramsc: Upon further inspection of the gitops docs [1][2], I don't see a need for the list generators as we don't generate parameters from a fixed list of cluster information,  and lacking the former - we don't need the merge generator.
[1] https://argo-cd.readthedocs.io/en/stable/operator-manual/applicationset/
[2] https://argo-cd.readthedocs.io/en/stable/operator-manual/applicationset/Generators-Merge/

components/segment-bridge/development/kustomization.yaml
Omeramsc: NVM my latest comment. 
While we configure the dev-env, we don't really need access to the sandbox operator as we install nothing. the only resource we really create there is the service account. I don't believe we need to change anything.

components/segment-bridge/base/serviceaccount.yaml
```diff
@@ -0,0 +1,38 @@
+
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: segment-bridge-read-access
```
gbenhaim, line 5: missing namespace
you can set one for all the resources on the kustomization file.

gbenhaim (changes requested):

components/segment-bridge/development/kustomization.yaml
Omeramsc: @gbenhaim  you mean that the `toolchain-host-operator` namespace won't exist for the resources in the serviceaccount.yaml to be created in? Since we don't need the operator itself in development, and there is the issue I mentioned on my first comment, why not create a dummy namespace in development? unless I'm missing something about how to install it

components/segment-bridge/base/serviceaccount.yaml
Omeramsc: fixed
will set in the kustomization file once the test will pass... for now I prefer it separated in case changes will be needed (e.g my former comment regarding creating the namespace in dev env)

Omeramsc:
Since there's a chance the toolchain namespace won't be set in development (due to a flag configuration), leading to a sync issue, and as we don't need the development environment in the foreseeable future for this component - I've added the component to the delete overlay of development

components/segment-bridge/development/kustomization.yaml
Omeramsc: Fixed by adding a delete overlay

gbenhaim (approved):

components/segment-bridge/base/serviceaccount.yaml
```diff
@@ -29,9 +29,99 @@ roleRef:
...
+roleRef:
+  kind: Role
+  name: segment-bridge-read-member-sa-secret
+  apiGroup: rbac.authorization.k8s.io
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: segment-bridge-read-access
+  namespace: toolchain-host-operator
```
Michkov, line 68: This role should not be here. I would put it into components/authentication. CRs should not be created in namespaces which are not managed by the component.

Omeramsc: In that case I should put all the `namespace: toolchain-host-operator` CRs there, right?

components/authentication/base/segment-bridge.yaml
```diff
@@ -0,0 +1,64 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
gbenhaim, line 1: Can you place it on a dedicated argocd application set? we are trying to reduce the manifests in the `authentication` component (you can do it in a different PR).

gbenhaim (approved):

PR #2087, opened 2023-07-11
GitOps Service: deploy to staging only, update components/gitops/staging/base/kustomization.yaml

jgwest:
Cherry pick https://github.com/redhat-appstudio/infra-deployments/pull/2080/commits to staging, only


jgwest:
Merging as the HAC-E2E tests are optional, and failing equally on other PRs as well.

PR #2097, opened 2023-07-12
Enable generic Grafana dashboards

skabashnyuk:
- Generic operator SDK dashboard
 - Generic go process dashboard


components/monitoring/grafana/base/kustomization.yaml
```diff
@@ -15,6 +15,7 @@ resources:
 - release/
 - performance/
 - pipeline-service/
+- dashboards/
```
amisstea, line 18: Can we rename this directory? `dashboards` is too generic IMO. All other directories referenced here contain dashboards.

amisstea (commented):
I'd like to get @avi-biton's input on this. The new guidelines are to create dashboards in team specific repos but that might be too restrictive when dealing with generic dashboards like these.

components/monitoring/grafana/base/kustomization.yaml
avi-biton: I think that `generic-dashboards` would be a good name for the directory

avi-biton:
@amisstea although the guidelines are to create dashboards in team specific repos, but as you mentioned, this is a generic dashboard and not a team's dashboard, I believe it can be created the way it is (in `infra-deployments`).

avi-biton (commented):
/lgtm

amisstea (approved):
/lgtm

PR #2101, opened 2023-07-12
Deploy proxy plugin as part of toolchain for development clusters

jrichter1:
As per [this doc](https://docs.google.com/document/d/1_VanTqldsChZXhUEG7n4gMyKiOL_lGGTNNIQM6xe6Ag/edit#heading=h.gn1t899dfpgm) a proxy plugin is needed to access the tekton-results API.

Adding one so it works out of the box on development clusters.


prietyc123:
@jrichter1 I see you are adding proxy-plugin for `tekton-results` here. Doesn't this affect the results test in e2e https://github.com/redhat-appstudio/e2e-tests/blob/main/tests/build/build_templates.go#L155-L172 . In case it affects the test would you mind updating those test as well accordingly.

prietyc123:
> @jrichter1 I see you are adding proxy-plugin for tekton-results here. Doesn't this affect the results test in e2e https://github.com/redhat-appstudio/e2e-tests/blob/main/tests/build/build_templates.go#L155-L172 . In case it affects the test would you mind updating those test as well accordingly.

I am holding it until we test it with the `pipeline` labeled e2e-test as this change is directly related to pipeline. Since on Infra-deployment AFAIK we don't test with `pipeline` label, ci might pass. My other concerns is it might break the other ci for ex periodic ci, after merge if we don't test the related test. 

/hold

jrichter1:
closing in favour of https://github.com/openshift/release/pull/41724

PR #2105, opened 2023-07-12
Update Pipeline Service

sayan-biswas:
## Changes
- Remove tekton results service monitor configs from infra
- Bump pipeline service which includes tekton result service monitor configs upstream

Manual PR for #2104


gabemontero:
looks like an unrelated toolchain startup problem

/test appstudio-e2e-tests

PR #2114, opened 2023-07-16
chore(RHTAPWATCH-376): segment-bridge member-cluster component

Omeramsc:
Deploy a component for the User Journey effort as segment-bridge. This is needed in order to have a Service account with read permissions to the Namespaces resources from the member clusters. Those resources will be used in a script that'll map between events and cluster's owner

### DIFF
--- a/components/segment-bridge/OWNERS
+++ b/components/segment-bridge/OWNERS
@@ -1,0 +1,6 @@
+# See the OWNERS docs: https://go.k8s.io/owners
+
+approvers:
+- Omeramsc
+- ifireball
+
--- a/components/segment-bridge/README.md
+++ b/components/segment-bridge/README.md
@@ -1,0 +1,7 @@
+# SEGMENT-BRIDGE
+
+Component used to allow reading resources from the host and member clusters
+
+Serves as part of the User Journey effort,
+related information can be found at the [segment-bridge repository](https://github.com/redhat-appstudio/segment-bridge)
+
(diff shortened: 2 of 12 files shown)


components/segment-bridge/OWNERS
```diff
@@ -0,0 +1,5 @@
+# See the OWNERS docs: https://go.k8s.io/owners
+
+approvers:
+- Omeramsc
+- ifireball
```
ifireball, line 5: missing EOF newline

Same goes for multiple other files

Omeramsc: fixed

components/segment-bridge/README.md
```diff
@@ -0,0 +1,7 @@
+# SEGMENT-BRIDGE
+
+Component used to allow reading resources from the staging member cluster
```
amisstea, line 3:
```suggestion
Component used to allow reading resources from the host and member clusters.
```

I see you've got duplicate changes in https://github.com/redhat-appstudio/infra-deployments/pull/2054/files#diff-442738682d1dc0a832b81da44904ca09c3a5c47b534ffcd462f00ac81226be80R3. You might as well make one version accurate.

Omeramsc: I guess you're right for this specific line, otherwise I preferred each change to be atomic, as rebasing will need to happen either way...

Omeramsc:
Since there's a chance the toolchain namespace won't be set in development (due to a flag configuration), leading to a sync issue, and as we don't need the development environment in the foreseeable future for this component - I've added the component to the delete overlay of development

gbenhaim (approved):

PR #2128, opened 2023-07-19
Add topology spread constraint

stuartwdouglas:
This should distribute the load between the nodes more evenly.


stuartwdouglas:
This is passing, but while we still have a mix of node types this might actually make things worse (as it will evenly distribute between the smaller and larger nodes).

stuartwdouglas:
Closing for now, as I don't think we want this unless we have all worked nodes sized in a similar manner.

PR #2136, opened 2023-07-21
Fix outdated paths in hack/destroy-cluster.sh

tkdchen:


hack/destroy-cluster.sh
```diff
@@ -24,16 +24,15 @@ echo
 echo "Starting with removing application, you can see progress $ARGO_CD_URL"
 echo "If there is running Sync then cancel it manually"
 echo "Remove Argo CD Applications:"
-kubectl delete -f $ROOT/argo-cd-apps/app-of-apps/all-applications-staging.yaml
-#kubectl -k $ROOT/argo-cd-apps/overlays/staging
+kubectl delete -k "$ROOT/argo-cd-apps/app-of-app-sets/staging"
```
tkdchen, line 27: Not sure if there is a correct replacement. Any idea?

PR #2149, opened 2023-07-24
monitoring: Add GitOps Service's Argo CD Dashboard

jgwest:
The contents of the dashboard that monitors the Argo CD instance in `gitops-service-argocd` is already part of the ConfigMap, we just need to create the corresponding `GrafanaDashboard` CR.


jgwest:
Merging as tests are passing, and it's a dashboard change only.

PR #2158, opened 2023-07-26
Fix(RHTAPWATCH-8): Upgrade to Grafana Operator v5

avi-biton:
We are upgrading the Grafana Operator from v4 to v5 as part of a fix to [RHTAPWATCH-8](https://issues.redhat.com//browse/RHTAPWATCH-8) (Somtimes the Grafana insnace is not reachable)

Upgrade to Grafana Operator v5:

- Update Grafana Operator to v5.3.0
- Update Grafana deployment yaml
- Update the APIs in all CRs (Datasources, Dashboards)
- Remove datasource definition from `generic-dashboards`
- Update README file

Notes
Dashboards that their source is in the team's repository were not updated to fit the new API.  
After this PR will get merge these dashboards would not be presented in Grafana.  
We will update the teams' dashboards in their repositories.  
These are the dashboards and repositories:

- Integration - https://github.com/redhat-appstudio/integration-service/blob/main/config/grafana/dashboard.yaml
- Performance - https://github.com/redhat-appstudio/perfscale/blob/main/grafana/dashboard.yaml
- Pipeline-service - https://github.com/openshift-pipelines/pipeline-service/blob/main/operator/gitops/argocd/grafana/dashboard.yaml
- Release - https://github.com/redhat-appstudio/release-service/blob/main/config/grafana/dashboard.yaml

### DIFF
--- a/components/monitoring/grafana/base/grafana-operator.yaml
+++ b/components/monitoring/grafana/base/grafana-operator.yaml
@@ -7,12 +7,11 @@ metadata:
   name: grafana-operator
   namespace: appstudio-grafana
 spec:
-  channel: v4
+  channel: v5
   installPlanApproval: Automatic
   name: grafana-operator
   source: community-operators
   sourceNamespace: openshift-marketplace
-  startingCSV: grafana-operator.v4.10.0
 ---
 apiVersion: operators.coreos.com/v1
 kind: OperatorGroup
--- a/components/monitoring/grafana/base/kustomization.yaml
+++ b/components/monitoring/grafana/base/kustomization.yaml
@@ -26,3 +26,20 @@ namespace: "appstudio-grafana"
 
 configurations:
   - cm-dashboard.yaml
+
+# Migration patch from Grafana v4 to v5, should be removed when all projects are migrated
+patchesJson6902:
+  - target:
+      group: integreatly.org
+      version: v1alpha1
+      kind: GrafanaDashboard
+      name: '.*'
+    patch: |-
+      - op: replace
+        path: "/apiVersion"
+        value: grafana.integreatly.org/v1beta1
+      - op: add
+        path: /spec/instanceSelector
+        value:
+          matchLabels:
+            dashboards: "appstudio-grafana"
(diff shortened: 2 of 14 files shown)


avi-biton:
@gbenhaim @amisstea please take a look

components/monitoring/grafana/base/kustomization.yaml
```diff
@@ -16,6 +16,7 @@ resources:
 - pipeline-service/
 - generic-dashboards/
 
+
```
yftacherzog, line 19: Do we need this change?

components/monitoring/grafana/development/kustomization.yaml
```diff
@@ -2,8 +2,8 @@ apiVersion: kustomize.config.k8s.io/v1beta1
...
-patches:
-  - path: auto-assign-role-patch.yaml
-    target:
-      name: grafana-oauth
-      kind: Grafana
+# patches:
+#   - path: auto-assign-role-patch.yaml
+#     target:
+#       name: grafana-oauth
+#       kind: Grafana
```
yftacherzog, line 9: Can this be deleted instead?

components/monitoring/grafana/base/kustomization.yaml
avi-biton: Removed

components/monitoring/grafana/development/kustomization.yaml
avi-biton: I removed the comments and added it back

components/monitoring/grafana/base/grafana-operator.yaml
```diff
@@ -7,12 +7,12 @@ metadata:
...
   namespace: appstudio-grafana
 spec:
-  channel: v4
+  channel: v5
   installPlanApproval: Automatic
   name: grafana-operator
   source: community-operators
   sourceNamespace: openshift-marketplace
-  startingCSV: grafana-operator.v4.10.0
+  startingCSV: grafana-operator.v5.3.0
```
Michkov, line 15: I would avoid setting this, in the past I could see that for new installations the stratingCSV version was installed and then updated to latest. So it's better to install newest version from the channel.

Michkov (approved):

components/monitoring/grafana/base/grafana-operator.yaml
avi-biton: I removed it

Michkov (approved):

gbenhaim (approved):

PR #2163, opened 2023-07-28
RHTAPSRE-64: allow limited set of users access to the cluster entitlement secrets

gabemontero:
@ernesgonzalez33 @dfodorRH @gbenhaim @brianwcook @arewm @ralphbean FYI you are being given view access to the entitlement secrets on the various RHTAP clusters based on https://issues.redhat.com/browse/RHTAPSRE-64 and the discussions in https://redhat-internal.slack.com/archives/C02CTEB3MMF/p1688479083511429


gabemontero:
Note - this level of RBAC changes does *NOT* require e2e testing.

gabemontero:
/hold

there is some disagreement in https://redhat-internal.slack.com/archives/C02CTEB3MMF/p1688479083511429 on whether we should do this, vs. just let users discover their entitlement secrets through other means and copy them into their namespace

PR #2179, opened 2023-07-31
managed-gitops update

jgwest:
Bring production container image up-to-date with dev/staging from https://github.com/redhat-appstudio/infra-deployments/pull/2155


jgwest:
Merging as tests are passing, and this was already approved as part of #2155 (upon which this PR is based)

PR #2237, opened 2023-08-10
Initial prototype of Multi Arch Controller

stuartwdouglas:


hack/preview.sh
```diff
@@ -201,6 +201,11 @@ sed -i.bak "s/rekor-server.enterprise-contract-service.svc/$rekor_server/" $ROOT
 [ -n "${IMAGE_CONTROLLER_IMAGE_TAG}" ] && yq -i e "(.images.[] | select(.name==\"quay.io/redhat-appstudio/image-controller\")) |=.newTag=\"${IMAGE_CONTROLLER_IMAGE_TAG}\"" $ROOT/components/image-controller/base/kustomization.yaml
 [[ -n "${IMAGE_CONTROLLER_PR_OWNER}" && "${IMAGE_CONTROLLER_PR_SHA}" ]] && yq -i e "(.resources[] | select(. ==\"*github.com/redhat-appstudio/image-controller*\")) |= \"https://github.com/${IMAGE_CONTROLLER_PR_OWNER}/image-controller/config/default?ref=${IMAGE_CONTROLLER_PR_SHA}\"" $ROOT/components/image-controller/base/kustomization.yaml
 
+[ -n "${MULTI_ARCH_CONTROLLER_IMAGE_REPO}" ] && yq -i e "(.images.[] | select(.name==\"multi-arch-controller\")) |=.newName=\"${MULTI_ARCH_CONTROLLER_IMAGE_REPO}\"" $ROOT/components/multi-arch-controller/base/kustomization.yaml
```
sbose78, line 204: Would the local deployment be able to silently ignore this functionality if they wished to ( by not configuring any aws keys for example ) ?

stuartwdouglas: We can definitely make it just do nothing if no hosts are configured. At the moment the hosts config is just hard coded to the IP of two AWS instanced I have spun up. If one of them goes down or restarts the we will need to update infra-deployments, which does not seem ideal.

components/tekton-ci/base/repository.yaml
```diff
@@ -180,3 +180,10 @@ metadata:
   name: pipeline-service
 spec:
   url: "https://github.com/openshift-pipelines/pipeline-service"
+---
+apiVersion: pipelinesascode.tekton.dev/v1alpha1
+kind: Repository
+metadata:
+  name: multi-arch-controller
```
sbose78, line 187: Is this relevant?
https://github.com/redhat-appstudio/multi-arch-controller/blob/main/.tekton/multi-arch-controller-pull-request.yaml#L18

sbose78 (changes requested):

components/tekton-ci/base/repository.yaml
stuartwdouglas: Not yet, but if I have it here it gives me the option of setting up builds with automated PR creation in the same manner as other repositories.

stuartwdouglas: Building in RHTAP was ok for the initial PR, but if we start iterating rapidly on this it will likely be a lot more convenient to have the automation.

sbose78 (approved):

PR #2238, opened 2023-08-10
chore: segment-bridge has the right usersignup access

Omeramsc:
The apiGroups set in the segment-bridge configuration to access toolchain-host-operator were wrong.
This PR fix it by changing it to toolchain.dev.openshift.com.

### DIFF
--- a/components/authentication/base/segment-bridge.yaml
+++ b/components/authentication/base/segment-bridge.yaml
@@ -4,8 +4,8 @@ metadata:
   name: segment-bridge-read-access
   namespace: toolchain-host-operator
 rules:
-- apiGroups: [""]
-  resources: ["UserSignup"]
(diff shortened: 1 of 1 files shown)


components/authentication/base/segment-bridge.yaml
```diff
@@ -4,7 +4,7 @@ metadata:
   name: segment-bridge-read-access
   namespace: toolchain-host-operator
 rules:
-- apiGroups: [""]
+- apiGroups: ["toolchain.dev.openshift.com"]
   resources: ["UserSignup"]
```
Michkov, line 8: resources must be in plural and lower case -> usersignups (as far as I know)

Omeramsc: You're right, fixed.

Michkov (approved):

PR #2264, opened 2023-08-16
DO-NOT-MERGE: use my build-definintions experiment with timestamps and buildah debug

gabemontero:
@jhutar @pmacik let's see if this picks up my test commit per https://github.com/gabemontero/build-definitions/commits/main


gabemontero:
yeah @pmacik @jhutar patching the build definitions like this does not have the desired affect of picking up one of our pre-merged versions.

I think we need to sort out how to define `BuildPipelineSelector` overrides in our CI tests.

PR #2267, opened 2023-08-17
Install RHTAP locally to support Sprayproxy + Github App

xinredhat:


hack/deploy-sprayproxy.sh
```diff
@@ -0,0 +1,18 @@
...
+
+# ------------------------------------------------------------------------------------
+# Spray Proxy that is used for distrubuting Github App requests to ephemeral clusters
+# ------------------------------------------------------------------------------------
+SPRAYPROXY_QE_TEM_DIR=/tmp/prayproxy-qe
+
+echo
+echo "Deploying SprayProxy on the cluster"
+rm -rf ${SPRAYPROXY_QE_TEM_DIR} 2>/dev/null || true
+git clone --depth=1 https://github.com/xinredhat/sprayproxy-qe.git ${SPRAYPROXY_QE_TEM_DIR}
```
Michkov, line 11: we should not use personal git repository. But anyway it would be better to create development overlay in https://github.com/redhat-appstudio/infra-deployments/tree/main/components/sprayproxy and deploy in common way.

xinredhat: @Michkov this Sprayprox is not part of RHTAP, which is used as a proxy  between the Github App and test cluster to distribute GithuApp webhook's requests to Pac running on the test cluster. It doesn't need to deploy this Sprayproxy when launching RHTAP CI, as there is a permeant Sprayproxy server. So  it doesn't make sense to create development overlay in https://github.com/redhat-appstudio/infra-deployments/tree/main/components/sprayproxy . Can we put it to other repo under a particular Github org?

Michkov: Not seeing reason to develop something by RHTAP team which is not part of RHTAP. Global servers could be issues in general and my understading was that we are trying to avoid it. Also when the "qe" sprayproxy would be updated then it could break all the testing. I do not see reason why to handle SprayProxy as standalone server outside of RHTAP deployment.

CC @sbose78 @ralphbean @arewm @adambkaplan - WDYT?

adambkaplan: @xinredhat Sprayproxy is most definitely part of our current "multi-cluster" RHTAP solution. If you are deploying RHTAP with a "host cluster" for namespace/workspace management, Sprayproxy is used to forward Github webhooks to member clusters.

xinredhat: @adambkaplan  This Sprayproxy deployment is for local testing purpose.   There is a Sprayproxy running on a permanent OCP cluster, which is used to register and unregister backend clusters on fly to test RHTAP on multiple ephemeral clusters in e2e CI. When debugging Pac tests locally, the preview.sh will deploy this Sprayproxy on cluster. I don't think we can deploy it via  argocd.

hack/build/setup-pac-integration.sh
```diff
@@ -48,19 +52,21 @@ setup-pac-app() (
         payload="{ \"iss\": $PAC_GITHUB_APP_ID, \"iat\": ${now}, \"exp\": $((now+10)) }"
 
         webhook_secret=$(openssl rand -hex 20)
+        token=$(sign rs256 "$payload" "$(echo "$PAC_GITHUB_APP_PRIVATE_KEY" | base64 -d)")
+        # return  sprayproxy host in a loop until it is available
+        while ! kubectl get -n sprayproxy route/sprayproxy-route &>/dev/null; do
+                echo "Waiting for sprayproxy route to be available"
```
Michkov, line 58: must go to stderr, otherwise it's retuned as the webhook secret itself and secret is created this way:
```
Waiting for sprayproxy route to be available
Waiting for sprayproxy route to be available
Waiting for sprayproxy route to be available
Waiting for sprayproxy route to be available
Waiting for sprayproxy route to be available
Waiting for sprayproxy route to be available
Waiting for sprayproxy route to be available
Waiting for sprayproxy route to be available
Waiting for sprayproxy route to be available
Waiting for sprayproxy route to be available
Waiting for sprayproxy route to be available
Waiting for sprayproxy route to be available
Waiting for sprayproxy route to be available
$TOKEN
```
But anyway I am missing configuration of spray proxy to point to PAC. Maybe waiting here for the sprayproxy route is not needed and can be hardcoded.

xinredhat: >must go to stderr

fixed

xinredhat: > configuration of spray proxy to point to PAC

@Michkov regarding this, I am not sure where it's the best place. When e2e CI is launched , it needs to register Pac in a permanent Sprayprox. When e2e CI finish, we want to unregister the pac url from the Sprayprox. So maybe e2e repo is the best place to implement registering/unregistering Pac to sprayproxy.

Michkov: that echo still goes to stdout

xinredhat: fixed

hack/build/setup-pac-integration.sh
```diff
@@ -48,19 +52,21 @@ setup-pac-app() (
         payload="{ \"iss\": $PAC_GITHUB_APP_ID, \"iat\": ${now}, \"exp\": $((now+10)) }"
 
         webhook_secret=$(openssl rand -hex 20)
+        token=$(sign rs256 "$payload" "$(echo "$PAC_GITHUB_APP_PRIVATE_KEY" | base64 -d)")
+        # return  sprayproxy host in a loop until it is available
+        while ! kubectl get -n sprayproxy route/sprayproxy-route >/dev/null 2>&1 ; do
```
prietyc123, line 57: @xinredhat As your manifest points to https://github.com/redhat-appstudio/sprayproxy/blob/main/config ,here I don't see [sprayproxy route](https://github.com/redhat-appstudio/sprayproxy/blob/main/config/route.yaml) inside `sprayproxy` namespace. May be that could be reason you are not getting the expected route

prietyc123 (changes requested):

hack/build/setup-pac-integration.sh
```diff
@@ -48,19 +52,21 @@ setup-pac-app() (
...
+                echo "Waiting for sprayproxy route to be available" >&2
+                sleep 1
+        done
+        sprayproxy_host=$(kubectl get -n sprayproxy -o template --template="{{.spec.host}}" route/sprayproxy-route)
```
prietyc123, line 61: same as [comment](https://github.com/redhat-appstudio/infra-deployments/pull/2267/files#r1302529636)

prietyc123 (dismissed):

hack/build/setup-pac-integration.sh
xinredhat: @prietyc123 it's there, you can verify the route namespace by executing  `kubectl kustomize development` , it will print out all the final resources.  The follow is the route.
```
---
apiVersion: route.openshift.io/v1
kind: Route
metadata:
  annotations:
    haproxy.router.openshift.io/hsts_header: max-age=63072000
  name: sprayproxy-route
  namespace: sprayproxy
spec:
  port:
    targetPort: proxy
  tls:
    insecureEdgeTerminationPolicy: Redirect
    termination: reencrypt
  to:
    kind: Service
    name: sprayproxy
```

hack/build/setup-pac-integration.sh
xinredhat: same as above

hack/build/setup-pac-integration.sh
prietyc123: Ahh my mistake

hack/build/setup-pac-integration.sh
```diff
@@ -1,7 +1,11 @@
 #!/usr/bin/env bash
+set -x 
 
 PAC_NAMESPACE='openshift-pipelines'
 PAC_SECRET_NAME='pipelines-as-code-secret'
+SPRAYPROXY_NAMESPACE='sprayproxy'
+
+ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"/../..
```
enarha, line 8: Why do you need that? It's not used anywhere.
It looks like you are trying to capture the Git top level directory. If that's the case `git rev-parse --show-toplevel` is shorter and not dependable to the script path.

hack/build/setup-pac-integration.sh
enarha, line 57: Maybe run this a reasonable number of times (100?) instead of an infinite loop? Even for a local script it's better experience instead of leaving it to the user to decide how long to wait.

components/sprayproxy/development/sprayproxy-deployment-patch.yaml
```diff
@@ -0,0 +1,18 @@
...
+  value:
+    name: SPRAYPROXY_SERVER_INSECURE_SKIP_TLS_VERIFY
+    value: "true"
+- op: add
+  path: /spec/template/spec/containers/0/env/-
+  value:
+    name: SPRAYPROXY_SERVER_ENABLE_DYNAMIC_BACKENDS
+    value: "true"
+
+# remvoe GH_APP_WEBHOOK_SECRET variable
```
enarha, line 13: typo: remove

xinredhat: fixed

hack/build/setup-pac-integration.sh
xinredhat: fixed

hack/build/setup-pac-integration.sh
xinredhat: fixed.

hack/build/setup-pac-integration.sh
```diff
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+set -x
```
Michkov, line 2: is that needed?

xinredhat: fixed.

Roming22 (approved):
/lgtm

hack/preview-template.env
```diff
@@ -100,7 +100,6 @@ export DOCKER_IO_AUTH=""
 ### pipelines-as-code-secret is created by preview.sh
 export PAC_GITHUB_APP_PRIVATE_KEY= # Base64 encoded private key of the GitHub APP
 export PAC_GITHUB_APP_ID= # Application ID
-export PAC_GITHUB_APP_WEBHOOK_SECRET= # Webhook secret of the GitHub APP
```
Michkov, line 103: why this env var was removed when it's still used? Only description comment should be updated.

xinredhat: @Michkov my fault. I added it back in PR https://github.com/redhat-appstudio/infra-deployments/pull/2316

PR #2271, opened 2023-08-17
move sandbox-sre-admin RBAC to sandbox component

MatousJobanek:


alexeykazakov (approved):

Michkov:
This is failing because it's applying ClusterRoles from sandbox-sre-admins.yaml on member and host operator. That's causing conflict since both of them are on the same cluster. Possible solution is to create development overlay which will not apply the RBAC or it will be applying it only on eg. member operator.

mfrancisc:
@Michkov thanks for your suggestion. Could you please take a look at https://github.com/redhat-appstudio/infra-deployments/pull/2284 , If I'm not missing something this should apply the ClusterRole only on member-operator cluster when in development mode (which is the same as host operator cluster).

mfrancisc:
@MatousJobanek the changes got merged with https://github.com/redhat-appstudio/infra-deployments/pull/2284 , I think we can close this one.

MatousJobanek:
@mfrancisc thanks for taking care of it :pray: 
closing in favor of https://github.com/redhat-appstudio/infra-deployments/pull/2284

PR #2259, opened 2023-08-16
RBAC Restructuring, moving rbac(s) from `components/authentication/ba…

manish-jangra:
…se` to service specific directory

1. segment-bridge.yaml has been moved to `components/segment-bridge/base/rbac/`
2. has-admin.yaml and has.yaml have been moved to `components/has/base/rbac/`
3. dora-metrics.yaml has been moved to `components/dora-metrics/base/rbac/`
4. internal-services.yaml --> `components/internal-service`
5. cluster-registration-maintainer.yaml --> `components/cluster-registration`
6. integration.yaml --> `components/integration`
7. release-team.yaml --> `components/release`
8. spi.yaml --> `components/spi/base/rbac`

PR #2273, opened 2023-08-17
Fix RBAC on prod for HAS team

johnmcollier:
RBAC not getting applied to HAS component since https://github.com/redhat-appstudio/infra-deployments/pull/2259 was merged


kim-tsao (approved):
/lgtm

PR #2277, opened 2023-08-18
All the Service Specific Roles && RoleBindings have been moved to the…

manish-jangra:
…ir respective Service directories. Cluster Scoped Cluster Role and Cluster Role Bindings are still available in `components/authentication/base` and move sandbox-sre-admin RBAC to sandbox component


manish-jangra:
So not merge, work in progress

PR #2279, opened 2023-08-19
Allow GitOps team members to access jgwest-tenant namespace where it is available

jgwest:
- On staging MT, allow GitOps Service Team members group to access `jgwest-tenant`
- On prod MT, allow only  `jgwest` github user id to access `jgwest-tenant`


keithchong (commented):
LGTM

jgwest:
Thank @keithchong! Merging as only UI tests are failing, which are optional (and it's very unlikely this PR would cause issues with them).

PR #2295, opened 2023-08-22
move inspect-pods clusterrole to sandbox component

mfrancisc:
Move `authentication/base/inspect-pods.yaml` to `sandbox` component and deploy it only on member clusters for all environements (stage, prod and dev)


manish-jangra (approved):
looks good, all the tests are also passing

bamachrn (approved):
/lgtm

PR #2312, opened 2023-08-24
RHTAP-1495: AppSet for RH users configurations

gbenhaim:
Add a new application set that will generate application for RH each tenant defined in https://github.com/redhat-appstudio/tenants-config Each tenant has a directory with the name if his/hers namespace. The generate application will be associated with the "tenants-config" project that restricts which resources can be deployed, to which namespaces and clusters. The project is defined using app-interface.

### DIFF
--- a/argo-cd-apps/base/tenants-config/tenants-config.yaml
+++ b/argo-cd-apps/base/tenants-config/tenants-config.yaml
@@ -1,0 +1,36 @@
+apiVersion: argoproj.io/v1alpha1
+kind: ApplicationSet
+metadata:
+  name: tenants-config
+spec:
+  generators:
+    - git:
+        repoUrl: https://github.com/redhat-appstudio/tenants-config
+        revision: main
+        directories:
+          - path: auto-generated/cluster/stone-prd-rh01/*
+  template:
+    metadata:
+      name: '{{path.basename}}-{{path[2]}}'
+    spec:
+      # The project is created using app-interface
+      project: tenants-config
+      source:
+        path: '{{path}}'
+        repoURL: https://github.com/redhat-appstudio/tenants-config
+        targetRevision: main
+      destination:
+        namespace: '{{path.basename}}'
+        name: '{{path[2]}}'
+      syncPolicy:
+        automated:
+          prune: true
+          selfHeal: true
+        syncOptions:
+          - CreateNamespace=true
+        retry:
+          limit: 10
+          backoff:
+            duration: 10s
+            factor: 2
+            maxDuration: 3m
--- a/argo-cd-apps/overlays/production/kustomization.yaml
+++ b/argo-cd-apps/overlays/production/kustomization.yaml
@@ -6,6 +6,7 @@ resources:
   - ../../base/all-clusters
   - ../../base/tekton-ci
   - ../../base/ci-workspaces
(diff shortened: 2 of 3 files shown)


argo-cd-apps/overlays/production/kustomization.yaml
```diff
@@ -6,6 +6,7 @@ resources:
   - ../../base/all-clusters
   - ../../base/tekton-ci
   - ../../base/ci-workspaces
+  - ../../base/tenants-config
```
Michkov, line 9: does not exist

Michkov (dismissed):

argo-cd-apps/overlays/production/kustomization.yaml
gbenhaim: thanks, fixed.

argo-cd-apps/base/tenants-config/tenants-config.yaml
```diff
@@ -0,0 +1,36 @@
+apiVersion: argoproj.io/v1alpha1
+kind: ApplicationSet
+metadata:
+  name: tenants-config
+spec:
+  generators:
+    - git:
+        repoUrl: https://github.com/redhat-appstudio/tenants-config
+        revision: main
```
Michkov, line 9: should be pinned to specific version

gbenhaim: We  don't want to pin it to specific version. We want ArgoCD to track main for this application.

argo-cd-apps/base/tenants-config/tenants-config.yaml
```diff
@@ -0,0 +1,36 @@
...
+        namespace: '{{path.basename}}'
+        name: '{{path[2]}}'
+      syncPolicy:
+        automated:
+          prune: true
+          selfHeal: true
+        syncOptions:
+          - CreateNamespace=true
+        retry:
+          limit: -1
```
ernesgonzalez33, line 32: Why do we need unlimited retries?

gbenhaim: fixed

argo-cd-apps/base/tenants-config/tenants-config.yaml
Michkov: None of pull-requests in that project has been reviewed before merging, so I do not trust that project.

Michkov: (but not blocking this PR)

My understading is that we do not want to have some random unreviewed changed, so that's why we are always pinning to specific sha.

CC: @sbose78 @arewm

gbenhaim: pr will get reviewd on https://github.com/redhat-appstudio/tenants-config
I just added a simple directory structure to it so we can get things rolling.

scoheb (approved):

argo-cd-apps/base/tenants-config/tenants-config.yaml
scoheb: new PR with required review: https://github.com/redhat-appstudio/tenants-config/pull/4

PR #2313, opened 2023-08-24
fix: authentication is not having sandbox-sre-admins.yaml

Michkov:
sandbox-sre-admins.yaml has been moved to
./components/sandbox/base/rbac/, this is causing fail of sync of authentication component.


manish-jangra (approved):
looks good

PR #2328, opened 2023-08-28
Fix staging sprayproxy duplicated GH_APP_WEBHOOK_SECRET

hugares:
GH_APP_WEBHOOK_SECRET is defined in the config at the repo level so using a "patch add" was resulting in duplicated variable:

```
spec:
  template:
    spec:
      containers:
      - args:
        env:
        - name: GH_APP_WEBHOOK_SECRET
        - name: GH_APP_WEBHOOK_SECRET
          valueFrom:
            secretKeyRef:
              key: webhook.secret
              name: pipelines-as-code-secret
```

Instead, patch the varialble using a strategic merge. In addition to fix the duplicate, this methodology do not rely on container or variable indices so also change the SPRAYPROXY_SERVER_BACKEND to use same mechanism for robusness and consistency reasons.

### DIFF
--- a/components/sprayproxy/staging/add-backends.yml
+++ b/components/sprayproxy/staging/add-backends.yml
@@ -1,6 +1,0 @@
----
-- op: add
-  path: /spec/template/spec/containers/0/env/0/value
-  value: >
-    https://pipelines-as-code-controller-openshift-pipelines.apps.stone-stg-m01.7ayg.p1.openshiftapps.com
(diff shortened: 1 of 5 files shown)


enarha:
@hugares thanks for the fix. Personally I also prefer the more explicit patches instead of the replace directives.

hugares:
@enarha Could you re-apply LGTM and approved label

hugares:
@sayan-biswas @enarha This is the last kube-linter issue for spray proxy, could you help getting it in then I will promote the change to prod once I see it working on staging

gabemontero (approved):
/lgtm

PR #2331, opened 2023-08-29
Fix kube-linter issues for cluster-registration component

hugares:
Set readOnlyRootFilesystem and runAsNonRoot to true for cluster-registration container.


hugares:
component is being removed(https://github.com/redhat-appstudio/infra-deployments/pull/2352), no more need for this MR

PR #2336, opened 2023-08-30
HACBS-2536: create new RH managed-workspace config project

scoheb:


argo-cd-apps/base/rh-managed-workspaces-config/rh-managed-workspaces-config.yaml
```diff
@@ -0,0 +1,36 @@
...
+          - path: auto-generated/cluster/stone-prd-rh01/*
+  template:
+    metadata:
+      name: '{{path.basename}}-{{path[2]}}'
+    spec:
+      # The project is created using app-interface
+      project: rh-managed-workspaces-config
+      source:
+        path: '{{path}}'
+        repoURL: https://github.com/redhat-appstudio/rh-managed-workspaces-config
```
gbenhaim, line 20: I think it would be better to use https://github.com/redhat-appstudio/tenants-config and just use another subfolder. The reason for that is that we need to maintain CI for validating the manifests, for example that they were built before submitting, run kube-lint, and probably other checks (for example ensuring that one tenants tried to deploy to another tenants workspace).

We can use different sub folders for each project:

 auto-generated/cluster/stone-prd-rh01/tenant/* - for user workspaces
 auto-generated/cluster/stone-prd-rh01/managed/* - for managed workspaces

PR #2339, opened 2023-08-31
fix: avoid deploying sandbox clusterroles in dev mode

mfrancisc:
Fix for https://issues.redhat.com/browse/SANDBOX-365

Toolchain operators are optional in dev mode, and only deployed if `--toolchain` option is specified. And these roles are not needed in dev mode anyway.


alexeykazakov (approved):

MatousJobanek (approved):
/lgtm

PR #2348, opened 2023-09-01
Add horreum to components

dustinblack:
This change add Horreum to the components via an operator. This currently includes dedicated postgresql and keycloak pods as a proof-of-concept. There is also a dedicated grafana pod, but that is vestigial and will be removed in a later update.

*Currently pending needed updates to the Horreum operator to fix an SSL issue, remove grafana, and correct the pod command for keycloak.*


jhutar:
I think that given a recent changes in the project, we will not need this. Current plan is t use Red Hat internal Horreum instance. We can reopen if plans change again :-/ Thank you Dustin for a great work!

PR #2352, opened 2023-09-04
Removing cluster-registration-operator

Katka92:
The `cluster-registration-operator` seems tobe a leftover from past times and is not used anymore. Removing it as it is currently also breaking OSD support (when the RHTAP deployed on OSD) as it is able to be scheduled on the infra nodes.

Related support case: [03584774](https://access.redhat.com/support/cases/#/case/03584774)


manish-jangra:
@gbenhaim should be the right person to review this.

psturc:
failure caused by an issue fixed [here](https://github.com/openshift/release/pull/43250)
/retest-required

amfred:
Looks like this PR will remove it from development and staging now, and then we will need a follow up to remove it from production.
In the architecture call, people said it should not longer be used, so this should be OK.

amfred:
@Katka92 Don't you also need to update the kustomization file in the overlays/development directory to add this?
```

patchesStrategicMerge:
  - delete-applications.yaml
```

hugares:
> patchesStrategicMerge:
>   - delete-applications.yaml

@amfred This is already in the overlay/development/kustomization.yaml file. The reason it is added in staging is because we did not already have a delete-applications.yaml file for that overlay

kasemAlem:
failed due to blocker [RHTAPBUGS-769](https://issues.redhat.com/browse/RHTAPBUGS-769)

hugares (approved):

hugares:
@Katka92 will you check on staging to make sure everything is ok and then do the follow up PR to remove cluster-registration-operator from prod?

Katka92:
Hi sure, I'll run some tests in dev mode or RHTAP and also check the staging. I'll leave it for several days and then create a PR for removal in prod if the stage is still okay.

hugares:
> Hi sure, I'll run some tests in dev mode or RHTAP and also check the staging. I'll leave it for several days and then create a PR for removal in prod if the stage is still okay.

@Katka92 Can we proceed with the removal of cluster-registration-operator from prod?

hugares:
> > Hi sure, I'll run some tests in dev mode or RHTAP and also check the staging. I'll leave it for several days and then create a PR for removal in prod if the stage is still okay.
> 
> @Katka92 Can we proceed with the removal of cluster-registration-operator from prod?

https://github.com/redhat-appstudio/infra-deployments/pull/2550

PR #2364, opened 2023-09-06
remove references to chains ns

ramessesii2:
The tekton-chains namespace is obsolete with the Tekton-Chains controller
      deployed with Openshift-Pipelines operator.

### Pipeline-Service fix - https://github.com/openshift-pipelines/pipeline-service/pull/751


gabemontero (approved):
/lgtm

components/cluster-secret-store/base/appsre-stonesoup-vault-secret-store.yaml
```diff
@@ -33,7 +33,6 @@ spec:
         - quality-dashboard
         - spi-system
         - group-sync-operator
-        - tekton-chains
```
gabemontero, line 36: Just realized @ramessesii2 @Roming22 - this will update production as well, since it is the base overlay, correct?

I think we want to isolate the removal solely to staging, no?

components/pipeline-service/base/rbac/kustomization.yaml
```diff
@@ -4,5 +4,4 @@ resources:
   - cluster-role
   - openshift-pipelines
   # - pipelines-as-code
-  # - tekton-chains
```
gabemontero, line 7: oh wait ... since this was commented out, we already have updated production in some sense ?

ramessesii2: That's what my understanding is as well. 
FYI @manish-jangra. Please let me know if that's no the case.

components/cluster-secret-store/base/appsre-stonesoup-vault-secret-store.yaml
ramessesii2: I didn't bother to isolate it because `tekton-chains` is already deployed via `openshift-pipelines` operator in the production as well and so I though it wouldn't make a difference in the production.
Anyways, to be on the safe side, I'll do a separate PR to address it. Thanks @gabemontero

ramessesii2: Or maybe you'd like to see how this works out in staging? So, I'll do some patch so that it's only reflected in staging.

ramessesii2: Update: https://github.com/redhat-appstudio/infra-deployments/pull/2364/files#diff-db4d5ee061a063bfe1c269a1e0c5709c900bb3e199bd1f24bb3e7c4a87d8b955R6

manish-jangra (approved):
Looks good to me

gabemontero:
Yeah @ramessesii2 based on how @enarha PR for pipelines-as-code went, I think we are OK, but let's get @enarha and @Roming22 to chime in, just in case I am missing something, and then we can lgtm this one

components/cluster-secret-store/staging/cluster-secret-store-patch.yaml
```diff
@@ -0,0 +1,26 @@
+# cluster-secret-store-patch.yaml
```
Roming22, line 1: I don't understand this change. Why is this file needed? Why is it added now (I would have expected an update that removes a reference to tekton-chains instead)? Why is it referencing things that are outside our control?

It feels to me that the changes in this dir should have been a separate PR.

ramessesii2: I'll try to clarify why a new file is added. So, first we want to rollout the changes around removal of `tekton-chains` ns in the staging. 
i. Now, this [ClusterSecretStore `appsre-stonesoup-vault`](https://github.com/redhat-appstudio/infra-deployments/blob/main/components/cluster-secret-store/base/appsre-stonesoup-vault-secret-store.yaml) is in the base of the kustomization and therefore it allows the `tekton-chains` ns both in staging and in cluster to use the vault.
ii. If I just remove [this line](https://github.com/redhat-appstudio/infra-deployments/blob/main/components/cluster-secret-store/base/appsre-stonesoup-vault-secret-store.yaml#L36), although that would make sure that `tekton-chains` ns which will be gone in future won't be allowed to use vault but at the same time it's going to affect both prod & stag.
iii. And therefore in accordance with [Gabe's ask](https://github.com/redhat-appstudio/infra-deployments/pull/2364#discussion_r1317257338) I came up with this patch. I don't know if there's an easier way to patch the resource to just exclude a specific ns.

This change will be temporary though. As soon as we verify this in staging, I'll raise a PR to update base file to remove `tekton-chains` ns.

Roming22 (approved):

PR #2374, opened 2023-09-07
Ignore differences for build request annotation

gbenhaim:
In order for configuring components usings ArgoCD, we need to ignore changes to the
build.appstudio.openshift.io/request annotation (it's being removed by the build controller).

### DIFF
--- a/argo-cd-apps/base/tenants-config/tenants-config.yaml
+++ b/argo-cd-apps/base/tenants-config/tenants-config.yaml
@@ -22,6 +22,11 @@ spec:
       destination:
         namespace: '{{path.basename}}'
         name: '{{path[2]}}'
+      ignoreDifferences:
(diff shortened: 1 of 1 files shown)


argo-cd-apps/base/tenants-config/tenants-config.yaml
```diff
@@ -22,6 +22,11 @@ spec:
       destination:
         namespace: '{{path.basename}}'
         name: '{{path[2]}}'
+      ignoreDifferences:
+        - group: appstudio.redhat.com
+          kind: Component
+          jsonPointers:
+            - /metadata/annotations/build.appstudio.openshift.io~1request
```
scoheb, line 29: is this desired: `~1request`?

scoheb (approved):

PR #2381, opened 2023-09-11
Allow build team to list and view SA and list secrets

tkdchen:
[STONEBLD-1611](https://issues.redhat.com//browse/STONEBLD-1611)

Build team members are added to allow list and view ServiceAccount and list secrets across namespaces.

### DIFF
--- a/components/build-service/base/rbac/build-maintainer.yaml
+++ b/components/build-service/base/rbac/build-maintainer.yaml
@@ -3,61 +3,23 @@ apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: build-maintainer
 rules:
-  - apiGroups:
-      - operators.coreos.com
-    resources:
-      - installplans
-    verbs:
+  - verbs:
       - get
       - list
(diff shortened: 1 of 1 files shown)


rcerven (approved):

components/build-service/base/rbac/kustomization.yaml
```diff
@@ -1,5 +1,6 @@
 resources:
 - build-maintainer.yaml
+- build-team-can-view-sa-secret.yaml
```
mmorhun, line 3: I think we cannot view secrets.

rcerven: yeah but that doesn't help with debugging tbh

tkdchen: only verb list is granted for Secret resource

Michkov (approved):

mmorhun (approved):

tkdchen:
Conflict is resolved.

PR #2385, opened 2023-09-11
Simplify RBACs for pipeline-service

Roming22:
@manish-jangra is trying to sanitize the RBACs in the repository.

I can't recall why we decided to have 2 sets of roles, but at the moment I believe we need just one for the team that grants us access to all tekton related resources, and gives us admin rights on the namespace we manage (so we can restart pods for example).

@sbose78 @jduimovich Please let me know if you do not agree with the content of this PR.


sbose78 (approved):

sbose78:
If its a small group with access, this should suffice.

manish-jangra:
looks good to me if you want to keep one cluster role 👍

components/pipeline-service/base/rbac/cluster-role/pipeline-service-sre.yaml
```diff
@@ -1,62 +1,31 @@
...
-  - eventlisteners
-  - triggerbindings
-  - triggers
-  - triggertemplates
+  - "*"
   verbs:
-  - get
-  - list
-  - watch
   - create
```
manish-jangra, line 23: need `create` access? Shouldn't that be handled by ArgoCD?

manish-jangra (changes requested):
need `create` access for tekton resources? Shouldn't that be handled by ArgoCD?

Roming22:
Force merge.

The content has been validated multiple times by now, and there's no reason for it to impact the tests since the changes are non-functional.

PR #2393, opened 2023-09-13
Create rhtap-admins cluster role and binding

hugares:
The managed built-in role dedicated-admins grants access to secrets in all namespaces and we do not want that. Create a new cluster role which will used instead of dedicated-admins and that role does not have the right to list secrets.

The initial set or rules proposed in this change was put together by inspecting the dedicated-admins role, which is an aggregated cluster role. Some permissions might be missing as this is an initial proposal only and we can add missing ones as we find them.

[RHTAPSRE-168](https://issues.redhat.com//browse/RHTAPSRE-168)

### DIFF
--- a/components/authentication/base/rhtap-admins.yaml
+++ b/components/authentication/base/rhtap-admins.yaml
@@ -1,0 +1,254 @@
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: rhtap-admins
+rules:
+  - apiGroups:
+      - appstudio.redhat.com
+    resources:
+      - applications
+      - buildpipelineselectors
+      - componentdetectionqueries
+      - components
+      - enterprisecontractpolicies
+      - environments
+      - integrationtestscenarios
+      - promotionruns
+      - releaseplanadmissions
+      - releaseplans
+      - releases
+      - releasestrategies
+      - snapshotenvironmentbindings
+      - snapshots
+      - spiaccesschecks
+      - spiaccesstokenbindings
+      - spifilecontentrequests
+      - remotesecrets
+    verbs:
+      - '*'
+  - apiGroups:
+      - managed-gitops.redhat.com
+    resources:
+      - gitopsdeploymentmanagedenvironments
+      - gitopsdeployments
+      - gitopsdeploymentsyncruns
+      - operations
+    verbs:
+      - '*'
+  - apiGroups:
+      - jvmbuildservice.io
+    resources:
+      - artifactbuilds
+      - dependencybuilds
+      - rebuiltartifacts
+      - systemconfigs
+      - tektonwrappers
+      - jbsconfigs
+    verbs:
+      - '*'
+  - apiGroups:
+      - pipelinesascode.tekton.dev
+    resources:
+      - repositories
+    verbs:
+      - '*'
+  - apiGroups:
+      - ''
+    resources:
+      - clusterpolicybindings
+      - clusterresourcequotas
+      - images
+      - netnamespaces
+      - nodes
+      - persistentvolumes
+      - securitycontextconstraints
+    verbs:
+      - get
+      - list
+      - watch
+  - apiGroups:
+      - apiserver.openshift.io
+    resources:
+      - apirequestcounts
+    verbs:
+      - get
+      - list
+      - watch
+  - apiGroups:
+      - machineconfiguration.openshift.io
+    resources:
+      - machineconfigs
+      - machineconfigpools
+    verbs:
+      - get
+      - list
+      - watch
+  - apiGroups:
+      - machine.openshift.io
+    resources:
+      - machinesets
+      - machines
+    verbs:
+      - get
+      - list
+      - watch
+  - apiGroups:
+      - network.openshift.io
+    resources:
+      - clusternetworks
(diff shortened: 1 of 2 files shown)


components/authentication/base/rhtap-admins.yaml
```diff
@@ -0,0 +1,215 @@
...
+    name: bamachrn
+  - kind: User
+    apiGroup: rbac.authorization.k8s.io
+    name: gbenhaim
+  - kind: User
+    apiGroup: rbac.authorization.k8s.io
+    name: hugares
+  - kind: User
+    apiGroup: rbac.authorization.k8s.io
+    name: manish-jangra
```
hugares, line 209: should we list infra team one by one or we should grant access to "Infra Team" github group instead?

components/authentication/base/rhtap-admins.yaml
```diff
@@ -0,0 +1,215 @@
...
+# dedicated-admins on stg:
+#   hugares, sbose78, amfred, MatousJobanek, gbenhaim, gabemontero, xcoulon, eisraeli
+# dedicated-admins on prod:
+#   skabashnyuk, sbose78, hugares, amfred, Michkov, MatousJobanek, gabemontero, gbenhaim, xcoulon, scoheb, eisraeli
```
hugares, line 213: here are the current dedicated admins on stg and prod, I guess I should grant rhtap-admins to the prod list as it is a superset of stg.

components/authentication/base/rhtap-admins.yaml
amfred: I think we'll need to factor out individuals' Github user IDs over time into another repo, and a good first step along that path would be to grant access to a group instead of individuals.

components/authentication/base/rhtap-admins.yaml
hugares: Done

components/authentication/base/rhtap-admins.yaml
hugares: Done, as per other comment

components/authentication/base/rhtap-admins.yaml
```diff
@@ -0,0 +1,197 @@
...
+    resources:
+      - repositories
+    verbs:
+      - '*'
+  - apiGroups:
+      - ''
+    resources:
+      - buildconfigs/webhooks
+      - buildlogs
+      - configmaps
```
amfred, line 60: I thought configmaps could have secrets in them?

components/authentication/base/rhtap-admins.yaml
```diff
@@ -0,0 +1,197 @@
...
+  - apiGroups:
+      - route.openshift.io
+    resources:
+      - routes
+    verbs:
+      - '*'
+  - apiGroups:
+      - rbac.authorization.k8s.io
+    resources:
+      - clusterrolebindings
```
amfred, line 163: Hm, do we really need this?  Because if we could edit clusterrolebindings then we could give ourselves access to the tenant secrets again, right?

amfred (changes requested):

components/authentication/base/rhtap-admins.yaml
hugares: good point, removed

components/authentication/base/rhtap-admins.yaml
```diff
@@ -0,0 +1,196 @@
...
+      - deploymentconfigs/scale
+      - egressnetworkpolicies
+      - endpoints
+      - imagestreamimages
+      - imagestreammappings
+      - imagestreams
+      - imagetags
+      - limitranges
+      - persistentvolumeclaims
+      - pods
```
bamachrn, line 71: this would also allow the `pod/logs` right? we would need that.

components/authentication/base/rhtap-admins.yaml
```diff
@@ -0,0 +1,196 @@
...
+      - pods/attach
+      - pods/exec
+      - pods/port-forward
+      - processedtemplates
+      - replicationcontrollers
+      - replicationcontrollers/scale
+      - resourcequotas
+      - rolebindings
```
bamachrn, line 79: may be we should also remove edit access for `rolebindings` as well

components/authentication/base/rhtap-admins.yaml
```diff
@@ -0,0 +1,196 @@
...
+    resources:
+      - deploymentconfigs
+      - deploymentconfigs/scale
+    verbs:
+      - '*'
+  - apiGroups:
+      - authorization.openshift.io
+    resources:
+      - roles
+      - rolebindings
```
bamachrn, line 133: may be we should only allow editing rolebinding with argocd only.

components/authentication/base/rhtap-admins.yaml
```diff
@@ -0,0 +1,196 @@
...
+  - apiGroups:
+      - route.openshift.io
+    resources:
+      - routes
+    verbs:
+      - '*'
+  - apiGroups:
+      - rbac.authorization.k8s.io
+    resources:
+      - rolebindings
```
bamachrn, line 163: same as above editing rolebindings should not be required.

amfred (approved):

PR #2398, opened 2023-09-14
fix(RHTAPWATCH-515): stop forwarding argocd metrics to rhobs

yftacherzog:
argocd_app_info metrics was accidentally remote-written to RHOBS which caused alerts to be fired for it. This change stops that.


avi-biton (approved):

amisstea (approved):

avi-biton (approved):
/approved
/lgtm

PR #2428, opened 2023-09-20
Grant rhtap-admins access to operators.coreos.com resources

hugares:


components/authentication/base/rhtap-admins.yaml
```diff
@@ -197,6 +205,12 @@ rules:
       - imagetags
     verbs:
       - '*'
+  - apiGroups:
+      - operators.coreos.com
```
amfred, line 209: Normally we only change the operators via infra-deployments.  What do we need this access for?

hugares: maybe this is a case of being added temporarily to the cluster/dedicated-admins. I want to edit temporarily a custerserviceversion to increase the memory limit of the pod spawned by the operator. If we prefer to not grant this to the rhtap admins, I will remove this section.

bamachrn: I would suggest we keep it for the rhtap-admins. This will allow to handle emergency cases.

bamachrn (approved):

components/authentication/base/rhtap-admins.yaml
hugares: I would also keep it. In fact, we never want to do a change manually but instead we do them via infra-deployments. All the write access we grant to rhtap-admins is only to do temporary changes in order to debug.

amfred: OK, let's do it.

amfred (approved):

PR #2437, opened 2023-09-22
Set connectionIdleTimeout for the Default Ingress Controller

manish-jangra:
Set connectionIdleTimeout for the Default Ingress Controller to 30 Minutes (Rosa Clusters) to match the Sandbox Cluster Configuration for the Default Ingress Controller.


bamachrn (approved):
/lgtm

PR #2447, opened 2023-09-25
RHTAPREL-435: fix missing overlay

scoheb:
- forgot to remove reference to ci-workspace


scoheb:
forced merge since it affects production and CI does not catch this.

PR #2455, opened 2023-09-26
Fix RBACs for Pipeline Service

Roming22:
* Removed user accounts from the RBACs so everything is controlled via the group.
* Moved the RoleBindings for namespaces into their own folder.
* Fixed the group name in the RoleBindings.


sbose78 (approved):
/lgtm

PR #2473, opened 2023-09-28
Add a cron job to send notification about cluster/dedicated-admins

hugares:
Both cluster-admins and dedicated-admins groups should have no permanent members. We add people to those groups as a temporary measure only. This cron checks if there are members in those groups and if there are, send a notification to team-rhtap-infra slack private channel.

To send the notification, the cron calls a slack workflow URL which accept clusterName, userNames aand groupName as parameters. The workflow then compose a message and send it to the team-rhtap-slack channel, e.g.:

```
Hi, this is a friendly reminder to not forget to remove the following
admin access:
cluster name: https://console-openshift-console.apps.somecluster.com
group: cluster-admins
user names: hares
```

[RHTAPSRE-29](https://issues.redhat.com//browse/RHTAPSRE-29)

### DIFF
--- a/components/authentication/base/admin-checker/cronjob.yaml
+++ b/components/authentication/base/admin-checker/cronjob.yaml
@@ -1,0 +1,51 @@
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: check-cluster-admins
+spec:
+  schedule: "0 7 * * 1" # every Monday at 7:00
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          containers:
+            - name: admin-checker
+              image: registry.redhat.io/openshift4/ose-cli:v4.12
+              env:
+                - name: ADMIN_CHECKER_WORKFLOW_URL
+                  valueFrom:
+                    secretKeyRef:
+                      name: rhtap-infra-secrets
+                      key: admin-checker-workflow-url
+              command:
+                - /bin/sh
+                - -c
+                - |
+                  validate_and_notify_for_group(){
+                    local group="$1"
+                    local users
+                    users=$(oc get group "${group}" -o jsonpath='{.users}' | tr -d '"[]')
+                    if [[ -n "${users}" ]]; then
+                      echo "The list of ${group} is: ${users}"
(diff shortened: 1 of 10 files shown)


components/authentication/base/admin-checker/external-secrets/rhtap-infra-secrets.yaml
```diff
@@ -0,0 +1,21 @@
+apiVersion: external-secrets.io/v1beta1
+kind: ExternalSecret
+metadata:
+  name: rhtap-infra-secrets
+  annotations:
+    argocd.argoproj.io/sync-options: SkipDryRunOnMissingResource=true
+    argocd.argoproj.io/sync-wave: "-1"
```
amfred, line 7: You might be able to use a later sync-wave here, like the default of 0.

components/authentication/base/admin-checker/cronjob.yaml
```diff
@@ -0,0 +1,51 @@
+apiVersion: batch/v1
```
amfred, line 1: Can you also remove this from the development overlay? 
Base will give us staging + production.  We don't want notifications from dev clusters.

amfred (changes requested):

components/authentication/base/admin-checker/cronjob.yaml
hugares: This is already the case, the whole authentication component is removed in the development overlay: https://github.com/redhat-appstudio/infra-deployments/blob/main/argo-cd-apps/overlays/development/delete-applications.yaml#L5

components/authentication/base/admin-checker/external-secrets/rhtap-infra-secrets.yaml
hugares: I did not check them all but all the ExternalSecret I did check use sync-wave -1 so I used -1 as well. I guess that if we use same sync wave as the one creating the conjob, it might complains that secret is not defined. Let me test that.

amfred (approved):

amfred (approved):

PR #2474, opened 2023-09-28
Clean-up Pipeline Service RBACs

Roming22:


sbose78 (approved):
/lgtm

components/pipeline-service/staging/base/team-support-rbac.yaml
```diff
@@ -1,57 +0,0 @@
----
-kind: ClusterRole
```
gabemontero, line 2: @sayan-biswas refresh my memory - does this ClusterRole play a part in how we were getting the permissions we needed to use results to get logs, etc. for users on stage

I know the results specific resources are not here, but did we need the other items from here ?

gabemontero: yeah still need clarification form @sayan-biswas to see if https://issues.redhat.com/browse/PLNSRVCE-1381 was inadvertently getting some of the permissions it needs from the ClusterRole you have moved

I suspect the answer is "no" but I am not 100% certain, even after your helpful clarifications

gabemontero: bump @sayan-biswas ... especially in light of the discrepancy you are seeing in being able to use the CLI in staging but not in prod

PR #2481, opened 2023-09-29
Fix subscription namaspace

bamachrn:
Restoring the namespace change to solve degraded state

### DIFF
--- a/configs/enable-dvo-for-all-cluster/install-dvo.yaml
+++ b/configs/enable-dvo-for-all-cluster/install-dvo.yaml
@@ -20,12 +20,13 @@ metadata:
   name: deployment-validation-operator
   labels:
     name: deployment-validation-operator
+  namespace: deployment-validation-operator
 spec:
   channel: alpha
   installPlanApproval: Automatic
   name: deployment-validation-operator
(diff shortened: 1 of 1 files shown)


configs/enable-dvo-for-all-cluster/install-dvo.yaml
```diff
@@ -20,12 +20,13 @@ metadata:
...
   labels:
     name: deployment-validation-operator
+  namespace: deployment-validation-operator
 spec:
   channel: alpha
   installPlanApproval: Automatic
   name: deployment-validation-operator
   source: community-operators
-  sourceNamespace: deployment-validation-operator
+  sourceNamespace: openshift-marketplace
```
hugares, line 29: if we use community-operators catalog, do we still need to define our own catalog source(line 1 to 15)?

hugares: I tested on my dev cluster without the custom catalog source and the operator installed without any issue.

bamachrn: done

PR #2483, opened 2023-09-29
Fix olm-restart kube-linter issues

hugares:
* set cpu/memory requests/limits
* run container with read only file system
* run container as non root user
* Change image as using latest tag is prohibited and image used only had latest tag


amfred (approved):
Please make sure this feedback gets to the Sandbox team as well.

hugares:
@xcoulon @amfred Can you re-apply lgtm label

amfred (approved):

PR #2503, opened 2023-10-04
chore(build-definitions): update owners

MartinBasti:
Build-definitions repository is mainatained by multiple teams and those teams members should be also approvers for build-pipeline-selector to reduce load on the current approvers.

Approve plugin doesn't currently support filters, so the only way how to achieve this was to make a subdirectory.
https://www.kubernetes.dev/docs/guide/owners/


MartinBasti:
I have to update [build-definitions tekton task](https://github.com/redhat-appstudio/build-definitions/blob/main/.tekton/push.yaml#L99)   as well, so please let me know if you agree with naming of subdirectory

mmorhun:
To me `build-definitions` doesn't reflect the selection of the build pipeline. Also, it confuses with the `build-definitions` repository. What about `build-pipeline-selectors` directory name?

MartinBasti:
updated

components/build-service/base/build-pipeline-selectors/OWNERS
```diff
@@ -0,0 +1,34 @@
...
+- tisutisu
+- chmeliik
+- dirgim
+- hongweiliu17
+- jsztuka
+- jinqi7
+- sonam1412
+- MartinBasti
+- Josh-Everett
+- 14rcole
```
sonam1412, line 17: Should we please also add Dheeraj in the list. Github user account: dheerajodha

components/build-service/base/build-pipeline-selectors/OWNERS
```diff
@@ -0,0 +1,34 @@
...
+
```
sonam1412, line 34: Should we please also add Dheeraj in the list. Github user account: dheerajodha

mmorhun (approved):

components/build-service/base/build-pipeline-selectors/OWNERS
MartinBasti: added

MartinBasti:
Follow up PR: https://github.com/redhat-appstudio/build-definitions/pull/608

Please `/lgtm` this

sonam1412 (approved):
LGTM

PR #2515, opened 2023-10-05
fix(STONEINTG-532): fix grafana panel display err

jencull:
for latency of release creation in seconds panel


jsztuka:
Seems like updating branch is needed.
So the fix is done by updating to newer version?

dirgim (approved):
/lgtm

PR #2524, opened 2023-10-06
Remove rights for rhtap-admin on internalrequests

ralphbean:
The InternalRequest resource permits elevated access outside of the RHTAP member cluster. Only the release pipeline service account should be managing these.


hugares (approved):
/lgtm

PR #2525, opened 2023-10-06
Run admin checker cron every day

hugares:
Run it every day at 11:30 UTC, which translate to 7:30 EDT and 6:30 EST.


amfred (approved):
Thanks. The security team would probably prefer that we check this daily.

amfred:
I removed the lgtm for a minute to allow Hugo to exclude weekends

amfred:
That's OK, new PR it is.  :-)

PR #2526, opened 2023-10-06
Do not run admin checker cron on weekends

hugares:


amfred (approved):
I like quiet weekends.

PR #2539, opened 2023-10-10
managed-gitops: cherrypick #2529 to staging update components/gitops/staging/base/kustomization.yaml

jgwest:
Deploy #2529 to staging only, to verify functionality.


jgwest:
LGTM

PR #2540, opened 2023-10-11
fix(RHTAPWATCH-542): Add status label to the spi metric on app-sre

hmariset:
status label was missing in the LabelKeep list so it couldn't be accessed for the spi metric on the app-sre grafana. This change enables the status label.


amisstea (approved):
/lgtm
/approve

PR #2568, opened 2023-10-17
fix: normalise dora metrics overlays

albarbaro:
Normalise dora metrics overlays into separate for each environment.

### DIFF
--- a/components/dora-metrics/development/kustomization.yaml
+++ b/components/dora-metrics/development/kustomization.yaml
@@ -1,4 +1,10 @@
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
-- ../base
+- https://github.com/redhat-appstudio/dora-metrics/deploy/base?ref=630e08a427f4a6cd3a5fd869d224de34d03874c2
+- ../base/external-secrets
+
+images:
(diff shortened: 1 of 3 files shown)


components/dora-metrics/development/kustomization.yaml
```diff
@@ -2,3 +2,8 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
 - ../base
+
+images:
+  - name: quay.io/redhat-appstudio/dora-metrics
+    newName: quay.io/redhat-appstudio/dora-metrics
+    newTag: 630e08a427f4a6cd3a5fd869d224de34d03874c2
```
hugares, line 9: Having the resource URL in the base and override the image tag in the overlay might not be ideal. Like in this commit, you end up having the image of latest (63e08a) but the config of an old revision (dadd13). I checked other components and most of them do not include the resource URL in the base and put them in each overlay.

albarbaro: Yes, was about to fix that. I was just waiting on the my local tests to finish before commit. Should be fine now. Thanks for the tip!

hugares (approved):
/lgtm

PR #2570, opened 2023-10-17
managed-gitops: Revert #2560 from staging, "update components/gitops/staging/base/kustomization.yaml (#2560)"

jgwest:
#2560 was delivered to staging to verify the functionality, however, there are additional errors that are occuring, so reverting to fix.

This reverts commit 89971ca4b6fa1f8ab872c8b7e5fa02d797d4c89d.


jgwest:
Merging without tests as this is a revert of a commit that went in <5 minutes ago, and that commit gitops service on staging.

PR #2574, opened 2023-10-18
managed-gitops: cherry pick #2557 to staging only, update components/gitops/staging/base/kustomization.yaml

jgwest:
Deploy #2557 to staging, only


jgwest:
```
Summarizing 1 Failure:
  [FAIL] [chains-suite Tekton Chains E2E tests] test creating and signing an image and task verify-enterprise-contract task [It] succeeds when policy is met [ec, HACBS, pipeline]
  /tmp/tmp.Zh2bPTGyD2/tests/build/chains.go:241
```

/retest

PR #2578, opened 2023-10-18
chore(RHTAPREL-647): releaseserviceconfig rbac

scoheb:
- allow rhtap-admin to view/edit releaseserviceconfigs
- allow release team to view/edit releaseserviceconfigs

### DIFF
--- a/components/authentication/base/rhtap-admins.yaml
+++ b/components/authentication/base/rhtap-admins.yaml
@@ -19,6 +19,7 @@ rules:
       - releaseplanadmissions
       - releaseplans
       - releases
+      - releaseserviceconfigs
       - releasestrategies
       - snapshotenvironmentbindings
       - snapshots
--- a/components/release/release-service-configurator-role.yaml
+++ b/components/release/release-service-configurator-role.yaml
@@ -1,0 +1,12 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: release-service-configurator
+rules:
+- apiGroups:
+  - appstudio.redhat.com
+  resources:
+  - releaseserviceconfigs
+  verbs:
+    - '*'
--- a/components/release/release-team.yaml
+++ b/components/release/release-team.yaml
@@ -11,3 +11,17 @@ roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
   name: component-maintainer
+---
+kind: RoleBinding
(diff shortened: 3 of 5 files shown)


components/release/release-team.yaml
```diff
@@ -11,3 +11,17 @@ roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
   name: component-maintainer
+---
+kind: RoleBinding
```
hugares, line 15: ClusterRole is cluster wide as the name state but using it in a RoleBinding grant the access to the resources only in the namespace the RoleBinding belongs to, `release-service` in this case.

If you only want to be able to view/edit `releaseserviceconfigs` in the `release-service`  namesapce, I suggest to use a Role instead of a ClusterRole. If you want to be able to view/edit  `releaseserviceconfigs` in any namespace, then use  a ClusterRoleBinding instead or a RoleBinding.

components/authentication/base/rhtap-admins.yaml
```diff
@@ -19,6 +19,7 @@ rules:
       - releaseplanadmissions
       - releaseplans
       - releases
+      - releaseserviceconfigs
```
hugares, line 22: Could be done in another PR or this one but we should add `releaseserviceconfigs` resource to the `view-appstudio` ClusterRole in "infra-deployments/components/authentication/base/everyone-can-view.yaml".

We grant view access to any RHTAP resources to all RHTAP developers in that manifest.

components/release/release-team.yaml
scoheb: thanks for the clarification and advice. Fixed.

components/authentication/base/rhtap-admins.yaml
scoheb: added.

components/release/release-service-configurator-role.yaml
```diff
@@ -0,0 +1,12 @@
+---
```
hugares, line 1: I missed before but testing the kustomize build now, I see that you also need to include this resources "release-service-configurator-role.yaml" in the kustomization.yaml file, the one in the same folder (component/release/kustomization.yaml)

scoheb: fixed

hugares: looks better, I see the role in the output of kustomize build now

hugares (approved):
/lgtm

PR #2630, opened 2023-10-26
add network policy for proxy-metrics

ranakan19:
add network policy to allow traffic from appstudio-monitoring to toolchain-host-operator


amisstea (approved):
I don't know enough about this change to say if it's necessary but I also don't see a reason to object. LGTM.

xcoulon (approved):

PR #2636, opened 2023-10-27
feat: Adding perf team that maintains perf-team-prometheus-reader NS

jhutar:
This is for Perf&Scale team to be able to read secret created for the service account.


jhutar:
Hello future reviewers :)

It is aligned to following team - is it correct?

https://github.com/orgs/redhat-appstudio/teams/performance

I'm asking because I also noticed I'm in this organization as well:

https://github.com/redhat-appstudio-sre

manish-jangra:
Looks good to me

hugares (approved):
/lgtm

PR #2633, opened 2023-10-26
PLNSRVCE-1470: add the metrics/labels pipeline service wants scraped by app sre grafana

gabemontero:
Since the metrics pipeline service uses for its app sre alerts originate from the openshift-pipelines namespace, they are considered in-cluster prometheus metrics and must be officially declared to be scraped by app sre grafana

@redhat-appstudio/pipeline-service @amisstea  fyi / ptal

PR #2642, opened 2023-10-27
PLNSRVCE-1470: minor fix to add the metrics/labels pipeline service wants scraped by app sre

gabemontero:
Minor fix to https://github.com/redhat-appstudio/infra-deployments/pull/2633

the namespace in the regex refers to the namespace label of our metrics, which is where the pipelinerun is running, not where the metrics-exporter or tekton controller is running

so we are trying to include all tenants plus the `tekton-ci` namespace where all of appstudio's CI run on in the RHTAP prod cluster

@redhat-appstudio/pipeline-service FYI


amisstea (approved):
/lgtm

PR #2649, opened 2023-10-30
sandbox: enable prune during sync

MatousJobanek:
so the obsolete manifests are removed


mfrancisc (approved):
Thanks for fixing it!

PR #2677, opened 2023-11-02
doc(RHTAPWATCH-554): troubleshooting missing metrics and labels

yftacherzog:
Adding a new section to the monitoring stack documentation that should explain why metrics or labels might be missing from AppSRE Grafana and how to troubleshoot that.

### DIFF
--- a/components/monitoring/prometheus/README.md
+++ b/components/monitoring/prometheus/README.md
@@ -56,7 +56,7 @@ Prometheus instance deployed by the RHOBS
 [Observability Operator](https://github.com/rhobs/observability-operator).
 
 #### Platform Prometheus
-Scrapes generic metrics produced by built-in exports such as cAdvisor,
+Mainly scrapes generic metrics produced by built-in exports such as cAdvisor,
 kube-state-metrics, etc.
 
 #### User Workload Monitoring (UWM) Prometheus
@@ -107,7 +107,7 @@ namespaces).
 For those reasons, another instance is needed to federate the other instances, and
 write metrics to RHOBS.
 
-This instance collect selected metrics from Platform Prometheus and UWM Prometheus, and
+This instance collects selected metrics from Platform Prometheus and UWM Prometheus, and
 remote-writes selected labels for those metrics to RHOBS, which in turn, makes the
 metrics accessible to AppSRE Grafana.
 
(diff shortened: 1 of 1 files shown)


avi-biton (approved):
lgtm

components/monitoring/prometheus/README.md
```diff
@@ -107,14 +107,77 @@ namespaces).
...
+default, and the metrics it is configured to scrape for services are generally of
+constant cardinality. I.e. the number of time series stored for a given service does
+not grow based on the service load. For this reason, we allow all metrics it scrapes
+to be federated by the OBO Prometheus instance (and reach RHOBS).
+
+All **metrics** reaching the OBO instance are remote-written to RHOBS, but not all
+**labels** are. This means that it might be that time series visible in AppSRE Grafana
+will not include some of the labels the same time series have on the data plane
+clusters. The OBO instance is configured to remote-write only specific labels, and if
+the presence of a new label is required in AppSRE, then this label should be added to
```
amisstea, line 146:
```suggestion
the presence of a new label is required in alerting rules or AppSRE grafana dashboards, then this label should be added to
```

components/monitoring/prometheus/README.md
```diff
@@ -107,14 +107,77 @@ namespaces).
...
+the configurations.
+
+##### Troubleshooting Missing Metrics and Labels
+
+There are a few steps to follow when specific metrics or labels are required for new
+alerting rules or Grafana dashboards, but are not present in AppSRE Grafana.
```
amisstea, line 152:
```suggestion
alerting rules or Grafana dashboards, but are not present in RHOBS.
```

amisstea (commented):
Looks great - just a few minor corrections.

components/monitoring/prometheus/README.md
yftacherzog: The thing here is that there's no easy way for us to see if the metrics/labels are reaching RHOBS, so we check AppSRE Grafana instead, assuming that the issue is not between RHOBS and AppSRE Grafana.

components/monitoring/prometheus/README.md
yftacherzog: done

components/monitoring/prometheus/README.md
amisstea: Ah, good point. LGTM then :+1:

amisstea (approved):

components/monitoring/prometheus/README.md
yftacherzog: I'll add a note

yftacherzog: @amisstea please take another look

amisstea: Even better! Still LGTM

PR #2707, opened 2023-11-08
STONEBLD-1869 Deploy the OTP service in production

stuartwdouglas:


arewm (approved):
LGTM -- deploying to production and removing from stage?

stuartwdouglas:
No, it is moved to base, so will apply to both stage and prod now.

stuartwdouglas:
Let's get this in, I will test it out today while the rest of the world is asleep.

PR #2716, opened 2023-11-08
Fix UID and dashboard name for Remote Secret dashboard

mshaposhnik:
Fix UID and dashboard name for Remote Secret dashboard, because duplicate UID causes Grafana to error state.


flacatus:
@mshaposhnik seems like is failing on build service tests. Can you create a jira issue?

PR #2717, opened 2023-11-08
Remove auto-update. Tide will take care of updating pr with main in tests

flacatus:
Basically this Workflows was created when we had suggested updated branches. fir testing I purpose to remove that because Tide will merge with main in PRs tests.


johnmcollier (approved):
/lgtm

Approving based off of voting on slack

PR #2727, opened 2023-11-10
fix(STONEINTG-608): create secret for pac used by integration-service

hongweiliu17:
* create pipelines-as-code-secret under integration-service NS to be used by integration test pipeline, I copy the vault path from https://github.com/redhat-appstudio/infra-deployments/tree/main/components/pipeline-service because we decide to use the vault from pipeline-service team
* add base/development/staging/production folder under components/integration
* update argo-cd-apps/base/member/integration/integration.yaml to get value from environement

### DIFF
--- a/components/integration/base/external-secrets/kustomization.yaml
+++ b/components/integration/base/external-secrets/kustomization.yaml
@@ -1,0 +1,5 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- pipelines-as-code-secret.yaml
+namespace: integration-service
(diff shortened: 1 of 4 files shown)


dirgim:
@hongweiliu17 I'm wondering if we should split up these changes to first update the development and staging environments, and after making sure that it works ok in staging, we could follow up with the update to production, wdyt?

hongweiliu17:
> @hongweiliu17 I'm wondering if we should split up these changes to first update the development and staging environments, and after making sure that it works ok in staging we could follow up with the update to production, wdyt?

Good idea. I was just thinking to create pac secret under `integration-service`, then after we have verified the secret works well then update the namespace of the secret in integration-service code. So let me remove the change in production yaml file. Thanks.

PR #2744, opened 2023-11-14
Set Image Controller quay token refresh interval to 15 minutes

mmorhun:
Quay.io doesn't allow preserving old token when new one is created. To avoid long outages, setting ExternalSecret refresh time to 15 minutes.


chmeliik (approved):
LGTM :+1:

mkosiarc (approved):

mkosiarc:
Do we want do it for our other secrets as well?

mmorhun:
@mkosiarc I see no point in frequent synchronization of secrets which we can rotate without outage.

arewm (approved):

PR #2776, opened 2023-11-17
fix(STONEINTG-561): add prefix to release latency metric

jencull:
Add integration_svc prefix to release_latency_seconds metric


MartinBasti:
Closing this PR in favor of #2784 which contains both changes

PR #2785, opened 2023-11-20
DO-NOT-MERGE test OSP nightly deployment

enarha:
This is to test the OSP nightly deployment in the infra-deployments CI.


arewm:
@enarha , do we still need this PR open?

PR #2798, opened 2023-11-22
Add horreum to components

jhutar:
This change add Horreum to the components via an operator. This currently includes dedicated postgresql and keycloak pods as a proof-of-concept. There is also a dedicated grafana pod, but that is vestigial and will be removed in a later update.

_Currently pending needed updates to the Horreum operator to fix an SSL issue, remove grafana, and correct the pod command for keycloak._


argo-cd-apps/base/all-clusters/infra-deployments/horreum/horreum.yaml
```diff
@@ -0,0 +1,41 @@
+apiVersion: argoproj.io/v1alpha1
```
gbenhaim, line 1: This should be placed under argo-cd-apps/base/member/infra-deployments (you can copy/paste the kustomize file from another directory in that path and just change the file names). 

Right now it will deploy also to the host cluster, but we want to deploy only to the members.

gbenhaim (changes requested):

argo-cd-apps/base/all-clusters/infra-deployments/horreum/horreum.yaml
jhutar: Thank you! Should be resolved with https://github.com/redhat-appstudio/infra-deployments/pull/2798/commits/686faa6dc2e242148390bb7a825878e8e43b1b72

arewm:
@jhutar , are you still trying to get this PR in?

jhutar:
Sorry for a late reply. This is not needed now.

PR #2810, opened 2023-11-24
Add admins for build-templates-e2e namespace

mkosiarc:
Currently the namespace does not have any admins. Previously it was managed manually with the role of cluster wide admin. We need people with admin access in the cluster to setup the secrets there - they weren't managed by infra-deployments configuration before but only manually, so we don't the names of the secrets currently in place.

### DIFF
--- a/components/build-templates/base/e2e/role.yaml
+++ b/components/build-templates/base/e2e/role.yaml
@@ -29,3 +29,31 @@ rules:
   - patch
   - update
   - watch
+---
+kind: Role
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: build-admin
+  namespace: build-templates-e2e
+rules:
+  - apiGroups:
+      - ''
+    resources:
+      - secrets
+    verbs:
+      - get
+      - list
+      - watch
+      - update
+      - patch
+      - delete
+
+  - apiGroups:
+      - ''
+    resources:
+      - pods
+    verbs:
+      - get
+      - list
+      - watch
+      - delete
--- a/components/build-templates/base/e2e/rolebinding.yaml
+++ b/components/build-templates/base/e2e/rolebinding.yaml
@@ -50,3 +50,22 @@ roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
   name: appstudio-pipelines-runner
+---
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: build-admins
+  namespace: build-templates-e2e
+subjects:
+  - apiGroup: rbac.authorization.k8s.io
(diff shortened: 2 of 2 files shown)


mmorhun (approved):

components/build-templates/base/rbac/build-admin.yaml
```diff
@@ -0,0 +1,46 @@
...
+---
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: build-admins
+  namespace: build-templates-e2e
+subjects:
+  - apiGroup: rbac.authorization.k8s.io
+    kind: Group
+    name: Build Admins team
```
chmeliik, line 38: Are there any QEs in this group? If not, should we add them as well?

components/build-templates/base/kustomization.yaml
```diff
@@ -1,6 +1,7 @@
 resources:
 - 0-namespace.yaml
 - e2e/
+- rbac
```
chmeliik, line 4: I think you could put the build-admin.yaml file in the e2e/ folder, since it's related only to the build-templates-e2e namespace

components/build-templates/base/rbac/build-admin.yaml
mkosiarc: Nobody from QE is there, they would have to create a new group in our github organization. But the question is also who owns this namespace? From the history, it seems only Michal took care of it.

mkosiarc: But I believe we can resolve this later.

chmeliik: I see, OK let's leave that for some other time

components/build-templates/base/kustomization.yaml
mkosiarc: moved it to the e2e folder and changed the kustomization files

chmeliik: Looking at the new diff, maybe just edit the existing role.yaml and rolebinding.yaml? :sweat_smile: 

```
- rolebinding.yaml
- role.yaml
- rbac
```

mkosiarc: yeah..that makes sense :D

chmeliik (approved):
LGTM, just need to `TODO: remove mkosiarc` now :grin:

mkosiarc:
> LGTM, just need to `TODO: remove mkosiarc` now 😁

Actually, I planned  to deploy like this - do my thing and only then remove it

chmeliik:
> LGTM, just need to `TODO: remove mkosiarc` now 😁

Or later, if we're not sure that the group will work

tkdchen (approved):

PR #2817, opened 2023-11-24
Fix patching of backup ExternalSecret

hugares:
PatchStrategicMerge does not support merge for all type of resources. This is the case for ExternalSecret. Instead of replacing the key in the extract object, it replaces the entire extract object with the one from the patch resulting in removing onversionStrategy and decodingStrategy.

Fix this by using a PatchesJson6902 instead of a PatchStrategicMerge.

Original ExternalSecret resource:
```
spec:
  dataFrom:
  - extract:
      conversionStrategy: Default
      decodingStrategy: None
      key: CHANGE_ME
```

Patched with PatchStrategicMerge:
```
spec:
  dataFrom:
  - extract:
      key: integrations-output/terraform-resources/stonesoups01ue1/stonesoup-infra-stage-backup/backup-stone-stg-m01
```

Patched with PatchesJson6902
```
spec:
  dataFrom:
  - extract:
      conversionStrategy: Default
      decodingStrategy: None
      key: integrations-output/terraform-resources/stonesoups01ue1/stonesoup-infra-stage-backup/backup-stone-stg-m01
```

### DIFF
--- a/components/backup/production/stone-prd-m01/backup-s3-credentials-patch.yaml
+++ b/components/backup/production/stone-prd-m01/backup-s3-credentials-patch.yaml
@@ -1,0 +1,3 @@
+- op: replace
+  path: /spec/dataFrom/0/extract/key
+  value: integrations-output/terraform-resources/stonesoupp01ue1/stonesoup-infra-prod-backup/backup-stone-prd-m01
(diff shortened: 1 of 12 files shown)


manish-jangra (approved):
looks good

PR #2831, opened 2023-11-28
Fix installation order of backup components

hugares:
Make sure backup components can install without having to intervene. ArgoCD tried to create dpa before the oadp was installed causing the whole backup application to be stuck as syncing.

Set sync-wave of the install in the following order: OperatorGroup, Subscription, ExternalSecret and then DataProtectionApplication.


manish-jangra (approved):
Approved

amfred (approved):

PR #2846, opened 2023-11-30
chore(RHTAPWATCH-552): sync production cluster's for segment-bridge

Omeramsc:
The segment-bridge component had sync issues on some of its objects. The cause seem to be the toolchain-host-operator setting, which moved into the segment-bridge directory and so got applied to all of its cluster's settings.

Applying the settings to the RH clusters in which
the toolchain-host-operator component does not exist is the cause for the unsync issue.

This was previously fixed for the staging environment; this patch apply the same logic to the production environment.

### DIFF
--- a/argo-cd-apps/base/member/infra-deployments/segment-bridge/segment-bridge-member.yaml
+++ b/argo-cd-apps/base/member/infra-deployments/segment-bridge/segment-bridge-member.yaml
@@ -19,18 +19,22 @@ spec:
                   values.clusterDir: stone-stg-m01
                 - nameNormalized: stone-stg-rh01
                   values.clusterDir: stone-stg-rh01
+                - nameNormalized: stone-prd-m01
+                  values.clusterDir: stone-stg-m01
+                - nameNormalized: stone-prd-rh01
+                  values.clusterDir: stone-stg-rh01
   template:
     metadata:
       name: segment-bridge-member-{{nameNormalized}}
     spec:
       project: default
       source:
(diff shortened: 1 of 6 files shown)


argo-cd-apps/base/member/infra-deployments/segment-bridge/segment-bridge-member.yaml
```diff
@@ -19,18 +19,22 @@ spec:
...
       project: default
       source:
-        path: '{{values.sourceRoot}}/{{values.environment}}/{{values.clusterDir}}'
+        path: "{{values.sourceRoot}}/{{values.environment}}/{{values.clusterDir}}"
         repoURL: https://github.com/redhat-appstudio/infra-deployments.git
         targetRevision: main
       destination:
         namespace: segment-bridge
-        server: '{{server}}'
+        server: "{{server}}"
```
hugares, line 37: do you want to keep those unrelated changes? i.e. changing the single quote to double. I am ok with them

hugares (approved):

argo-cd-apps/base/member/infra-deployments/segment-bridge/segment-bridge-member.yaml
Omeramsc: usually I'll remove it to shrink unnecessary updates out of this change (linter artifact), but since most of the review (tests) steps have passed successfully, I don't mind leaving it in.

PR #2848, opened 2023-11-30
Add backup schedule for tenants namespaces

hugares:
Is it not supported to use wildcards in the includedNamespaces so exclude the namespaces we do not want instead. That way, we might backup more than needed but that is better than less.

Schedule the backup twicee a day, at 1:30 UTC which is outside of working hours from GMT-8(5:30PM) to 
 GMT+5:30(7:00AM) and at 13:30 UTC which is 12 hours after first backup to minimise work loss in case on an
incident.

Move the ExternalSecret to the -2 sync wave, since it can be done at same time operator is being installed. Move DPA to -1 sync wave as we need this to be created before backup schedule.

### DIFF
--- a/components/backup/base/external-secret.yaml
+++ b/components/backup/base/external-secret.yaml
@@ -3,7 +3,7 @@ kind: ExternalSecret
 metadata:
   name: backup-s3-credentials
   annotations:
-    argocd.argoproj.io/sync-wave: "-1"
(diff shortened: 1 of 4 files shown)


manish-jangra (approved):
looks good to me, we might want to change the ttl for backups.

amfred (approved):

PR #2849, opened 2023-11-30
PLNSRVCE-1537: add new combined/filtered pipeline service alert metrics to app sre in-cluster prometheus list

gabemontero:
We have combined the 2 seperate histograms used each for our execution and overhead alert queries into one metric.  This reduces the number metric queries required for our alerts by half, and also allows us to filter pipelieruns that incorrectly corrupt our overall averages (because they only take a few seconds to complete)

Note - I will remove the original metrics we added for the pipeline-service after the app sre grafana panels have been updated in both app sre stage and prod.

@redhat-appstudio/pipeline-service FYI

### DIFF
--- a/components/monitoring/prometheus/base/monitoringstack/monitoringstack.yaml
+++ b/components/monitoring/prometheus/base/monitoringstack/monitoringstack.yaml
@@ -87,6 +87,10 @@ spec:
   endpoints:
   - params:
       'match[]': # scrape only required metrics from in-cluster prometheus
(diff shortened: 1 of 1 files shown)


gabemontero:
fyi if we hit the same mirroring / registry.redhat.io problems in the e2e's that @flacatus identified in another PR of mine earlier today I will force merge

gabemontero:
there is also a quay.io / red hat registries outage as I type and the e2e's are still running

PR #2850, opened 2023-11-30
Upgrade OADP operator to channel 1.3

hugares:
Version 1.3.0 was released so upgrade to that version since we did not start to take backups. Let's upgrade now and test backup with latest version.


manish-jangra (approved):
looks good. approving

PR #2854, opened 2023-12-01
feat(STONEINTG-608) add pipelines-as-code-secret to prod yaml

hongweiliu17:


components/integration/production/kustomization.yaml
```diff
@@ -13,3 +14,11 @@ namespace: integration-service
 
 components:
   - ../k-components/manager-resources-patch
+
+patches:
+  - path: pipelines-as-code-secret-patch.yaml
```
dirgim, line 19: This doesn't seem to match the filename of the patch `pipelines-as-code-secret.yaml` that's also added in this PR.

hongweiliu17: Yes.....Thanks!

PR #2861, opened 2023-12-04
Remove status section from backup schedule

hugares:
This section is updated by the application, we should not have it in the object definition.


bamachrn (approved):
LGTM

PR #2873, opened 2023-12-05
feat(RHTAPWATCH-629): Add a gosmee server

ifireball:
Run a gosmee server on the host cluster to allow it to forward webhooks to internal clusters.

This is only deploying on staging clusters ATM.

### DIFF
--- a/argo-cd-apps/base/host/smee/kustomization.yaml
+++ b/argo-cd-apps/base/host/smee/kustomization.yaml
@@ -1,0 +1,6 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - smee.yaml
+components:
+  - ../../../k-components/inject-infra-deployments-repo-details
--- a/components/smee/base/kustomization.yaml
+++ b/components/smee/base/kustomization.yaml
@@ -1,0 +1,4 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources: []
--- a/components/smee/production/kustomization.yaml
+++ b/components/smee/production/kustomization.yaml
@@ -1,0 +1,4 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources: []
--- a/components/smee/staging/deployment.yaml
+++ b/components/smee/staging/deployment.yaml
@@ -1,0 +1,36 @@
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: gosmee
+  labels:
+    app: gosmee
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: gosmee
+  template:
+    metadata:
+      labels:
+        app: gosmee
+    spec:
+      containers:
+        - image: "ghcr.io/chmouel/gosmee:v0.20.2"
+          imagePullPolicy: Always
+          name: gosmee
+          args: ["server", "--address", "0.0.0.0"]
+          ports:
+            - name: "gosmee-http"
+              containerPort: 3333
+              protocol: TCP
+          securityContext:
+            readOnlyRootFilesystem: true
+            runAsNonRoot: true
+          resources:
+            limits:
+              cpu: 1
+              memory: 32Mi
+            requests:
+              cpu: 1
+              memory: 32Mi
--- a/components/smee/staging/kustomization.yaml
+++ b/components/smee/staging/kustomization.yaml
@@ -1,0 +1,7 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- deployment.yaml
+- route.yaml
+- service.yaml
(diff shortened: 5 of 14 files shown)


components/smee/staging/kustomization.yaml
```diff
@@ -0,0 +1,7 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
```
hugares, line 3: Maybe we should create all those resources in their own namespace, i.e not the default one? If you agree, then put namespace in this file and it will be applied to all resources.
```
namespace: someNamespace
```

components/smee/production/kustomization.yaml
```diff
@@ -0,0 +1,4 @@
+---
```
hugares, line 1: (optional) I assume you are provisioning for production eventually but currently this file is not used, you could remove the production folder.

components/smee/base/kustomization.yaml
```diff
@@ -0,0 +1,4 @@
+---
```
hugares, line 1: (optional) I assume you are provisioning for production and staging using the same resources eventually but currently this file is not used, you could remove the base folder. Another option would be to move all the resources from staging folder to the base folder and include base in staging kustomize file.

components/smee/staging/kustomization.yaml
hugares: humm, that is a good point. Let me try this and confirm is this is the case.

components/smee/staging/deployment.yaml
```diff
@@ -0,0 +1,26 @@
+---
+apiVersion: apps/v1
```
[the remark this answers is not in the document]
ifireball: I think at leas the 2nd line in that list is irrellevant for OpenShift that uses random UIDs by default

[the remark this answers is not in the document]
hugares: You are correct, this is a no-op change when running on OCP but  it does not harm to set "runAsNonRoot" to true, it makes kube-linter happy.

components/smee/staging/kustomization.yaml
hugares: I just tested this and you are correct, having the namespace at the application level is enough. You can ignore this comment and sorry for the noise :)

argo-cd-apps/base/host/smee/kustomization.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
```
hugares, line 1: While testing this change in my dev environment for the namespace question, I realized that you are missing to include this folder/kustomization.yaml in `argo-cd-apps/base/host/kustomization.yaml`, without this, the application is not created for the host

ifireball: I only want this in stage for now, I'll more it to base once I test it for a while.

I think using delete-application is overkill at this early phase.

components/smee/staging/deployment.yaml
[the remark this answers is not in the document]
ifireball: I think I resolved all issues now

argo-cd-apps/base/host/smee/kustomization.yaml
hugares: ok, but if you do not add it in `argo-cd-apps/base/host/kustomization.yaml`, the application will not even be in staging

hugares: If I build the staging overlay, the smee application is not there:
```
kustomize build argo-cd-apps/overlays/staging/ | grep smee
```

components/smee/staging/deployment.yaml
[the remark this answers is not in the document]
hugares: thanks!

argo-cd-apps/base/host/smee/kustomization.yaml
gbenhaim: Just a note, without the delete application patch it will get deployed to all environments, not just staging.

gbenhaim (approved):

argo-cd-apps/base/host/smee/kustomization.yaml
hugares: I just tested and it is deployed to dev from staging component folder, it will also be deployed to prod from staging folder. I unresolved the thread so you see it. If you want to proceed as is, let me know and I will resolve the thread and put lgtm label.

ifireball: @hugares Are you seeing the resources get deployed or just empty applications? I don't want the resources to be deployed in production, and I don't particularly care about dev ATM. 

@gbenhaim I thought only having stuff in `staging` directory under the component would be sufficient, otherwise what is the point of having those directories and all the complex Application resource that lead to them?

ifireball: Ok, I think I've found what you meant - pade sure to apply the overlay patch in production and the deletion patch in dev.

PR #2877, opened 2023-12-05
Promote integration-service from staging to production

14rcole:


components/integration/production/kustomization.yaml
```diff
@@ -1,14 +1,13 @@
-apiVersion: kustomize.config.k8s.io/v1beta1
-kind: Kustomization
+apiVersion: kustomize.config.k8s.io/v1beta1 kind: Kustomization
```
dirgim, line 1: This doesn't seem like a valid change,, the `kind: Kustomization` should be in a new line.

components/integration/production/kustomization.yaml
```diff
@@ -1,14 +1,13 @@
...
 resources:
 - ../base
 - ../base/external-secrets
-- https://github.com/redhat-appstudio/integration-service/config/default?ref=86d6da5140bf9f3e6394a6bf860e4ae503f83bf9
+- https://github.com/redhat-appstudio/integration-service/config/default?ref=dd5a5e7abc4e575de00d5b568b95a8f74a13a50d
```
dirgim, line 5: This sha doesn't match what's [currently in staging](https://github.com/redhat-appstudio/infra-deployments/blob/5832d72250048f080b549c0c643f121321ff751b/components/integration/staging/kustomization.yaml#L6).

14rcole: Looks like I grabbed an older sha.  Fixed

dirgim (approved):
/lgtm
/approve

PR #2881, opened 2023-12-06
RHTAPWATCH-619: Create RBAC configuration for a logging-admin

eisraeli:


components/monitoring/logging/rbac/kustomization.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
```
hugares, line 1: You need to include this new rbac folder/kustomization in `monitoring/logging/base/kustomization.yaml` if you want this in all overlays or in `monitoring/logging/staging/kustomization.yaml`  and `monitoring/logging/production/kustomization.yaml`  for staging and prod only

hugares (changes requested):

components/monitoring/logging/rbac/kustomization.yaml
eisraeli: @hugares I moved it under base as I want to apply it for both staging and production.

amisstea (approved):

PR #2882, opened 2023-12-06
Remove PaC config for ec-cli repo

simonbaird:
This is causing problems as we onboard EC to RHTAP. RHTAP is saying "Git repository is already handled by Pipelines as Code" and refusing to create the custom build pipeline PR.

IIUC the existing PaC pipelines in this repo were experimental and should not be mission critical. The tekton bundles being created by those are also pushed by the GitHub actions workflow.

Ref: [EC-289](https://issues.redhat.com/browse/EC-289)


simonbaird:
:pray:

PR #2883, opened 2023-12-06
Add ec CLI downloads to OpenShift Console

zregvart:
With this the ec CLI referenced from the EC Tekton Task bundle, i.e. the `verify_ec_task_bundle` key of the `ec-defaults` ConfigMap, will be used to download all `ec` binaries from all images in the multi-arch ec CLI image, feed to a emptyDir volume and served via Apache HTTPD running on the cluster. The URL those will be propagated to the `ConsoleCLIDownload` resource, which in turn will make those downloads available on the OpenShift Console.

Ref: https://issues.redhat.com/browse/EC-278

### DIFF
--- a/components/enterprise-contract/download-service.yaml
+++ b/components/enterprise-contract/download-service.yaml
@@ -1,0 +1,275 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: ec-cli-download-role
+  namespace: enterprise-contract-service
+rules:
+  - apiGroups:
+      - route.openshift.io
+    resourceNames:
+      - ec-cli-download-route
+    resources:
+      - routes
+    verbs:
+      - get
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: ec-cli-download-role-binding
+  namespace: enterprise-contract-service
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: ec-cli-download-role
+subjects:
+  - kind: ServiceAccount
+    name: default
+    namespace: enterprise-contract-service
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: ec-cli-download-cluster-role
+  namespace: enterprise-contract-service
+rules:
+  - apiGroups:
+      - console.openshift.io
+    resources:
+      - consoleclidownloads
+    verbs:
+      - create
+      - patch
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: ec-cli-download-cluster-role-binding
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: ec-cli-download-cluster-role
+subjects:
+  - kind: ServiceAccount
+    name: default
+    namespace: enterprise-contract-service
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  annotations:
+    argocd.argoproj.io/sync-wave: 2
+    ignore-check.kube-linter.io/latest-tag: The openshift/tools:latest ImageStream only has the latest tag, and that is the version we want here
+    ignore-check.kube-linter.io/no-read-only-root-fs: The httpd init scripts modify its configuration
+  labels:
+    app.kubernetes.io/name: ec-cli-download
+    app.kubernetes.io/part-of: enterprise-contract
+  name: ec-cli-download-deployment
+  namespace: enterprise-contract-service
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: ec-cli-download
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/name: ec-cli-download
+    spec:
+      containers:
+        - image: image-registry.openshift-image-registry.svc:5000/openshift/httpd:2.4-ubi9
+          name: http
+          ports:
+            - containerPort: 8443
+              name: https
+              protocol: TCP
+          resources:
+            limits:
+              cpu: 100m
+              memory: 100Mi
+            requests:
+              cpu: 100m
+              memory: 10Mi
+          securityContext:
+            #readOnlyRootFilesystem: true
+            runAsNonRoot: true
+          volumeMounts:
+            - mountPath: /var/www/html
+              name: wwwdata
+              readOnly: true
+            - mountPath: /opt/app-root/httpd-ssl/private
+              name: tls-key
+              readOnly: true
+            - mountPath: /opt/app-root/httpd-ssl/certs
+              name: tls-certificate
+              readOnly: true
+      initContainers:
+        - args:
+            - |
+              #!/usr/bin/bash
+
+              set -o errexit
+              set -o nounset
+              set -o pipefail
+
+              function handle_error {
+                sleep infinity
+              }
+              trap handle_error ERR
+
+              imgref=$({
+                dir="$(mktemp --directory --tmpdir=.)"
+                cd "${dir}"
+                oc image extract "$EC_TASK_BUNDLE_REF"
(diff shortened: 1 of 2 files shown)


components/enterprise-contract/download-service.yaml
```diff
@@ -0,0 +1,234 @@
...
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: ec-cli-download-cluster-role
+subjects:
+  - kind: ServiceAccount
+    name: default
+    namespace: enterprise-contract-service
+---
+apiVersion: apps/v1
+kind: Deployment
```
hugares, line 59: This deployment is is adding 13 kube-linter issues. This should be detected at the PR level but we did not enable it yet, we need to fix all existing issue before we can enable it. Can you address those please:
```
kustomize build components/enterprise-contract/ | kube-linter lint -
KubeLinter v0.6.4-0-g4f0b1e01fd

<standard input>: (object: enterprise-contract-service/ec-cli-download-deployment apps/v1, Kind=Deployment) The container "download" is using an invalid container image, "image-registry.openshift-image-registry.svc:5000/openshift/tools:latest". Please use images that are not blocked by the `BlockList` criteria : [".*:(latest)$" "^[^:]*$" "(.*/[^:]+)$"] (check: latest-tag, remediation: Use a container image with a specific tag other than latest.)

<standard input>: (object: enterprise-contract-service/ec-cli-download-deployment apps/v1, Kind=Deployment) container "download" does not have a read-only root file system (check: no-read-only-root-fs, remediation: Set readOnlyRootFilesystem to true in the container securityContext.)

<standard input>: (object: enterprise-contract-service/ec-cli-download-deployment apps/v1, Kind=Deployment) container "http" does not have a read-only root file system (check: no-read-only-root-fs, remediation: Set readOnlyRootFilesystem to true in the container securityContext.)

<standard input>: (object: enterprise-contract-service/ec-cli-download-deployment apps/v1, Kind=Deployment) container "download" is not set to runAsNonRoot (check: run-as-non-root, remediation: Set runAsUser to a non-zero number and runAsNonRoot to true in your pod or container securityContext. Refer to https://kubernetes.io/docs/tasks/configure-pod-container/security-context/ for details.)

<standard input>: (object: enterprise-contract-service/ec-cli-download-deployment apps/v1, Kind=Deployment) container "http" is not set to runAsNonRoot (check: run-as-non-root, remediation: Set runAsUser to a non-zero number and runAsNonRoot to true in your pod or container securityContext. Refer to https://kubernetes.io/docs/tasks/configure-pod-container/security-context/ for details.)

<standard input>: (object: enterprise-contract-service/ec-cli-download-deployment apps/v1, Kind=Deployment) container "download" has cpu request 0 (check: unset-cpu-requirements, remediation: Set CPU requests and limits for your container based on its requirements. Refer to https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#requests-and-limits for details.)

<standard input>: (object: enterprise-contract-service/ec-cli-download-deployment apps/v1, Kind=Deployment) container "download" has cpu limit 0 (check: unset-cpu-requirements, remediation: Set CPU requests and limits for your container based on its requirements. Refer to https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#requests-and-limits for details.)

<standard input>: (object: enterprise-contract-service/ec-cli-download-deployment apps/v1, Kind=Deployment) container "http" has cpu request 0 (check: unset-cpu-requirements, remediation: Set CPU requests and limits for your container based on its requirements. Refer to https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#requests-and-limits for details.)

<standard input>: (object: enterprise-contract-service/ec-cli-download-deployment apps/v1, Kind=Deployment) container "http" has cpu limit 0 (check: unset-cpu-requirements, remediation: Set CPU requests and limits for your container based on its requirements. Refer to https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#requests-and-limits for details.)

<standard input>: (object: enterprise-contract-service/ec-cli-download-deployment apps/v1, Kind=Deployment) container "download" has memory request 0 (check: unset-memory-requirements, remediation: Set memory requests and limits for your container based on its requirements. Refer to https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#requests-and-limits for details.)

<standard input>: (object: enterprise-contract-service/ec-cli-download-deployment apps/v1, Kind=Deployment) container "download" has memory limit 0 (check: unset-memory-requirements, remediation: Set memory requests and limits for your container based on its requirements. Refer to https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#requests-and-limits for details.)

<standard input>: (object: enterprise-contract-service/ec-cli-download-deployment apps/v1, Kind=Deployment) container "http" has memory request 0 (check: unset-memory-requirements, remediation: Set memory requests and limits for your container based on its requirements. Refer to https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#requests-and-limits for details.)

<standard input>: (object: enterprise-contract-service/ec-cli-download-deployment apps/v1, Kind=Deployment) container "http" has memory limit 0 (check: unset-memory-requirements, remediation: Set memory requests and limits for your container based on its requirements. Refer to https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#requests-and-limits for details.)

Error: found 13 lint errors

```

components/enterprise-contract/download-service.yaml
```diff
@@ -0,0 +1,234 @@
...
+                html { font-family: "RedHatText", "Overpass", overpass, helvetica, arial, sans-serif; line-height: 1.5rem; }
+                ul { list-style: none; padding-left: 0; }
+                </style>
+                <body>
+                <ul>'
+                echo "${manifests}" | jq -r ".[] | \"<li><a href=\\\"\(.config.os)_\(.config.architecture)/ec_\(.config.os)_\(.config.architecture).tar.gz\\\">Download ec CLI for \(.config.os)/\(.config.architecture)</a></li>\""
+                echo '</ul>
+                </body>
+                </html>'
+              } > index.html
```
simonbaird, line 168: It feels a bit untidy to have all the echo ... | oc apply and and html created in here. Would it be nicer to make a little custom image for the httpd server? Just an idea, not a blocker.

Ps, does this resemble the way that TAS does it for their cli downloads?

zregvart: My main goal here was not to produce a new image at all. From what I understand RHTAS builds [an image](https://github.com/securesign/sigstore-ocp/blob/914bdbc5f29c40dcc8fc73989dd3561663324758/images/Dockerfile-clientserver) with the CLIs baked in.

components/enterprise-contract/download-service.yaml
zregvart: Thanks, I needed to add two ignores, so with that the lint issues should be resolved.

hugares: Thanks

components/enterprise-contract/download-service.yaml
```diff
@@ -0,0 +1,275 @@
...
+  - apiGroups:
+      - console.openshift.io
+    resources:
+      - consoleclidownloads
+    verbs:
+      - create
+      - patch
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
```
lcarva, line 46: Is a ClusterRoleBinding needed because ConsoleCLIDownloads resources are cluster resources?

components/enterprise-contract/download-service.yaml
lcarva: It looks like they provide CLI binaries for Windows, Mac, and Linux for x86_64(amd64). I don't think this approach is going to work for the Windows and Mac CLIs. :disappointed: 

Linux ppc64le and s390x are really low in their priority list. They may never provide those.

I think we should follow suit.

components/enterprise-contract/download-service.yaml
zregvart: Yes

components/enterprise-contract/download-service.yaml
simonbaird: FWIW I think it's okay if that comes in a future PR.

lcarva (approved):
/lgtm

PR #2905, opened 2023-12-11
feat: Add environment configmaps to integration service

14rcole:
Add configmaps to the integration service so that the service can tell what environment it's running in.  This will be used to selectively enable pprof profiling.


MartinBasti:
How is this configmap mounted to service? Does kustomize do any magic to mount it? Or do you plan to access that via API directly?

14rcole:
Closing this for the time being. Will open a new PR if adding profiling becomes a priority.

PR #2907, opened 2023-12-11
Reorganize backup component to allow backing up host clusters

hugares:
Split base folder into subfolder, all-clusters, host and members and move resources into the proper subfolder.


bamachrn (approved):
/lgtm

PR #2908, opened 2023-12-11
fix(STONEINTG-694): tekton-ci - clair-in-ci-db token

MartinBasti:
Migrating clair-in-ci-db github token into separate key withint integration-service away from build-service namespace.


dirgim (approved):
/lgtm
/approve

PR #2917, opened 2023-12-13
STONEBLD-1999: use new registry.redhat.io secret for e2e-tests

tkdchen:
[STONEBLD-1999](https://issues.redhat.com//browse/STONEBLD-1999)

- New service account rhtap-e2e-tests-pull-secret is created in customer
  portal Registry Service Accounts management page. This service account
  contains the token required for authentication by registry.redhat.io.
- Secret registry-redhat-io-pull-secret is created under vault path
  production/build/build-templates-e2e. Its content is copied from the
  service account manually.
- The token is syned to production cluster via an ExternalSecret
  resource.
- ServiceAccount is updated to link the token.
- Since e2e-tests runs on production cluster only, all these changes
  apply to production only. The ServiceAccount for staging is updated to
  not link the secret.

### DIFF
--- a/components/build-templates/production/kustomization.yaml
+++ b/components/build-templates/production/kustomization.yaml
@@ -3,4 +3,5 @@ kind: Kustomization
 resources:
   - ../base
   - e2e-quay-push-secret.yaml
+  - e2e-registry-redhat-io-pull-secret.yaml
   - e2e-serviceaccount.yaml
--- a/components/build-templates/production/e2e-registry-redhat-io-pull-secret.yaml
+++ b/components/build-templates/production/e2e-registry-redhat-io-pull-secret.yaml
@@ -1,0 +1,25 @@
+apiVersion: external-secrets.io/v1beta1
+kind: ExternalSecret
+metadata:
+  name: registry-redhat-io-pull-secret
+  namespace: build-templates-e2e
+  annotations:
+    argocd.argoproj.io/sync-options: SkipDryRunOnMissingResource=true
+    argocd.argoproj.io/sync-wave: "-1"
+spec:
+  dataFrom:
+    - extract:
+        key: production/build/build-templates-e2e/registry-redhat-io-pull-secret
+  refreshInterval: 15m
+  secretStoreRef:
+    kind: ClusterSecretStore
+    name: appsre-stonesoup-vault
+  target:
+    creationPolicy: Owner
+    deletionPolicy: Delete
+    name: registry-redhat-io-pull-secret
+    template:
+      engineVersion: v2
+      type: kubernetes.io/dockerconfigjson
+      data:
+        .dockerconfigjson: "{{ .config }}"
--- a/components/build-templates/production/e2e-serviceaccount.yaml
+++ b/components/build-templates/production/e2e-serviceaccount.yaml
@@ -5,8 +5,7 @@ metadata:
   namespace: build-templates-e2e
 secrets:
   - name: quay-push-secret
-  - name: 6340056-stonesoup-build-definitions-e2e-pull-secret
+  - name: registry-redhat-io-pull-secret
 imagePullSecrets:
   - name: quay-push-secret
-  # TODO: manage this secret properly via an ExternalSecret
-  - name: 6340056-stonesoup-build-definitions-e2e-pull-secret
+  - name: registry-redhat-io-pull-secret
--- a/components/build-templates/staging/e2e-serviceaccount.yaml
+++ b/components/build-templates/staging/e2e-serviceaccount.yaml
@@ -7,4 +7,3 @@ secrets:
   - name: quay-push-secret
 imagePullSecrets:
   - name: quay-push-secret
-  # TODO: get a serviceaccount for registry.redhat.io pull access


components/build-templates/base/e2e/serviceaccount.yaml
```diff
@@ -2,9 +2,9 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: appstudio-pipeline
-  namespace: build-templates-e2e
```
rhtap-build-sa, line 5: I think this is needed, otherwise it will be created in the build-templates namspace

Also there was a reason why we did not keep the service account in base, but duplicated it for production and stage. The reason is that it the SA file will then be deployed to dev env as well. But it will fail to find the secrets, causing problems.

mkosiarc, line 5: I think this is needed, otherwise it will be created in the build-templates namspace

Also there was a reason why we did not keep the service account in base, but duplicated it for production and stage. The reason is that it the SA file will then be deployed to dev env as well. But it will fail to find the secrets, causing problems.

tkdchen: > I think this is needed, otherwise it will be created in the build-templates namspace

namespace is specified in the `kustomization.yaml`

mkosiarc: I see that, but in the second part of the comment I mentioned that the SA should not be put into base, as it will try to deploy it to the dev env as well. In stage/prod kustomization files, the namespace is not mentioned. It loads the ../base kustomization file first, not the ../base/e2e so the namespace is ambiguous and does not work (or at least it didn't previously - output of kustomize build also does not show the correct namespace in that case)

mkosiarc: https://github.com/redhat-appstudio/infra-deployments/commit/b7ac9be6d926e1bc35ccd8a912411acc3016c99f

tkdchen: Updated.

components/build-templates/production/kustomization.yaml
```diff
@@ -4,3 +4,4 @@ resources:
   - ../base
   - e2e-quay-push-secret.yaml
   - e2e-registryredhatio-pull-secret.yaml
+  - serviceaccount.yaml
```
chmeliik, line 7: nipick: these could be called e2e-serviceaccount.yaml, to make it clear that they're not for the build-templates namespace but for the build-templates-e2e namespace

chmeliik (approved):
LGTM, just a small nitpick

components/build-templates/production/kustomization.yaml
chmeliik: (a bigger refactoring is in order, but IIRC Max already filed a story to get rid of the build-templates namespace so that's covered)

components/build-templates/production/e2e-registryredhatio-pull-secret.yaml
```diff
@@ -0,0 +1,25 @@
+apiVersion: external-secrets.io/v1beta1
+kind: ExternalSecret
+metadata:
+  name: registryredhatio-pull-secret
```
tisutisu, line 4: will it be good to name it in a more user friendly way like `registry-redhat-io-pull-secret`

components/build-templates/staging/e2e-registryredhatio-pull-secret.yaml
```diff
@@ -0,0 +1,25 @@
+apiVersion: external-secrets.io/v1beta1
```
tisutisu, line 1: we are running e2e-tests in production cluster only, does it make sense to duplicate in staging as well? Are you keeping it for any future use case?

mmorhun: Even if we keep it, we need to have a comment (here and in Vault if possible) that the credentials are not used now.

mmorhun (approved):

components/build-templates/staging/e2e-registryredhatio-pull-secret.yaml
mkosiarc: yeah, I would remove it as part of the refactoring story.

components/build-templates/production/kustomization.yaml
mkosiarc: I agree, I see no point in renaming the SA now

components/build-templates/base/e2e/serviceaccount.yaml
mkosiarc: After your change the commit message is not up to date. This part 
```
All above are done for both production and staging. As a result, the
production and staging serviceaccount YAML files are same, then move it
to base/e2e/ to remove the dupliate.
```
Is no longer valid.

And a little nitpick, as in this case it does not matter because the commits in the PR will be squashed before they are merged. You made a change in the 1 commit but reversed it in the second commit. For future, simple amend or rebase would be better for reviewers.

chmeliik: > You made a change in the 1 commit but reversed it in the second commit. For future, simple amend or rebase would be better for reviewers.

:100:

components/build-templates/staging/e2e-registryredhatio-pull-secret.yaml
tkdchen: @tisutisu you had commented this thought to the rotation document and we talked there already. The story mentions staging. It's up to you. If it is really unnecessary, I'll remove the update to staging.

components/build-templates/production/e2e-registryredhatio-pull-secret.yaml
tkdchen: updated

components/build-templates/production/kustomization.yaml
tkdchen: Renamed the filename.

tisutisu:
LGTM

chmeliik (approved):
LGTM :+1: 

The commits just need a cleanup

mmorhun (approved):

tkdchen:
Squashed all the commits. PR description is synced with the resulting commit. No change to the existing changes.

mkosiarc (approved):

PR #2918, opened 2023-12-13
Deploy backup component on prd member clusters

hugares:
This change will install oadp and configure it to backup appstudio resources in tenant's namespaces.

### DIFF
--- a/components/backup/production/stone-prd-m01/dpa-kmskeyid-patch.yaml
+++ b/components/backup/production/stone-prd-m01/dpa-kmskeyid-patch.yaml
@@ -1,0 +1,3 @@
+- op: replace
+  path: /spec/backupLocations/0/velero/config/kmsKeyId
+  value: 42cb559c-b70f-49fa-ab41-c4bc5d3049be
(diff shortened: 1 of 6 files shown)


manish-jangra (approved):
looks good to me

components/backup/production/stone-prd-m01/dpa-kmskeyid-patch.yaml
```diff
@@ -0,0 +1,3 @@
+- op: replace
+  path: /spec/backupLocations/0/velero/config/kmsKeyId
+  value: 42cb559c-b70f-49fa-ab41-c4bc5d3049be
```
bamachrn, line 3: is this going to change frequently? I do we need to rotate it. may be we pull it from vault?

bamachrn (commented):
rest looks good to me

components/backup/production/stone-prd-m01/dpa-kmskeyid-patch.yaml
hugares: We had a discussion about that in the original change where we created the backup for the 1 cluster, at least the part to put this in vault. The outcome was that since this is only the id of the key and not the key itself, we did not need to put this in vault.

I have no clue if this will change frequently. We could discuss this with the team and if we decide to move this to vault, I am ok to do ti but I would prefer to merge this change first and do another PR to move all key id to vault together.

amfred (approved):

PR #2920, opened 2023-12-13
fix(STONEINTG-694): fix broken kustomize for tekton-ci

MartinBasti:
Clair-in-ci-db secrets is not creeated due the error in kustomize


hugares (approved):
/lgtm

PR #2923, opened 2023-12-13
Apply prd overlay patch to both segment-bridge applications

hugares:
Without this, the productions applications were deployed from the staging folder.


Omeramsc (approved):
LGTM

PR #2954, opened 2023-12-19
Use PaC secret from pipeline service vault

mkosiarc:
both services share the same secret, it should be loaded from the pipeline service vault now.

https://issues.redhat.com/browse/STONEBLD-1893


mmorhun (approved):
LGTM

PR #2959, opened 2023-12-19
fix: update release role

davidmogar:
ReleaseStrategies are deprecated so it needs to be removed from the Role. In addition, we need to delete InternalRequests. The Role contains now that verb for this resource.


scoheb:
reference the associated JIRA issue in the PR description?

scoheb (approved):

johnbieren (approved):

PR #2960, opened 2023-12-19
Deploy backup component to stone-stg-m01

hugares:
This is the last cluster to deploy backup component to so now that they all have it, include oadp in all-clusters kustomize file.


manish-jangra (approved):
looks good

PR #2988, opened 2024-01-02
fix: image-contoller and build-service overlay changes for refactoring dev/stage and prod 

tisutisu:
The changes are part of story: https://issues.redhat.com/browse/STONEBLD-1924


components/build-service/development/kustomization.yaml
```diff
@@ -2,6 +2,17 @@ apiVersion: kustomize.config.k8s.io/v1beta1
...
+
+namespace: build-service
+
+images:
+- name: quay.io/redhat-appstudio/build-service
+  newName: quay.io/redhat-appstudio/build-service
+  newTag: 45e8fe7c3b036d1df3eee11c1d83e0633e451c07
+
+commonAnnotations:
+  argocd.argoproj.io/sync-options: SkipDryRunOnMissingResource=true
```
chmeliik, line 15: Does `commonAnnotations` need to be here as well, or is it enough that it's already in `base/kustomization.yaml`?

components/build-service/development/kustomization.yaml
```diff
@@ -2,6 +2,17 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
 - ../base
+- https://github.com/redhat-appstudio/build-service/config/default?ref=45e8fe7c3b036d1df3eee11c1d83e0633e451c07
+
+namespace: build-service
```
chmeliik, line 7: Same question as `commonAnnotations`

components/build-service/staging/base/kustomization.yaml
```diff
@@ -3,3 +3,14 @@ kind: Kustomization
 resources:
```
chmeliik, line 3: I noticed that staging has two different kustomization.yaml files:
* components/build-service/staging/base/kustomization.yaml
* components/build-service/staging/stone-stage-p01/kustomization.yaml

^ should we update the second one as well?

components/image-controller/development/kustomization.yaml
```diff
@@ -2,8 +2,17 @@ apiVersion: kustomize.config.k8s.io/v1beta1
...
+
+images:
+- name: quay.io/redhat-appstudio/image-controller
+  newName: quay.io/redhat-appstudio/image-controller
+  newTag: 8208ff372b880773b911062dc828f050e45c5312
+
+namespace: image-controller
 
 patches:
+  - path: ./manager_resources_patch.yaml
```
chmeliik, line 15: Does the patch need to be copied to each environment or could it be only in `base`?

components/build-service/staging/base/kustomization.yaml
chmeliik: Hmm, actually - wouldn't it be possible to keep the resources in `base` so that development and staging would take them from `base`, and override them only in production?

components/build-service/development/kustomization.yaml
tisutisu: we need to add it here as well, since resource `https://github.com/redhat-appstudio/build-service/config/default?ref=45e8fe7c3b036d1df3eee11c1d83e0633e451c07` is being moved to individual overlays

components/image-controller/development/kustomization.yaml
tisutisu: since the patch is being applied to one of the resource coming from `https://github.com/redhat-appstudio/image-controller/config/default?ref=8208ff372b880773b911062dc828f050e45c5312` resource and we moved it to the individual overlays. 
When trying to apply the patch referring from base directory was getting error like `Error: trouble configuring builtin PatchTransformer with config: security; file 'manager_resources_patch.yaml' is not in or below`, 
I think there is a way to bypass this error, by using parameter 
`--load-restrictor string   if set to 'LoadRestrictionsNone', local kustomizations may load files from outside their root. This does, however, break the relocatability of the kustomization. (default "LoadRestrictionsRootOnly")`
when running `kustomize build`, but since it breaks `the relocatability of the kustomization` , I think it would be good to keep the patch inside each overlays

components/build-service/staging/base/kustomization.yaml
tisutisu: > ^ should we update the second one as well?
`components/build-service/staging/stone-stage-p01/kustomization.yaml` is referring the `- ../base` in the resources, so both is covered.
> Hmm, actually - wouldn't it be possible to keep the resources in base so that development and staging would take them from base, and override them only in production?
Could you please explain a bit more on this point? Not able to get what is change that you are suggesting.

components/image-controller/development/kustomization.yaml
chmeliik: Ack, makes sense

components/build-service/staging/base/kustomization.yaml
chmeliik: I was thinking that most things could stay in `components/build-service/base/kustomization.yaml`.
In `prod/kustomization.yaml`, we would only override the ref for
```
- https://github.com/redhat-appstudio/build-service/config/default?ref=45e8fe7c3b036d1df3eee11c1d83e0633e451c07

images:
- name: quay.io/redhat-appstudio/build-service
  newName: quay.io/redhat-appstudio/build-service
  newTag: 45e8fe7c3b036d1df3eee11c1d83e0633e451c07
```
But it looks like that doesn't work, so nevermind :sweat_smile:

.tekton/tasks/promote-component.yaml
```diff
@@ -0,0 +1,276 @@
...
+        - name: shared-dir
+          mountPath: /shared
+      workingDir: /shared
+      env:
+        - name: GH_REPO
+          value: "$(params.GH_REPO)"
+        # - name: ORIGIN_REPO
+        #   value: $(params.ORIGIN_REPO)
+      script: |
+        # REPO_NAME=$(echo $ORIGIN_REPO | cut -f5 -d/)
```
chmeliik, line 66: The commented-out parts can be removed

.tekton/tasks/promote-component.yaml
```diff
@@ -0,0 +1,276 @@
...
+    - name: GIT_IMAGE
+      description: Image reference containing the git command
+      default: registry.redhat.io/openshift-pipelines/pipelines-git-init-rhel8:v1.8.2-8@sha256:a538c423e7a11aae6ae582a411fdb090936458075f99af4ce5add038bb6983e8
+    # per https://kubernetes.io/docs/concepts/containers/images/#imagepullpolicy-defaulting
+    # the cluster will set imagePullPolicy to IfNotPresent
+     # also per direction from Ralph Bean, we want to use image digest based tags to use a cue to automation like dependabot or renovatebot to periodially submit pull requests that update the digest as new images are released.
+    - name: SCRIPT_IMAGE
+      description: Image reference for SCRIPT execution
+      # this image is built using https://github.com/redhat-appstudio/build-tasks-dockerfiles/blob/main/update-infra-deployments-task-scripts-image/Dockerfile
+      default: quay.io/redhat-appstudio/update-infra-deployments-task-script-image@sha256:2748f1a4f1af4e35214745aed4e56a9d06f6bdbd30572e7ade13729e67f23cc9
```
chmeliik, line 32: Maybe we can use the `update-infra-deployments-task-script-image` as the GIT_IMAGE as well? It looks like it contains `git`. That way we wouldn't have to depend on the pipelines-git-init image

chmeliik (approved):
LGTM, just minor nitpicks :+1:

.tekton/tasks/promote-component.yaml
tisutisu: Done.

.tekton/tasks/promote-component.yaml
tisutisu: Good point. Merged two parameters `GIT_IMAGE` and `SCRIPT_IMAGE` into one parameter `IMAGE` now and using `update-infra-deployments-task-script-image` for both the steps.

components/build-service/staging/base/kustomization.yaml
```diff
@@ -3,3 +3,14 @@ kind: Kustomization
...
+
+namespace: build-service
+
+images:
+- name: quay.io/redhat-appstudio/build-service
+  newName: quay.io/redhat-appstudio/build-service
+  newTag: 45e8fe7c3b036d1df3eee11c1d83e0633e451c07
+
+commonAnnotations:
+  argocd.argoproj.io/sync-options: SkipDryRunOnMissingResource=true
```
mmorhun, line 16: missing new line

components/image-controller/staging/kustomization.yaml
```diff
@@ -3,3 +3,14 @@ kind: Kustomization
...
+
+images:
+- name: quay.io/redhat-appstudio/image-controller
+  newName: quay.io/redhat-appstudio/image-controller
+  newTag: 8208ff372b880773b911062dc828f050e45c5312
+
+namespace: image-controller
+
+patches:
+  - path: ./manager_resources_patch.yaml
```
mmorhun, line 16: missing new line

.tekton/build-service-prod-overlay-update.yaml
```diff
@@ -0,0 +1,30 @@
+apiVersion: tekton.dev/v1beta1
+kind: PipelineRun
+metadata:
+  name: build-service-prod-overlay-update
+  namespace: rhtap-promotion-staging
+  annotations:
+     pipelinesascode.tekton.dev/on-cel-expression: |
+      event == "push" && target_branch == "main" && "components/build-service/staging/base/kustomization.yaml".pathChanged()
```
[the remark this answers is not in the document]
tisutisu: It is base under the staging directory`staging/base`

components/build-service/staging/base/kustomization.yaml
tisutisu: the pr got merged now.. will address this in a future PR

components/image-controller/staging/kustomization.yaml
tisutisu: the pr got merged now.. will address this in a future PR

PR #2991, opened 2024-01-02
Switch IBM Cloud for multi-platform to new account

arewm:
The new ssh secrets have alreay been added to the vault.

PR #2994, opened 2024-01-03
Revert "Switch IBM Cloud for multi-platform to new account (#2991)"

arewm:
This reverts commit 9b26ed394de3066be586ce41ada1028812764a97.

The PR #2991 was merged which results in an error with the OTP service. This needs to be backed out until the issue can be resolved.

Errors encountered were of the flavor:
```
STEP-BUILD

mkdir -p ~/.ssh
if [ -e "/ssh/error" ]; then
  #no server could be provisioned
  cat /ssh/error
  exit 1
elif [ -e "/ssh/otp" ]; then
 curl --cacert /ssh/otp-ca -XPOST -d @/ssh/otp $(cat /ssh/otp-server) >~/.ssh/id_rsa
 echo "" >> ~/.ssh/id_rsa
else
  cp /ssh/id_rsa ~/.ssh
fi
failed to get SSH secret, system may not be configured correctly
```


mmorhun (approved):

stuartwdouglas:
It is likely a missing label on the relevant secrets.

PR #3003, opened 2024-01-04
Fail kube-linter check if issues are found

hugares:
Now that all issues have been fixed, fail the check if new kube-linter issues are found.

[RHTAPSRE-139](https://issues.redhat.com//browse/RHTAPSRE-139)


psturc (approved):
/lgtm

PR #3005, opened 2024-01-04
Revert pipeline-service

Roming22:
There seems to be a critical bug in Tekton Results when check_owner is set to false.

rh-pre-commit.version: 2.0.3


sayan-biswas (approved):

Roming22:
Forcing the merge because of the impact of the issue.

PR #3013, opened 2024-01-05
Fix build-service production application path

hugares:
In f505a550e2eab004dd0c846533501a55d9c73579, the build-service application clusterDir was set to base to allow overwriting it for private clusters, i.e. stone-stage-p01.

The problem is that the application path was change which affects both production and staging but in the component folder, only staging files were moved to nested base folder.

Fix issue by moving production files to nested base folder.


gbenhaim:
thanks @hugares

gbenhaim (approved):

PR #3027, opened 2024-01-10
RHTAP-2116: Webhook target for the internal cluster

gbenhaim:
The internal cluster fetches webhook events through the go-smee server, thus the Webhook url target should be for the go-smee server.


mmorhun (approved):

hugares:
re-applying /lgtm, was removed because of rebase

PR #3035, opened 2024-01-10
feat: Allow integration team to delete snapshots

14rcole:
Users are increasingly running into issues with their snapshot quota and the integration team is currently best suited too recognize these issues as they occur.  Giving the integration team permission to delete snapshots across all namespaces will decrease customer complaints and prevent them from having to recognize and resolve the issue themselves. Once Snapshot GC has been implemented this permmission can be revoked.

### DIFF
--- a/components/integration/base/delete-snapshots.yaml
+++ b/components/integration/base/delete-snapshots.yaml
@@ -1,0 +1,25 @@
+kind: ClusterRole
+apiVersion: authorization.openshift.io/v1
+metadata:
+  name: delete-snapshots
+rules:
+  - apiGroups:
+    - "appstudio.redhat.com"
+    resources:
+      - "snapshots"
+    verbs:
+      - "delete"
+      - "deletecollection"
+---
+kind: ClusterRoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: delete-snapshots
+subjects:
+  - kind: Group
+    apiGroup: rbac.authorization.k8s.io
+    name: Integration Team
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: delete-snapshots
--- a/components/integration/base/kustomization.yaml
+++ b/components/integration/base/kustomization.yaml
@@ -1,6 +1,7 @@
 resources:
 - allow-argocd-to-manage.yaml
 - argocd-permissions.yaml
+- delete-snapshots.yaml
 - integration.yaml
 
 apiVersion: kustomize.config.k8s.io/v1beta1


components/integration/base/delete-snapshots.yaml
```diff
@@ -0,0 +1,42 @@
...
+#      - "delete"
+#      - "deletecollection"
+---
+kind: RoleBinding
+#apiVersion: authorization.openshift.io/v1
+apiVersion: authorization.openshift.io/v1
+metadata:
+  name: delete-snapshots
+  namespace: integration-service
+subjects:
```
hugares, line 20: instead of listing users, can you use a github group?

I see this group: https://github.com/orgs/redhat-appstudio/teams/integration-team which is missing people but I could help you to add people and we could use that one instead.

components/integration/base/delete-snapshots.yaml
```diff
@@ -0,0 +1,42 @@
+#kind: Role
```
gbenhaim, line 1: Why those line were commented out?

components/integration/base/delete-snapshots.yaml
dheerajodha: Since that link gives a 404, we have a group for our team here: https://github.com/orgs/redhat-appstudio/teams/test-team which we can rename, if others agree

hugares: The other gives a 404 because it is a private group and only member can see it. I add all the integration team member if you want or even better, I can grant maintainer role to someone in integration team who will be able to add other members.

hugares: Or simply use test-team if this team already represents the integration team.

components/integration/base/delete-snapshots.yaml
14rcole: Commented them out for testing, forgot to uncomment.  Fixed now

components/integration/base/delete-snapshots.yaml
14rcole: test-team does have some people who aren't on the integration team, so we'll have to add or remove to/from one of the lists.  However, Kruno is the only maintainer of test-team and he's on PTO for the time being, so in the interest of speed @hugares can you add me as a maintainer to the integration-team group?  I can take care of the rest from there.

hugares: @14rcole Done, you are a maintainer of https://github.com/orgs/redhat-appstudio/teams/integration-team

components/integration/base/delete-snapshots.yaml
```diff
@@ -0,0 +1,26 @@
...
+---
+kind: RoleBinding
+#apiVersion: authorization.openshift.io/v1
+apiVersion: authorization.openshift.io/v1
+metadata:
+  name: delete-snapshots
+  namespace: integration-service
+subjects:
+  - kind: Group
+    name: integration-team
```
hugares, line 22: The name of the team is "Integration team"
```suggestion
  - apiGroup: rbac.authorization.k8s.io
    kind: Group
    name: Integration team
```

components/integration/base/delete-snapshots.yaml
```diff
@@ -0,0 +1,26 @@
...
+  - apiGroups:
+    - "appstudio.redhat.com"
+    resources:
+      - "snapshots"
+    verbs:
+      - "delete"
+      - "deletecollection"
+---
+kind: RoleBinding
+#apiVersion: authorization.openshift.io/v1
```
hugares, line 15: Remove commented line
```suggestion
```

components/integration/base/delete-snapshots.yaml
```diff
@@ -0,0 +1,26 @@
...
+rules:
+  - apiGroups:
+    - "appstudio.redhat.com"
+    resources:
+      - "snapshots"
+    verbs:
+      - "delete"
+      - "deletecollection"
+---
+kind: RoleBinding
```
hugares, line 14: If the intent is to allow integration team to delete snapshots in all namespaces as mentioned in the commit msg/PR description, you need to do a ClusterRoleBinding otherwise a RoleBinding as you did will only allow to delete snapshots in integration-service namespace.

```suggestion
kind: ClusterRoleBinding
```

components/integration/base/delete-snapshots.yaml
```diff
@@ -0,0 +1,26 @@
...
+#apiVersion: authorization.openshift.io/v1
+apiVersion: authorization.openshift.io/v1
+metadata:
+  name: delete-snapshots
+  namespace: integration-service
```
hugares, line 19: Remove this, goes with comment at line 14
```suggestion
```

hugares (changes requested):

components/integration/base/delete-snapshots.yaml
```diff
@@ -0,0 +1,24 @@
...
+      - "delete"
+      - "deletecollection"
+---
+kind: ClusterRoleBinding
+apiVersion: authorization.openshift.io/v1
+metadata:
+  name: delete-snapshots
+subjects:
+  - kind: Group
+    name: Integration Team
```
hugares, line 20: Can you add:
```suggestion
    apiGroup: rbac.authorization.k8s.io
    name: Integration Team
```

hugares (approved):

components/integration/base/delete-snapshots.yaml
```diff
@@ -0,0 +1,25 @@
+kind: ClusterRole
+apiVersion: authorization.openshift.io/v1
```
hugares, line 2: The CI is failing because ArgoCD is syncing in a loop. I deployed this PR to my test cluster and I reproduce the issue. Looks like kustomize does not handle well the "authorization.openshift.io/v1" api as it is adding the namespace to the ClusterRole even though it should not as this is a global resource. This is causing the sync loop...

I missed this at review but you need to use:

```suggestion
rbac.authorization.k8s.io/v1
```

components/integration/base/delete-snapshots.yaml
```diff
@@ -0,0 +1,25 @@
...
+  - apiGroups:
+    - "appstudio.redhat.com"
+    resources:
+      - "snapshots"
+    verbs:
+      - "delete"
+      - "deletecollection"
+---
+kind: ClusterRoleBinding
+apiVersion: authorization.openshift.io/v1
```
hugares, line 15: Same reason as above
```suggestion
rbac.authorization.k8s.io/v1
```

components/integration/base/delete-snapshots.yaml
```diff
@@ -0,0 +1,25 @@
...
+metadata:
+  name: delete-snapshots
+subjects:
+  - kind: Group
+    apiGroup: rbac.authorization.k8s.io
+    name: Integration Team
+roleRef:
+  apiGroup: authorization.openshift.io
```
hugares, line 23:
```suggestion
  apiGroup: rbac.authorization.k8s.io
```

hugares (changes requested):

components/integration/base/delete-snapshots.yaml
hugares: Sorry, my suggestion was not ok, it should have been: 
```suggestion
apiVersion: rbac.authorization.k8s.io/v1
```

14rcole:
> This change is failing to deploy on staging and production, the reason is one comment was not addressed: [#3035 (review)](https://github.com/redhat-appstudio/infra-deployments/pull/3035#pullrequestreview-1822246135)

Looks like there was a push issue.  I've fixed it in #3038

PR #3044, opened 2024-01-11
Remove unused controller image reference

lcarva:
The enterprise-contract-controller repository produces a container image. This is meant to run a Kubernetes controller. However, the controller is never deployed in konflux. (It is actually not deployed anywhere!). The repo is solely used to host the EC CRDs.

This commit removes references to the unused controller.


lcarva:
@zregvart, looking for your input here. I executed `kustomize build components/enterprise-contract` with and without the removed content and it produced the same output.

zregvart:
Missed this, looking at this now

zregvart (approved):

zregvart:
The `images` works as a search-replace, if it finds `name` in one of the image references supported it would replace it with `${newName}:${newTag}`.

So, if we had used the `quay.io/redhat-appstudio/enterprise-contract-controller` image somewhere, say in a `Deployment`, then we would see the change, since we have not used it anywhere the change will be nil, so this is expected.

PR #3052, opened 2024-01-12
Fail kustomize build step on failure

hugares:
The kustomize build step was considered a success even if kustomize build failed. Change implementation to fail the step if at least one of kustomize build commands fail.

Exclude the k-components from the search as some of them are used for sharing patches between overlays. Those components do not build on their own and they get built through the included overlays.

Instead of using a for loop as before, change implementation to use xargs so kustomize build command can run in parallel and print error message so we know which folder did not build. This brings down the execution time from 3 minutes to 1.

[RHTAPSRE-309](https://issues.redhat.com//browse/RHTAPSRE-309)

### DIFF
--- a/.github/workflows/kube-linter.yaml
+++ b/.github/workflows/kube-linter.yaml
@@ -24,7 +24,8 @@ jobs:
 
       - name: Run kustomize build
         run: |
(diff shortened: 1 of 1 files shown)


hugares:
working as expected:
image

PR is ready but I need to fix all existing kustomize build issues before we can merge it.

gbenhaim:
thanks @hugares lgtm

amfred (approved):

amfred:
I agree w/Gal; we can merge it when the builds are passing.

PR #3073, opened 2024-01-16
enterprise contract update

zregvart:


zregvart:
Exactly the same as #3072, hoping that the "RHTAP Staging" check can be triggered this way.

zregvart:
@redhat-appstudio/enterprise-contract can I get a `/ok-to-test` `/lgtm` and `/approve` here, for good measure?

PR #3078, opened 2024-01-17
fix: replace authorization.openshift.io in integration base

14rcole:
Kustomization YAML file that allows the integration team to delete snapshots is not working because the `authorization.openshift.io` labe needs to be replaced with `rbac.authorization.k8s.io`.

### DIFF
--- a/components/integration/base/delete-snapshots.yaml
+++ b/components/integration/base/delete-snapshots.yaml
@@ -1,0 +1,25 @@
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: delete-snapshots
+rules:
+  - apiGroups:
+    - "appstudio.redhat.com"
+    resources:
+      - "snapshots"
+    verbs:
+      - "delete"
+      - "deletecollection"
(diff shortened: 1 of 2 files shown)


dirgim:
The changes look good to me, but needs a rebase first

MartinBasti:
Commit says "replace", what are we replacing?

MartinBasti:
Probably fix for #3035

hugares:
> Probably fix for #3035

yes it is. It looks like this PR is now base on master before 3035 went in, which explains why we see the whose yaml as added instead of just replacing the type. rebasing this PR should do the trick

hugares:
> > Probably fix for #3035
> 
> yes it is. It looks like this PR is now base on master before 3035 went in, which explains why we see the whole yaml as added instead of just replacing the type. rebasing this PR should do the trick

Here is a new PR to fix this, I was not sure I could rebase on someone else's fork: https://github.com/redhat-appstudio/infra-deployments/pull/3114

14rcole:
Closing in favor of #3114

PR #3088, opened 2024-01-18
STONEBLD-2010: Added build-team as maintainer tekton-ci component

tisutisu:
This PR addresses story: https://issues.redhat.com/browse/STONEBLD-2010


mkosiarc:
Now that we have group based permissions I would remove the individual users. 
Pavel is still in our group and he said it should be safe to remove jingi

tisutisu:
> Now that we have group based permissions I would remove the individual users. Pavel is still in our group and he said it should be safe to remove jingi

@mkosiarc  I missed the discussion, updated the PR as you suggested. please do check once again, if it looks good now.

psturc (approved):

mkosiarc (approved):

PR #3094, opened 2024-01-18
fix(integration): define resources per deployments

MartinBasti:
This will allows us to set different resources per deployment.

It also allows to do rolling changes from dev to stage to prod.


jsztuka (approved):
lgtm

PR #3109, opened 2024-01-22
enable ssl/tls for tekton results db connections

avinal:
- Adds required YAML config for enabling db ssl connections
- Uses public certificates for RDS available here: https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.SSL.html#UsingWithRDS.SSL.RegionCertificates
- Enables dbssl for staging environment

Merge after openshift-pipelines/pipeline-service#906 is pulled into infra-deployment

Signed-off-by: Avinal Kumar <avinal@redhat.com>


enarha:
Shouldn't you update the deploy.yaml files? This should only update the staging clusters right?

enarha:
How this PR related to https://github.com/redhat-appstudio/infra-deployments/pull/3436 ? Are they related? Is there an order to be merged?

avinal:
> Shouldn't you update the deploy.yaml files? 

Ah I missed it, I will update.

> This should only update the staging clusters right?

Yes

> How this PR relates to https://github.com/redhat-appstudio/infra-deployments/pull/3436 ? Are they related? Is there an order to be merged?

Ideally they should be merged together so that there is no dependency on each other. Let us discuss this in daily.

gabemontero:
> > Shouldn't you update the deploy.yaml files?
> 
> Ah I missed it, I will update.
> 
> > This should only update the staging clusters right?
> 
> Yes
> 
> > How this PR relates to #3436 ? Are they related? Is there an order to be merged?
> 
> Ideally they should be merged together so that there is no dependency on each other. Let us discuss this in daily.

yeah you need to pull the commit from #3436 into this PR

we also need to discuss how the cert is injected here

components/pipeline-service/base/certificates/tekton-results/tekton-results-db-ssl.yaml
```diff
@@ -0,0 +1,132 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: rds-root-crt
+data:
+# Public certificate for RDS instances are pubically available to download here: 
+# https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.SSL.html#UsingWithRDS.SSL.RegionCertificates
+  tekton-results-db-ca.pem: |-
+    -----BEGIN CERTIFICATE-----
```
gabemontero, line 9: don't certs like this need to be injected into vault and accessed by external secrets?

components/monitoring/grafana/base/dashboards/pipeline-service/kustomization.yaml
```diff
@@ -2,4 +2,4 @@
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
-  - https://github.com/openshift-pipelines/pipeline-service/operator/gitops/argocd/grafana/?ref=542db3fcc168c426d39dbed231a4230b101f8a2a
+  - https://github.com/openshift-pipelines/pipeline-service/operator/gitops/argocd/grafana/?ref=5e38cd44d8455e77def659ea116a684c5cfa4320
```
gabemontero, line 5: matches https://github.com/redhat-appstudio/infra-deployments/pull/3525/files

components/pipeline-service/development/kustomization.yaml
```diff
@@ -8,8 +8,8 @@ commonAnnotations:
   argocd.argoproj.io/sync-options: SkipDryRunOnMissingResource=true
 
 resources:
-  - https://github.com/openshift-pipelines/pipeline-service.git/developer/openshift/gitops/argocd/pipeline-service?ref=542db3fcc168c426d39dbed231a4230b101f8a2a
-  - https://github.com/openshift-pipelines/pipeline-service.git/developer/openshift/gitops/argocd/pipeline-service-storage?ref=542db3fcc168c426d39dbed231a4230b101f8a2a
+  - https://github.com/openshift-pipelines/pipeline-service.git/developer/openshift/gitops/argocd/pipeline-service?ref=5e38cd44d8455e77def659ea116a684c5cfa4320
+  - https://github.com/openshift-pipelines/pipeline-service.git/developer/openshift/gitops/argocd/pipeline-service-storage?ref=5e38cd44d8455e77def659ea116a684c5cfa4320
```
gabemontero, line 12: matches https://github.com/redhat-appstudio/infra-deployments/pull/3525/files

components/pipeline-service/staging/base/kustomization.yaml
```diff
@@ -8,11 +8,12 @@ commonAnnotations:
   argocd.argoproj.io/sync-options: SkipDryRunOnMissingResource=true
 
 resources:
-  - https://github.com/openshift-pipelines/pipeline-service.git/operator/gitops/argocd/pipeline-service?ref=542db3fcc168c426d39dbed231a4230b101f8a2a
+  - https://github.com/openshift-pipelines/pipeline-service.git/operator/gitops/argocd/pipeline-service?ref=5e38cd44d8455e77def659ea116a684c5cfa4320
```
gabemontero, line 11: matches https://github.com/redhat-appstudio/infra-deployments/pull/3525/files

gabemontero (approved):
/lgtm

gabemontero:
was about to ask @avinal if you ran https://github.com/redhat-appstudio/infra-deployments/blob/main/hack/generate-deploy-config.sh ... see you force-pushed something ..

avinal:
Yes I ran the deploy.sh, you can see the changes in the respective deploy.yaml files. The kube lint was failing so I fixed that and pushed.

gabemontero:
@avinal pipeline service is out of sync and degraded in the latest e2e run https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/redhat-appstudio_infra-deployments/3109/pull-ci-redhat-appstudio-infra-deployments-main-appstudio-e2e-tests/1774797750232158208 

the api server deployment is not happy if you look at https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/redhat-appstudio_infra-deployments/3109/pull-ci-redhat-appstudio-infra-deployments-main-appstudio-e2e-tests/1774797750232158208/artifacts/appstudio-e2e-tests/gather-extra/artifacts/deployments.json :
```
            "status": {
                "conditions": [
                    {
                        "lastTransitionTime": "2024-04-01T14:05:51Z",
                        "lastUpdateTime": "2024-04-01T14:05:51Z",
                        "message": "Deployment does not have minimum availability.",
                        "reason": "MinimumReplicasUnavailable",
                        "status": "False",
                        "type": "Available"
                    },
                    {
                        "lastTransitionTime": "2024-04-01T14:15:52Z",
                        "lastUpdateTime": "2024-04-01T14:15:52Z",
                        "message": "ReplicaSet \"tekton-results-api-777f7bf5bd\" has timed out progressing.",
                        "reason": "ProgressDeadlineExceeded",
                        "status": "False",
                        "type": "Progressing"
                    }
```

for whatever reason I cannot find a watcher deployment.

nor do I see watcher pods listed at https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/redhat-appstudio_infra-deployments/3109/pull-ci-redhat-appstudio-infra-deployments-main-appstudio-e2e-tests/1774797750232158208/artifacts/appstudio-e2e-tests/gather-extra/artifacts/pods/

I suspect  you will need to run an `oc kustomize` against the developer overlay and see what is happening wrt the missing watcher deployment

avinal:
@gabemontero I figured out the issue. The certs generation was not added in infra-deployments, I was not aware of the test flow, so I missed it. Xin helped me understand it. Now I have added the cert generation, it is mostly copy-paste from the pipeline-service, but a quick look would be great.

hack/secret-creator/create-plnsvc-secrets.sh
```diff
@@ -5,6 +5,7 @@ main() {
     create_namespace
     create_db_secret
     create_s3_secret
+    create_db_cert_secret_and_configmap
```
gabemontero, line 8: ok so the analogous items here like the db and s3 secrets are pulled from vault and external secrets for stage and prod @avinal , which is why it is OK that this script creates them for the developer overlay

but there is no analogous vault / external secret item for this public cert

so even though things are not right as I type, even if we got them right, this would only work for the developer overlay and not stage/prod

I think the solution @scoheb outlined in the email you sent, where it adds the public cert to the system root CA when building the image, is the way to go, at least short term, since we have a Dockerfile we can update in the results downstream fork.

Now, when we move to the openshift pipelines operator version of results, we'll have to see if that same approach works ...i.e. is there a Dockerfile somewhere in the CPaaS pipeline that would could do the analogous thing for.

From @scoheb email:
- https://github.com/redhat-appstudio/managed-gitops/blob/main/Dockerfile#L49
- https://github.com/redhat-appstudio/segment-bridge/blob/22e23754649e150239d5e86f40dac7f99456d8e6/data/ca-trust/2022-IT-Root-CA.pem#L2

gabemontero: /hold

while this script ^^  has progressed us along, I don't think that can be the final solution.

@enarha FYI

gabemontero:
the watcher and api server now come up with these changes, but the postgresql gitops application stays out of sync because of the cert related changes

I'll attach a screen shot

per discussion in email and slack, and with this solution not working for stage/prod, a pivot to changing the api server image to add the amazon cert to the root SA per pointer from @scoheb appears the be the path we need to take for the time being

/hold

hack/secret-creator/create-plnsvc-secrets.sh
gabemontero: I realized that such a script could be moved to a `Job` like what @Roming22  did for the chains secret over in https://github.com/openshift-pipelines/pipeline-service/blob/main/operator/gitops/argocd/pipeline-service/openshift-pipelines/chains-secrets-config.yaml#L46-L111

Now, I also know @enarha is hoping to rework how we do the chains secret (though I don't remember for sure if that rework included removing this Job)

avinal: I am going through the mails and the threads. To summarize:
1. The RDS public cert can be added to the Dockerfile itself, this is how we are already doing
2. For development, we may want to use a Job, as we are doing in chains. 

If that is all, we can proceed with the discussion.

enarha:
Replaced by https://github.com/redhat-appstudio/infra-deployments/pull/3900 .
Closing.

PR #3115, opened 2024-01-22
Replace `RHTAP Admins` group by `konflux-admins`

hugares:


amfred:
Were you able to test this somehow @hugares ? Are you sure it won't break anything?

gbenhaim:
@amfred see my comment here https://github.com/redhat-appstudio/infra-deployments/pull/3116#issuecomment-1907375167

It's safe to merge it.

PR #3116, opened 2024-01-22
Replace `Infra Team` group by `konflux-infra`

hugares:


amfred:
Were you able to test this one somehow @hugares ?

gbenhaim:
@amfred the old groups are not getting deleted so it's ok to merge it and see if the new group was created as expected and then delete the old one.

gbenhaim:
I verified and the role binding was created as expected.

hugares:
> Were you able to test this one somehow @hugares ?

I did check before opening this PR that the new group was synced to clusters. I did not test the new role binding before but I was confident it would not cause any issue.

PR #3134, opened 2024-01-23
Replace dora-metrics RBAC users by `konflux-qe-admins` group

hugares:


psturc (approved):
/retest-required

PR #3150, opened 2024-01-25
chore(integration-service): update controller resources

MartinBasti:
Currently in production integration-service is using ~600Mi of memory and ~200mi of CPU.

Updating requests accordingly.


jsztuka (approved):
lgtm

kasemAlem (approved):

PR #3151, opened 2024-01-25
Replace `Release Team` group by `konflux-release-team`

hugares:
Use new group name which is aligned with new service name but also with the name of the Rover group. Having both GitHub and Rover group names aligned will allow to have clusters auth with either GitHub or Red Hat SSO without having any difference in the RBACs.


hugares:
@davidmogar I did add everyone to new GH team that are in Rover group except Ariel Chernetsky, he is not in the redhat-appstudio organization yet. If you provide is GH username, I will add him.

PR #3167, opened 2024-01-28
RHTAPSRE-351: Add manifests for the internal production deployment

gbenhaim:


manish-jangra:
The changes look good to me. I saw you have added commits per configuration, which looks good too.

PR #3168, opened 2024-01-28
RHTAPWATCH-666: Deploy smee server for prod

gbenhaim:
Deploy the smee server to the production host cluster. It will be used for our internal production environment.


components/smee/production/ip-allow-list.yaml
```diff
@@ -0,0 +1,15 @@
+---
```
ifireball, line 1: We should copy the comment about where the addresses came from here.

ifireball (changes requested):

components/smee/production/ip-allow-list.yaml
gbenhaim: done

ifireball (approved):
Lgtm

hugares:
@gbenhaim You need to rebase this one to get test passing. You are getting an issue that was fixed yesterday.

PR #3170, opened 2024-01-29
Change refresh interval for Build Service and Image Controller to red…

mmorhun:
…uce downtime after credentials rotation


Roming22 (approved):
/lgtm

PR #3171, opened 2024-01-29
Use GH group for RBACs in integration component

hugares:

RHTAPSRE-287


MartinBasti (approved):

14rcole (approved):

MartinBasti:
I still don't see test results here :|

PR #3173, opened 2024-01-29
Refactor authentication component to allow to add new auth type

hugares:
Authentication component only worked with GitHub. Refactor the component to split what is common to all auth types vs what is specific to GitHub. This will allow to support other auth type such RH IdP(RH SSO).

Move files common to all auth types in base/all and the one specific to GH under base/github.

For now, both prod and staging clusters all use GH but a new auth type will be added in a separate change to support RH IdP.


hugares:
I validated this change by building staging and production overlays before and after this change and the content is 100% the same

PR #3179, opened 2024-01-30
Use RH IdP authentication for stone-stage-p01

hugares:
First, add support for Red Hat IdP in authentication component. Group sync is syncing rover groups through ldap. Any Rover groups starting with "konklux-" will be synced.

Rework the authentication application to use a base folder within staging and prod and override this folder for clusters using RH IdP, i.e. stone-stage-p01 for now.

One difference between GH and RH IdP is in the latter, we do not have a single group with all people allowed to have read only roles on Konflux clusters. Instead, grant the read only roles to all the teams. Follow up changes will be done to do the same way for GH, this will allow us to not have to maintain a global group.


components/authentication/base/rh-idp/group-sync/konflux-rover-groups.yaml
```diff
@@ -0,0 +1,38 @@
...
+          tolerateMemberOutOfScopeErrors: true
+          groupUIDAttribute: dn
+          groupMembershipAttributes:
+            - uniqueMember
+          userNameAttributes:
+            - uid
+          groupsQuery:
+            baseDN: 'ou=adhoc,ou=managedGroups,dc=redhat,dc=com'
+            derefAliases: never
+            filter: (&(objectClass=rhatRoverGroup)(cn=konflux-*))
```
hugares, line 33: @gbenhaim Here we have 2 options, the one implemented is to sync any groups starting with "konflux-". Another one is to explicitly mention all the groups to sync:
```
whitelist:
  - 'cn=konflux-build,ou=adhoc,ou=managedGroups,dc=redhat,dc=com'
  - 'cn=konflux-core,ou=adhoc,ou=managedGroups,dc=redhat,dc=com'
  -  ...
  - 'cn=konflux-spi,ou=adhoc,ou=managedGroups,dc=redhat,dc=com'
```
We need to decide which way we prefer.

gbenhaim: I think we need to mention all groups explicitly since there is nothing blocking other people form creating a groups that start with `konflux-*`

gbenhaim: can be done in a different pr

hugares: If an unwanted konflux-* group is created, e.g. konflux-intruders, it will be sync but no RBACs will use it so the only harm is to have a unused k8s group.

I went with current approach because I find this cool that all you have to do for a new group is to create a rover group, then nothing else apart using it in desired RBACs. That said, I was hesitating between that as whitelisting them explicitly, thus the question.

components/authentication/base/rh-idp/group-sync/konflux-rover-groups.yaml
```diff
@@ -0,0 +1,38 @@
+apiVersion: redhatcop.redhat.io/v1alpha1
+kind: GroupSync
+metadata:
+  name: konflux-rover-groups
+  annotations:
+    argocd.argoproj.io/sync-options: SkipDryRunOnMissingResource=true
+spec:
+  providers:
+    - ldap:
```
hugares, line 9: @gbenhaim Another question, should we prune the groups that are deleted from rover?

components/authentication/base/rh-idp/group-sync/konflux-rover-groups.yaml
gbenhaim: we can leave it with the wildcard.

components/authentication/base/rh-idp/group-sync/konflux-rover-groups.yaml
gbenhaim: yes

components/authentication/base/rh-idp/group-sync/konflux-rover-groups.yaml
hugares: Thanks!

components/authentication/base/rh-idp/group-sync/konflux-rover-groups.yaml
hugares: Done in https://github.com/redhat-appstudio/infra-deployments/pull/3187

PR #3186, opened 2024-01-31
Deploy authentication component to stone-prod-p01

hugares:
Authentication component now supports Red Hat IdP so deploy it to stone-prod-p01 to get rover groups synced and  grant permissions to different groups based on existing RBACs.


components/authentication/production/base/kustomization.yaml
```diff
@@ -2,6 +2,8 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
   - ../../base/github
+components:
+  - ../k-components/rhtap-infra-secrets-patch
```
gbenhaim, line 6: what is the benefit of putting it in a component?

hugares: that patch is used in both `production/base` and `production/stone-prod-p01`, I did not want to duplicate the patch

hugares: hold on, I copied this from somewhere to share the patch but IIUC your comment, I could do the same with a resources. Let me test something

hugares: If I convert the rhtap-infra-secrets-patch from a component to a patch, it does not work. The patch is applied before the resource to be patched is loaded.

PR #3202, opened 2024-02-03
fix(sandbox stage): add missing kustomization entry for toolchainconfig 

mfrancisc:
forgot to add the entry in the kustomization.yaml ( as usual 🤦 )


hugares (approved):
/lgtm

PR #3203, opened 2024-02-04
chore(RHTAPWATCH-765): pipeline-service can manage SpaceRequests

Omeramsc:
In order to create ephemeral namespaces for
testing konflux components, the pipeline-service
service account needs to be able to manage (create, delete, get) the SpaceRequest custom resource.

~Related PR:  https://github.com/codeready-toolchain/member-operator/pull/529~
Blocking: https://issues.redhat.com/browse/RHTAPWATCH-764

### DIFF
--- a/components/build-templates/base/e2e/rolebinding.yaml
+++ b/components/build-templates/base/e2e/rolebinding.yaml
@@ -54,6 +54,19 @@ roleRef:
 kind: RoleBinding
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
+  name: appstudio-pipelines-namespace-creator-rolebinding
+subjects:
+  - kind: ServiceAccount
+    name: appstudio-pipeline
+    namespace: build-templates-e2e
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: appstudio-pipelines-namespace-creator
+---
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
   name: build-admins
   namespace: build-templates-e2e
 subjects:
--- a/components/promotion/base/appstudio-pipelines-runner-rolebinding.yaml
+++ b/components/promotion/base/appstudio-pipelines-runner-rolebinding.yaml
@@ -10,3 +10,16 @@ subjects:
 - kind: ServiceAccount
   name: appstudio-pipeline
   namespace: rhtap-promotion-staging
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: appstudio-pipelines-namespace-creator-rolebinding
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: appstudio-pipeline-namespace-creator
+subjects:
+- kind: ServiceAccount
+  name: appstudio-pipeline
+  namespace: rhtap-promotion-staging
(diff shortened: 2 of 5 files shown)


components/tekton-ci/base/appstudio-pipelines-runner-rolebinding.yaml
```diff
@@ -10,3 +10,16 @@ subjects:
 - kind: ServiceAccount
   name: appstudio-pipeline
   namespace: tekton-ci
+---
+apiVersion: rbac.authorization.k8s.io/v1
```
gbenhaim, line 14: why this change is needed?

components/promotion/base/appstudio-pipelines-runner-rolebinding.yaml
```diff
@@ -10,3 +10,16 @@ subjects:
 - kind: ServiceAccount
   name: appstudio-pipeline
   namespace: rhtap-promotion-staging
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
```
gbenhaim, line 15: why this change is needed?

components/build-templates/base/e2e/rolebinding.yaml
```diff
@@ -53,6 +53,19 @@ roleRef:
 ---
 kind: RoleBinding
 apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: appstudio-pipelines-namespace-creator-rolebinding
+subjects:
```
gbenhaim, line 58: why this change is needed?

Omeramsc: I applied the namespace-creator to wherever there are configurations related to appstudio-pipelines for uniformity. should those three be removed? @gbenhaim

gbenhaim: I suggest to wait with this PR for now. I don't think it's needed but maybe I'm missing something.

Omeramsc: Speaking with the pipeline-services team, it seems like `appstudio-pipelines-runner-rolebinding.yaml` is linking the role (I added in https://github.com/codeready-toolchain/member-operator/pull/529) to the user itself, so I'm not sure if it can work without. I'll wait on the pipelines\build review anyway

Omeramsc:
@gbenhaim I added the `appstudio-namespace-creator_role.yaml` to this PR as it wasn't relevant in the member-operator repository. 
We should probably not wait with this PR any longer as it is currently blocking https://issues.redhat.com/browse/RHTAPWATCH-764 as we don't have the means to test the Task without access to SpaceRequest.

components/tekton-ci/base/appstudio-pipelines-runner-rolebinding.yaml
```diff
@@ -10,3 +10,16 @@ subjects:
...
+kind: RoleBinding
+metadata:
+  name: appstudio-pipelines-namespace-creator-rolebinding
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: appstudio-pipelines-namespace-creator
+subjects:
+- kind: ServiceAccount
+  name: appstudio-pipeline
```
mmorhun, line 24: I don't think that build pipeline should be able to create namespaces.

Omeramsc: I wonder though, even if removed here - if all `appstodio-pipeline` SA are gonna get created with the ability to create namespaces, wouldn't it give it this permission anyway? or am I missing something

Omeramsc:
Not relevant for now, closing.

PR #3208, opened 2024-02-05
feat(STONEINTG-711): create slack webhook secret for build-definitions

hongweiliu17:
* Create slack webhook for build-definitions


hongweiliu17:
We may just add a new key for slack webhook of STONEINTG-711 to existing `slack-webhook-notification-secret`

PR #3214, opened 2024-02-05
Use GH group for RBACs in segment-bridge component

hugares:


amisstea (approved):
/lgtm

PR #3226, opened 2024-02-06
Use GH group for RBACs in sandbox component

hugares:

Replace users by new group named konflux-core, defined here[1] for GH and there[2] for Rover.

[1]https://github.com/orgs/redhat-appstudio/teams/konflux-core
[2]https://rover.redhat.com/groups/group/konflux-core


hugares:
@mfrancisc I granted the RBACs to konflux-core which is every one in the team (@alexeykazakov @mfrancisc @sadlerap @filariow ). If you do not want the entire team to have those access and prefer a subset of the team, I can create another group called konflux-core-admins and use this group in the RBACs instead.

PR #3229, opened 2024-02-07
add space provisioner config for new cluster

mfrancisc:


components/sandbox/toolchain-host-operator/staging/stone-stage-p01/space-provisioner-configs.yaml
```diff
@@ -0,0 +1,26 @@
+apiVersion: toolchain.dev.openshift.com/v1alpha1
+kind: SpaceProvisionerConfig
+metadata:
+  name: member-stone-stg-m01.7ayg.p1.openshiftapps.com
+  namespace: toolchain-host-operator
+spec:
+  toolchainCluster: member-stone-stg-m01.7ayg.p1.openshiftapps.com
```
hugares, line 7: The private cluster (stone-stage-p01) in another konflux instance. We deploy both host and member components on it. We even have UI deployed there. It is a full konflux instance in one cluster.

I reviewed too quickly first time, I did not see that those file were for private cluster. Why are we referring to the public member clusters?

mfrancisc: I'm asking for the name of the `toolchaincluster` config here https://redhat-internal.slack.com/archives/C02CTEB3MMF/p1707308636848989?thread_ts=1707219257.932319&cid=C02CTEB3MMF , this file needs to be updated, but I don't have this name yet.

mfrancisc: updated see [76fb8c2](https://github.com/redhat-appstudio/infra-deployments/pull/3229/commits/76fb8c222b3165668ff59473985f6b83942734c6)

PR #3260, opened 2024-02-12
Rename plnscv ExternalSecret target

Roming22:
This change completes the migration of the Chains signing secret to Vault in staging.
Once validated a similar change will be promoted to production.

rh-pre-commit.version: 2.1.0

### DIFF
--- a/components/pipeline-service/staging/base/kustomization.yaml
+++ b/components/pipeline-service/staging/base/kustomization.yaml
@@ -15,6 +15,12 @@ resources:
   - ../../base/rbac
 
 patches:
+  - path: tekton-chains-signing-secret-name.yaml
+    target:
+      name: tekton-chains-signing-secret
+      group: external-secrets.io
+      version: v1beta1
+      kind: ExternalSecret
   - path: metrics-exporter-trace.yaml
     target:
       kind: Deployment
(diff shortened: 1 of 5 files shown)


components/pipeline-service/staging/base/kustomization.yaml
```diff
@@ -15,6 +15,12 @@ resources:
   - ../../base/rbac
 
 patches:
+  - path: tekton-chains-signing-secret-name.yaml
+    target:
+      name: tekton-chains-signing-secret
```
gabemontero, line 20: shouldn't we explicitly cite the namespace here?

gabemontero (commented):
one minor question / possible needed change

components/pipeline-service/staging/base/kustomization.yaml
Roming22: We don't do it [here](https://github.com/redhat-appstudio/infra-deployments/blob/main/components/pipeline-service/staging/stone-stg-m01/resources/kustomization.yaml).

If you want to make it explicit, I would have no issues with it, but that does not seem necessary at the moment.

gabemontero: ok sure if staging is OK let's keep it consistent

my question was from lack of certainty to be completely honest

feel free to resolve

PR #3263, opened 2024-02-13
RHTAPSRE-351: Fix smee-client component path

gbenhaim:
- There was a missing patch for production
- The path to the component in the ApplicationSet was wrong.

### DIFF
--- a/argo-cd-apps/base/smee-client/smee-client.yaml
+++ b/argo-cd-apps/base/smee-client/smee-client.yaml
@@ -24,7 +24,7 @@ spec:
     spec:
       project: default
       source:
-        path: '{{values.sourceRoot}}'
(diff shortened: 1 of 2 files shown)


argo-cd-apps/base/smee-client/smee-client.yaml
```diff
@@ -24,7 +24,7 @@ spec:
     spec:
       project: default
       source:
-        path: '{{values.sourceRoot}}'
+        path: '{{values.sourceRoot}}/{{values.environment}}/{{values.clusterDir}}'
```
ifireball, line 27: Why do we need those directories in the path string if they don't actually exist in the repo?

gbenhaim: they exist

hugares (approved):

PR #3266, opened 2024-02-13
Grant grafana component maintainer to o11y team

hugares:
Component maintainer is an existing ClusterRole giving permissions extra permissions on installplans, serviceaccounts, secrets, pods and deployments, see exact definition here[1].

Grant components-maintainer role to konflux-o11y for the grafana components, i.e. appstudio-grafana namespace.

Also add the delete deployments permission in component-maintainer as this is needed for grafana but it make sense to add for all components maintainers.

[1]https://github.com/redhat-appstudio/infra-deployments/blob/main/components/authentication/base/all/component-maintainer.yaml

### DIFF
--- a/components/monitoring/grafana/base/rbac/grafana-maintainers.yaml
+++ b/components/monitoring/grafana/base/rbac/grafana-maintainers.yaml
@@ -1,0 +1,12 @@
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: appstudio-grafana-maintainers
+  namespace: appstudio-grafana
+subjects:
+  - kind: Group
+    name: konflux-o11y
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: component-maintainer
--- a/components/authentication/base/all/component-maintainer.yaml
+++ b/components/authentication/base/all/component-maintainer.yaml
@@ -59,3 +59,4 @@ rules:
     verbs:
       - get
       - patch
+      - delete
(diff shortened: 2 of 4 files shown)


components/monitoring/grafana/base/rbac/grafana-maintainers.yaml
```diff
@@ -0,0 +1,12 @@
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: appstudio-grafana-maintainers
+  namespace: appstudio-grafana
+subjects:
+  - kind: Group
+    name: konflux-o11y-admins
```
hugares, line 8: Here I proposed to restrict this to a subset of o11y team. This group currently has @amisstea @eisraeli and @ifireball . I can use the konflux-o11y group also if you want to grant these permissions to the entire team. Let me know what you want to do

avi-biton: Its a very short list of people, and since we will all start supporting this domain I think we all should be granted with permission

hugares: I do not mind changing this, I copied the pattern other groups use but that said, if o11y want to grant this to all members, works for me :)

amisstea: I see little issue with using the non-admin group in this instance since it's scoped to just grafana and not something with more sensitive data or in the critical path of the end user's workflow.

amisstea (approved):
/lgtm

components/monitoring/grafana/base/rbac/grafana-maintainers.yaml
hugares: Done, I used konflux-o11y team

avi-biton (approved):
/lgtm

gbenhaim (approved):

PR #3276, opened 2024-02-14
Rename plnscv ExternalSecret target in prd

Roming22:
This change completes the migration of the Chains signing secret to Vault in production.

Additional info:
* Staging specific config is removed as part of the prod activation.
* `signing-secrets` needs to be manually deleted for the gitops sync to be successful.
* `signing-secrets-vault` needs to be manually cleaned up.

rh-pre-commit.version: 2.1.0


hugares (approved):
/lgtm

PR #3285, opened 2024-02-16
fix(hack): integration service has multiple config refs

MartinBasti:
Original code has been replacing everything into defult, but we have more configs paths than just default


dirgim (approved):
/lgtm
/approve

PR #3288, opened 2024-02-16
Set pipeline signing-secrets to orphan

hugares:
Even thought the secret is created by the ExternalSecret, the
TektonInstallerSet was taking over the secret ownership preventing the
ExternalSecret from managing the secret.

Set secret as orphan as we want the value to be updated if if change in
vault but we do not want to take ownership of the secret as it is
created by TektonInstallerSet.

### DIFF
--- a/components/pipeline-service/base/external-secrets/openshift-pipelines/chains-signing-secrets.yaml
+++ b/components/pipeline-service/base/external-secrets/openshift-pipelines/chains-signing-secrets.yaml
@@ -15,6 +15,5 @@ spec:
     kind: ClusterSecretStore
     name: appsre-stonesoup-vault
   target:
-    creationPolicy: Owner
-    deletionPolicy: Delete
+    creationPolicy: Orphan
     name: signing-secrets
--- a/components/pipeline-service/production/stone-prd-m01/deploy.yaml
+++ b/components/pipeline-service/production/stone-prd-m01/deploy.yaml
@@ -1745,8 +1745,7 @@ spec:
     kind: ClusterSecretStore
     name: appsre-stonesoup-vault
   target:
-    creationPolicy: Owner
-    deletionPolicy: Delete
+    creationPolicy: Orphan
     name: signing-secrets
 ---
 apiVersion: external-secrets.io/v1beta1
--- a/components/pipeline-service/production/stone-prd-rh01/deploy.yaml
+++ b/components/pipeline-service/production/stone-prd-rh01/deploy.yaml
@@ -1745,8 +1745,7 @@ spec:
     kind: ClusterSecretStore
     name: appsre-stonesoup-vault
   target:
-    creationPolicy: Owner
-    deletionPolicy: Delete
+    creationPolicy: Orphan
     name: signing-secrets
 ---
 apiVersion: external-secrets.io/v1beta1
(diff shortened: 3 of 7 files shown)


Roming22 (commented):
If we set it to `immutable` will the ExternalSecret operator be able to delete/recreate the secret when the source value will change?

hugares:
> hmm .... it seemed like in our examinations @hugares that the secret was getting set to imutable anyway when the external secret won the race

That is not the case, immutable is false by default. The only place where immutable is set to true is on prd-m01 and I cannot find an explanation apart that it might have been added by someone.

@Roming22 
> If we set it to immutable will the ExternalSecret operator be able to delete/recreate the secret when the source value will change?

That is a really good question. I assumed that yes while doing this PR... I just tested that and if we set this to immutable, the ExternalSecret operator does not delete/recreate the secret when source change in Vault...

gabemontero:
> > hmm .... it seemed like in our examinations @hugares that the secret was getting set to imutable anyway when the external secret won the race
> 
> That is not the case, immutable is false by default. The only place where immutable is set to true is on prd-m01 and I cannot find an explanation apart that it might have been added by someone.
>
yeah I was under the opposite assumption and that external secrets did it by default but if there is an imutable field your reasoning makes sense
 
> @Roming22
> 
> > If we set it to immutable will the ExternalSecret operator be able to delete/recreate the secret when the source value will change?
> 
> That is a really good question. I assumed that yes while doing this PR... I just tested that and if we set this to immutable, the ExternalSecret operator does not delete/recreate the secret when source change in Vault...

yep which means this is a deal breaker for long term support, and this just gets us out of jail potentially for now, correct?

hugares:
> > @Roming22
> > > If we set it to immutable will the ExternalSecret operator be able to delete/recreate the secret when the source value will change?
> > 
> > 
> > That is a really good question. I assumed that yes while doing this PR... I just tested that and if we set this to immutable, the ExternalSecret operator does not delete/recreate the secret when source change in Vault...
> 
> yep which means this is a deal breaker for long term support, and this just gets us out of jail potentially for now, correct?

Ok then, another option to try instead of using `creationPolicy: owner`, we could change to `creationPolicy: merge`, see https://external-secrets.io/latest/api/spec/#external-secrets.io/v1beta1.ExternalSecretCreationPolicy

Roming22:
@hugares After reading the docs you linked, I'd rather we use `Orphan`. That way the secret stays put even if the ExternalSecret is removed.

gabemontero:
> @hugares After reading the docs you linked, I'd rather we use `Orphan`. That way the secret stays put even if the ExternalSecret is removed.

if our assumptions that someone manually set immutable to true in that one instance and no controllers/operators are, and that both owner/orphan will properly modify the secret if it already exists, then yes @hugares I agree with @Roming22

hugares:
I just tested orphan it does exactly what we want IFAIK. I will update the PR

hugares:
> I just tested orphan it does exactly what we want IFAIK. I will update the PR

Done

gabemontero (approved):
/lgtm

Roming22 (commented):
/lgtm

Roming22 (approved):

Roming22 (approved):
/lgtm

PR #3294, opened 2024-02-19
Add utility to generate deploy.yaml

Roming22:
rh-pre-commit.version: 2.1.0

### DIFF
--- a/hack/generate-deploy-config.sh
+++ b/hack/generate-deploy-config.sh
@@ -1,0 +1,49 @@
+#!/bin/bash
+set -o errexit
+set -o nounset
+set -o pipefail
+
+ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"/..
+
+# Print help message
+function print_help() {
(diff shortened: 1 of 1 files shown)


gabemontero (commented):
IIRC from a recent arch call it appears the other Konflux components are not going to the model put into place for pipeline-service anytime soon ....is that correct?

if so, perhaps rename the script to `generate-pipeline-service-deploy-config.sh

but if other components are going to employ the model recently established for pipeline-service, sure keep the name generic

otherwise, if I parse the bash correctly, I think this achieves what we have been doing manually, so 

lgtm

Roming22:
We don't know what the future holds. It won't do anything on components that do not follow the `deploy.yaml` model.

So I'd rather keep the name generic so that we don't have to rename it in the future.

gabemontero:
> We don't know what the future holds. It won't do anything on components that do not follow the `deploy.yaml` model.
> 
> So I'd rather keep the name generic so that we don't have to rename it in the future.

Ok I'm fine with that.  Thanks for clarifying.

gabemontero (approved):
/lgtm

PR #3310, opened 2024-02-21
Backup Configuration for stone-prod-p01 and stone-stage-p01

manish-jangra:


components/backup/base/privatelink/backup-tenants-schedule.yaml
```diff
@@ -0,0 +1,55 @@
+apiVersion: velero.io/v1
```
hugares, line 1: This one looks very similar to the schedule we use on public clusters, can we reuse the same one?

components/backup/base/privatelink/backup-toolchain-host-schedule.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: velero.io/v1
```
hugares, line 1: This one looks very similar to the schedule we use on public clusters, can we reuse the same one?

components/backup/base/privatelink/backup-toolchain-member-schedule.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: velero.io/v1
```
hugares, line 1: This one looks very similar to the schedule we use on public clusters, can we reuse the same one?

components/backup/base/privatelink/backup-tenants-schedule.yaml
hugares: I checked and they are exactly the same. I have a proposal to remove duplication, let me open another PR with one commit on top of yours for you to see. If you agree with the change, you can apply the changes in this PR.

hugares: here it is: https://github.com/redhat-appstudio/infra-deployments/pull/3315

manish-jangra: thanks @hugares, added what you suggested.

hugares:
@manish-jangra I cannot believe I missed this at review... The backup ArgoCD application is not deployed on downstream clusters. We need another PR to deploy backup application on downstream-production.

manish-jangra:
> @manish-jangra I cannot believe I missed this at review... The backup ArgoCD application is not deployed on downstream clusters. We need another PR to deploy backup application on downstream-production.

Sorry, I missed it too. I see that backup application is added into `delete-applications.yaml` with `$patch: delete`

PR #3320, opened 2024-02-22
feat(RHTAPREL-795): add release cleanup cronjob

theflockers:
This PR adds the remove-expired-release CronJob to remove releases after specific period.


theflockers:
Relates to https://github.com/redhat-appstudio/release-service/pull/344

components/release/base/cronjobs/remove-expired-releases.yaml
```diff
@@ -0,0 +1,75 @@
...
+                  function deleteAndLog() {
+                    release=${1%:*}
+                    namespace=${1#*:}
+                    kubectl delete release $release -n $namespace |while read logLine; do
+                      echo "INFO: namespace=${namespace} log=${logLine}"
+                    done                                                                                                  
+                  }                                                                                                       
+                  export -f deleteAndLog
+                  xargs -a ${EXPIRED_RELEASES_FILE} -i -P ${MAX_PROCS} bash -c 'deleteAndLog "{}"'
+              imagePullPolicy: Always
```
davidmogar, line 53: I don't think we need Always given that you are not pointing to a floating tag.

components/release/base/cronjobs/remove-expired-releases.yaml
```diff
@@ -0,0 +1,75 @@
...
+              command:
+                - /bin/bash
+                - -c
+                - |
+                  set -o pipefail
+                  PATH="/bin:/usr/bin:/usr/local/bin"
+                  NOW=$(date +%s)
+                  kubectl get release --all-namespaces \
+                  --sort-by=.status.expirationTime \
+                  --template '{{range .items}}{{.metadata.name}}{{"\t"}}{{.metadata.namespace}}{{"\t"}}{{.status.expirationTime}}{{"\n"}}{{end}}' > /tmp/kubectl-out
```
mmalina, line 28: Why is `/tmp/kubectl-out` used directly and `EXPIRED_RELEASES_FILE` is defined as an env var? I would probably suggest defining both at the top of the script - below `NOW=`.

theflockers: Sure. Just thought the "eventually" we would like to put this stuff in some configmap. I will move it to the script.

PR #3324, opened 2024-02-22
konflux(stage): move toolchain OLM configuration

mfrancisc:
Move toolchain operators OLM config from sandbox-sre.

Jira: https://issues.redhat.com/browse/ASC-496


components/sandbox/toolchain-host-operator/staging/stone-stage-p01/toolchain-host-operator.yaml
```diff
@@ -0,0 +1,35 @@
+apiVersion: operators.coreos.com/v1alpha1
```
hugares, line 1: this file is the same for stone-stage-p01 and stone-stg-host, to not duplicate it, I suggest to move it in `components/sandbox/toolchain-host-operator/base/` folder

mfrancisc: yes you're right. TBH I'm not sure what `stone-stage-p01 ` contains and if we want it to be in synch with `stone-stg-host`, so I'm taking the opportunity to discuss this here. 

If you think it makes sense and does will be in synch I can move the config up one level.

mfrancisc: BTW I don't have access to the new "internal clusters" ( both stage and prod ) so I cannot check.

mfrancisc: @hugares @gbenhaim can you confirm that is fine to move this file up one level since it should be the same for both stone-stage-p01 and stone-stg-host ?

hugares: we should put anything common to host in `components/sandbox/toolchain-host-operator/base/` and only keep in specific cluster folder things that are specific to a cluster, e.g. a patch to change some info that are cluster specific...

hugares: I tested the change locally of 

> @hugares @gbenhaim can you confirm that is fine to move this file up one level since it should be the same for both stone-stage-p01 and stone-stg-host ?

I tested this locally before suggesting it so yes, I think it is fine to do it.

mfrancisc: going to update the PR. Thanks for checking and confirming!

mfrancisc: I now realized we want to keep stage and prod separate. Should I have a components/sandbox/toolchain-host-operator/staging/base/ instead for this PR

hugares: What do you mean by separate?

1. You want to roll out staging first then production but the files will be the same in the end
2.  You will have completely different files

hugares: Instead of having to create yet another folder for staging which would be temporary, I suggest we keep PR as-is. We duplicate files in all staging cluster folders so only staging will get the changes first and then to roll out to prod, we move the file to base folder, removing duplication at the same time. 
WDYT?

mfrancisc: Yes, let's do that. And for prod I'll move the file up to base folder.

hugares (approved):
/lgtm

PR #3340, opened 2024-02-27
increase Tekton Results pruner gracePeriod for Dev env

ramessesii2:
Following Konflux-bug 348, increasing the time period before Results Pruner kicks in to prune executed PipelineRuns from 10m to 2h.

There have been few instances where the PipelineRuns fail but we can't find the exact cause of failure because PipelineRun is already pruned. Konflux e2e demands to have PipelineRuns stored in the cluster for a longer period in order to proceed with tests, and further debugging by devs if needed.

Signed-off-by: Satyam Bhardwaj <sabhardw@redhat.com>

rh-pre-commit.version: 2.2.0


Roming22 (approved):
/lgtm

PR #3348, opened 2024-02-28
enterprise contract update

zregvart:


zregvart:
Identical to #3347 with different author to circumvent the Red Hat Konflux check

PR #3350, opened 2024-02-28
pipeline-service update manual

enarha:
After switching one of the staging clusters (stone-stg-rh01) to the OSP nightly builds, now we switch the remaining two clusters (stone-stg-m01 and stone-stage-p01).


Roming22 (approved):
/lgtm

PR #3358, opened 2024-02-28
feat(STONE-812): IS mapping to console URL

jsztuka:
Integration Service needs to establish a way to configure a console url to be used for reporting links in gitHub.
This is the way how we would create ConfigMap containing link to specific environment which will be then fetched by pipeline adapter in integration service.

### DIFF
--- a/components/integration/development/kustomization.yaml
+++ b/components/integration/development/kustomization.yaml
@@ -10,6 +10,11 @@ images:
   newName: quay.io/redhat-appstudio/integration-service
   newTag: 55b2bae85419df8cd8b9e633d212826a7b9567ee
 
+configMapGenerator:
+- name: console-url
+  literals:
+    - CONSOLE_URL=""
+
 namespace: integration-service
 
 patches:
--- a/components/integration/production/base/kustomization.yaml
+++ b/components/integration/production/base/kustomization.yaml
@@ -11,6 +11,11 @@ images:
   newName: quay.io/redhat-appstudio/integration-service
   newTag: c6ceab5e54a3b57e1a699d534ad1179ac60029f2
 
+configMapGenerator:
+- name: console-url
+  literals:
+    - CONSOLE_URL="https://console.redhat.com/preview/application-pipeline/ns/{{ .Namespace }}/pipelinerun/{{ .PipelineRunName }}"
+
 namespace: integration-service
 
 patches:
--- a/components/integration/staging/base/kustomization.yaml
+++ b/components/integration/staging/base/kustomization.yaml
@@ -11,6 +11,11 @@ images:
   newName: quay.io/redhat-appstudio/integration-service
   newTag: 55b2bae85419df8cd8b9e633d212826a7b9567ee
 
+configMapGenerator:
+- name: console-url
+  literals:
+    - CONSOLE_URL="https://console.dev.redhat.com/preview/application-pipeline/ns/{{ .Namespace }}/pipelinerun/{{ .PipelineRunName }}"
+
 namespace: integration-service
 
 patches:
--- a/components/integration/development/manager_resources_patch.yaml
+++ b/components/integration/development/manager_resources_patch.yaml
@@ -15,3 +15,10 @@ spec:
           requests:
             cpu: 100m
             memory: 20Mi
+        env:
+        - name: CONSOLE_URL
+          valueFrom:
+              configMapKeyRef:
+                name: console-url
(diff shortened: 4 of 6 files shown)


jsztuka:
Users are capable of setting the custom **console_url** values using console-config-url.yaml file.

components/integration/staging/base/kustomization.yaml
```diff
@@ -11,7 +11,16 @@ images:
...
+configMapGenerator:
+- name: console-url
+  literals:
+    - CONSOLE_URL=""
+
 namespace: integration-service
 
 patches:
   - path: manager_resources_patch.yaml
+  - path: config-console-url.yaml
```
MartinBasti, line 23: I don't understand why are we patching something that's defined in the same dir? Why not to define it directly?

jsztuka: Removed, direct definition is sufficient enough for this case.

components/integration/development/kustomization.yaml
```diff
@@ -10,6 +10,11 @@ images:
   newName: quay.io/redhat-appstudio/integration-service
   newTag: c6ceab5e54a3b57e1a699d534ad1179ac60029f2
 
+configMapGenerator:
+- name: console-url
+  literals:
+    - CONSOLE_URL="devUrlChangeInFuture"
```
dirgim, line 16: I think this might break when getting processed on the [integration-service side](https://github.com/redhat-appstudio/integration-service/pull/609/files#diff-e47e5b0d6143994b257b1b2d54a0525a6470ac5636ffebc2452517a7e367abbdR213).
Can we simply not set this envvar in development instances for starters?

components/integration/production/base/kustomization.yaml
```diff
@@ -11,6 +11,11 @@ images:
   newName: quay.io/redhat-appstudio/integration-service
   newTag: b37833f64c8017c942ba95ca648d3b880b39af62
 
+configMapGenerator:
+- name: console-url
+  literals:
+    - CONSOLE_URL="https://console.redhat.com/preview/application-pipeline/ns/%s/pipelinerun/%s"
```
dirgim, line 17: I just tried this URL out, but it doesn't seem to work for me :/

jsztuka: Here is an example for stage env.
https://console.dev.redhat.com/preview/application-pipeline/ns/mbasti-tenant/pipelinerun/test-python-gitlab-kmgvd-fhcws
It should redirects.

dirgim: Yup, you're right, I was using the prod instance and the redirect took ages, but it worked in the end.

components/integration/staging/base/kustomization.yaml
```diff
@@ -11,6 +11,11 @@ images:
   newName: quay.io/redhat-appstudio/integration-service
   newTag: 55b2bae85419df8cd8b9e633d212826a7b9567ee
 
+configMapGenerator:
+- name: console-url
+  literals:
+    - CONSOLE_URL="https://console.dev.redhat.com/preview/preview/application-pipeline/ns/%s/pipelinerun/%s"
```
dirgim, line 17: Just noticed there are two previews in this link, should it be `preview/preview` -> `preview/`?

jsztuka: Good catch!

components/integration/staging/base/kustomization.yaml
```diff
@@ -11,6 +11,11 @@ images:
   newName: quay.io/redhat-appstudio/integration-service
   newTag: 55b2bae85419df8cd8b9e633d212826a7b9567ee
 
+configMapGenerator:
+- name: console-url
+  literals:
+    - CONSOLE_URL="https://console.dev.redhat.com/preview/application-pipeline/ns/%s/pipelinerun/%s"
```
MartinBasti, line 17: as we discussed, can we keep consistency with PAC and use templating instead of format string?

https://github.com/redhat-appstudio/infra-deployments/blob/6141189cea62ccdd9827dd29f4fa3eddf92b73e6/components/pipeline-service/production/base/update-tekton-config-pac.yaml#L8

jsztuka: Needed to make namespace > .Namespace and pr > .PipelineRunName to work properly.

dirgim (approved):
/lgtm
/approve

PR #3371, opened 2024-03-04
Update sprayproxy prod to 2 replicas

jkhelil:


Roming22 (approved):

Roming22:
I've approved so that the PR can move forward and be merged. But I wonder why we chose to duplicate the configuration in both `staging` and `production` instead of moving the configuration from `staging` to `base`.

PR #3375, opened 2024-03-04
chore(RHTAPWATCH-838): promote snapshot gc to prod

yftacherzog:
Promote snapshots garbage collection to production


yftacherzog:
@dirgim @MartinBasti please take a look

PR #3378, opened 2024-03-04
KFLUXBUGS-863: bump stg results watcher controller tuning

gabemontero:
Similar to what we did for the tekton pipelines controller, we are bumping the knative tuning parameters for controller worker thread count to the recommended, higher scale, non development, settings. Unfortunately, results/watcher bypasses the api clientqps/burst command line args that knative provides, and it is at an old enough knative level that it does provide the environment variable overrides for those settings. So we bypass that tuning at this time.  We are also bumping cpu request/limits and mem limit (again) based on recent obsevations.

We are finally moving forward with the worker thread bump since we have the goroutine/memory lead sorted out.

FYI the recently discovered pruning issue in stage https://issues.redhat.com/browse/PLNSRVCE-1679 is unrelated to the mem leak fix (though we may temporarily patch results downstream to bypass that bug while PaC gets sorted out).

I am not setting these values in the development overlay as I was worried about developer and e2e cluster size varying enough.

Based on what we see in staging, if need be I'll revert this and then start vetting through the developer overlay path.

rh-pre-commit.version: 2.2.0

### DIFF
--- a/components/pipeline-service/staging/base/bump-results-watcher-mem.yaml
+++ b/components/pipeline-service/staging/base/bump-results-watcher-mem.yaml
@@ -1,0 +1,7 @@
+---
+- op: replace
+  path: /spec/template/spec/containers/1/resources/limits/memory
+  value: "4Gi"
+- op: replace
+  path: /spec/template/spec/containers/1/resources/requests/memory
+  value: "4Gi"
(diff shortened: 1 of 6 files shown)


gabemontero:
rollout in stage successful

PR #3387, opened 2024-03-06
RHTAPSRE-404: Update build-service to use auto generated ca bundle

gbenhaim:
Openshift will inject the ca bundle data into the trusted-ca configmap. Mount this configmap into the build-service pod.


components/build-service/base/add-ca-bundle-patch.yaml
```diff
@@ -0,0 +1,21 @@
+---
```
MartinBasti, line 1: Why these files have been added to base? If it's intentional, it missed entries in kustomize

gbenhaim: mistake. I removed them.

components/build-service/components/rh-certs/add-rh-certs-patch.yaml
```diff
@@ -1,30 +1,20 @@
----
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: build-service-controller-manager
+  name: controller-manager
```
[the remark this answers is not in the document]
gbenhaim: mistake, I removed this change.

gbenhaim:
thanks for the review @MartinBasti

MartinBasti (approved):

PR #3391, opened 2024-03-06
KONFLUX-2040: Add robot push token for quay.io/konflux-ci

ralphbean:
The purpose is to support a migration of container images from quay.io/redhat-appstudio to quay.io/konflux-ci.


hugares (approved):
/lgtm

PR #3402, opened 2024-03-07
ArgoCD: Ignore the diff in appstudio-pipeline SA

hugares:
Prior to this change, this resources sync status was toggling between sync and out-of-sync. The reason is ArgoCD is fighting with OCP. ArgoCD is creating the SA with:
```
imagePullSecrets:
  - name: quay-push-secret
```

Then ocp adds `appstudio-pipeline-dockercfg-gl9vz`:
```
imagePullSecrets:
  - name: quay-push-secret
  - name: appstudio-pipeline-dockercfg-gl9vz
```

Then  ArgoCD removes `appstudio-pipeline-dockercfg-gl9vz` and then OCP put put it back and so on.

Configure the build-template application to ignore the difference.


bamachrn (approved):
/lgtm

PR #3414, opened 2024-03-11
Grant konflux-admins to konflux-sre team

hugares:
Create a new GH/rover group named konflux-sre to include SRE people and grant them konflux-admins.

Another option would have been to add them to konflux-admins group but I prefer to not do that. In fact, I would like to delete that group and grant konflux-admins role to teams who should have it, like konflux-sre.

[RHTAPSRE-425](https://issues.redhat.com//browse/RHTAPSRE-425)


eisraeli (approved):
LGMT

eisraeli (approved):
LGTM

PR #3421, opened 2024-03-12
Update setup-nfs-quickcluster.sh to work with RHEL-8

Tojaj:
If restart of nfs.service fails, try to restart nfs-server.service instead.


dirgim:
The change looks reasonable to me, but I'll defer to @scoheb and @johnbieren as the original authors of the script.

scoheb (approved):

kasemAlem (approved):

hugares:
There is a job "Red Hat Konflux" cannot run... Forcing this change in as this script is not tested by the failing checks

PR #3427, opened 2024-03-13
RHTAPSRE-429: Changing destination namespace

manish-jangra:
This PR will change the destination namespace from `argocd-staging` to `konflux-public-argocd-staging` for the ArgoCD staging instance. 

The destination namespace is `konflux-public-argocd-staging` to host the Public ArgoCD Staging Instance on the OSD Cluster.


gbenhaim:
@manish-jangra because the namespace name changes (which is a bit different than the last migration we did, I suggest the following approach):

1. submit a pr to disable auto-sync for staging. you should also disable it for the environemnt overlay (argo-cd-apps/overlays/staging/kustomization.yaml)

2. Create a new overlay for the new namespace - `argo-cd-apps/overlays/konflux-public-staging` its content should be the same as the content of `argo-cd-apps/overlays/staging/` except for the namespace in the kustomization file.

3. After merging 2 (assuming the namespaces in app-interface were already made) we would end up with 2 argocd instances each use a different overlay but deploys the same content to different namespaces (and autosync is disabled)

4. If we see that the new argocd is ok, submit a pr to app-interfaces for removing the old argocd.

5. enable back auto sync.

manish-jangra:
@gbenhaim I'm following the steps you suggested above.

gbenhaim (approved):

PR #3429, opened 2024-03-13
fix: release-expiration cronjob incorrect directory

theflockers:
This PR changes the output temp directory used by the cronjob `remove-expired-releases`

### DIFF
--- a/components/release/base/cronjobs/remove-expired-releases.yaml
+++ b/components/release/base/cronjobs/remove-expired-releases.yaml
@@ -30,7 +30,7 @@ spec:
                   NOW=$(date +%s)
                   kubectl get release --all-namespaces \
                   --sort-by=.status.expirationTime \
-                  --template '{{range .items}}{{.metadata.name}}{{"\t"}}{{.metadata.namespace}}{{"\t"}}{{.status.expirationTime}}{{"\n"}}{{end}}' > /tmp/kubectl-out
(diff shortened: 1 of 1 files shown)


components/release/base/cronjobs/remove-expired-releases.yaml
```diff
@@ -30,7 +30,7 @@ spec:
                   NOW=$(date +%s)
                   kubectl get release --all-namespaces \
                   --sort-by=.status.expirationTime \
-                  --template '{{range .items}}{{.metadata.name}}{{"\t"}}{{.metadata.namespace}}{{"\t"}}{{.status.expirationTime}}{{"\n"}}{{end}}' > /tmp/kubectl-out
+                  --template '{{range .items}}{{.metadata.name}}{{"\t"}}{{.metadata.namespace}}{{"\t"}}{{.status.expirationTime}}{{"\n"}}{{end}}' > /var/tmp/kubectl-out
```
mmalina, line 33: You will also need to change the other place that reads the file. But also, I'd suggest using `mktemp`.
```
OUTPUT=$(mktemp)
.. > $OUTPUT
...
awk ... $OUTPUT
```

mmalina: So you're saying plain `mktemp` won't work here, right? Fine in that case.

mmalina (approved):

components/release/base/cronjobs/remove-expired-releases.yaml
theflockers: exactly. thanks!

happybhati (approved):
/lgtm

PR #3470, opened 2024-03-20
Revert sprayproxy migration

jkhelil:
There is an issue pushing images to quay actually

``` 
Error: POST https://quay.io/v2/redhat-appstudio/pull-request-builds/blobs/uploads/: UNAUTHORIZED: access to the requested resource is not authorized; map[]
main.go:74: error during command execution: POST https://quay.io/v2/redhat-appstudio/pull-request-builds/blobs/uploads/: UNAUTHORIZED: access to the requested resource is not authorized; map[]
```

We suspect the issue coming from https://github.com/redhat-appstudio/infra-deployments/pull/3435
We trying to revert https://github.com/redhat-appstudio/infra-deployments/pull/3435 here https://github.com/redhat-appstudio/infra-deployments/pull/3469

If the revert is merged we should also revert this

### DIFF
--- a/components/sprayproxy/production/kustomization.yaml
+++ b/components/sprayproxy/production/kustomization.yaml
@@ -2,13 +2,13 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
   - ../base
-  - https://github.com/konflux-ci/sprayproxy/config?ref=f1afe07688592565986ded70ca912d80629c04a0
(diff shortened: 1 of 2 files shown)


jkhelil:
@enarha  I hink you are right, We may not need this, so initially i thought we need the new secret to pull images from quay.io/konflux-ci(new repo) but doesnt seem to be the case. I will close this one then

PR #3471, opened 2024-03-20
adjust exporter mem to account for increased prod load

gabemontero:
monitoring exporter for the last few days, I still see GC occurring and memory reductions, but it simply day to day is closer to the current 2Gi memory limit than it used to be

rh-pre-commit.version: 2.2.0

### DIFF
--- a/components/pipeline-service/production/base/bump-exporter-mem.yaml
+++ b/components/pipeline-service/production/base/bump-exporter-mem.yaml
@@ -1,7 +1,7 @@
 ---
 - op: replace
   path: /spec/template/spec/containers/0/resources/limits/memory
-  value: "2Gi"
+  value: "3Gi"
 - op: replace
   path: /spec/template/spec/containers/0/resources/requests/memory
-  value: "2Gi"
+  value: "3Gi"
--- a/components/pipeline-service/production/stone-prd-m01/deploy.yaml
+++ b/components/pipeline-service/production/stone-prd-m01/deploy.yaml
@@ -1254,10 +1254,10 @@ spec:
         resources:
           limits:
             cpu: 500m
-            memory: 2Gi
+            memory: 3Gi
           requests:
             cpu: 250m
-            memory: 2Gi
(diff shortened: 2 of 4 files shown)


gabemontero:
@hugares - as the exporter quit oom'ing / crash back looping during the evening US Eastern after to load on prod-rh01 had died down somewhat, I suspect what was happening was that there were enough live pipelineruns yesterday during US business hours yesterday that when the export first established its watches and controller cache, there was enough coming in that it was hitting its memory limit.

gabemontero:
> @hugares - as the exporter quit oom'ing / crash back looping during the evening US Eastern after to load on prod-rh01 had died down somewhat, I suspect what was happening was that there were enough live pipelineruns yesterday during US business hours yesterday that when the export first established its watches and controller cache, there was enough coming in that it was hitting its memory limit.

The only other theory I could come up with was that pod was pinned to a node where there was not 2Gi available.

Do you think that is at all possible @hugares ?  

I think I remember you saying that eventually the controllers would get a dedicated node pool, but didn't think that was imminent.

hugares:
> > @hugares - as the exporter quit oom'ing / crash back looping during the evening US Eastern after to load on prod-rh01 had died down somewhat, I suspect what was happening was that there were enough live pipelineruns yesterday during US business hours yesterday that when the export first established its watches and controller cache, there was enough coming in that it was hitting its memory limit.

I know there was a huge backlog of pipelineruns to process yesterday, I do not know the internals of that process but the backlog hypothesis make sense.

> 
> The only other theory I could come up with was that pod was pinned to a node where there was not 2Gi available.

That should not happen in theory. requests.memory is guaranteed and used by the scheduler to place pod on a node. 

> Do you think that is at all possible @hugares ?
> 
> I think I remember you saying that eventually the controllers would get a dedicated node pool, but didn't think that was imminent.

We want to do this but the work is not started so you are correct, this is not imminent

hugares (approved):

PR #3472, opened 2024-03-20
Configure ArgoCD to not prune signing-secrets secret

hugares:
We use Vault to store signing-secrets as a way to backup this secret and get it recreated on the cluster it its get deleted. To get it back in the cluster, ExternalSecret is used as usual to extract value from Vault and put in in signing-secrets secret.

Since this secret is created by created/managed by TektonInstallerSets, the ExternalSecret is configured to create the secret as orphan if it does not exists, meaning that it creates it without setting the owner reference so ES and TIS do not fight over the ownership.

Because ES is not setting ownership, the secret created was getting pruned by ArgoCD as soon as ES was creating it.

Fix this issue by adding the annotation
`argocd.argoproj.io/sync-options: Prune=false` to the secret being created by ES.


hugares:
I forgot to regenerate deployments...

PR #3475, opened 2024-03-21
Remove former argo-cd-apps/overlays/staging folder

hugares:
Delete folder used by public argocd before we migrate to konflux-public-staging one.

Adjust staging-downstream to use import konflux-public-staging but make sure to set namespace back to argocd-staging to preserve backward compatibility.

Adjust production-downstream to use import konflux-public-staging as it was already importing staging and not production.

This change is also setting back the auto-sync policy on all staging-downstream and production-downstream applications.


manish-jangra:
> Adjust production-downstream to use import konflux-public-staging as it was already importing staging and not production

I tried looking into it but didn't understand why both of them are using staging instead of production-downstream using production.

manish-jangra (approved):
looks good to me

hugares:
> > Adjust production-downstream to use import konflux-public-staging as it was already importing staging and not production
> 
> I tried looking into it but didn't understand why both of them are using staging instead of production-downstream using production.

I think this is a mistake but I did not wanted to change it to prod as part of this change. Changing it to prod is adding 2 more applications so I will do this in a separate PR and review with @gbenhaim if we need those 2 additional app or we delete them

PR #3477, opened 2024-03-21
Adding Hugo access to manage konflux user

dperaza4dustbit:
Infra team needs access to manage konflux users as it affects konflux capacity in prod


mfrancisc (approved):
/lgtm

PR #3465, opened 2024-03-19
Increase memory of production build controller

hugares:
On rh01, the pod is getting OOMKilled. This change is temporary to increase the memory limit. Once we have the working number, the change will be done in build-service repository.

PR #3483, opened 2024-03-21
Promote Build Service from Staging to Production

mmorhun:


rcerven (approved):

arewm:
@hugares , the changes from the build controller should be included in this so I don't think we would need the override configured in https://github.com/redhat-appstudio/infra-deployments/pull/3465 anymore.

PR #3485, opened 2024-03-21
go back to 1 watcher replica, asses impact on prod-rh01 results RDS hang

gabemontero:
During triage of the prod-rh01 results RDS lock contention query hand issue, noticed for the first time since we enabled multiple watcher replicas that both replicas were processing objects, vs. one acting and the primary and processing object, and the other replica acting as a standby.

We could not find an instance of a pipeline run or task run being processed by both replicas, but there are too many to effectively search for.

Removing this variable of 2 replicas manipulating the same DB entries leading to the lock contention.

@hugares @sayan-biswas FYI

rh-pre-commit.version: 2.2.0


hugares (approved):
/lgtm

PR #3489, opened 2024-03-22
Add cronjobs for pruning repositories:

rcerven:
redhat-appstudio-tekton-catalog/pull-request-builds redhat-appstudio/pull-request-builds

[STONEBLD-2202](https://issues.redhat.com//browse/STONEBLD-2202)
[STONEBLD-2204](https://issues.redhat.com//browse/STONEBLD-2204)


tkdchen (commented):
Could you please add more information in the commit message about this solution?

After going through the changes for the first time, there are a few questions in my mind:

* `requests.exceptions` is imported, why not just using `requests` to make HTTP requests, whereas `urllib` is used instead?

* Why an overlay with specific name `stone-prd-rh-01` is required?

* Is it possible to reuse the existing prune_images.py script under image-controller by making it more configurable?

* Are those tokens newly created or reused from existing one? Are the quay tokens documented at anywhere?

* Why this prune_images.py script can't be shipped with image-controller like the existing one `config/registry_image_pruner/`?

For the secret name, it should be good to mention `-quay-token` for clarity.

rcerven:
> Could you please add more information in the commit message about this solution?
> 
> After going through the changes for the first time, there are a few questions in my mind:
> 
>     * `requests.exceptions` is imported, why not just using `requests` to make HTTP requests, whereas `urllib` is used instead?
I was using prune_images.py as base, I don't think this really matters though

> 
>     * Why an overlay with specific name `stone-prd-rh-01` is required?
this script is for removing specific repositories, so we want it to run just on 1 production cluster

> 
>     * Is it possible to reuse the existing prune_images.py script under image-controller by making it more configurable?
I don't want to mix those 2 scripts, they are doing different kind of pruning

> 
>     * Are those tokens newly created or reused from existing one? Are the quay tokens documented at anywhere?
new
not sure where are documented any current tokens

> 
>     * Why this prune_images.py script can't be shipped with image-controller like the existing one `config/registry_image_pruner/`?
like I've said this is only for 1 prod cluster, so it is better to keep it only in infra repo

> 
> 
> For the secret name, it should be good to mention `-quay-token` for clarity.

I think that is sufficient having there full repo name:
                    name: redhat-appstudio-tekton-catalog-pr-token
                    key: quaytoken
with key quaytoken

components/image-controller/production/stone-prd-rh01/resources/image_pruner/prune_images.py
```diff
@@ -0,0 +1,182 @@
...
+import re
+import time
+
+from collections.abc import Iterator
+from http.client import HTTPResponse
+from typing import Any, Dict, List
+from urllib.parse import urlencode
+from urllib.request import Request, urlopen
+from datetime import datetime
+from requests.exceptions import HTTPError
```
chmeliik, line 15: We can't import third-party libraries

```bash
$ podman run --rm -ti registry.redhat.io/rhel8/python-39:1-120.1684740828 python -c 'import requests'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'requests'
```

argo-cd-apps/base/member/infra-deployments/image-controller/image-controller.yaml
```diff
@@ -12,9 +12,11 @@ spec:
               values:
                 sourceRoot: components/image-controller
                 environment: staging
-                clusterDir: ""
+                clusterDir: base
           - list:
-              elements: []
+              elements:
+                - nameNormalized: stone-prd-rh01
+                  values.clusterDir: stone-prd-rh01
```
chmeliik, line 19: What do these changes do? Could you explain in the commit message?

chmeliik:
> > Why this prune_images.py script can't be shipped with image-controller like the existing one config/registry_image_pruner/?

> like I've said this is only for 1 prod cluster, so it is better to keep it only in infra repo

That's quite chaotic. Can we keep the pruning jobs in a consistent location? If the pruning is redhat-specific, it should probably in infra-deployments. If every deployment of image-controller should come with the pruning jobs, they should live in image-controller.

components/image-controller/production/stone-prd-rh01/resources/image_pruner/prune_images.py
```diff
@@ -0,0 +1,182 @@
...
+    image_digests = [image["manifest_digest"] for image in tags]
+
+    for tag in tags:
+        if (match := tag_regex.match(tag["name"])) is not None:
+            if f"sha256:{match.group(1)}" not in image_digests:
+                if dry_run:
+                    LOGGER.info("Leftover image %s from %s/%s should be removed", tag["name"], namespace, name)
+                else:
+                    LOGGER.info("Removing leftover image %s from %s/%s", tag["name"], namespace, name)
+                    delete_image_tag(quay_token, namespace, name, tag["name"])
```
chmeliik, line 98: >  * Is it possible to reuse the existing prune_images.py script under image-controller by making it more configurable?

> I don't want to mix those 2 scripts, they are doing different kind of pruning

_Are_ they doing different kinds of pruning? Isn't this exactly what the existing pruner does?

rcerven:
> > > Why this prune_images.py script can't be shipped with image-controller like the existing one config/registry_image_pruner/?
> 
> > like I've said this is only for 1 prod cluster, so it is better to keep it only in infra repo
> 
> That's quite chaotic. Can we keep the pruning jobs in a consistent location? If the pruning is redhat-specific, it should probably live in infra-deployments. If every deployment of image-controller should come with the pruning jobs, they should live in image-controller.

That is exactly how it is.
Pruner in image-controller is pruning (removing obsolete artifacts after image removal)  ALL user namespace in all clusters

These pruners are cleaning specific non-user production repositories (which should be done just once on one cluster),
also apart from leftovers it is also actually cleaning images based on age

components/image-controller/production/stone-prd-rh01/resources/image_pruner/prune_images.py
rcerven: yeah one is doing ONLY leftovers other is doing more, and like I've said bellow those scripts are for different purposes

rcerven:
> > not sure where are documented any current tokens
> 
> `https://<redhat-internal-docs-host>/display/RHTAPBUILD/Secrets`

ok will add it there

components/image-controller/production/stone-prd-rh01/resources/image_pruner/prune_images.py
rcerven: ok importing it now from urllib

mmorhun (approved):
LGMT

argo-cd-apps/base/member/infra-deployments/image-controller/image-controller.yaml
```diff
@@ -12,9 +12,11 @@ spec:
               values:
                 sourceRoot: components/image-controller
                 environment: staging
-                clusterDir: ""
+                clusterDir: base
```
hugares, line 15: Changing the default `clusterDir` affect both prod and staging. You moved all the files from `components/image-controller/production` to `components/image-controller/production/base` which is fine but you need to do the same for staging, i.e. move `components/image-controller/staging` into `components/image-controller/staging/base`

hugares (changes requested):

components/image-controller/production/stone-prd-rh01/resources/redhat-appstudio-pr-token.yaml
```diff
@@ -0,0 +1,19 @@
...
+metadata:
+  name: redhat-appstudio-pr-token
+  annotations:
+    argocd.argoproj.io/sync-options: SkipDryRunOnMissingResource=true
+    argocd.argoproj.io/sync-wave: "-1"
+spec:
+  dataFrom:
+    - extract:
+        key: production/build/redhat-appstudio-pr-token
+  refreshInterval: 5m
```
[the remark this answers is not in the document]
rcerven: ok changing to 20h, as job will run daily

components/image-controller/production/stone-prd-rh01/resources/redhat-appstudio-tekton-catalog-pr-token.yaml
```diff
@@ -0,0 +1,19 @@
...
+metadata:
+  name: redhat-appstudio-tekton-catalog-pr-token
+  annotations:
+    argocd.argoproj.io/sync-options: SkipDryRunOnMissingResource=true
+    argocd.argoproj.io/sync-wave: "-1"
+spec:
+  dataFrom:
+    - extract:
+        key: production/build/redhat-appstudio-tekton-catalog-pr-token
+  refreshInterval: 5m
```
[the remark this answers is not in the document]
rcerven: ok changing to 20h, as job will run daily

argo-cd-apps/base/member/infra-deployments/image-controller/image-controller.yaml
rcerven: will do

hugares (approved):

mmorhun (changes requested):
The script is working even if the token is not valid. Please check token validity before performing any action (e.g. fetching all tags)

mmorhun (approved):
/lgtm

PR #3493, opened 2024-03-25
feat(STONEINTG-844): add configuration template of task URL

dirgim:
* Adds a new configuration template of task URL to a new envvar so it can be used by integration service
* Override the default stage console URL configuration to account for the different console base URL


jsztuka (approved):
lgtm

jsztuka:
E2e should not be affected by this change, its failing on something else.

jsztuka (approved):
/lgtm

PR #3494, opened 2024-03-25
RHTAPSRE-489: Deploy repository-validator

gbenhaim:
The repository validator is an admission webhook for PAC repositories. It will deny the creation of those if their URL isn't allowed in the configuration of the webhook.

It's needed for the internal clusters, where we need to control which repositories are onboarded.

Webhook repo - https://github.com/konflux-ci/repository-validator


components/repository-validator/development/kustomization.yaml
hugares, line 1: This development folder is not used because the application is not part of development overlay. I would say either we add this to development overlay or delete this folder

gbenhaim: I added it to the overlay. thanks!

hugares (approved):
/lgtm

gbenhaim:
@hugares I needed to make modification to the linting action (I explained why in the second commit message).
Please let me know what you think.

PR #3504, opened 2024-03-26
RHTAPSRE-353: RBAC for creating workspaces in internal clusters

gbenhaim:
For creating new workspaces in the internal cluster a keycloak user should be created. Provide the permissions to manage keycloak users to the `konflux-workspace-admins` group.

### DIFF
--- a/argo-cd-apps/base/keycloak/keycloak.yaml
+++ b/argo-cd-apps/base/keycloak/keycloak.yaml
@@ -1,0 +1,49 @@
+apiVersion: argoproj.io/v1alpha1
+kind: ApplicationSet
+metadata:
+  name: keycloak
+spec:
+  generators:
+    - merge:
+        mergeKeys:
+          - nameNormalized
+        generators:
+          - clusters:
+              values:
+                sourceRoot: components/keycloak
+                environment: staging
+                clusterDir: base
+              selector:
+                matchLabels:
+                  appstudio.redhat.com/internal-member-cluster: "true"
+          - list:
+              elements: []
+  template:
+    metadata:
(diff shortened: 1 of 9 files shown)


argo-cd-apps/base/keycloak/keycloak.yaml
```diff
@@ -0,0 +1,49 @@
...
+        repoURL: https://github.com/redhat-appstudio/infra-deployments.git
+        targetRevision: main
+      destination:
+        namespace: rhtap-auth
+        server: '{{server}}'
+      ignoreDifferences:
+        - group: keycloak.org
+          kind: KeycloakRealm
+          jsonPointers:
+            - /spec/realm/identityProviders/0/config/clientSecret
```
gbenhaim, line 37: this is going to be required when I add the manifest for deploying keycloak itself.

components/keycloak/production/base/kustomization.yaml
hugares, line 1: Why putting this file in a base sub folder? It could be in components/keycloak/production as we have no cluster specific files.
(feel free to resolve this one, I am ok keeping it as-is if you prefer the way it is now)

components/keycloak/staging/base/kustomization.yaml
hugares, line 1: Why putting this file in a base sub folder? It could be in components/keycloak/staging as we have no cluster specific files.
(feel free to resolve this one, I am ok keeping it as-is if you prefer the way it is now)

hugares (approved):

gbenhaim:
The failed tests has nothing to do with this change.

PR #3509, opened 2024-03-26
chore: project-controller deployment in production

hmariset:
Update the production files to include project-controller deployment in production.

### DIFF
--- a/argo-cd-apps/overlays/konflux-public-production/kustomization.yaml
+++ b/argo-cd-apps/overlays/konflux-public-production/kustomization.yaml
@@ -12,9 +12,6 @@ resources:
 
 namespace: konflux-public-production
 
-patchesStrategicMerge:
-  - delete-applications.yaml
-
 patches:
   - path: production-overlay-patch.yaml
     target:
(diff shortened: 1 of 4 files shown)


argo-cd-apps/overlays/konflux-public-production/kustomization.yaml
```diff
@@ -12,9 +12,6 @@ resources:
 
 namespace: konflux-public-production
 
-patchesStrategicMerge:
-  - delete-applications.yaml
-
 patches:
```
hugares, line 15: I forgot to tell you before, you need to add a path in the list below for project-controller:

```
- path: production-overlay-patch.yaml
    target:
      kind: ApplicationSet
      version: v1alpha1
      name: project-controller
```

and you need the same in argo-cd-apps/overlays/production-downstream/kustomization.yaml

hugares (changes requested):

argo-cd-apps/overlays/konflux-public-production/kustomization.yaml
hmariset: Added

hugares (approved):
/lgtm

PR #3513, opened 2024-03-27
RHTAPSRE-467: Webhook configuration for internal production

gbenhaim:
The webhook configuration will allow to set the webhook target based on the repository url. Thus, external SCMs will emit webhooks to the smee-server, while internal SCMs will emit webhooks directly to the PAC endpoint on the cluster.


mmorhun (approved):
/lgtm

PR #3516, opened 2024-03-27
Remove capacityThresholds from the ToolchainConfig on staging

metlos:
This has been replaced by the dedicated SpaceProvisionerConfig CRs and is being removed from the ToolchainConfig CRD altogether.

Note that the deployed version of the sandbox host operator no longer uses this configuration and only looks at the configuration provided by the SpaceProvisionerConfig so this should be safe to do. Nevertheless, this PR only updates the staging cluster configuration and if all goes well, the same can be applied to prod.

There is a separate PR doing the same thing on prod: https://github.com/redhat-appstudio/infra-deployments/pull/3517

### DIFF
--- a/components/sandbox/toolchain-host-operator/staging/stone-stage-p01/toolchainconfig.yaml
+++ b/components/sandbox/toolchain-host-operator/staging/stone-stage-p01/toolchainconfig.yaml
@@ -12,9 +12,6 @@ spec:
         defaultSpaceTier: 'appstudio'
       automaticApproval:
         enabled: true
-      capacityThresholds:
-        resourceCapacityThreshold:
-          defaultThreshold: 90
       spaceConfig:
         spaceRequestEnabled: true
         spaceBindingRequestEnabled: true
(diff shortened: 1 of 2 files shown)


mfrancisc (approved):

metlos:
Seems like the CI didn't like my fork being called `appstudio-infra-deployments`. Let's see if renaming it to just `infra-deployments` will help...

metlos:
@mfrancisc would you happen to know why all the tests are failing? https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/redhat-appstudio_infra-deployments/3516/pull-ci-redhat-appstudio-infra-deployments-main-appstudio-e2e-tests/1772981580659167232#1:build-log.txt%3A1210 seems like an infrastructure issue?

mfrancisc:
@metlos TBH I'm not 100% either :( . I see multiple errors in the logs , but I don't think those are related to your change. I usually retry only the required jobs.

PR #3520, opened 2024-03-28
Enable auto-approval for internal production

gbenhaim:
Will help to reduce the load from the migration team to approve users.


components/sandbox/toolchain-host-operator/production/stone-prod-p01/toolchainconfig.yaml
```diff
@@ -72,4 +72,4 @@ spec:
       memberStatus:
         gitHubSecret:
           ref: member-operator-secret
-          accessTokenKey: github.access.token
```
ernesgonzalez33, line 75: I think you deleted the last line of the file without noticing

gbenhaim: it's just the addition of a new line

PR #3556, opened 2024-04-08
chore(RHTAPWATCH-891): Update o11y prod commit sha

hmariset:
Updated commit sha for o11y grafana production


amisstea (approved):
I verified that we now have the `grafana_ds_up` metric available in our stage RHOBS tenant.

/lgtm

PR #3559, opened 2024-04-09
chore: deploy obo on internal staging cluster

yftacherzog:
The observability operator was assumed to be deploy out of the box on all non-dev clusters as it's built into ROSA, but this is not the case for the internal stage cluster. This change is installing it explicitly to that cluster.


manish-jangra (approved):
looks good to me

eisraeli (approved):

PR #3560, opened 2024-04-09
KFLUXINFRA-470: Replace stonesoups01ue1 with appsres05ue1 in Secret Paths

manish-jangra:
We're migrating all the resources from `stonesoups01ue1` to `appsres05ue1`. The app interface merge request has been merged, therefore we need to update the secret paths in infra-deployments.


hugares (approved):
/lgtm

PR #3568, opened 2024-04-15
Move stone-stg-host authentication to rh-idp

hugares:
RH sso is configured as an identity provider already on that cluster. Change the authentication component to use rh-idp folder instead of github which will sync the groups from rover instead of github.

Since all the clusters will use ldap connection to sync rover groups, move the ldap cert to base folder. Also change the default ldap URL to the external facing one and path the internal clusters to use the internal LDAP.

External LDAP requires authentication and internal does not but supports it. To make things more simple, configure both to use authentication.

[KFLUXINFRA-132](https://issues.redhat.com//browse/KFLUXINFRA-132)


hugares:
This PR is not tested by e2e tests, I will by-pass CI

PR #3571, opened 2024-04-16
KFLUXINFRA-471: Replacing stonesoupp01ue1 with appsrep05ue1 in Secret Paths

manish-jangra:
We're migrating all the Konflux resources from `stonesoupp01ue1` to `appsrep05ue1`. This requires to update the secret paths or configurations containing `stonesoupp01ue1` with `appsrep05ue1`.


bamachrn (approved):
LGTM

PR #3573, opened 2024-04-16
Create common app for Kubesaw to avoid collisions

dperaza4dustbit:
Fixes Bug: ASC-503

When host operator and member operator are deployed in same cluster there are two applications synching the same resource which is causing ArgoCD to trigger a degradation warning. Moving common resources of host and members to their own application that gets deployed on all clusters.


argo-cd-apps/base/all-clusters/infra-deployments/kustomization.yaml
```diff
@@ -11,5 +11,6 @@ resources:
   - disable-csvcopy-for-all-cluster
   - enable-dvo-for-all-cluster
   - backup
+  - kubesaw-common
```
hugares, line 14: Adding this application here is OK as we want it in staging and production but we should not deploy this in development overlay. Currently, host and member application exists in dev but they point to a folder that contains nothing which is not the case for kubesaw-common.
I opened this PR(https://github.com/redhat-appstudio/infra-deployments/pull/3576) to remove host and member from development overlay so we should get it merge, rebase this PR and delete kubesaw-common from dev overlay

components/sandbox/kustomization.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- ../base
```
hugares, line 4: this is wrong path, kustomize build is failing. I suggest instead of adding this kustomize file, that we rename base folder to kubesaw-common and change `argo-cd-apps/base/all-clusters/infra-deployments/kubesaw-common/kubesaw-common.yaml` application to point to `components/sandbox/kubesaw-common` folder.

Having a folder named the same way as the ArgoCD application is kind of a pattern we use, making it simpler to identify the folder associated with the application

hugares (changes requested):

components/sandbox/kustomization.yaml
dperaza4dustbit: OK, I just directly pointed to base in worked, going to rename base

argo-cd-apps/base/all-clusters/infra-deployments/kustomization.yaml
dperaza4dustbit: OK waiting for PR(https://github.com/redhat-appstudio/infra-deployments/pull/3576) to merge. I assume I will remove kubesaw-common from dev overlay in another PR right?

components/sandbox/kustomization.yaml
hugares: I like that !

argo-cd-apps/base/all-clusters/infra-deployments/kustomization.yaml
hugares: In this PR, In fact you can do it without waiting for my PR to go in, just follow the example from my PR

hugares: My PR passed the tests, can you approve and lgtm it, it will merge with approvals

components/sandbox/kustomization.yaml
dperaza4dustbit: done

hugares (approved):

mfrancisc (approved):

PR #3576, opened 2024-04-16
Delete toolchain host and member application from dev overlay

hugares:
Both applications point to an empty folder, kustomize resources wise. Remove both ArgoCD applications and they empty folders to remove the possible confusion. Toolchain is not deployed using ArgoCD in development overlay.


dperaza4dustbit (approved):
/lgtm

PR #3580, opened 2024-04-17
try enabling chains deep inspection via infra change only given tekto…

enarha:
to test the konflux e2e's against openshift-pipelines/pipeline-service#985 before merging openshift-pipelines/pipeline-service#985


gabemontero:
/lgtm

@yashvardhannanavati  @lcarva FYI

lcarva:
It sounds like we should have a single place to define the Chains config?

gabemontero:
> It sounds like we should have a single place to define the Chains config?

that is a long term goal @lcarva not just for chains, but for all tekton related config, given how truly dead "tekton as a service" is; I

n fact I recently prototyped a first step in unwinding things, and having a "infra-deployments only" gitops / kustomize approach, and @enarha and I have had some preliminary discussions on the details of it; however, that prototype could be cleaned up even more, and then we have to worry about the feeders into pipeline-service repo
- consumption of the openshift pipeline nightlies as part of Red Hat's "Service First" initiative
- our downstream results fork (while the operator's version still lags too far behind for Konflux's needs)
- the metrics exporter (for the stats we are still iterating on, where upstreaming again is too long of a loop)

PR #3590, opened 2024-04-18
Fix deployment of kubesaw-common application

hugares:
The deployment was failing due to the missing namespace. Create the namespace resource and also add the namespace directive in the kustomization file to create all resources in the namespace.

[ASC-503](https://issues.redhat.com//browse/ASC-503)


dperaza4dustbit (commented):
This looks good to me, just one comment before approve should we set this to false on the kubesaw-common application:

syncOptions:
          - CreateNamespace=true

hugares:
> This looks good to me, just one comment before approve should we set this to false on the kubesaw-common application:
> 
> syncOptions: - CreateNamespace=true

Good point, done in latest revision

dperaza4dustbit (approved):
/lgtm

PR #3593, opened 2024-04-18
RHTAPSRE-296: Keycloak deployment

gbenhaim:
- Deploy Keycloak on internal staging.

- Prepare the manifests production but don't deploy them yet.

- Don't deploy to the development environment yet, since it requires additional rework of the deployment scripts.

- Set the sync policy of the Keycloak app to manual. It will help to revert quickly in case the change will break the existing manual deployment.

- A development overlay was added to the ui component, but it isn't deployed by default since it requires additional rework of the deployment scripts.


components/ui/development/set-hostname.yaml
```diff
@@ -0,0 +1,4 @@
+---
+- op: add
+  path: /spec/host
+  value: @TBA
```
hugares, line 4: Kustomize build is failing because of the "@"
```suggestion
  value: TBA
```

gbenhaim: I added quotes. I want to keep the @ sign since it would be useful later when adjusting the dev scripts.

hugares: You forgot to push the update?

gbenhaim: I pushed the update now

PR #3595, opened 2024-04-18
Renaming kubesaw-common namespace to bypass validation rules

dperaza4dustbit:
There is a velidation webhook that blocks namespaces that contains string kube.


hugares (approved):
/lgtm

PR #3598, opened 2024-04-19
Remove unsupported defaultTier from ToolchainConfig

hugares:
This field is causing ArgoCD to be out of sync as it add the field but OCP is removing it as it is not part of CRD definition.


dperaza4dustbit (approved):
/lgtm

PR #3601, opened 2024-04-19
chore: add ci-helper-app Repository CR

psturc:
https://issues.redhat.com/browse/KONFLUX-2245

First part of moving manifests for ci-helper-app to infra-deployments

### DIFF
--- a/components/tekton-ci/base/repository.yaml
+++ b/components/tekton-ci/base/repository.yaml
@@ -103,3 +103,10 @@ metadata:
   name: quality-dashboard
 spec:
   url: "https://github.com/redhat-appstudio/quality-dashboard"
+---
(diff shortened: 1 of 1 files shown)


components/tekton-ci/base/repository.yaml
```diff
@@ -103,3 +103,10 @@ metadata:
   name: quality-dashboard
 spec:
   url: "https://github.com/redhat-appstudio/quality-dashboard"
+---
```
tisutisu, line 106: Will it interfere with the existing [tekton pipelines](https://github.com/redhat-appstudio/ci-helper-app/tree/main/.tekton) in the repo?

psturc: good question, these pipelines manifests will be removed in a follow-up PR

PR #3629, opened 2024-04-25
Build config for build pipelines for build-service,

rcerven:
will be used in future instead of build selectors

[STONEBLD-2276](https://issues.redhat.com//browse/STONEBLD-2276)

### DIFF
--- a/components/build-service/base/build-pipeline-config/build-pipeline-config.yaml
+++ b/components/build-service/base/build-pipeline-config/build-pipeline-config.yaml
@@ -1,0 +1,13 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: build-pipeline-config
+  namespace: build-service
+data:
+  config.yaml: |
+    default-pipeline-name: docker-builder
+    pipelines:
+    - name: fbc-builder
+      bundle: quay.io/redhat-appstudio-tekton-catalog/pipeline-fbc-builder:032a8745d43a942a247f365fc890b06023ccd67d
+    - name: docker-builder
+      bundle: quay.io/redhat-appstudio-tekton-catalog/pipeline-docker-build:032a8745d43a942a247f365fc890b06023ccd67d
--- a/components/build-service/base/rbac/build-config.yaml
+++ b/components/build-service/base/rbac/build-config.yaml
@@ -1,0 +1,15 @@
+# role in build-service https://github.com/konflux-ci/build-service/pull/296
+---
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: build-pipeline-config-read-only-binding
+  namespace: build-service
+subjects:
+  - apiGroup: rbac.authorization.k8s.io
+    kind: Group
+    name: system:authenticated
+roleRef:
+  kind: Role
+  name: build-pipeline-config-read-only
+  apiGroup: rbac.authorization.k8s.io
--- a/components/build-service/base/kustomization.yaml
+++ b/components/build-service/base/kustomization.yaml
@@ -1,6 +1,7 @@
 resources:
 - allow-argocd-to-manage.yaml
 - build-pipeline-selectors/build-pipeline-selector.yaml
+- build-pipeline-config/build-pipeline-config.yaml
 - monitoring.yaml
 - rbac
 
--- a/components/build-service/base/rbac/kustomization.yaml
+++ b/components/build-service/base/rbac/kustomization.yaml
@@ -1,6 +1,7 @@
 resources:
 - build-maintainer.yaml
 - build-admin.yaml
+- build-config.yaml
 
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization


components/build-service/base/build-pipeline-config/build-pipeline-config.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: build-pipeline-config
```
mmorhun, line 4: To me current name is confusing. We are listing rather `build-pipelines` than a config of a pipeline(s). Please rename CM and the file

hugares (approved):

components/build-service/base/build-pipeline-config/build-pipeline-config.yaml
mmorhun: If we are going to have extended config (not just key-value pairs), then we should consider following [Tekton resolver structure](https://tekton.dev/docs/pipelines/resolution-getting-started/#getting-started-with-resolvers): `resolver`+`params`, where `resolver` could be:
 - `bundles`
 - `git`
 - `cluster`
 - ...

This, in turn, would require more advanced handling on Build Service side.

components/build-service/base/build-pipeline-config/build-pipeline-config.yaml
```diff
@@ -0,0 +1,16 @@
...
+config.yaml: |
+  pipelines:
+  - name: fbc-builder
+    bundle: quay.io/redhat-appstudio-tekton-catalog/pipeline-fbc-builder:032a8745d43a942a247f365fc890b06023ccd67d
+  - name: docker-builder
+    bundle: quay.io/redhat-appstudio-tekton-catalog/pipeline-docker-build:032a8745d43a942a247f365fc890b06023ccd67d
+  - name: java-builder
+    bundle: quay.io/redhat-appstudio-tekton-catalog/pipeline-java-builder:032a8745d43a942a247f365fc890b06023ccd67d
+  - name: nodejs-builder
+    bundle: quay.io/redhat-appstudio-tekton-catalog/pipeline-nodejs-builder:032a8745d43a942a247f365fc890b06023ccd67d
```
arewm, line 16: Did we have a decision on whether we were creating these?

components/build-service/base/build-pipeline-config/build-pipeline-config.yaml
```diff
@@ -0,0 +1,16 @@
...
+metadata:
+  name: build-pipeline-config
+  namespace: build-service
+data:
+config.yaml: |
+  pipelines:
+  - name: fbc-builder
+    bundle: quay.io/redhat-appstudio-tekton-catalog/pipeline-fbc-builder:032a8745d43a942a247f365fc890b06023ccd67d
+  - name: docker-builder
+    bundle: quay.io/redhat-appstudio-tekton-catalog/pipeline-docker-build:032a8745d43a942a247f365fc890b06023ccd67d
```
arewm, line 12: When can we get away from using the term "docker"? 😅

rcerven: suggest something else

components/build-service/base/build-pipeline-config/build-pipeline-config.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: build-pipeline-config
+  namespace: build-service
+data:
+config.yaml: |
+  pipelines:
```
mmorhun, line 8: Does it make sense to have default pipeline specified, so UI can show it by default? E.g.
```yaml
  default-pipeline-name: docker-builder
  pipelines:
     ...
```

components/build-service/base/build-pipeline-config/build-pipeline-config.yaml
mmorhun: I think we can omit s2i since they are not used, AFAIK.

components/build-service/base/build-pipeline-config/build-pipeline-config.yaml
rcerven: added

components/build-service/base/build-pipeline-config/build-pipeline-config.yaml
rcerven: removed

mmorhun (approved):

components/build-service/base/rbac/build-config.yaml
```diff
@@ -0,0 +1,26 @@
+kind: Role
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: build-pipeline-config-read-only
+  namespace: build-service
+rules:
+- apiGroups: [""]
+  resources: ["configmaps"]
+  resourceNames: ["build-pipeline-config"]
+  verbs: ["get"]
```
chmeliik, line 10: Maybe this file should be in the build-service repo? That's where the BuildPipelineSelector role lives too https://github.com/konflux-ci/build-service/blob/main/config/rbac/buildpipelineselector_viewer_role.yaml

mmorhun: Yes, let's move it into Build Service, since the role might be needed for community edition of Konflux.

rcerven: moved role to build service

mmorhun (approved):
/lgtm

PR #3631, opened 2024-04-26
chore: move ci-helper-app Repo CR to konflux-ci

psturc:
[KONFLUX-2245](https://issues.redhat.com//browse/KONFLUX-2245)

### DIFF
--- a/components/konflux-ci/base/repository.yaml
+++ b/components/konflux-ci/base/repository.yaml
@@ -5,3 +5,10 @@ metadata:
   name: sprayproxy
 spec:
   url: "https://github.com/konflux-ci/sprayproxy"
+---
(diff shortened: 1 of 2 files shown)


psturc:
@ralphbean is there a plan to store anything related to dev environment in `components/konflux-ci`?

I'm wondering if we should turn off triggering openshift-ci e2e jobs for changes made only in `components/konflux-ci` directory (i.e. to update [this regexp](https://github.com/openshift/release/blob/335c62472139fdd498e55afc773eced3fb4b8bc4/ci-operator/config/redhat-appstudio/infra-deployments/redhat-appstudio-infra-deployments-main.yaml#L51))

ralphbean:
@psturc good catch. No further plans for components/konflux-ci. I think it can be safely added to that regex.

PR #3644, opened 2024-04-30
pin results image until watcher mem leak fix is deemed ok

gabemontero:
An idea I had @enarha to bypass the results bump I'm getting performance QE to help me test with https://github.com/redhat-appstudio/infra-deployments/pull/3616

Rather than pick up the results image bumps like you see in https://github.com/redhat-appstudio/infra-deployments/pull/3616/files#diff-e961bdc3a405dfda1b39ed17ee82576bb68e2a35e9288e7de208e695e4ee5944R1497

We could force dev/stage/prod to specific sha's of the results image from here in infra deployments, and effectively pin the version here.

That way, if your operator bump is ready to merge before the testing completes with https://github.com/redhat-appstudio/infra-deployments/pull/3616 we could include something like this in the PR that does your operator bump without having to revert the results bump in openshift-pipelines/pipeline-service.

What do you think?

rh-pre-commit.version: 2.2.0

### DIFF
--- a/components/pipeline-service/development/kustomization.yaml
+++ b/components/pipeline-service/development/kustomization.yaml
@@ -12,6 +12,12 @@ resources:
   - https://github.com/openshift-pipelines/pipeline-service.git/developer/openshift/gitops/argocd/pipeline-service-storage?ref=e52f83e174efb8f308f6e32d1e0fc9f8eb9ed893
   - ../base/rbac
 
+images:
+  - name: quay.io/redhat-appstudio/tekton-results-api
+    newTag: bae7851ff584423503af324200f52cd28ca99116
+  - name: quay.io/redhat-appstudio/tekton-results-watcher
+    newTag: bae7851ff584423503af324200f52cd28ca99116
+
 patches:
   - path: chains-tekton-config-patches.yaml
     target:
--- a/components/pipeline-service/production/base/kustomization.yaml
+++ b/components/pipeline-service/production/base/kustomization.yaml
@@ -14,6 +14,12 @@ resources:
   - ../../base/testing
   - ../../base/rbac
 
+images:
+  - name: quay.io/redhat-appstudio/tekton-results-api
+    newTag: bae7851ff584423503af324200f52cd28ca99116
+  - name: quay.io/redhat-appstudio/tekton-results-watcher
+    newTag: bae7851ff584423503af324200f52cd28ca99116
+
 patches:
   # - path: scale-down-exporter.yaml
   #   target:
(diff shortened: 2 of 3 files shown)


enarha:
> An idea I had @enarha to bypass the results bump I'm getting performance QE to help me test with #3616
> 
> Rather than pick up the results image bumps like you see in https://github.com/redhat-appstudio/infra-deployments/pull/3616/files#diff-e961bdc3a405dfda1b39ed17ee82576bb68e2a35e9288e7de208e695e4ee5944R1497
> 
> We could force dev/stage/prod to specific sha's of the results image from here in infra deployments, and effectively pin the version here.
> 
> That way, if your operator bump is ready to merge before the testing completes with #3616 we could include something like this in the PR that does your operator bump without having to revert the results bump in openshift-pipelines/pipeline-service.
> 
> What do you think?
> 

We can do that. But I would like to pin the prod versions as well (to whatever version), so we do not accidentally update prod when we promote the operator bump to prod.

gabemontero:
> > An idea I had @enarha to bypass the results bump I'm getting performance QE to help me test with #3616
> > Rather than pick up the results image bumps like you see in https://github.com/redhat-appstudio/infra-deployments/pull/3616/files#diff-e961bdc3a405dfda1b39ed17ee82576bb68e2a35e9288e7de208e695e4ee5944R1497
> > We could force dev/stage/prod to specific sha's of the results image from here in infra deployments, and effectively pin the version here.
> > That way, if your operator bump is ready to merge before the testing completes with #3616 we could include something like this in the PR that does your operator bump without having to revert the results bump in openshift-pipelines/pipeline-service.
> > What do you think?
> 
> We can do that. But I would like to pin the prod versions as well (to whatever version), so we do not accidentally update prod when we promote the operator bump to prod.

Totally agree wrt prod @enarha .  This was just a quick demonstration.  Bottom line, let's apply this technique across dev/stage/prod when we are ready to bump the nightlies, in the nightly bump PR,  if the results testing I've commissioned performance QE to help with has not completed. 

Closing this out.

Thanks.

gabemontero:
OK @enarha I'm making this official given the performance test results from staging over the weekend with that latest form of the mem leak / cancel context fix

Per my recent email, I'm back to reworking that 

This will allow us to iterate in the pipeline service repo without reverting results bumps there.  We'll simply undo dev, stage, then prod here as we vet the next version of the fix.

enarha:
/lgtm
Just to confirm, you tried to regenerate the deploy.yaml files, but it did not yield any diff, right?

gabemontero:
> /lgtm Just to confirm, you tried to regenerate the deploy.yaml files, but it did not yield any diff, right?

correct @enarha

PR #3646, opened 2024-05-01
WIP: Local upstream 3204 copy

gabemontero:
a copy of https://github.com/redhat-appstudio/infra-deployments/pull/3204 @enarha so that we can manually rebase and see fixes elsewhere are needed for the build controller and multi-platform controller start issues, and prove unconditionally they are somehow caused by the


enarha:
Thanks. I usually rebase the original PR (in this case #3204). I already did once, few days ago with earlier nightly version.
I also just tested that specific nightly version (5.0.5-307) on ROSA cluster and it works well. I remember it also passing the pipeline-service CI, but I see it now red, not sure what happened. I can give it a manual try again tomorrow.

PR #3656, opened 2024-05-03
Bump hac-dev, chrome service, chrome frontend tags

ashwindasr:
Mainly to get https://github.com/openshift/hac-dev/pull/926 in, but bumping the others as well.

### DIFF
--- a/components/ui/production/kustomization.yaml
+++ b/components/ui/production/kustomization.yaml
@@ -11,7 +11,7 @@ images:
   # chrome service
   - name: quay.io/cloudservices/chrome-service
     newName: quay.io/cloudservices/chrome-service
-    newTag: 0574186
+    newTag: 1f6a4c1
   # hac
   - name: quay.io/cloudservices/hac-core-frontend
     newName: quay.io/cloudservices/hac-core-frontend
--- a/components/ui/production/kustomization.yaml
+++ b/components/ui/production/kustomization.yaml
@@ -7,11 +7,11 @@ images:
   # chrome frontend
   - name: quay.io/cloudservices/insights-chrome-frontend
     newName: quay.io/cloudservices/insights-chrome-frontend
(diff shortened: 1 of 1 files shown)


ashwindasr:
Thanks @hugares, since I don't have the permissions, you may have to force push like last time https://github.com/redhat-appstudio/infra-deployments/pull/3632

hugares:
> Thanks @hugares, since I don't have the permissions, you may have to force push like last time #3632

Sure, I can do that. There is an issue with kustomize build of prod overlay:
```
Error: invalid Kustomization: json: cannot unmarshal number into Go struct field Image.images.newTag of type string
```

ashwindasr:
Yeah looks we we got the tag as all numbers https://quay.io/repository/cloudservices/chrome-service?tab=tags

ashwindasr:
when you're force pushing, can you use the tag before this one, thats alpha-numberic so should be good

hugares:
> when you're force pushing, can you use the tag before this one, thats alpha-numberic so should be good

Can you comment in the file with the exact tags I should use, so I put the right ones

ashwindasr:
updated PR, pushed the new tag

hugares (approved):
/lgtm

hugares:
> Thanks @hugares, since I don't have the permissions, you may have to force push like last time

Looks like this will not be needed. Since I put the /ok-to-test and you pushed again, the check that used to block things is not there

ashwindasr:
looks like `appstudio-e2e-tests` failure is not a flake. any ideas @hugares

ashwindasr:
/hold

Since @sahil143 had mentioned that there are other big changes that were merged this week, which are still yet to be tested on staging.

ashwindasr:
/unhold

Since testing has finished

PR #3672, opened 2024-05-08
Update application service references

thepetk:
Updates refererences of `redhat-appstudio/application-service` to `konflux-ci/application-service` as part of the migration of the application-service repos to the konflux-ci orgs.


thepetk:
@johnmcollier I haven't updated the references to `quay.io/redhat-appstudio/application-service` image yet. We should do that after we have set up the new component right?

PR #3675, opened 2024-05-08
Deploy backup on clusters with backup label only

hugares:
With the addition of EaaS clusters, the backup need to deployed only on host and member clusters, in other words all clusters except Eaas.

ArgoCD ApplicationSet cluster generators does not support or condition to match `host` or `member` labels. For that reason, a new label named 'appstudio.redhat.com/backup' was added to all clusters where we need backup, i.e. host and members, see [1].

Change backup ApplicationSet to deploy on clusters with that new label. Since all application set defined in argo-cd-apps/base/all-clusters are getting a selector applied with Kustomize, adding the matchLabels in backup ApplicationSet only was getting overwritten by the all-cluster selector.

To work around that issue, move backup folder out of argo-cd-apps/base/all-clusters and include it in both staging and prod overlays.

[KFLUXINFRA-554](https://issues.redhat.com//browse/KFLUXINFRA-554)

[1]https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/105341


manish-jangra (commented):
If we're removing backup from all-clusters, shouldn't we add this to `staging-downstream` and `production-downstream` overlays as well?

hugares:
> If we remove the backup from all-clusters, shouldn't we add this to the `staging-downstream` and `production-downstream` overlays too? Isn't it the same thing, removing from all clusters and adding backup manually to each overlay, it sounds like duplicating. I'm sorry if I'm confusing.
both downstream overlays are including the public ones so no need to add backup, it is already there:
```
```

manish-jangra:
looks good to me :)

PR #3682, opened 2024-05-09
Create first application/component on eaas cluster

hugares:
This change add a new eaas folder where we can put all ArgoCD applications specific to that cluster type.

To have a complete skeleton, add a dummy application/component named `some-component`, to be replaced with real one.


amisstea (approved):
/lgtm

manish-jangra (approved):
looks good to me

PR #3683, opened 2024-05-09
Do not create openshift-logging namespace

hugares:
This was added to the prerequisite folder for stone-stage-p01 when it was HCP cluster but on ROSA clusters, the namespace already exists and cannot be tampered with as it is protected by admission web hook.

This is the reason why logging application is always out of sync on stone-prod-p01 and now the same is happening on kflux-stg-es01 as it will happen on stone-stage-p01 once ongoing re-installation is completed.


amisstea (approved):
/lgtm

PR #3686, opened 2024-05-10
feat: allow integration team to manage integrationtestscenarios

dirgim:
* The integration team is currently best suited to deal with issues with IntegrationTestScenario resources
* Giving the integration team permission to manage IntegrationTestScenarios across all namespaces will decrease customer complaints and prevent them from having to recognize and resolve the issue themselves


hugares (approved):
/lgtm

PR #3693, opened 2024-05-13
Enable Tekton StepActions in dev and staging

amisstea:
This is a workaround for passing artifacts between tasks in an integration pipeline where workspaces backed by persistent volumes are not available. It was one of the suggestions proposed during last week's community architecture call on the topic of passing around ephemeral cluster credentials after they've been provisioned.

### DIFF
--- a/components/pipeline-service/development/enable-step-actions.yaml
+++ b/components/pipeline-service/development/enable-step-actions.yaml
@@ -1,0 +1,8 @@
+---
+apiVersion: operator.tekton.dev/v1alpha1
+kind: TektonConfig
+metadata:
+  name: config
+spec:
+  pipeline:
+    enable-step-actions: true
--- a/components/pipeline-service/development/kustomization.yaml
+++ b/components/pipeline-service/development/kustomization.yaml
@@ -52,3 +52,7 @@ patches:
       kind: Deployment
       namespace: tekton-results
       name: tekton-results-watcher
+  - path: enable-step-actions.yaml
+    target:
+      kind: TektonConfig
+      name: config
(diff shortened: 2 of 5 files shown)


lcarva:
@amisstea, I'm curious how StepActions would help in passing data between tasks.

amisstea:
> @amisstea, I'm curious how StepActions would help in passing data between tasks.

The credentials for authenticating to ephemeral clusters won't reside in the tenant's namespace but rather on another cluster which the pipeline runner service account will be able to access. A StepAction will allow integration pipeline authors to easily acquire the kubeconfig and place it in an `emptyDir` workspace shared across the TaskRun. If the credentials are required in another task, the StepAction can be reused. So, in effect, this doesn't allow the user to pass data between tasks but eliminates the need in this specific case.

lcarva:
> > @amisstea, I'm curious how StepActions would help in passing data between tasks.
> 
> The credentials for authenticating to ephemeral clusters won't reside in the tenant's namespace but rather on another cluster which the pipeline runner service account will be able to access. A StepAction will allow integration pipeline authors to easily acquire the kubeconfig and place it in an `emptyDir` workspace shared across the TaskRun. If the credentials are required in another task, the StepAction can be reused. So, in effect, this doesn't allow the user to pass data between tasks but eliminates the need in this specific case.

Thank you for the thorough explanation. It makes sense to me. We have another effort, Trusted Artifacts, that could also leverage StepActions. Depending on how your experience is with StepActions, we may use it as well.

gabemontero (approved):
/lgtm

PR #3701, opened 2024-05-14
Only deploy ci-helper-app ApplicationSet on public staging

hugares:
ApplicationSet was added to development overlay to then be deleted. Simplify this by not adding in the first place.

Since both staging and production downstream are based on public staging overlay, they inherit the ci-helper-app ApplicationSet but we do not have a cluster labeled with `appstudio.redhat.com/ci-helper-app: "true"` so not application are created. Delete the ApplicationSet from those overlays.


argo-cd-apps/overlays/development/kustomization.yaml
```diff
@@ -12,7 +12,6 @@ resources:
   - ../../base/quality-dashboard
```
psturc, line 12: should we do the same for quality-dashboard?

hugares: yes, I will do a separate PR for it

argo-cd-apps/overlays/development/kustomization.yaml
dheerajodha, line 1: should we also remove this block of code that's preset at the end of this file?
```
- path: development-overlay-patch.yaml
  target:
      kind: ApplicationSet
      version: v1alpha1
      name: ci-helper-app
```

argo-cd-apps/overlays/development/kustomization.yaml
hugares: https://github.com/redhat-appstudio/infra-deployments/pull/3702

argo-cd-apps/overlays/development/kustomization.yaml
dheerajodha: thank you!

dheerajodha (approved):

PR #3703, opened 2024-05-15
feat: Add ability to run the e2e-tests from rhtap-build-tenant ns

tisutisu:
`rhtap-build-tenant:appstudio-pipeline` SA needs the same level of privileges that our `tekton-ci:appstudio-pipeline` sa has, to be able to run the e2e-tests in `build-templates-e2e` namespace of production cluster.
This PR adds required role-binding for `rhtap-build-tenant:appstudio-pipeline` SA to run e2e-tests in build-templates-e2e ns.
Part of the story: https://issues.redhat.com/browse/STONEBLD-2353
After the source-build image is built, we will run the e2e-tests through the integration-pipeline added in [this PR](https://github.com/redhat-appstudio/build-tasks-dockerfiles/pull/85)

### DIFF
--- a/components/build-templates/base/e2e/rolebinding.yaml
+++ b/components/build-templates/base/e2e/rolebinding.yaml
@@ -14,6 +14,9 @@ subjects:
 - kind: ServiceAccount
   name: appstudio-pipeline
   namespace: konflux-ci
+- kind: ServiceAccount
+  name: appstudio-pipeline
+  namespace: rhtap-build-tenant
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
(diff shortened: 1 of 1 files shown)


tisutisu:
> Could you improve the description to explain why exactly this is needed?
> 
> If I understand correctly, it's because we onboarded the https://github.com/redhat-appstudio/build-tasks-dockerfiles/ repo to konflux, so now all the pipelines in that repo have to run in the rhtap-build-tenant namespace. And we want to run e2e tests in that repo as a Tekton pipeline, not via Prow (why?).

We wanted to use a konflux integration test for running the e2e-tests after source-build image is built, use similar approach like the `build-definitions` CI but it will be a integration-pipeline instead of build-pipeline so the `rhtap-build-tenant:appstudio-pipeline` needs the same level of privileges that our `tekton-ci:appstudio-pipeline` sa has, to be able to run the e2e-tests. Prow was not picked because we are migrating away from it and will be moving to all the konflux-ci in future. Does it makes sense?

chmeliik (approved):
/lgtm

Ack,thanks for the explanation :

PR #3706, opened 2024-05-15
chore: qd: konflux-qe-admins perms for pods

psturc:
we want to be able to restart pods for QD and ci-helper-app on staging cluster in cases when for example secrets get updated in vault


hugares (approved):
/lgtm

PR #3708, opened 2024-05-15
Improve deployment

gbenhaim:
Sync waves for Keycloak
    
    Deploy the namespace, operator group, subscription, and then
    all the rest. Otherwise CRs are failed to get deployed.

    Improve sandbox deployment
    
    1. Create the namespace for the host and member components.
    2. Use sync waves because there are dependency between the resources.


hugares (approved):
/lgtm

PR #3709, opened 2024-05-15
Update routes for internal staging and update Keycloak IDP config

gbenhaim:
PAC: Update console routes for internal staging

    UI: Update routes for staging

    Keycloak: Update staging routes

    Add missing config to the OCP IDP
    
    The sync mode which decides how users are imported was missing.
    FORCE means the user will get updated in Keycloak every time he/she
    logins.


hugares (approved):
/lgtm

PR #3717, opened 2024-05-16
Update Keycloak routes for the registration-service int stage

gbenhaim:


gbenhaim:
This change isn't tested in the e2e test suite.

PR #3727, opened 2024-05-16
Additional sandbox configurations for int staging

gbenhaim:
Router and UI proxy access for the registration service
    
    The OCP router needs access to the host operator namespace since
    it has routes in it.
    
    In addition, the proxy used by the UI needs to access the registration
    service.

    Update routes for int stage space provisioner


hugares (approved):
/lgtm

PR #3738, opened 2024-05-20
Reverting to Old AMIs

manish-jangra:
Blessed AMI(s) are encrypted with KMS Keys from Another AWS Account.

That would require some spike story to use them for the Multi-Platform Controller.


bamachrn (approved):
/lgtm

PR #3752, opened 2024-05-21
fix: sync snapshotgc reference in integration component

dirgim:
The snapshot garbage collector reference has gone out of sync so it's not being updated by the promotion script anymore.


MartinBasti (approved):

MartinBasti:
https://github.com/redhat-appstudio/infra-deployments/pull/3750/files

ahh, so "only" prod is affected

PR #3768, opened 2024-05-23
Remove gitops prod

gbenhaim:


gbenhaim:
@psturc can you please run smoke tests on statging so we would make sure the removal of gitops didn't break anything and then we can remove it from prod as well?

hugares (approved):

gbenhaim:
testing in prod was completed successfully.
Merging directly since e2e test doesn't changes for prod.

PR #3773, opened 2024-05-23
chore: add cronjob to cleanup stale applications in build-templates-e2e ns

tisutisu:
Story: https://issues.redhat.com/browse/STONEBLD-2323

### DIFF
--- a/components/tekton-ci/production/cleanup-cronjob.yaml
+++ b/components/tekton-ci/production/cleanup-cronjob.yaml
@@ -1,0 +1,31 @@
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: cleanup-stale-resources
+spec:
+  schedule: "0 1 * * *" #Daily at 01:00
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          serviceAccountName: appstudio-pipeline
+          containers:
+          - name: remove-stale-resources
+            image: registry.redhat.io/openshift4/ose-cli:v4.12
+            command:
+            - /bin/sh
+            - -c
+            - |
+              date; echo Cleaning up the stale applications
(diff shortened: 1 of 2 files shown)


components/tekton-ci/production/cleanup-cronjob.yaml
```diff
@@ -0,0 +1,31 @@
...
+          serviceAccountName: appstudio-pipeline
+          containers:
+          - name: remove-stale-resources
+            image: registry.redhat.io/openshift4/ose-cli:v4.12
+            command:
+            - /bin/sh
+            - -c
+            - |
+              date; echo Cleaning up the stale applications
+              oc get application -n build-templates-e2e -o go-template --template '{{range .items}}{{.metadata.name}} {{.metadata.creationTimestamp}}{{"\n"}}{{end}}' | awk '$2 <= "'$(date -d'now-3 days' -Ins --utc | sed 's/+0000/Z/')'" { print $1 }' | xargs --no-run-if-empty oc delete application -n build-templates-e2e
```
chmeliik, line 20: Was it supposed to be `sed 's/+00:00/Z/'`? Or is just my `date` that formats it differently?

```bash
$ date -d'now-3 days' -Ins --utc
2024-05-25T16:03:09,745810498+00:00
```

We probably also don't want to include nanoseconds, the `metadata.creationTimestamp` doesn't include them either (i.e. use `--iso-8601=seconds` instead of `-Ins`

chmeliik (commented):
Details in the bash script, overall LGTM

Thanks :tada:

components/tekton-ci/production/cleanup-cronjob.yaml
tisutisu: Thanks for pointing them out. Addressed them now.

chmeliik (approved):
LGTM :+1:

PR #3776, opened 2024-05-23
KFLUXINFRA-581: Configuring New AWS Account for Multi Platform Build

manish-jangra:
Configuring New AWS Account for Multi-Platform Build Private Production Overlay

PR #3779, opened 2024-05-23
Revert "KFLUXINFRA-581: Configuring New AWS Account for Multi Platform Build"

hugares:
Reverts redhat-appstudio/infra-deployments#3776

arm64 builds are failing since this change was merged


hugares:
this is a revert for prod so merging right away

PR #3782, opened 2024-05-26
Router and UI proxy access for the registration service

gbenhaim:
The OCP router needs access to the host operator namespace since it has routes in it.

In addition, the proxy used by the UI needs to access the registration service.


gbenhaim:
this change is not tested by the e2e suite, merging.

PR #3783, opened 2024-05-26
Skip dry run on missing resources for sandbox host

gbenhaim:
Without it, the host crds are failed to be synced.


gbenhaim:
change are not tested in the e2e suite, merging.

PR #3784, opened 2024-05-26
sandbox-host: Allow traffic from olm namespace

gbenhaim:
The pods in the olm namespace need to access the catalog source pod which run on the toolchain-host-operator namespace.


hugares (approved):
/lgtm

PR #3787, opened 2024-05-26
Skip dry run on missing resources for backup

gbenhaim:
Without it, the backup crds are failed to be synced.


gbenhaim:

PR #3792, opened 2024-05-27
Updating Secret Path for stone-prod-p02

manish-jangra:
Updating Secret Path for stone-prod-p02


gbenhaim:
changes are not tested by e2e suite. merging.

PR #3793, opened 2024-05-27
updating secrets prefix for spi and remote secret controller

manish-jangra:
updating secrets prefix for spi and remote secret controller


gbenhaim:
the change is not tested in the e2e suite. merging.

PR #3729, opened 2024-05-16
initial version of mintmaker service

scoheb:
- This is just a simple skeleton controller such
  that we can get mintmaker integrated.

PR #3798, opened 2024-05-28
Introduce Mintmaker

gnaponie:
Mintmaker is a new service that will provide the Renovate's functionalities inside Konflux.

ref: [CLOUDWF-10108](https://issues.redhat.com/browse/CLOUDWF-10108)


components/mintmaker/base/mintmaker-config.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: mintmaker-config
```
qixiang, line 4: The controller expects a configmap named `renovate-config`.

components/mintmaker/base/mintmaker-config.yaml
```diff
@@ -0,0 +1,12 @@
...
+  namespace: mintmaker
+data:
+  config.js: |
```
qixiang, line 7: The file must be named `renovate.json` and be in JSON format. The controller will patch the following keys

- platform
- endpoint
- username
- gitAuthor

Here is an example configmap we can use at this moment: https://gist.github.com/qixiang/915389b2c26b8ca5dbe479e2a5038bb0

gnaponie: Ok, I've taken something from there, I don't think we want to keep them all based on what I've understood by stakeholders. But I've taken what I think it makes sense.

components/mintmaker/base/mintmaker-config.yaml
```diff
@@ -0,0 +1,37 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: renovate-config
+  namespace: mintmaker
+data:
+  renovate.json: |
+    module.exports = {
```
qixiang, line 8: This is not in json format, the `module.exports =` should be removed, otherwise the controller is unable to  patch this renovate config.

components/mintmaker/base/mintmaker-config.yaml
```diff
@@ -0,0 +1,37 @@
...
+      "onboarding": false,
+      "requireConfig": "ignored",
+      "platform": "github",
```
qixiang, line 11: The controller will patch the `platform` accordingly, so it's unnecessary to include this.

components/mintmaker/base/mintmaker-config.yaml
```diff
@@ -0,0 +1,37 @@
...
+      "platformCommit": true,
+      "autodiscover": false,
+      "tekton": {
```
qixiang, line 14: I think we should use the same `tekton` manager config from https://gist.github.com/qixiang/915389b2c26b8ca5dbe479e2a5038bb0, that's the exact `tekton` manager config from build-service.

gnaponie: But should we use the same? To be honest some of their configuration don't seem very reasonable to me (or we are unaware of the reasons behind it). I wouldn't keep it as it is. We can discuss about it in more details in the future if you want, but I think that this as an initial version should be ok.

qixiang: > But should we use the same?

I believe the answer is yes, once MintMaker is deployed, there are actually two Renovate, one from build-service, one from MintMaker, and both are running the Tekton manager (with same account), we should not expect different behavior between the two Renovate.

scoheb:
can you also verify that this is no longer an issue: https://github.com/redhat-appstudio/infra-deployments/pull/3729#discussion_r1604862077

components/mintmaker/base/mintmaker-config.yaml
gnaponie: These are the settings I don't like and why:
* dependencyDashboard: the dependency dashboard should be enabled in my opinion
* branchName: I don't think we should specify a different branch, also because RHEL wants to specify a branch -- perhaps they don't care about tekton, but I don't want to maintain managers that are that different
* commitBody: this should be already be added if the commit is signed, which is something we have to do
* commitMessageTopic: I dont' see why change that
* recreateWhen and rebaseWhen: I don't think we want to change that, we want to decide that when we'll have the on-demand

qixiang:
> can you also verify that this is no longer an issue: [#3729 (comment)](https://github.com/redhat-appstudio/infra-deployments/pull/3729#discussion_r1604862077)

It has been addressed in latest code. This PR is verified by deploying in development env (by `./hack/bootstrap-cluster.sh preview`), after the deployment, the controller is running well in mintmaker namespace and is able to create jobs to run Renovate commands against components.

However, there may be updates required for staging and production environments, but I believe it is acceptable to postpone these updates as we are not making much changes to these two environments in this pull request.

@scoheb could you help to review this?

scoheb:
we need to address the kube-linter errors here: https://github.com/redhat-appstudio/infra-deployments/actions/runs/9369177927/job/25842546983?pr=3798

qixiang:
Hi @scoheb thank you for reviewing, the kube-linter reported errors should be fixed by latest push, however I'm not able to trigger or approve the test.

scoheb (approved):

PR #3778, opened 2024-05-23
SRVKP-4361: enable results from sidecar logs for pipelines in prod

gabemontero:
https://github.com/redhat-appstudio/infra-deployments/pull/3699 has been running in stage for a week

will consider this after https://github.com/redhat-appstudio/infra-deployments/pull/3775 from @enarha is sorted out 

rh-pre-commit.version: 2.2.0
rh-pre-commit.check-secrets: ENABLED

PR #3818, opened 2024-05-30
chore(RHTAPWATCH-1021): enable Tekton StepActions in production

Omeramsc:
Continuing the work on #3693 , this changeenable Tekton StepActions in the production cluster.

It is a workaround which allows passing artifacts between tasks in an integration pipeline where workspaces backed by persistent volumes are not available

### DIFF
--- a/components/pipeline-service/production/stone-prd-m01/deploy.yaml
+++ b/components/pipeline-service/production/stone-prd-m01/deploy.yaml
@@ -1883,6 +1883,7 @@ spec:
     enable-git-resolver: true
     enable-hub-resolver: true
     enable-tekton-oci-bundles: true
+    enable-step-actions: true
     options:
       configMaps:
         config-logging:
--- a/components/pipeline-service/production/stone-prd-rh01/deploy.yaml
+++ b/components/pipeline-service/production/stone-prd-rh01/deploy.yaml
@@ -1883,6 +1883,7 @@ spec:
     enable-git-resolver: true
     enable-hub-resolver: true
     enable-tekton-oci-bundles: true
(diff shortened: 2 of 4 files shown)


amisstea (approved):
/lgtm

gabemontero:
@Omeramsc @amisstea I would suggest having a revert PR ready prior to his merging just in case something unexpected happens turning on this still alpha fetaure in  v0.59.0 (our current version of tekton in prod) that we did not see in stage

reminder: the support contract for alpha features is only best effort from an openshift pipelines / upstream tekton perspective

this feature does move from alpha to beta in v0.60.0 (we won't be v0.60.0 until pre-release versions of 1.16 of openshift pipelines are available to pick up ... feature freeze for that 1.16 is not until August).

as a point of comparison, we plan to have the "revert PR" ready just in case for https://github.com/redhat-appstudio/infra-deployments/pull/3778 once we finish vetting the move to that alpha feature in the arch call next week.

I'll add the approve label once I hear back from you here wrt the backup revert plan.

@hugares @gbenhaim @manish-jangra @bamachrn FYI

Omeramsc:
Thanks @gabemontero 
I added the following revert PR #3819

gabemontero:
> Thanks @gabemontero I added the following revert PR #3819

thanks folks .... I've started https://redhat-internal.slack.com/archives/C02CTEB3MMF/p1717079314200699 to nail down the particulars.  The thinking is to label this to merge but also put a hold, and then @Omeramsc can cancel the hold first thing for him Sunday morning and verify the feature is doing what is desired.

PR #3819, opened 2024-05-30
fix(RHTAPWATCH-1021): revert #3818 enable Tekton StepActions in prod

Omeramsc:
a revert PR in case anything unexpected happends while turning
  the StepAction feature on while still an alpha feature in v0.59.9
  
(  as it is built upon an unmerged change, it appears as no files were changed)

### DIFF
--- a/components/pipeline-service/production/stone-prd-m01/deploy.yaml
+++ b/components/pipeline-service/production/stone-prd-m01/deploy.yaml
@@ -1883,6 +1883,7 @@ spec:
     enable-git-resolver: true
     enable-hub-resolver: true
     enable-tekton-oci-bundles: true
+    enable-step-actions: true
     options:
       configMaps:
         config-logging:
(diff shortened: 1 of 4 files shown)


gabemontero:
> a revert PR in case anything unexpected happends while turning the StepAction feature on while still an alpha feature in v0.59.9
> 
> ( as it is built upon an unmerged change, it appears as no files were changed)

yep you'll have to rebase / do the revert after the merge of #3818

gabemontero:
@Omeramsc @amisstea best as I can tell there has been no problems reported that are related to https://github.com/redhat-appstudio/infra-deployments/pull/3818

unless you all know something I don't, I think we can close this 

next step I suppose will be if you or someone else starts modifying build-definitions to use StepActions, is that correct?

amisstea:
@Omeramsc was able to verify StepActions are working in prod so I agree we can close this.

@gabemontero I already have some StepActions to add to the build-definitions repo. I should have a PR for them sometime this week.

PR #3832, opened 2024-05-31
SRVKP-4361: enable results from sidecar logs for pipelines in prod - p01 only

gabemontero:
see https://redhat-internal.slack.com/archives/C02CTEB3MMF/p1717120576571899 

@scoheb @lcarva @stuartwdouglas 

rh-pre-commit.version: 2.2.0

### DIFF
--- a/components/pipeline-service/production/base/update-tekton-config-features.yaml
+++ b/components/pipeline-service/production/base/update-tekton-config-features.yaml
@@ -1,0 +1,12 @@
+---
+- op: add
+  path: /spec/pipeline/results-from
+  # default upstream setting
+  # value: termination-message
+  value: sidecar-logs
+
+- op: add
+  path: /spec/pipeline/max-result-size
+  # default upstream setting
+  # value: 4096
+  value: 12288
--- a/components/pipeline-service/production/base/kustomization.yaml
+++ b/components/pipeline-service/production/base/kustomization.yaml
@@ -13,6 +13,7 @@ resources:
   - ../../base/external-secrets
   - ../../base/testing
   - ../../base/rbac
+  - tekton-pipelines-controller-pods-log-access-rbac.yaml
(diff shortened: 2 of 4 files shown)


components/pipeline-service/production/base/update-tekton-config-features.yaml
```diff
@@ -0,0 +1,12 @@
...
+  path: /spec/pipeline/results-from
+  # default upstream setting
+  # value: termination-message
+  value: sidecar-logs
+
+- op: add
+  path: /spec/pipeline/max-result-size
+  # default upstream setting
+  # value: 4096
+  value: 12288
```
gabemontero, line 12: @lcarva is looking at the bootc pipelinerun logs to see if this size is sufficient

lcarva: Comparing what is in taskResults vs the container termination message, it looks like the result almost made it. Only the last image ref was cut off. The new size should be enough. Looking at the SBOM would be an extra guarantee that there aren't any other entries that are just not showing up, but I don't have access to it.

I think we should go ahead with this change. It won't make the size smaller. We may have to bump the size again, but that seems like a reasonable thing to do.

gabemontero:
/hold

@lcarva had an idea about tweaking the bootc dockerfile to work around the result size issue that @scoheb is going to try first before we merge this

scoheb:
Containerfile update to use the same registries throughout has fixed the results too large issue.

gabemontero:
per discussion on the June 4 Konflux community calls we are going to hit the pause button on enabling sidecar in prod, most likely in favor of https://issues.redhat.com/browse/STONEBLD-2326 @lcarva @ralphbean 

I'll revert the setting in stage soon

@jkhelil FYI

PR #3880, opened 2024-06-13
Fix fast pool AMI

stuartwdouglas:
There was a copy/paste error that used an AMI with the wrong arch.


chmeliik (approved):
/lgtm

hugares:
This one is affecting production and is not tested by e2e, merging now

PR #3882, opened 2024-06-13
Draft: Add OTel collector install & configuration in the staging environment

mike-kingsbury:


hugares:
The `tracing-workload-otel-collector` ApplicationSet will not be deployed on staging, you need to remove code deleting it from the overlay, here is the change  to do:
```
diff --git a/argo-cd-apps/overlays/konflux-public-staging/delete-applications.yaml b/argo-cd-apps/overlays/konflux-public-staging/delete-applications.yaml
index 47e78222..3cb9b66d 100644
--- a/argo-cd-apps/overlays/konflux-public-staging/delete-applications.yaml
+++ b/argo-cd-apps/overlays/konflux-public-staging/delete-applications.yaml
@@ -4,10 +4,3 @@ kind: ApplicationSet
 metadata:
   name: gitops
 $patch: delete
----
-# otel-collector is dev only.
-kind: ApplicationSet
-metadata:
-  name: tracing-workload-otel-collector
-$patch: delete

```

components/tracing/otel-collector/base/otel-collector.yaml
```diff
@@ -0,0 +1,56 @@
+apiVersion: argoproj.io/v1alpha1
+kind: ApplicationSet
```
hugares, line 2: I am a bit confused about the fact that we have an applicationSet here in the component directory.

The pattern is within argo-cd-apps folder, we have an applicationSet that will create Applications and each Application will target a specific cluster and the content of the Application is a folder in components where we have the k8s resources to deploy on that cluster part of that application.

mike-kingsbury: Yeah, I'm struggling a bit with the Helm chart and getting proper values passed into it.  I think I know the direction to go; the various layers make this a bit challenging.

PR #3892, opened 2024-06-18
Update renovate's configuration

gnaponie:
ref: [CWFHEALTH-2950](https://issues.redhat.com//browse/CWFHEALTH-2950)

### DIFF
--- a/components/mintmaker/base/renovate-config.yaml
+++ b/components/mintmaker/base/renovate-config.yaml
@@ -7,19 +7,14 @@ data:
   renovate.json: |
     {
       "onboarding": false,
-      "requireConfig": "ignored",
+      "requireConfig": "optional",
       "platformCommit": true,
       "autodiscover": false,
+      "enabledManagers": ["tekton", "dockerfile", "lockFileMaintenance"],
       "tekton": {
         "fileMatch": ["\\.yaml$","\\.yml$"],
         "includePaths": [".tekton/**"],
         "packageRules": [
-        {
-            "matchPackagePatterns": [
-              "[*]"
-            ],
-            "enabled": false
-          },
           {
             "matchPackagePatterns": [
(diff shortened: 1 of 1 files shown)


staticf0x (approved):
LGTM

components/mintmaker/base/renovate-config.yaml
```diff
@@ -28,8 +22,31 @@ data:
             "matchDepPatterns": [
               "^quay.io/redhat-appstudio-tekton-catalog/"
             ],
-            "semanticCommits": "enabled"
+            "groupName": "RHTAP references",
```
mmorhun, line 25: Probably we need `Konflux`

components/mintmaker/base/renovate-config.yaml
```diff
@@ -28,8 +22,31 @@ data:
...
+            "branchName": "konflux/references/{{baseBranch}}",
+            "commitBody": "Signed-off-by: {{{gitAuthor}}}",
+            "commitMessageTopic": "RHTAP references",
```
mmorhun, line 28: `Konflux`

mmorhun (approved):

components/mintmaker/base/renovate-config.yaml
```diff
@@ -7,19 +7,13 @@ data:
   renovate.json: |
```
qixiang, line 7: Will there be another PR to enable only the three managers later, after we change the Renovate image to our custom image that supports the RPM manager?

gnaponie: No, I just forgot :) I've updated this PR to have only those 3 managers.
BTW I didn't add a preset on purpose, I see our (rebuilds guild) preset might not be ok with this conf. So let's just skip it for now, I don't really see the benefit at this point. We can see if we need a preset in the future.

PR #3893, opened 2024-06-18
remove production overlay for mintmaker

scoheb:
- remove until mintmaker has been validated in staging


hugares (approved):
/lgtm

hugares:
to remove an applicationSet from prod, you need a delete path, like this one: https://github.com/redhat-appstudio/infra-deployments/blob/main/argo-cd-apps/overlays/konflux-public-production/delete-applications.yaml#L9

hugares:
this PR is not tested by e2e and we need this in, merging it right away

PR #3900, opened 2024-06-19
Update pipeline-service and enable Results DB TLS

enarha:
Enabling Results DB TLS requires both changes to both pipeline-service and infra-deployments repos to be promoted together, thus the manual PR. It also includes the latest pipeline-service updates.

Included PRs:
- https://github.com/openshift-pipelines/pipeline-service/pull/1006
- https://github.com/openshift-pipelines/pipeline-service/pull/1004
- https://github.com/openshift-pipelines/pipeline-service/pull/999


hack/secret-creator/create-plnsvc-secrets.sh
```diff
@@ -64,6 +65,48 @@ stringData:
...
+        > /dev/null
+    chmod og-rwx ".tmp/tekton-results/tls.key"
+    openssl x509 -req -text -days 7 -CAcreateserial \
+        -extfile <(printf "subjectAltName=DNS:postgres-postgresql.tekton-results.svc.cluster.local") \
+        -in ".tmp/tekton-results/tls.csr" \
+        -CA ".tmp/tekton-results/ca.crt" \
+        -CAkey ".tmp/tekton-results/ca.key" \
+        -out ".tmp/tekton-results/tls.crt" \
+        > /dev/null
+    cat ".tmp/tekton-results/ca.crt" > ".tmp/tekton-results/tekton-results-db-ca.pem"
```
avinal, line 101: Do we only need ca.crt? My understanding is to combine both ca.crt and tls.crt

avinal (commented):
One question. rest looks good to me.

hack/secret-creator/create-plnsvc-secrets.sh
enarha: The client only needs the ca.crt (which signed the server.crt), so it can verify the server.crt. From the documentation:
```server.crt and server.key should be stored on the server, and root.crt should be stored on the client so the client can verify that the server's leaf certificate was signed by its trusted root certificate. root.key should be stored offline for use in creating future certificates.```

avinal: Understood

avinal (approved):
Looks good to me

avinal (approved):
/lgtm
/approve

PR #3903, opened 2024-06-19
Add cronjob to delete old DependencyUpdateCheck CRs

qixiang:
An old DependencyUpdateCheck CR means:
1. The CR was created more than 48 hours ago
2. The CR has already been processed by the controller


components/mintmaker/base/cronjobs/delete-dependency-update-checks.yaml
```diff
@@ -0,0 +1,68 @@
+---
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: delete-dependencyupdatechecks
+  namespace: mintmaker
+spec:
+  schedule: "0 1,13 * * *"
```
gnaponie, line 8: This seems a weird cronjob schedule. Any reason behind this? Perhaps put a comments about what that means.

gnaponie (approved):

components/mintmaker/base/cronjobs/delete-dependency-update-checks.yaml
qixiang: Actually there is no special reason. The schedule `0 1,13 * * *` runs the job twice a day, at 1:00 AM and 1:00 PM. This timing was chosen to avoid the more common scheduling slots of 12:00 AM and 12:00 PM, which are often used for other cron jobs and might lead to higher system load or contention. It's totally fine to change it to once a day or just `0 */12 * * *` and etc.

gnaponie: That makes sense. Thanks for the explanation.

PR #3905, opened 2024-06-19
Add external secret for Mintmaker

qixiang:
This new external secret is the registry.redhat.io pull secret which will be used by Renovate in Mintmaker to check container images.


scoheb:
> no pipelinerun found for component rhtap-demo-component

hugares:
This change is only affecting staging overlay which is not tested by e2e, merging it

PR #3919, opened 2024-06-21
KFLUXINFRA-635: Restore s390x and add a100 node to internal-stage hostconfig

bamachrn:


stuartwdouglas (approved):

components/multi-platform-controller/staging-downstream/host-config.yaml
```diff
@@ -89,3 +89,9 @@ data:
   host.ppc1.user: "root"
   host.ppc1.secret: "internal-stage-ibm-ssh-key"
   host.ppc1.concurrency: "4"
+
+  host.ibm-gpu-intel64.address: "10.130.81.14"
+  host.ibm-gpu-intel64.platform: "linux-ibm-gpu/intel64"
```
stuartwdouglas, line 94: one minor thing, shouldn't this be amd64 not intel64?

bamachrn: its intel machine not AMD so I thought lets make it intel64.

stuartwdouglas: The architecture is called amd64 though

bamachrn: yes, let me change it

PR #3927, opened 2024-06-21
KFLUXBUGS-863: repin watcher image to mem leak version

gabemontero:
While the now third version of the mem leak and cancelled context  fixes held up better in stage by moving log storage off the reconciler thread, the ongoing grpc/http2 performance issues showed enough performance regression from the mem leak fix, coupled with no improvement with cancelled context (log storage still got interrupted) that I am resetting the watcher version to what is currently in prod.

We are actively working the prototypes to either do log retrieval in the api server, or preferrably, move S3 interaction to the watcher.

I was able to successfully test the older version of the watcher with the latest version of the api server, so I am leaving the api server at the more recent version to get its performance tuning improvements.

long term we need https://github.com/openshift-pipelines/pipeline-service/pull/1010 in place so that future pipeline service updates to not accidentally bump the version of the watcher image

I also left the enablement of pprof for the watcher, so that we can get thread dumps if necessary; per go doc enablement of pprof is safe in prod as long as you don't cpu profile

@enarha PTAL ... once this is in place, then along with https://github.com/redhat-appstudio/infra-deployments/blob/main/components/pipeline-service/production/base/kustomization.yaml#L17-L21 (though we can undo the api server pin if things look OK in stage) you should be able to move forward with TLS related changes for results without disrupting how the watcher is currently working in prod.

rh-pre-commit.version: 2.3.0

### DIFF
--- a/components/pipeline-service/development/kustomization.yaml
+++ b/components/pipeline-service/development/kustomization.yaml
@@ -12,6 +12,10 @@ resources:
   - https://github.com/openshift-pipelines/pipeline-service.git/developer/openshift/gitops/argocd/pipeline-service-storage?ref=ad974044255725077080ace5e4be8840ee85b2da
   - ../base/rbac
 
+images:
+  - name: quay.io/redhat-appstudio/tekton-results-watcher
+    newTag: bae7851ff584423503af324200f52cd28ca99116
+
 # generate a new configmap with updated values (logs api, db ssl mode) and replace the default one
 configMapGenerator:
   - behavior: replace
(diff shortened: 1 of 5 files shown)


enarha (approved):
/lgtm

gabemontero:
@enarha I've tagged for merge, but put a hold on for now while I work a couple of other items.  I'll release the hold so this merges when I can monitor stage uninterrupted and confirm results is OK afterward before I leave for PTO.

thanks

PR #3943, opened 2024-06-25
Increase memory allocation for mintmaker controller in production

qixiang:
This is same to 0de62426 which was for staging.


hugares (approved):
/lgtm

PR #3968, opened 2024-06-28
Adding tracing component to the secret store allow list

mftb:


manish-jangra:
@gbenhaim is that the only file to be edited?

components/cluster-secret-store/base/appsre-stonesoup-vault-secret-store.yaml
```diff
@@ -52,3 +52,4 @@ spec:
         - integration-service
         - ci-helper-app
         - mintmaker
+        - tracing
```
hugares, line 55: the namespace is `konflux-otel `, not `tracing`.
```suggestion
        - konflux-otel 
```

hugares:
> @gbenhaim is that the only file to be edited?

yes it is, we are only missing the `konflux-otel` namespaces to be allowed to access the `appsre-stonesoup-vault` ClusterSecretStore

components/cluster-secret-store/base/appsre-stonesoup-vault-secret-store.yaml
mftb: TYVM, fixing

mftb:
@hugares I've updated to use the proper namespace

hugares (approved):
/lgtm

hugares:
flaky test and this change is not tested by e2e, merging

PR #3973, opened 2024-06-28
feat: allow rootful MPC builds to disable entitlement mounts

ralphbean:
By default, podman will mount secrets declared in
/usr/share/containers/mounts.conf, even if those secrets do not exist on the host system.  This means that it is impossible to mount in entitlements provided by the user, because the subscription management tools will choose the empty entitlement mounted in over the creds mounted in by the user.


ralphbean:
If this works as expected, I'll return and make it consistent across the other environments.

scoheb (approved):

PR #3974, opened 2024-06-28
Fix service name for Opentelemetry Collector

mike-kingsbury:
The ApplicationSet name with the cluster name normalized creates a service URL with the cluster name embedded in it.  By overriding the fullName in the Helm chart, the same name will be used across all deployments.

### DIFF
--- a/argo-cd-apps/base/member/infra-deployments/tracing-workload-otel-collector/tracing-workload-otel-collector.yaml
+++ b/argo-cd-apps/base/member/infra-deployments/tracing-workload-otel-collector/tracing-workload-otel-collector.yaml
@@ -33,6 +33,8 @@ spec:
                 value: otel/opentelemetry-collector-k8s
               - name: "mode"
                 value: deployment
+              - name: "fullnameOverride"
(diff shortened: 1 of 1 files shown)


gbenhaim:
Do we want to use the same service name for all deployments (prod, stage, external, internal) ?

mftb:
@gbenhaim personally I don't mind it having the same name unless there are some guidelines that tells us otherwise. @mike-kingsbury what do you think?

gbenhaim:
what is the purpose of that name? how it's being used?

mike-kingsbury:
Yes, it’s intentional to have the same service name across all the
clusters; it’s only being accessed by services within the cluster.

mike-kingsbury:
Its the Openshift service name for the Opentelemetry collector.  Any
service within that Openshift cluster can send trace data to that
endpoint.  The configuration of the Opentelemetry collector contains
settings regarding dev/prod/stage, there's no need for it within the
service name.  By keeping the name consistent across all the clusters and
deployments, it simplifies its usage by other services.

PR #3996, opened 2024-07-02
Patch the tracing ApplicationSets to transform stage to prod

mike-kingsbury:
This will patch the otel collector and tracing ApplicationSets to transform the environment from staging to prod.


hugares (approved):
/lgtm

PR #4005, opened 2024-07-04
Patch RENOVATE_IMAGE env variable

gnaponie:
For testing purposes, the staging environment should use the `latest` mintmaker-renovate-image, and production should use a specific digest instead, so we can make sure that the latest changes are well tested before going into production.

ref: [CWFHEALTH-3100](https://issues.redhat.com//browse/CWFHEALTH-3100)

### DIFF
--- a/components/mintmaker/production/base/manager_patches.yaml
+++ b/components/mintmaker/production/base/manager_patches.yaml
@@ -15,3 +15,6 @@ spec:
           requests:
             cpu: 100m
             memory: 256Mi
+        env:
+        - name: RENOVATE_IMAGE
+          value: "quay.io/konflux-ci/mintmaker-renovate-image:40af057"
--- a/components/mintmaker/production/base/kustomization.yaml
+++ b/components/mintmaker/production/base/kustomization.yaml
@@ -22,4 +22,4 @@ patches:
       kind: ExternalSecret
       group: external-secrets.io
       version: v1beta1
-  - path: manager_resources_patch.yaml
+  - path: manager_patches.yaml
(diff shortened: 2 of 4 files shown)


gnaponie:
@qixiang since this was a small patch I thought it could be added to the other existing one. Not sure if that's a good practice. It's my first patch in kustomize. Anyhow, I did test this with `kustomize build` and the patch works.

components/mintmaker/production/base/manager_patches.yaml
```diff
@@ -15,3 +15,6 @@ spec:
           requests:
             cpu: 100m
             memory: 256Mi
+        env:
+        - name: RENOVATE_IMAGE
+          value: "quay.io/konflux-ci/mintmaker-renovate-image:067c096e9e0e7de05041eb0fe27f329d8c1c993cb8c73e00637981d299408b2d"
```
qixiang, line 20: The current image being used in production is a newer one (`40af0570a79be63de3487670e1d6172ba00b7dcc`) released yesterday. To improve traceability and ease of reference, I recommend using the git commit hash tag instead of the SHA digest hash. The git commit hash tag directly indicates which commit the image corresponds to, making it more intuitive to track changes and debug issues.

qixiang (changes requested):

components/mintmaker/production/base/manager_patches.yaml
scoheb: I agree

scoheb:
looks good. I would just make the change that @qixiang recommended.

gnaponie:
Thanks for the review. Suggestions applied.
@qixiang please take another look

qixiang (approved):
/approve
/lgtm

PR #4009, opened 2024-07-05
Disable automatic approval on stone-prod-p01

hugares:
The plan is to decommission this cluster so disable automatic approval to prevent new users from coming in.

[KFLUXINFRA-648](https://issues.redhat.com//browse/KFLUXINFRA-648)


manish-jangra (approved):
looks good

PR #4021, opened 2024-07-08
Update SPI to the next OAuth-less version on production

mshaposhnik:
Update SPI to the version without OAuth being deployed


hugares (approved):
/lgtm

PR #4052, opened 2024-07-11
fix: correct ReleasePlan ignoreDifferences

scoheb:
- This will prevent Argo from constantly trying to Sync since the Release Service wants to update that value


hugares (approved):
/lgtm

hugares:
This is not tested by e2e tests, merging

PR #4061, opened 2024-07-12
workspaces: fix namespace replacement

filariow:


components/workspaces/staging/kustomization.yaml
```diff
@@ -1,6 +1,5 @@
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
-namespace: konflux-workspaces
```
hugares, line 3: For my own understanding, the namespace we want most of resource is "workspaces-system", is that correct?

filariow: Right, most of the resources will be there. At the same time we need to deploy some Roles and RoleBinding in toolchain-host-operator.

hugares (approved):
/lgtm

PR #4074, opened 2024-07-15
Fixing issue with TemplateTier metadata

dperaza4dustbit:
Removing the creationtimestamp field causing synch issue with TemplateTiers


hugares (approved):
/lgtm

PR #4085, opened 2024-07-16
Improve nginx deployment

gbenhaim:
1. For 60 seconds, check if no containers in the pod crash loop and only then declare that the pod ready.

2. Add readiness probe. Traffic will be forwarded only to pods that pass the readiness check.

3. Adjust probes to start only after 60 seconds to match the change mention in 1.

4. During upgrade, allow an additional pod to be scheduled for the new replicaset.

5. Don't allow the deployment to have less that 3 pod ready, even during upgrade.

### DIFF
--- a/components/ui/base/proxy/proxy.yaml
+++ b/components/ui/base/proxy/proxy.yaml
@@ -12,10 +12,11 @@ spec:
   selector:
     matchLabels:
       app: proxy
+  minReadySeconds: 60
   strategy:
     rollingUpdate:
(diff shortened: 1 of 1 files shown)


filariow (approved):
Nice!

sadlerap (approved):
Looks good to me :tada:

hugares (approved):
/lgtm

dperaza4dustbit (approved):
/lgtm

PR #4090, opened 2024-07-17
promote hac-dev to prod

sahil143:


components/ui/production/stone-prod-p02/kustomization.yaml
hugares, line 1: let's change p01 also, until we decommission it

sahil143: Done. Updated p01 as well.

PR #4101, opened 2024-07-17
Update mintmaker-renovate-image digest

gnaponie:


gnaponie:
@qixiang @staticf0x @nvtnlucie @querti PTAL

PR #4120, opened 2024-07-18
Fix synch loop in staging for TierTemplates

dperaza4dustbit:
Making sure TierTemplates have spec.template.metadata is set to empty object {}. This should fix the constant sync loop in our staging environments


hugares (approved):
/lgtm

PR #4126, opened 2024-07-18
Disable automated sync for toolchain-member-operator

hugares:
There is an issue in prod were NS was deleted and we do not want ArgoCD to automatically sync to allow us to restore things.

### DIFF
--- a/argo-cd-apps/base/toolchain-member/toolchain-member-operator.yaml
+++ b/argo-cd-apps/base/toolchain-member/toolchain-member-operator.yaml
@@ -30,15 +30,3 @@ spec:
       destination:
         namespace: toolchain-member-operator
         server: '{{server}}'
-      syncPolicy:
-        automated:
-          prune: true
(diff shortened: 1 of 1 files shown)


argo-cd-apps/base/toolchain-member/toolchain-member-operator.yaml
```diff
@@ -30,12 +30,6 @@ spec:
       destination:
         namespace: toolchain-member-operator
         server: '{{server}}'
-      syncPolicy:
-        automated:
-          prune: true
-          selfHeal: true
-        syncOptions:
-          - CreateNamespace=false
         retry:
```
amisstea, line 33: This looks like it needs to be removed as well.

hugares: Done

amisstea (approved):

PR #4132, opened 2024-07-19
Disable automated pruning for toolchain operators

hugares:
We had an outage due to a back overlay change. The overlay change resulted in deleting the toolchain member applications thus all resources in toolchain-member-operator namespace. We were lucky that we recovered without data loss.

Since toolchain host and member namespaces contains user registration CRs that are not deployed by ArgoCD, disable the automated pruning to prevent accidental deletion. From now on, if a resource is deleted from those application, someone will need to manually start the sync with prune option after reviewing the resources to delete.


amisstea (approved):

arewm (approved):
lgtm

filariow (approved):

gbenhaim (approved):

PR #4144, opened 2024-07-22
feat(KONFLUX-3796): Test arm64 multi-arch builds with lots of VMs

jhutar:
Check if we can support this many of concurrent multi-arch builds.


ifireball (approved):
lgtm

PR #4146, opened 2024-07-22
Make gh tenants-config argo sync manual

ernesgonzalez33:
Draft PR set up as a start of the tenants-config migration to GitLab. Here I'm making the ArgoCD sync manual before making the GH read-only


gbenhaim:
This change isn't being tested by the e2e tests.

PR #4160, opened 2024-07-23
Increase tekton-pipelines-remote-resolvers mem limit

gbenhaim:
We see spikes in production that reaches the 4Gi limit.

### DIFF
--- a/components/pipeline-service/production/base/update-tekton-config-performance.yaml
+++ b/components/pipeline-service/production/base/update-tekton-config-performance.yaml
@@ -31,10 +31,23 @@
   path: /spec/pipeline/options/deployments/tekton-operator-proxy-webhook/spec/replicas
   # default pipeline-service setting is 1
   value: 2
-- op: replace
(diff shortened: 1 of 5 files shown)


components/pipeline-service/production/base/update-tekton-config-performance.yaml
```diff
@@ -31,10 +31,23 @@
...
-- op: replace
-  path: /spec/pipeline/options/deployments/tekton-pipelines-remote-resolvers/spec/replicas
-  # default pipeline-service setting is 1
-  value: 2
+- op: add
+  path: /spec/pipeline/options/deployments/tekton-pipelines-remote-resolvers/spec
+  # replicas - default pipeline-service setting is 1
+  # memory limit - default upstream setting is 4Gi
+  value:
+    replicas: 2
```
gabemontero, line 39: ooh ... didn't know you could patch this way with the `op: add` gitops approcah ... cool

gabemontero (approved):
/lgtm

PR #4163, opened 2024-07-23
workspaces: specify host for route

sadlerap:
This fixes a out-of-sync loop between what ArgoCD wants (an empty host) and what the route controller wants (a populated .spec.host field).


filariow (approved):

sadlerap:
@filariow I think I need a `/lgtm` as well for this to merge.  Approvals don't auto-lgtm on this repository.

PR #4171, opened 2024-07-24
enable OSP pruner again for resolver memory issues and results prune performance issues  seen on prod-rh01

gabemontero:
@enarha @gbenhaim @ralphbean PTAL

rh-pre-commit.version: 2.3.0

### DIFF
--- a/components/pipeline-service/production/base/update-tekton-config-performance.yaml
+++ b/components/pipeline-service/production/base/update-tekton-config-performance.yaml
@@ -31,6 +31,14 @@
   path: /spec/pipeline/options/deployments/tekton-operator-proxy-webhook/spec/replicas
   # default pipeline-service setting is 1
   value: 2
+- op: replace
+  path: /spec/pruner
+  value:
+    keep: 10
+    keep-since: 60
+    resources:
+      - pipelinerun
+    schedule: "*/15 * * * *"
 - op: add
   path: /spec/pipeline/options/deployments/tekton-pipelines-remote-resolvers/spec
   # replicas - default pipeline-service setting is 1
--- a/components/pipeline-service/production/stone-prd-m01/deploy.yaml
+++ b/components/pipeline-service/production/stone-prd-m01/deploy.yaml
@@ -2162,7 +2162,11 @@ spec:
             namespace }}/pipelinerun/{{ pr }}/logs/{{ task }}
   profile: all
   pruner:
-    disabled: true
+    keep: 10
+    keep-since: 60
+    resources:
+    - pipelinerun
+    schedule: '*/15 * * * *'
   targetNamespace: openshift-pipelines
   trigger:
     options:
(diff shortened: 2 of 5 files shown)


components/pipeline-service/production/base/update-tekton-config-performance.yaml
```diff
@@ -31,6 +31,13 @@
   path: /spec/pipeline/options/deployments/tekton-operator-proxy-webhook/spec/replicas
   # default pipeline-service setting is 1
   value: 2
+- op: replace
+  path: /spec/pruner
+  value:
+    keep: 10
+    resources:
+      - pipelinerun
+    schedule: 0 * * * *
```
gabemontero, line 40: should be once an hour per https://crontab.guru/every-1-hourhttps://crontab.guru/every-1-hour

gabemontero:
`keep-since` added @enarha 

that said, I'm not sure if the e2e's ran last time ... will try to launch manually

/test appstudio-e2e-tests

gabemontero:
is is running now

components/pipeline-service/production/base/update-tekton-config-performance.yaml
```diff
@@ -31,6 +31,14 @@
   path: /spec/pipeline/options/deployments/tekton-operator-proxy-webhook/spec/replicas
   # default pipeline-service setting is 1
   value: 2
+- op: replace
+  path: /spec/pruner
+  value:
+    keep: 10
+    keep-since: 15
```
enarha, line 38: I suggest we run the pruner more often (every 10 or 15 mins for example) and then set the keep-since to 60 (minutes). This way when the watcher pruning is lagging, we won't prune prs which finished soon after x:15.
The schedule will look like `schedule: */15 * * * *` and `keep-since: 60`. WDYT?

gabemontero: my yaml editor does not seem to like `*/15 * * * *` but both it and crontab.guru seem to abide `0/15 * * * *`

that also lines up with what Romain used to do i.e. https://github.com/openshift-pipelines/pipeline-service/commit/962f485476bb1c75a8f9f6a743a0cb953e587fbc#diff-7b0f46511f3c76f0deb347cc69126f54ac59102bee06eeee07471cf9fa02b8bdR58

I like your suggestion though ... you good with that slight adjustment @enarha ?

gabemontero: update pushed in the interim

enarha: The config is good, but I think you forgot to regenerate the yamls.

enarha: Not sure if it matters, but I see all examples putting the value in quotes "*/15 * * * *", so maybe add that as well.

gabemontero: both updates (gen deploy and quotes) implemented and pushed @enarha PTAL

gabemontero:
tests ran successfully .... details in https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/redhat-appstudio_infra-deployments/4171/pull-ci-redhat-appstudio-infra-deployments-main-appstudio-e2e-tests/1816154072160407552/artifacts/appstudio-e2e-tests/redhat-appstudio-e2e/build-log.txt

reminder pod logs are not collected on passing tests

/hold cancel

PR #4184, opened 2024-07-25
enable OSP pruner again

enarha:
enable OSP pruner again for resolver memory issues and results prune performance issues seen on prod-rh01 (#4171). The pruner is set to run every 10 minutes, pruning pipelineruns older than 60 minutes. It serves as a backup pruner for the Tekton Results pruner which ideally will be pruning the pipelineruns earlier up to 10 mins after completion.

This is a temporary measure until the Results pruner performance is improved.

### DIFF
--- a/components/pipeline-service/production/base/update-tekton-config-performance.yaml
+++ b/components/pipeline-service/production/base/update-tekton-config-performance.yaml
@@ -31,6 +31,14 @@
   path: /spec/pipeline/options/deployments/tekton-operator-proxy-webhook/spec/replicas
   # default pipeline-service setting is 1
   value: 2
+- op: replace
+  path: /spec/pruner
+  value:
+    disabled: false
+    keep-since: 60
+    resources:
+      - pipelinerun
+    schedule: "*/10 * * * *"
 - op: add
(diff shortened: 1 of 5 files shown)


gbenhaim (commented):
/approve
/lgtm

enarha:
The appstudio-e2e-tests seems to be broken. Not sure if it is disabled explaining why it does not start automatically. The error message is ```cat: /usr/local/konflux-ci-secrets-new/redhat-appstudio-qe/exodus_non_prod_key: No such file or directory``` and as a result the Konflux deployment does not even start on the testing cluster. @gbenhaim I believe we can just force merge the PR. I tested the configuration manually.

gabemontero (approved):

dheerajodha:
> The appstudio-e2e-tests seems to be broken. Not sure if it is disabled explaining why it does not start automatically. The error message is cat: /usr/local/konflux-ci-secrets-new/redhat-appstudio-qe/exodus_non_prod_key: No such file or directory and as a result the Konflux deployment does not even start on the testing cluster.

Yes, that was a blocker bug, thanks for noting. Fixed by [KFLUXBUGS-1487](https://issues.redhat.com/browse/KFLUXBUGS-1487)

PR #4189, opened 2024-07-25
Add konflux-eaas kubesaw tier

amisstea:
The user should only be allowed to create select resources (ClusterTemplateInstance, Secrets) in the provisioned namespace.


amisstea:
@dperaza4dustbit please review the new `konflux-eaas` tier. I'm also looking for some guidance about the rest.

I didn't intend these change but `ksctl generate nstemplatetiers` forces them on me:
- `metadata.creationTimestamp` is set to null on all generated resources. Perhaps this is a recent change with `ksctl` but they should be non-harmful.
- A new admin TierTemplate was created for the `appstudio` and `appstudiolarge` tiers. The previously generated versions lacked some RBAC rules for roles, rolebindings, serviceaccounts, and serviceaccount/tokens. I haven't removed the older admin TierTemplate considering I'd rather not touch these files to begin with. What should be done about this?

I'm running the latest `ksctl` version.
```
> ksctl --version
ksctl version commit: 'c6809574a6394a38284860ddd88999e2e30b073c-dirty', build time: '2024-07-25T18:58:24Z
```

dperaza4dustbit (commented):
I would not expect changes to the other tiers since the only change in source is on a new tier being defined. Since you don't have these changes to ksctl it is adding that null createTimestamp. https://github.com/kubesaw/ksctl/pull/50. I would recommend to build ksctl with this changes until I can merge the changes in ksctl.

amisstea:
@dperaza4dustbit @sadlerap @filariow please review. The issues with `ksctl` were resolved with #4190 (although I did manually delete the `null` creationTimestamps). Thanks for fixing that!

sadlerap (approved):

components/sandbox/tiers/staging/konflux-eaas/tiertemplate-konflux-eaas-clusterresources-222461918-222461918.yaml
```diff
@@ -0,0 +1,31 @@
...
+metadata:
+  name: konflux-eaas-clusterresources-222461918-222461918
+  namespace: toolchain-host-operator
+spec:
+  revision: 222461918-222461918
+  template:
+    apiVersion: template.openshift.io/v1
+    kind: Template
+    metadata:
+      creationTimestamp: null
```
dperaza4dustbit, line 16: This is still not good, you will need to use my changes to ksctl here https://github.com/kubesaw/ksctl/pull/50/files

components/sandbox/tiers/src/konflux-eaas/cluster.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: template.openshift.io/v1
+kind: Template
+metadata:
+  name: konflux-eaas-cluster-resources
```
dperaza4dustbit, line 4: Metadata should have empty object like in here https://github.com/redhat-appstudio/infra-deployments/blob/main/components/sandbox/tiers/src/appstudio/cluster.yaml#L3

dperaza4dustbit (changes requested):

components/sandbox/tiers/staging/konflux-eaas/tiertemplate-konflux-eaas-clusterresources-222461918-222461918.yaml
amisstea: That worked, thanks!

components/sandbox/tiers/src/konflux-eaas/cluster.yaml
amisstea: Fixed.

dperaza4dustbit (approved):
/lgtm

PR #4205, opened 2024-07-29
workspaces: correctly override image references

sadlerap:
In our top-level kustomize manifest for each staging environment, we were referring to the image name embedded in the deployment manifests directly, not the name that would be replacing it after processing by lower-level kustomize manifests.


sadlerap:
looks like prow is down again...

PR #4214, opened 2024-07-30
Enable EaaS in public staging

amisstea:


components/cluster-as-a-service/staging/external-secrets.yaml
```diff
@@ -0,0 +1,56 @@
+apiVersion: external-secrets.io/v1beta1
+kind: ExternalSecret
+metadata:
+  name: cluster-as-a-service-hypershift-credentials
+  namespace: clusters
+spec:
+  dataFrom:
+  - extract:
+      key: staging/platform/terraform/it-cloud-aws-konflux-preprod-eaas01
```
amisstea, line 9: @manish-jangra please review. I don't have access to this folder in vault so I can't verify some things like the key here or the ones referenced in `spec.target.template.data`.

components/cluster-as-a-service/staging/external-secrets.yaml
```diff
@@ -0,0 +1,56 @@
...
+  - extract:
+      key: staging/eaas/konflux-eaas-stage
+  refreshInterval: 5m
+  secretStoreRef:
+    kind: ClusterSecretStore
+    name: appsre-stonesoup-vault
```
amisstea, line 15: Should this be `appsre-vault` instead? We seem to have both in use.

components/cluster-as-a-service/staging/external-secrets.yaml
manish-jangra: vault path looks good

components/cluster-as-a-service/staging/external-secrets.yaml
manish-jangra: I don't think that makes any difference.

manish-jangra (approved):
looks good

PR #4221, opened 2024-07-30
Fix MintMaker tag digest

staticf0x:
For some reason the `c56e548` tag disappeared from quay?


staticf0x:
PTAL @qixiang @gnaponie

qixiang (approved):
I don't see the `c56e548` commit in mintmaker-renovate-image repo, so that digest was used by mistake.

PR #4153, opened 2024-07-22
Move staging eaas SpaceProvisionerConfig to public deployment

amisstea:

PR #4223, opened 2024-07-30
Add prod SpaceProvisionerConfig for EaaS

amisstea:


amisstea:
@dperaza4dustbit @sadlerap please review. This is the prod equivalent of #4153.

sadlerap (approved):
As far as I can tell, this is good to go, though I'd like a second pair of eyes from @dperaza4dustbit.

components/sandbox/toolchain-host-operator/production/stone-prd-host1/space-provisioner-configs.yaml
```diff
@@ -28,3 +28,19 @@ spec:
...
+apiVersion: toolchain.dev.openshift.com/v1alpha1
+kind: SpaceProvisionerConfig
+metadata:
+  name: member-kflux-prd-es01.1ion.p1.openshiftapps.com
+  namespace: toolchain-host-operator
+  annotations:
+    argocd.argoproj.io/sync-options: SkipDryRunOnMissingResource=true
+spec:
+  toolchainCluster: member-kflux-prd-es01.1ion.p1.openshiftapps.com
+  enabled: true
```
dperaza4dustbit, line 41: you don't want this enabled, as it will then be used when user is signed up, set to false. Since you are going to specify the member on the SpaceRequest it will pick the EaaS member only for space request which is what I think you need here

dperaza4dustbit (changes requested):

components/sandbox/toolchain-host-operator/production/stone-prd-host1/space-provisioner-configs.yaml
amisstea: Good to know! I've corrected it here as well as for staging.

dperaza4dustbit (approved):
/lgtm

PR #4224, opened 2024-07-30
workspaces: fix outdated namespace reference

sadlerap:
Workspaces is deployed in the `workspaces-system` namespace, not the `konflux-workspaces` namespace.  Update argocd's reference so that it doesn't have to manage a namespace that will never be populated.


filariow (approved):

dperaza4dustbit (approved):
/lgtm

hugares (approved):
/lgtm

PR #4225, opened 2024-07-30
workspaces: add production configuration

sadlerap:


sadlerap:
/unhold

This should be good to review.

sadlerap:
Tests are failing due to an issue with toolchain-e2e, see #4324 for context.

components/workspaces/production/stone-prd-host1/kustomization.yaml
```diff
@@ -0,0 +1,16 @@
...
+images:
+- name: quay.io/konflux-workspaces/workspaces-server
+  newTag: v0.1.0-alpha5
+- name: quay.io/konflux-workspaces/workspaces-operator
+  newTag: v0.1.0-alpha5
+
+configMapGenerator:
+- behavior: merge
+  literals:
+  - log.level=-4
```
filariow, line 15: -4 may be too low, I'd suggest `0` (https://pkg.go.dev/log/slog#Level)

components/workspaces/production/stone-prod-p01/kustomization.yaml
```diff
@@ -0,0 +1,15 @@
...
+images:
+- name: quay.io/konflux-workspaces/workspaces-server
+  newTag: v0.1.0-alpha5
+- name: quay.io/konflux-workspaces/workspaces-operator
+  newTag: v0.1.0-alpha5
+
+configMapGenerator:
+- behavior: merge
+  literals:
+  - log.level=-4
```
filariow, line 14: ditto

sadlerap:
/hold

Wait for #4224 to merge, since it might delete/recreate all resources.

components/workspaces/production/stone-prd-host1/kustomization.yaml
sadlerap: Fixed in 886ad32.

components/workspaces/production/stone-prod-p01/kustomization.yaml
sadlerap: Fixed in 886ad32.

components/workspaces/production/stone-prod-p01/kustomization.yaml
filariow, line 1: @hugares @manish-jangra we need to support this until full decommission, right?

components/ui/base/proxy/nginx.conf
```diff
@@ -58,6 +58,11 @@ http {
             proxy_read_timeout 30m;
         }
 
+        location /apis/workspaces.konflux-ci.dev/ {
```
filariow, line 61: shouldn't it be under `/api/k8s/`?

```suggestion
        location /api/k8s/apis/workspaces.konflux-ci.dev/ {
```

sadlerap: yes, you're right, we should align with how the other clusters are configured.

sadlerap: Had to do some path rewriting in nginx, but I've fixed it in a953f202.

sadlerap:
/unhold

#4224 has merged without any issue.

components/workspaces/production/stone-prod-p01/kustomization.yaml
hugares: yes

filariow (approved):

sadlerap:
/hold
Merge this on Monday; let's not break production before a weekend.

hugares (approved):
/lgtm

PR #4237, opened 2024-07-31
Fix ExternalSecret labels

amisstea:
Labels inside `spec.target.template` are ignored by the API causing the Application to stay out of sync. Labels inside `metadata` should be copied to the resulting Secret.


hmariset (approved):
/lgtm

PR #4238, opened 2024-07-31
chore(RHTAPWATCH-1188): remove tenants-config apps

scoheb:
- we no longer need to monitor and watch the GH tenants-config repositories.
- these resources were migrated to Gitlab.

Requires on https://gitlab.cee.redhat.com/releng/konflux-release-data/-/merge_requests/1033


scoheb:
> @scoheb how can we be sure this will not trigger a delete on cascade of the resources? Because that would delete everything and the other argo would recreate everything and then a lot of components will get new builds at the same time

I will research this.

in the meantime, I have set this PR to be a draft.

scoheb:
not the correct approach. will open a new PR

PR #4241, opened 2024-08-01
feat(KFLUXINFRA-739): freeze MPC version in PROD

ifireball:
Change production overlays for multi-platform-controller to specify
a fixed version so that a maunal update would be needed in ordr to push
changes to the production clusters.

### DIFF
--- a/components/multi-platform-controller/production-downstream/kustomization.yaml
+++ b/components/multi-platform-controller/production-downstream/kustomization.yaml
@@ -1,8 +1,17 @@
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
+
+namespace: multi-platform-controller
+
 resources:
 - ../base
 - host-config.yaml
 - external-secrets.yaml
 
-namespace: multi-platform-controller
+images:
+- name: multi-platform-controller
+  newName: quay.io/konflux-ci/multi-platform-controller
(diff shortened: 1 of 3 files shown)


components/multi-platform-controller/production-downstream/kustomization.yaml
```diff
@@ -1,8 +1,17 @@
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
+
+namespace: multi-platform-controller
+
 resources:
```
mshaposhnik, line 6: I propose to make the deploy resources tag distinctive too, that will allow to test changes on staging/development

ifireball: yeah... I meant to do that... but it seems it going to be ugly... hold on...

ifireball: Please check now

mshaposhnik (approved):

arewm (approved):
How was the version determined previously?

ifireball:
> How was the version determined previously?

The release pipeline is updating the `base` setup on every merge to the MPC repo. Since the production environment did not previously have overlays overriding it, the MPC version was simply being update automatically in all environments at once.

PR #4251, opened 2024-08-02
Enable kflux-stg-es01 SpaceProvisionerConfig

amisstea:
SpaceRequests cannot be provisioned unless this is enabled. The default targetClusterRole for user signup is tenant so the placementRoles on the SpaceProvisionerConfig should prevent it from being undesireably used for that flow.


amisstea:
@dperaza4dustbit @filariow @sadlerap @MatousJobanek please review and confirm the user signup behavior. According to [this](https://github.com/codeready-toolchain/host-operator/blob/9fee36cd8f1d1d10b03ba02a312de6ccac5c4a73/pkg/space/space.go#L29), new spaces for user signup always use the tenant placement role.

amisstea:
FWIW this was already enabled previously in #4153 and there were no reported issues in staging. It was only disabled recently in #4223 out of caution.

dperaza4dustbit (approved):
/lgtm

dperaza4dustbit:
I will do some additional testing in staging to make sure now workspace ends up there for signup. For the production PR will like feedback from the KubeSaw.

PR #4255, opened 2024-08-04
Allow ExternalSecrets in namespaces konflux-qe-team-tenant and rhtap-shared-tenant

flacatus:
RHTAP and Konflux QE store secrets in vault to run in multiple tenants e2e tests.

### DIFF
--- a/components/cluster-secret-store/base/appsre-stonesoup-vault-secret-store.yaml
+++ b/components/cluster-secret-store/base/appsre-stonesoup-vault-secret-store.yaml
@@ -55,3 +55,5 @@ spec:
         - konflux-otel
         - local-cluster
         - clusters
+        - konflux-qe-team-tenant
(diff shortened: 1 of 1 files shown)


components/cluster-secret-store/base/appsre-stonesoup-vault-secret-store.yaml
```diff
@@ -55,3 +55,5 @@ spec:
         - konflux-otel
         - local-cluster
         - clusters
+        - konflux-qe-team-tenant
```
manish-jangra, line 58: What kind of secrets will these two namespaces retrieve from the Hashicorp Vault?

flacatus: @manish-jangra will retrieve our qe secrets to run tests

PR #4295, opened 2024-08-07
workspaces: enable public-viewer support in public staging

filariow:
These changes enable support for public-viewer in public staging even if the feature is not completely merged in KubeSaw. This will avoid the loop in which KubeSaw host-operator deletes the SpaceBinding for the Public-Viewer user and the Konflux-Workspaces operator is creating it back.

RegistrationService is not going to be impacted by this change until https://github.com/codeready-toolchain/registration-service/pull/453 merges.


dperaza4dustbit (approved):
/lgtm

PR #4300, opened 2024-08-09
Increase staging limitRange CPU requests

hugares:
This is a temporary change to reduce node over commitment by increasing default CPU requests. Nodes are over committed by a factor of 10 to 20 so increase request by a factor of 10 to start with. Further adjustment might be done in follow up changes depending of the effect.

[KFLUXINFRA-770](https://issues.redhat.com//browse/KFLUXINFRA-770)


components/sandbox/tiers/staging/appstudio/nstemplatetier-appstudio.yaml
```diff
@@ -11,7 +11,7 @@ spec:
   clusterResources:
     templateRef: appstudio-clusterresources-809836689-809836689
   namespaces:
-  - templateRef: appstudio-tenant-199961605-199961605
```
hugares, line 14: @dperaza4dustbit Should I delete file `appstudio-tenant-199961605-199961605` as part of this PR or I must do it in a follow up PR?

dperaza4dustbit: > @dperaza4dustbit Should I delete file `appstudio-tenant-199961605-199961605` as part of this PR or I must do it in a follow up PR?

You should not change what ksctl generates, the previous objects are used by KubeSaw controller to determine if there are resources that have to be deleted. It does a diff between the new tier and the old tier. I would not remove the old tier. At some point we can send a cleanup pr removing all the old Tiers only without touching the current tiertemplates reference by the tier object

jhutar (approved):
/lgtm

PR #4301, opened 2024-08-09
Increase productionstaging limitRange CPU requests

hugares:
This is a temporary change to reduce node over commitment by increasing default CPU requests. Nodes are over committed by a factor of 10 to 20 so increase request by a factor of 10 to start with. Further adjustment might be done in follow up changes depending of the effect.

KFLUXINFRA-770


hugares:
This change is a patch to improve performance issue in production, merging now

PR #4312, opened 2024-08-12
PVCORE-4435 - Move otel image to Quay.  Pin versions of image and chart

mike-kingsbury:
- Moves the image to quay.io and pins it to a specific tag.
- Pins the chart to a specific version to avoid mismatches in the future.


hugares (approved):
/lgtm

PR #4321, opened 2024-08-13
Remove deployment related quotas from staging

hugares:
We no longer support users creating deployments in Konflux so no need to keep the deployment related quotas. The main one we want to get rid of is the pod count which can block build pipelines but delete all the deployment related ones since they are no longer needed.


gbenhaim:
I would like David to review as well

dperaza4dustbit (approved):
/lgtm

PR #4323, opened 2024-08-13
Disable autogeneration of TektonConfig

enarha:
When deploying OSP, the operator generates a default TektonConfig is one does not already exist. This creates a race condition with the TektonConfig managed in this repo and deployed on the cluster by ArgoCD. Setting the AUTOINSTALL_COMPONENTS=false variable in the OSP Subscription, prevents the generation of the TektonConfig, so always the one from this repo is used.


gabemontero (approved):
/lgtm

PR #1030, opened 2022-12-12
Added fluent bit logging setup for appstudio-stage

bamachrn:
This PR
* Adds fluent-bit daemonsets to the cluster
* Fluent-bit is deployed with hostpath volume mounts for tracking the logs in the nodes
* This forwards the logs to the AWSCloudwatch `us-east-2 ` region 

Signed-off-by: Bama Charan Kundu <bamachrn@gmail.com>

PR #4324, opened 2024-08-13
install ksctl during kubesaw deployment

sadlerap:
With [PR #1030] landing in toolchain-e2e, tests now will try to pair an open PR against toolchain-e2e with an open PR against [ksctl].  We're always going to run tests with the latest version of ksctl, so we should install it before deploying member-operator.

[PR #1030]: https://github.com/codeready-toolchain/toolchain-e2e/pull/1030
[ksctl]: https://github.com/kubesaw/ksctl/

### DIFF
--- a/hack/sandbox-development-mode.sh
+++ b/hack/sandbox-development-mode.sh
@@ -10,4 +10,5 @@ echo
 echo "Installing the Toolchain (Sandbox) operators in dev environment:"
 rm -rf ${TOOLCHAIN_E2E_TEMP_DIR} 2>/dev/null || true
 git clone --depth=1 https://github.com/codeready-toolchain/toolchain-e2e.git ${TOOLCHAIN_E2E_TEMP_DIR}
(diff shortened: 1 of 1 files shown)


lcarva (approved):

dperaza4dustbit (commented):
/lgtm

sadlerap:
First attempt was failing since it appears that `KSCTL_BIN_DIR` is only set if `USE_INSTALLED_KSCTL` is false.  By invoking `go install` directly, `ksctl` should be made accessible via the path.

sadlerap:
/test appstudio-hac-e2e-tests

This test shouldn't be affected by the changes in this PR, not sure why they flaked.

Dannyb48:
We can ignore the upgrade-tests. Those won't pass because we always start from main and main is where the problem is. So the upgrade test won't be fixed until this fix is merged. Plus we are planning to re-vamp the upgrade tests.

https://github.com/konflux-ci/e2e-tests/blob/ef7f3a7b3f5e1e495ecf415f151a5b573d633109/magefiles/magefile.go#L1274

MatousJobanek:
@sadlerap I didn't realize the consequence of the change on your test suite, sorry about that. 
This should fix it since you use the "latest" flag. https://github.com/codeready-toolchain/toolchain-e2e/pull/1034 after this PR, the pairing shouldn't be executed in your test suite and it will always use the latest version from master.

PR #4328, opened 2024-08-14
feat(KFLUXINFRA-741): PROD: MPC w/ "local" support

ifireball:
Bump the **production** version version of multi-platform-controller.
The major change included in this version is the support for "local" platforms.

To actually use the new feature, an additional change for configuring "local" platforms would be needed.


arewm (approved):

ifireball (commented):
/retest

PR #4333, opened 2024-08-14
KFLUXINFRA-786: Add resource pool for ppc64le in public prod

bamachrn:
This PR adds PPC64LE nodes reducing their concurrecy. This is to handle the issue of pods not getting assigned to the nodes.


hugares:
@bamachrn Thanks!

PR #4329, opened 2024-08-14
chore(KFLUXINFRA-743) Prod local platforms

ifireball:
Adding a set of MPC platforms that use the "local" definitions.
Also adding "mlarge" platforms as a migration path away from using MPC with generic platform names as a way to run on a VM.

This change would require #4328 to work.

PR #4347, opened 2024-08-15
Fix IBM a100 ssh key

hugares:
This VM has both stage and prod ssh key but the stage one is not available as a secret in the prod cluster.

This is causing the multi-platform-controller update task to fail for this node.


ifireball:
Can we get #4329 merged before this one so that I don't have to re-base it again?

ifireball:
> > Can we get #4329 merged before this one so that I don't have to re-base it again?
> 
> sure, I thought it would but I just saw that @arewm set hold label on that other one, we need to remove that label

it is there probably because it needed to wait for the MPC update that already landed, but we probably need @arewm to confirm

hugares:
@ifireball https://github.com/redhat-appstudio/infra-deployments/pull/4349 is merged, can you /lgtm this one now

ifireball (approved):
/lgtm

PR #4349, opened 2024-08-15
Remove compute-deployment quota from staging

hugares:
We no longer support users creating deployments in Konflux so no need to keep the compute-deployment quota.

KFLUXINFR-798


components/sandbox/tiers/src/appstudio/ns_tenant.yaml
```diff
@@ -27,21 +27,7 @@ objects:
...
-    name: compute-deploy
-    namespace: ${SPACE_NAME}-tenant
-  spec:
-    scopes:
-    - NotTerminating
-    hard:
-      limits.cpu: "20"
-      limits.memory: ${MEMORY_LIMIT}
-      requests.cpu: 1750m
-      requests.memory: ${MEMORY_REQUEST}
```
dperaza4dustbit, line 44: you only changed the src, did you run ksctl to generate?

hugares: nope, I forgot.. thanks. Fixed in latest revision

dperaza4dustbit (approved):
/lgtm

PR #4361, opened 2024-08-19
KFLUXINFRA-777: Config to Configure Worker Nodes with NVMe Disks

manish-jangra:
This Pull Request will do the following -

1. Create Namespace
2. ServiceAccount
3. ClusterRoleBinding
4. ConfigMap
5. DaemonSet

This DaemonSet will make use of shell scripts coming from the ConfigMap to configure NVMe Disks.

### DIFF
--- a/argo-cd-apps/base/member/infra-deployments/nvme-storage-configurator/nvme-storage-configurator.yaml
+++ b/argo-cd-apps/base/member/infra-deployments/nvme-storage-configurator/nvme-storage-configurator.yaml
@@ -1,0 +1,40 @@
+apiVersion: argoproj.io/v1alpha1
+kind: ApplicationSet
+metadata:
+  name: nvme-storage-configurator
+spec:
+  generators:
+    - merge:
+        mergeKeys:
+          - nameNormalized
+        generators:
+          - clusters:
+              values:
+                sourceRoot: configs/nvme-storage-configurator
+                environment: staging
+          - list:
+              elements: []
+  template:
+    metadata:
+      name: nvme-storage-configurator-{{nameNormalized}}
+    spec:
+      project: default
+      source:
+        path: '{{values.sourceRoot}}/{{values.environment}}'
+        repoURL: https://github.com/redhat-appstudio/infra-deployments.git
+        targetRevision: main
+      destination:
+        namespace: default
+        server: '{{server}}'
+      syncPolicy:
+        automated:
+          prune: true
+          selfHeal: true
+        syncOptions:
+        - CreateNamespace=true
+        retry:
+          limit: -1
+          backoff:
+            duration: 10s
+            factor: 2
+            maxDuration: 3m
--- a/argo-cd-apps/overlays/production-downstream/delete-applications.yaml
+++ b/argo-cd-apps/overlays/production-downstream/delete-applications.yaml
@@ -29,3 +29,9 @@ kind: ApplicationSet
 metadata:
   name: quality-dashboard
 $patch: delete
+---
+apiVersion: argoproj.io/v1alpha1
+kind: ApplicationSet
+metadata:
+  name: nvme-storage-configurator
+$patch: delete
--- a/configs/nvme-storage-configurator/base/daemonset.yaml
+++ b/configs/nvme-storage-configurator/base/daemonset.yaml
@@ -1,0 +1,61 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: nvme-storage-configurator
+  labels:
+    app: nvme-storage-configurator
+spec:
+  selector:
+    matchLabels:
+      app: nvme-storage-configurator
+  template:
+    metadata:
+      labels:
+        app: nvme-storage-configurator
+    spec:
+      nodeSelector:
+        konflux-ci.dev/storage: "nvme"
+      tolerations:
+        - key: konflux-ci.dev/storage
+          operator: "Equal"
+          value: "nvme"
+          effect: "NoSchedule"
+      hostPID: true
+      serviceAccount: nvme-storage-configurator
+      initContainers:
+        - name: install
+          image: "quay.io/openshift/origin-network-tools"
+          command: ['/bin/bash', '-c', '/bin/bash -x /scripts/install.sh']
+          env:
+          - name: NODE_NAME
+            valueFrom:
+              fieldRef:
+                apiVersion: v1
+                fieldPath: spec.nodeName
+          securityContext:
+            privileged: true
+          volumeMounts:
+            - mountPath: /scripts
+              name: script-config
+            - mountPath: /host-etc-systemd-dir
+              name: etc-systemd-dir
+            - mountPath: /host-usr-local-bin
+              name: usr-local-bin
+            - mountPath: /host-dev-dir
+              name: dev-dir
+      containers:
+        - name: pause
+          image: registry.redhat.io/rhel8/pause:8.10
+      volumes:
+        - name: etc-systemd-dir
+          hostPath:
+            path: /etc/systemd/system
+        - name: usr-local-bin
+          hostPath:
+            path: /usr/local/bin
+        - name: dev-dir
+          hostPath:
+            path: /dev
+        - name: script-config
+          configMap:
+            name: nvme-storage-configurator
--- a/configs/nvme-storage-configurator/base/install.sh
+++ b/configs/nvme-storage-configurator/base/install.sh
@@ -1,0 +1,24 @@
+#!/usr/bin/env bash
+
+# Enable debugging
+set -eux
+
+if [ -f /host-etc-systemd-dir/nvme-init-done ]; then
+    echo 'NVMe init already done'
+    oc adm uncordon $NODE_NAME
+    exit 0
+fi
+
+cp /scripts/mount.sh /host-usr-local-bin/mount.sh
+chmod 0755 /host-usr-local-bin/mount.sh
+
+cp /scripts/nvme-storage.service /host-etc-systemd-dir/nvme-storage.service
+chmod 0644 /host-etc-systemd-dir/nvme-storage.service
+
+oc adm cordon $NODE_NAME
+oc adm drain $NODE_NAME --delete-emptydir-data --ignore-daemonsets --grace-period=-1
+
+nsenter -t 1 -m -u -i -n -p -- systemctl daemon-reload
+nsenter -t 1 -m -u -i -n -p -- systemctl enable nvme-storage.service
+touch /host-etc-systemd-dir/nvme-init-done
+nsenter -t 1 -m -u -i -n -p -- systemctl reboot
--- a/configs/nvme-storage-configurator/base/kustomization.yaml
+++ b/configs/nvme-storage-configurator/base/kustomization.yaml
@@ -1,0 +1,14 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: nvme-storage-config
+resources:
+  - namespace.yaml
+  - serviceaccount.yaml
+  - clusterrolebinding.yaml
+  - daemonset.yaml
+configMapGenerator:
+- name: nvme-storage-configurator
+  files:
+  - install.sh
+  - mount.sh
+  - nvme-storage.service
--- a/argo-cd-apps/base/member/infra-deployments/kustomization.yaml
+++ b/argo-cd-apps/base/member/infra-deployments/kustomization.yaml
@@ -18,6 +18,7 @@ resources:
   - project-controller
   - spacerequest-cleaner
   - mintmaker
+  - nvme-storage-configurator
   - tracing-workload-otel-collector
   - tempo
   - notification-controller
--- a/argo-cd-apps/base/member/infra-deployments/nvme-storage-configurator/kustomization.yaml
+++ b/argo-cd-apps/base/member/infra-deployments/nvme-storage-configurator/kustomization.yaml
@@ -1,0 +1,4 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- nvme-storage-configurator.yaml
--- a/argo-cd-apps/overlays/konflux-public-production/delete-applications.yaml
+++ b/argo-cd-apps/overlays/konflux-public-production/delete-applications.yaml
@@ -11,3 +11,9 @@ kind: ApplicationSet
 metadata:
   name: tempo
 $patch: delete
+---
+apiVersion: argoproj.io/v1alpha1
+kind: ApplicationSet
+metadata:
+  name: nvme-storage-configurator
+$patch: delete
--- a/configs/nvme-storage-configurator/base/clusterrolebinding.yaml
+++ b/configs/nvme-storage-configurator/base/clusterrolebinding.yaml
@@ -1,0 +1,67 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: node-maintenance
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - nodes
+  verbs:
+  - get
+  - list
+  - watch
+  - patch
+- apiGroups:
+  - ""
+  resources:
+  - pods/eviction
+  verbs:
+  - create
+- apiGroups:
+  - ""
+  resources:
+  - pods
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - extensions
+  resources:
+  - daemonsets
+  verbs:
+  - get
+- apiGroups:
+  - apps
+  resources:
+  - daemonsets
+  verbs:
+  - get
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: node-maintenance
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: node-maintenance
+subjects:
+  - kind: ServiceAccount
+    name: nvme-storage-configurator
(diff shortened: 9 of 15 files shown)


configs/nvme-storage-configurator/daemonset.yaml
```diff
@@ -0,0 +1,51 @@
...
+  selector:
+    matchLabels:
+      app: nvme-storage-configurator
+  template:
+    metadata:
+      labels:
+        app: nvme-storage-configurator
+    spec:
+      nodeSelector:
+        node-role.kubernetes.io/konflux-tenants: ""
```
hugares, line 18: Question: IIUC, we will need to label the worker pool with `konflux-tenants` for the disk configuration to happen?

Why not running this on all the non infra workers? the daemon set will configure the nvme disk if there is one otherwise it will do nothing.  Two advantages of running this on all the non infra workers are:

- no need to label the worker pool
- if we ever start to use EC2 instances with nvme disk for the workers where Konflux will run eventually, no change will be required.

manish-jangra: Lets discuss this.

argo-cd-apps/base/member/infra-deployments/nvme-storage-configurator/nvme-storage-configurator.yaml
```diff
@@ -0,0 +1,41 @@
...
+  generators:
+    - merge:
+        mergeKeys:
+          - nameNormalized
+        generators:
+          - clusters:
+              values:
+                sourceRoot: configs/nvme-storage-configurator
+                environment: "staging"
+                clusterDir: ""
```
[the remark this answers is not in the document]
manish-jangra: removed

configs/nvme-storage-configurator/configmap.yaml
```diff
@@ -0,0 +1,72 @@
...
+
+    mount --bind ${CONTAINER_DIR} ${MOUNT_POINT}
+
+    echo "Filesystem setup and mounting complete."
+
+  nvme-storage.service: |
+    [Unit]
+    Description=Custom Service for NVMe Storage
+    DefaultDependencies=no
+    Before=kubelet.service network-pre.target
```
gbenhaim, line 66: I think there is a conflict here. Since this script is called using a daemonset, the unit can't start before the kubelet.

configs/nvme-storage-configurator/daemonset.yaml
gbenhaim: I suggest that we will first run it on a subset of nodes just to see it's stable. I would like also to verify the behavior of the daemonset after the node reboots and after cluster upgrade.

configs/nvme-storage-configurator/daemonset.yaml
```diff
@@ -0,0 +1,51 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: nvme-storage-configurator
+  labels:
+    app: nvme-storage-configurator
+  namespace: kube-system
```
gbenhaim, line 7: i don't think we should use system namespace. Is there a reason to use it?

configs/nvme-storage-configurator/daemonset.yaml
```diff
@@ -0,0 +1,51 @@
...
+        app: nvme-storage-configurator
+    spec:
+      nodeSelector:
+        node-role.kubernetes.io/konflux-tenants: ""
+      hostPID: true
+      serviceAccount: nvme-storage-configurator
+      initContainers:
+        - name: configure
+          image: "quay.io/openshift/origin-network-tools"
+          command: ['/bin/bash', '-c', 'mkdir /cache && cp /scripts/configure.sh /cache && chmod +x /cache/configure.sh && /bin/bash /cache/configure.sh && exit 0']
```
gbenhaim, line 24: you don't need the `exit 0` at the end.

gbenhaim: what is the purpose of this cache dir anyway? can we just call `/bin/bash /scripts/configure.sh` ?

configs/nvme-storage-configurator/kustomization.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+resources:
```
gbenhaim, line 4: should we set the `namespace` here?

argo-cd-apps/base/member/infra-deployments/nvme-storage-configurator/nvme-storage-configurator.yaml
```diff
@@ -0,0 +1,39 @@
...
+  name: nvme-storage-configurator
+spec:
+  generators:
+    - merge:
+        mergeKeys:
+          - nameNormalized
+        generators:
+          - clusters:
+              values:
+                sourceRoot: configs/nvme-storage-configurator
```
gbenhaim, line 13: I think that we should follow the pattern used in the other application sets so we can have different versions of this app in prod and stage (for testing purposes).

configs/nvme-storage-configurator/configmap.yaml
```diff
@@ -0,0 +1,72 @@
+apiVersion: v1
+kind: ConfigMap
```
gbenhaim, line 2: lets use kustomize configmap generator for generating this file:
1. then we can edit each script/unit file in its own file and have the benefit of IDEs.
2. when the content of the configmap changes, it will cause a roll out of the daemonset that will retrigger the `configure*` scripts.

gbenhaim (changes requested):
Thanks for working on it @manish-jangra 
1. I posted some comments
2. I would like us to wait with this change until we move pipelines to run on a dedicated node pool.
3. Can you also open a ticket for SREP requesting them to support the configuration of instances with NVME disks?

configs/nvme-storage-configurator/daemonset.yaml
manish-jangra: @gbenhaim We will run it on a new node pool with the help of NodeSelector. DaemonSet will only run its pod on the nod ematching the node slector label. We won't touch existing worker or infra nodes.

configs/nvme-storage-configurator/configmap.yaml
manish-jangra: @gbenhaim for the first time, DS is starting the service manually. After the node reboots, this service will run before kublet service. If it's fundmentally incorrect, let's discuss this.

configs/nvme-storage-configurator/daemonset.yaml
manish-jangra: It was a mistake, it should be `nvme-storage-config`. I have fixed it.

configs/nvme-storage-configurator/daemonset.yaml
manish-jangra: I have shortened it.

`'/bin/bash', '-c', 'chmod +x /scripts/configure.sh && /bin/bash /cache/configure.sh'`

configs/nvme-storage-configurator/kustomization.yaml
manish-jangra: Should we? If we should, which namespace? the one where DaemonSet will run?

configs/nvme-storage-configurator/daemonset.yaml
manish-jangra: I have tested the reboot behaviour. After I rebooted the node, I found the NVMe Disk intact.

configs/nvme-storage-configurator/kustomization.yaml
manish-jangra: removed the `namespace.yaml` and added namespace attribute in `kustomisation.yaml`

argo-cd-apps/base/member/infra-deployments/nvme-storage-configurator/nvme-storage-configurator.yaml
manish-jangra: changes made for handling staging and production overlays

jhutar (approved):
We measure this on the Stage and there was 34% improvement in build duration. For details see https://docs.google.com/document/d/1Z_J1-vM-4u7ThbnU7-dY0cwRNCOaIAaebDqCChdhDjs/edit#heading=h.hjqbmzauj6f0

configs/nvme-storage-configurator/daemonset.yaml
manish-jangra: fixed now.

configs/nvme-storage-configurator/configmap.yaml
manish-jangra: now using kustomize to generate the configmap

configs/nvme-storage-configurator/base/daemonset.yaml
```diff
@@ -0,0 +1,50 @@
...
+  selector:
+    matchLabels:
+      app: nvme-storage-configurator
+  template:
+    metadata:
+      labels:
+        app: nvme-storage-configurator
+    spec:
+      nodeSelector:
+        konflux-ci.dev/storage: "nvme"
```
gbenhaim, line 17: so we also need to add a node pool for those machines right?

configs/nvme-storage-configurator/base/configure.sh
```diff
@@ -0,0 +1,20 @@
...
+cp -f /scripts/configuration.sh /host-usr-local-bin/configuration.sh
+chmod 0755 /host-usr-local-bin/configuration.sh
+
+cp -f /scripts/nvme-storage.service /host-etc-systemd-dir/nvme-storage.service
+chmod 0644 /host-etc-systemd-dir/nvme-storage.service
+
+nsenter -t 1 -m -u -i -n -p -- systemctl daemon-reload
+nsenter -t 1 -m -u -i -n -p -- systemctl enable nvme-storage.service
+touch /host-etc-systemd-dir/nvme-init-done
+nsenter -t 1 -m -u -i -n -p -- systemctl reboot
```
gbenhaim, line 20: can we do this? don't we need to evacuate  the pods first from the node (and disable scheduling for it)?

gbenhaim: also, why do we need a reboot?

configs/nvme-storage-configurator/base/configure.sh
```diff
@@ -0,0 +1,20 @@
...
+
+# Enable debugging
+set -eux
+
+if [ -f /host-etc-systemd-dir/nvme-init-done ]; then
+    echo 'NVMe init already done'
+    exit 0
+fi
+
+cp -f /scripts/configuration.sh /host-usr-local-bin/configuration.sh
```
gbenhaim, line 11: why `-f` is needed?

configs/nvme-storage-configurator/base/daemonset.yaml
```diff
@@ -0,0 +1,50 @@
...
+        - name: pause
+          image: registry.redhat.io/rhel8/pause:8.10
+      volumes:
+        - name: etc-systemd-dir
+          hostPath:
+            path: /etc/systemd/system
+        - name: usr-local-bin
+          hostPath:
+            path: /usr/local/bin
+        - name: dev-dir
```
gbenhaim, line 45: why this mount is needed? when running a privileged pod does it get access to the hosts /dev dir?

configs/nvme-storage-configurator/base/configure.sh
gbenhaim, line 11: I find the names of the scripts confusing (maybe because they similar).
Lets give them meaningful names.

we can call the script that install the unit file and the script it runs `install.sh`
and the the script that does the mount `mount.sh`

configs/nvme-storage-configurator/base/kustomization.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: nvme-storage-config
+resources:
+  - serviceaccount.yaml
+  - clusterrolebinding.yaml
+  - daemonset.yaml
+configMapGenerator:
+- name: nvme-storage-configurator
```
gbenhaim, line 9: you need to mention the namespace that will be used for the reource in here. In addition you need to create a resource for the namespace itself.

configs/nvme-storage-configurator/base/daemonset.yaml
manish-jangra: Yes, we need to add this label for the new node pool with nvme disks.

configs/nvme-storage-configurator/base/configure.sh
manish-jangra: This configuration requires reboot as kubelet service has dependency on `/var/lib/kubelet/pods` and `/var/lib/containers`. We can add cordoning and draining option for the node and uncordon them after the reboot.

configs/nvme-storage-configurator/base/daemonset.yaml
manish-jangra: We're making use of hostPaths in order to access the file system of ec2 instance. We have mapped `/dev` to programatically the nvme device name rather than hard coding `nvme1n1`.

configs/nvme-storage-configurator/base/configure.sh
manish-jangra: ok lets change it. I will change them in next commit.

configs/nvme-storage-configurator/base/kustomization.yaml
manish-jangra: added namespae resource

configs/nvme-storage-configurator/base/configure.sh
manish-jangra: I have changed the secripts to `install.sh `and `mount.sh`.

configs/nvme-storage-configurator/base/configure.sh
manish-jangra: I have addeded the following commands to gracefully reboot the nodes --

```
oc adm cordon $NODE_NAME
oc adm drain $NODE_NAME --delete-emptydir-data --ignore-daemonsets --grace-period=0 --force
```

and after the node reboots 

`oc adm uncordon $NODE_NAME`

configs/nvme-storage-configurator/base/configure.sh
manish-jangra: I just added this out of habit of forcefully copying the files to host's file system. Please let me know if `-f` needs to go.

configs/nvme-storage-configurator/base/daemonset.yaml
gbenhaim: when running a privileged container, its /dev directory should be the same as the host, making this mount redundant.

configs/nvme-storage-configurator/base/install.sh
```diff
@@ -0,0 +1,24 @@
...
+fi
+
+cp -f /scripts/mount.sh /host-usr-local-bin/mount.sh
+chmod 0755 /host-usr-local-bin/mount.sh
+
+cp -f /scripts/nvme-storage.service /host-etc-systemd-dir/nvme-storage.service
+chmod 0644 /host-etc-systemd-dir/nvme-storage.service
+
+oc adm cordon $NODE_NAME
+oc adm drain $NODE_NAME --delete-emptydir-data --ignore-daemonsets --grace-period=0 --force
```
gbenhaim, line 19: 1. I think that we need to lt pods finish gracefully, other wise we would that pipelines fail because their pods are getting deleted.

2. what is the purpose of the --force flag? I didn't understand its description in the cli help message.

manish-jangra: removed --force flag and changed the grace-period to -1 to use the one coming from the pods.

manish-jangra: Now, Pods should be gracefully evicted.

configs/nvme-storage-configurator/base/configure.sh
gbenhaim: You need to decide if it's needed or not. a habit is not a reason to keep it :)

configs/nvme-storage-configurator/base/daemonset.yaml
manish-jangra: I have no idea about this. But when we tried running this without mouting the /dev directory, we couldn't access the /dev/xxxx. I assumed, we couln't not be accessed because it was not mounted.

configs/nvme-storage-configurator/base/configure.sh
manish-jangra: removed `-f` option

manish-jangra:
@gbenhaim could you please take a look at this whenever your schedule allows you?

Thanks :)

argo-cd-apps/overlays/production-downstream/delete-applications.yaml
```diff
@@ -29,3 +29,15 @@ kind: ApplicationSet
 metadata:
   name: quality-dashboard
 $patch: delete
+---
+apiVersion: argoproj.io/v1alpha1
+kind: ApplicationSet
+metadata:
+  name: workspaces
+$patch: delete
```
hugares, line 37: @manish-jangra this looks like a bad rebase, we should not be deleting workspaces in this PR.
```suggestion
---
```

manish-jangra: @hugares It seems that main branch doesn't have delete entry for workspaces component. [Main Branch](https://github.com/redhat-appstudio/infra-deployments/blob/main/argo-cd-apps/overlays/konflux-public-production/delete-applications.yaml)

hugares (approved):

PR #4369, opened 2024-08-20
SRVKP-5900: enable pprof for results thread dumps, prod

gabemontero:
As part of the deadlock alert / SOP, this will allow SRE to gather goroutine dumps so we can isolate where the deadlock is.

Been running on stage for a bit and I confirmed I can get goroutine dumps from both result api and watcher pods.

https://github.com/redhat-appstudio/infra-deployments/blob/5e89946f746273a5cee58d0753fc4cfeecec47e6/components/pipeline-service/staging/base/main-pipeline-service-configuration.yaml#L732

https://github.com/redhat-appstudio/infra-deployments/blob/5e89946f746273a5cee58d0753fc4cfeecec47e6/components/pipeline-service/staging/base/main-pipeline-service-configuration.yaml#L825

pprof enablement per golang is safe in production:  https://go.dev/doc/diagnostics

FYI there is an issue unfortunately with minimally tekton-pipelines-controller that I've opened and issue for.

rh-pre-commit.version: 2.3.1

### DIFF
--- a/components/pipeline-service/production/base/main-pipeline-service-configuration.yaml
+++ b/components/pipeline-service/production/base/main-pipeline-service-configuration.yaml
@@ -727,6 +727,8 @@ data:
     S3_MULTI_PART_SIZE=5242880
     GCS_BUCKET_NAME=
     STORAGE_EMULATOR_HOST=
+    PROFILING=true
+    PROFILING_PORT=6060
 kind: ConfigMap
 metadata:
   annotations:
@@ -818,6 +820,7 @@ metadata:
 ---
 apiVersion: v1
 data:
+  profiling.enable: "true"
   _example: |
     ################################
     #                              #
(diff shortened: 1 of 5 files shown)


gabemontero:
Also enabled in dev overlay:

https://github.com/redhat-appstudio/infra-deployments/blob/5e89946f746273a5cee58d0753fc4cfeecec47e6/components/pipeline-service/staging/base/main-pipeline-service-configuration.yaml#L732

https://github.com/redhat-appstudio/infra-deployments/blob/5e89946f746273a5cee58d0753fc4cfeecec47e6/components/pipeline-service/staging/base/main-pipeline-service-configuration.yaml#L825

gabemontero:
confirmed I can get thread dumps on the watcher in prod

api servers need to be restarted ... will see about that later this week unless a recycle happens naturally in the interim.

PR #4374, opened 2024-08-21
Fix ssh key for s390x stage ext

bamachrn:


components/multi-platform-controller/staging/external-secrets.yaml
```diff
@@ -89,3 +89,26 @@ spec:
...
+  namespace: multi-platform-controller
+  labels:
+    build.appstudio.redhat.com/multi-platform-secret: "true"
+  annotations:
+    argocd.argoproj.io/sync-options: SkipDryRunOnMissingResource=true
+    argocd.argoproj.io/sync-wave: "-1"
+spec:
+  dataFrom:
+    - extract:
+        key: production/build/multi-platform-controller/ibm-production-s390x-ssh-key
```
hugares, line 106: should we extract the staging key?

PR #4378, opened 2024-08-21
Replace chains public key Job by ExternalSecret

enarha:
Instead of the Job used to read the secret and create a new secret including only the public key, we now use ExternalSecret pointing to the same vault entry, but creating a secret including only the public key. This eliminates the race condition issues we had with the Job. This is also a more conformant way of managing the secret using the same ExternalSecret mechanism.
This will also make it easier to rotate the secret, updating the value in the vault will update both secrets (the one including private+public keys and the other one including only the public key).

Changes in the production kustomizations are to suppress the new ExternalSecret creation until we are ready to do the same change in production. Then we only need to remove the "delete" patches.

### DIFF
--- a/components/pipeline-service/production/stone-prd-m01/resources/kustomization.yaml
+++ b/components/pipeline-service/production/stone-prd-m01/resources/kustomization.yaml
@@ -21,3 +21,12 @@ patches:
       group: external-secrets.io
       version: v1beta1
       kind: ExternalSecret
+  - target:
+      kind: ExternalSecret
+      name: tekton-chains-public-key
+    patch: |
+      $patch: delete
+      apiVersion: external-secrets.io/v1beta1
+      kind: ExternalSecret
+      metadata:
+        name: tekton-chains-public-key
--- a/components/pipeline-service/base/external-secrets/openshift-pipelines/chains-public-key.yaml
+++ b/components/pipeline-service/base/external-secrets/openshift-pipelines/chains-public-key.yaml
@@ -1,0 +1,25 @@
+apiVersion: external-secrets.io/v1beta1
+kind: ExternalSecret
+metadata:
+  name: tekton-chains-public-key
+  namespace: openshift-pipelines
+  annotations:
+    argocd.argoproj.io/sync-options: SkipDryRunOnMissingResource=true
+    argocd.argoproj.io/sync-wave: "-1"
+spec:
+  data:
+    - secretKey: "cosign.pub"
+      remoteRef:
+        key: ""  # will be added by the overlays
+        property: "cosign.pub"  # IMPORTANT: without specifying the public key as a property, the whole information from the vault key (including the private key) will be included!
+  refreshInterval: 5m
+  secretStoreRef:
+    kind: ClusterSecretStore
+    name: appsre-stonesoup-vault
+  target:
+    creationPolicy: Orphan
+    name: public-key
+    template:
+      metadata:
+        annotations:
+          argocd.argoproj.io/sync-options: Prune=false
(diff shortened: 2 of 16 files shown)


enarha:
For the record, the change only affects the staging deployments, leaving the development overlay unchanged, so the CI checks add no real value. For what it's worth I tested the next ExternalSecret config on one of the staging clusters. FYI.

components/pipeline-service/production/stone-prd-m01/resources/kustomization.yaml
```diff
@@ -21,3 +21,12 @@ patches:
       group: external-secrets.io
       version: v1beta1
       kind: ExternalSecret
+  - target:
+      kind: ExternalSecret
+      name: tekton-chains-public-key
+    patch: |
+      $patch: delete
```
gabemontero, line 28: I'm unfamiliar if whether gitops delete complains enough if the entry is not there, and the gitops rollout is prevented / stays out of sync.

Do you have experience with this @enarha or somehow was able to confirm this via some form of local testing?

Maybe @hugares knows how gitops will react?

I suppose the alternative, which I suspect you were trying to avoid @enarha, is duplicating the `components/base/external-secrets/openshift-pipelines` entries in `staging/base` and then `production/base`

gabemontero (commented):
Just the one question I have below (and it is from my lack of knowledge vs. my knowing something is wrong)

Also, to add a bit more detail to @enarha 's comment https://github.com/redhat-appstudio/infra-deployments/pull/4378#issuecomment-2302387435 and leaving the Job in the development overlay, one of the key motivations from removing the Job in staging/production is that k8s does allow modification of Jobs, so we would hit intermittent issues during gitops roll out where pipeline-service would stay out of sync until we manually deleted the Job (sometimes multiple attempts were needed).

Of course, we don't have to worry about gitops rollout during the e2e tests runs that use the development overlay, so there is less motivation for removing there.  Plus, the timing with the one-off scripts in infra-deployments/hack that are called during bootstrap appear to be untenable for adding a script to replicate what the Job does.... those scripts are called too soon in the process.

components/pipeline-service/production/stone-prd-m01/resources/kustomization.yaml
enarha: kustomize does not complain if the target is not there. And yes, that I found the easier alternative than duplicating configuration and making the next change (enable prod) easier as well. Once the change is merged, it's easy to verify and we can quickly proceed with updating prod, effectively removing that bit in a day or two.

gabemontero: cool deal

lcarva:
+1 on the approach (as much as I understand it).

One question, is there something we can do to avoid a mistake that would accidentally copy the private key to the `public-key` Secret? A simple well-placed comment could be sufficient.

hugares (approved):
/lgtm

hugares:
/hold
To allow time to answer Luiz's question

hugares (approved):
/lgtm

enarha:
> +1 on the approach (as much as I understand it).
> 
> One question, is there something we can do to avoid a mistake that would accidentally copy the private key to the `public-key` Secret? A simple well-placed comment could be sufficient.

One mistake I can think of if removing the property specifying the public key (cosign.pub) in which case all the information from the vault referenced by the key will be included. I addressed that in a comment.

enarha:
> /hold To allow time to answer Luiz's question

Thanks. Please let me remove the hold myself when I'm available to monitor the process.

PR #4382, opened 2024-08-21
Send traces via honeycomb refinery for sampling purposes

mike-kingsbury:
Send traces via the RH Honeycomb Refinery instance to enable sampling of traces.


mftb:
Do we still need the honeycomb secrets to send the traces to the refinery?

mike-kingsbury:
> Do we still need the honeycomb secrets to send the traces to the refinery?

Yes.  I can go into more detail, but fundamentally the API key is needed on traces being sent to Refinery.

mftb:
Got it.

/lgtm
/approve

mftb (approved):

PR #4394, opened 2024-08-26
Add view permissions for CaaS and hypershift resources

amisstea:


amisstea:
> @amisstea , do we need to update our permissions ADR for this?

I don't think that's necessary. That [ADR](https://github.com/konflux-ci/architecture/blob/main/ADR/0011-roles-and-permissions.md#roles-and-permissions-table) lists the permissions for the Konflux workspace roles whereas this change grants permissions to engineers who can authenticate to the clusters.

hugares (approved):
/lgtm

PR #4397, opened 2024-08-27
KFLUXINFRA-802: Segregate Konflux and Tenants Workload using TektonConfig

manish-jangra:
This PR will help us schedule Konflux and Tenants Workload on dedicated Node Pools. This will make use of TektonConfig.

NodeSelector and Tolerations values are sample values only, we can change them if we want. We will need to create a new Node Pool in OCM (Manually or using Terraform) with the values of (NodeSelector and Tolerations ) we specifiy in the TektonConfig.


enarha (commented):
/lgtm

hugares (approved):
/lgtm

PR #4403, opened 2024-08-27
Replace chains public key Job by ExternalSecret - prod

enarha:
Instead of the Job used to read the secret and create a new secret including only the public key, we now use ExternalSecret pointing to the same vault entry, but creating a secret including only the public key. This eliminates the race condition issues we had with the Job. This is also a more conformant way of managing the secret using the same ExternalSecret mechanism.


enarha:
/hold

@gabemontero @hugares this is follow up on #4378 but for updating prod. The PR is ready for review. The reason I've put the hold label is to make sure it's merged when I'm around to monitor the process.

hugares (approved):
/lgtm

gabemontero (approved):
/lgtm

PR #4406, opened 2024-08-27
workspaces: don't delete applicationset from prod

sadlerap:
This enables workspaces in production environments.  Resources were added in PR #4393.


hugares (approved):
/lgtm

PR #4408, opened 2024-08-28
chore: Give access to integration service to access vault secrets for e2e tests

flacatus:
Integration service need access to QE secrets in order to run e2e tests in Konflux CI


jduimovich (commented):
LGTM

PR #4410, opened 2024-08-28
KFLUXINFRA-808: Set default pipelines timeout to 2 hours - prod

gbenhaim:
Increase the default timeout for pipelines from
1 hours (the default) to 2 hours.

The reason is that users complained about timeouts while their pipelines were waiting for resources.


jhutar (approved):
/lgtm

hugares (approved):
/lgtm

PR #4411, opened 2024-08-28
Migrate ClusterRole appstudio-pipelines-runner

scoheb:
- currently this ClusterRole is managed and deployed by the Sandbox host operator.
- Since it is a Konflux specific role, it should be migrated and maintained in this repo and deployed by Argo
- This role should be installed on member clusters

Needed for [KONFLUX-4171](https://issues.redhat.com//browse/KONFLUX-4171)

### DIFF
--- a/components/sandbox/toolchain-member-operator/base/rbac/appstudio-pipelines-runner.yaml
+++ b/components/sandbox/toolchain-member-operator/base/rbac/appstudio-pipelines-runner.yaml
@@ -1,0 +1,36 @@
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: appstudio-pipelines-runner
+rules:
+  - verbs:
+      - get
+    apiGroups:
+      - ''
+    resources:
+      - secrets
+  - verbs:
+      - use
+    apiGroups:
+      - security.openshift.io
+    resources:
+      - securitycontextconstraints
+    resourceNames:
+      - appstudio-pipelines-scc
+  - verbs:
+      - get
+      - list
+    apiGroups:
+      - appstudio.redhat.com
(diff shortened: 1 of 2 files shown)


gabemontero:
seems OK to me @scoheb 

I've added @enarha for awareness on the openshift-pipelines team

hugares:
@scoheb How do we tell the host operator to not create/update this resource once managed by ArgoCD?

components/sandbox/common/rbac/appstudio-pipelines-runner.yaml
hugares, line 1: do we need this on all clusters, i.e. both host and member ?

scoheb: my understanding is the host operator creates this once and then argo will reconcile the file?

I believe @dperaza4dustbit wanted to remove this Role from the host operator as well though.

scoheb:
> @scoheb How do we tell the host operator to not create/update this resource once managed by ArgoCD?

my understanding is the host operator creates this once and then argo will reconcile the file?

I believe @dperaza4dustbit wanted to remove this Role from the host operator as well though.

components/sandbox/common/rbac/appstudio-pipelines-runner.yaml
hugares: common component is deployed to both host and member, you need to move this to member folder: https://github.com/redhat-appstudio/infra-deployments/tree/main/components/sandbox/toolchain-member-operator

scoheb: fixed the path. thanks

hugares (approved):
/lgtm

PR #4415, opened 2024-08-29
KFLUXINFRA-802: Adding Node Selector and Tolerations in TektonConfig

manish-jangra:
Adding Node Selector and Tolerations in TektonConfig to make use of the default pod template for the PipelineRuns and TaskRuns.

Related Pull Request for Stage Member Cluster (m01) -
- #4397


enarha:
@manish-jangra Adding kustomization patches in the per cluster resources directory is good if you want to test on a single cluster or if you want to have different values on each cluster. If you have the same values per cluster type (stage/prod) or for all clusters, the way to do it if add it just once per environment, development (used by the CI and developer) https://github.com/redhat-appstudio/infra-deployments/blob/main/components/pipeline-service/development/main-pipeline-service-configuration.yaml#L1478 , staging https://github.com/redhat-appstudio/infra-deployments/blob/main/components/pipeline-service/staging/base/main-pipeline-service-configuration.yaml#L1294 and production https://github.com/redhat-appstudio/infra-deployments/blob/main/components/pipeline-service/production/base/main-pipeline-service-configuration.yaml#L1445. Once you add you changes, you use the generate script to update the deploy.yaml files for all staging and production clusters.

bamachrn:
I agree with @enarha lets right the nodeselector in one place and patch it the deployconfigs for other clusters. so that we dont repeat the same codes.

hugares (approved):
/lgtm

enarha:
@manish-jangra The changes to staging and production are good. Please add the change to the development environment as well (https://github.com/redhat-appstudio/infra-deployments/blob/main/components/pipeline-service/development/main-pipeline-service-configuration.yaml#L1478). This is what the Konflux CI and manual deployments use and we want to keep it as close as possible to prod/stage, even if the change does not have real impact. Once done, the change is good to be merged IMO.

enarha (commented):
/lgtm

PR #4417, opened 2024-08-29
More SPI leftovers cleanup

mshaposhnik:
More SPI leftovers cleanup for KFXINFRA-660


components/sandbox/tiers/src/appstudio/ns_tenant.yaml
```diff
@@ -120,18 +120,6 @@ objects:
   spec:
     hard:
       count/enterprisecontractpolicies.appstudio.redhat.com: "512"
-- apiVersion: v1
```
hugares, line 123: When changing the tiers sources, we need to regenerate them for staging and prod, this is documented in the README.adoc in the tiers folder. The problem is we need to use a version of ksctl with an un-merged  change: https://github.com/kubesaw/ksctl/pull/50

I can regenerate the tiers and amend this PR if you want

hugares: Done

hugares (approved):
/lgtm

components/sandbox/tiers/src/appstudio/ns_tenant.yaml
mshaposhnik: Thanks!

PR #4422, opened 2024-08-29
promote hac-dev latest changes

sahil143:


arewm:
This bring the HAC version up to what it is in the external production cluster. It includes one reverted commit from the current state.

/lgtm

hugares:
One day I would like to understand where this "Red Hat Konflux" check is coming from...

hugares:
> One day I would like to understand where this "Red Hat Konflux" check is coming from...

I amended the PR to get rid of that pending check...

hugares (approved):
/lgtm

PR #4427, opened 2024-08-29
ui: pass relevant requests to workspaces

sadlerap:
Workspaces is now live in production environments, but is currently inaccessible as it's not exposed.  In external environments, workspaces is configured to use a route, but on internal environments, it can piggy-back off of the nginx configuration already in place for UI components.


hugares (approved):
/lgtm

PR #4428, opened 2024-08-30
add Cloudability's metrics-agent to stage-p01

filariow:
This PR introduces an ApplicationSet for [Cloudability's metrics-agent](https://github.com/cloudability/metrics-agent/). The Agent is installed via [it's Helm
Chart](https://github.com/cloudability/metrics-agent/tree/master?tab=readme-ov-file#deploying-with-helm) and it will be initially deployed only in cluster stage-p01.


argo-cd-apps/base/all-clusters/infra-deployments/cloudability-metrics-agent/cloudability-metrics-agent.yaml
```diff
@@ -0,0 +1,59 @@
...
+          - clusters:
+              values:
+                sourceRoot: components/cloudability-metrics-agent
+                environment: staging
+                clusterDir: ""
+          - list:
+              elements:
+                # restrict to stage-p01 cluster for PoC
+                - nameNormalized: stone-stage-p01
+                  values.clusterDir: stone-stage-p01
```
sadlerap, line 23: You'll need to add a deletion patch to restrict the deployment environments, otherwise you'll end up deploying this to production.  See #4406 as an example of the reverse of what you need to add.

To test your change, you can build the kustomize manifest at `argo-cd-apps/overlays/production-downstream` and `argo-cd-apps/overlays/konflux-public-production` and see if `cloudability-metrics-agent` shows up.

manish-jangra:
I think, we should also have a `kustomisation.yaml` in `components/cloudability-metrics-agent/staging/stone-stage-p01` directory pointing to values file.

argo-cd-apps/base/all-clusters/infra-deployments/cloudability-metrics-agent/cloudability-metrics-agent.yaml
filariow: thanks! done in 771183f3de78e3a64d70b1119f66bef08d336aa5

filariow:
> I think, we should also have a `kustomisation.yaml` in `components/cloudability-metrics-agent/staging/stone-stage-p01` directory pointing to values file.

I took inspiration from [components/tracing/otel-collector/production](https://github.com/redhat-appstudio/infra-deployments/tree/main/components/tracing/otel-collector/production).
The [kustomization.yaml](https://github.com/redhat-appstudio/infra-deployments/blob/main/components/tracing/otel-collector/production/kustomization.yaml) file there is not referencing the `values.yaml` file, so I thought it's not needed here either.

Also, given that the `path` field is not used in the `ref-source`, this should not build what's in the folder so it seems to me it should be fine (cf. https://argo-cd.readthedocs.io/en/stable/user-guide/multiple_sources/#helm-value-files-from-external-git-repository). WDYT?

argo-cd-apps/base/all-clusters/infra-deployments/cloudability-metrics-agent/cloudability-metrics-agent.yaml
sadlerap: hmm, this will also be enabled in external staging, although it looks your intention is only to enable this for internal staging.  I think you'll need to add a deletion patch into `argo-cd-apps/overlays/konflux-public-staging` as well.

manish-jangra (approved):

manish-jangra:
@filariow lets create the the directory structure like other components 

- base
- staging
- production

will be easier to maintain configuration for different configurations for different overlays.

sadlerap (approved):

ralphbean:
@filariow this one needs a rebase and conflict resolution for tide to accept it.

filariow:
thanks @ralphbean, I'm waiting for cloudability's team to guarantee compatibility with OpenShift 4.15 and 4.16 (https://github.com/cloudability/metrics-agent/issues/270). I'll move this PR to draft.

PR #4430, opened 2024-08-30
set port when forwarding requests to workspaces

sadlerap:
The workspaces service listens on port 8000, so we need to explicitly communicate on that port when passing requests to it.


hugares (approved):
/lgtm

PR #4441, opened 2024-09-03
chore: label correspondingly development nodes

flacatus:
- Make sure all development cluster will have proper nodes labels

### DIFF
--- a/hack/preview.sh
+++ b/hack/preview.sh
@@ -141,6 +141,29 @@ if [ "$EC_DISABLE_DOWNLOAD_SERVICE" = "true" ]; then
   yq eval 'del(.resources[] | select(. == "download-service.yaml"))' -i  $ROOT/components/enterprise-contract/kustomization.yaml
 fi
 
+# After changes introduced in https://github.com/redhat-appstudio/infra-deployments/pull/4415/files the nodes need to be labeled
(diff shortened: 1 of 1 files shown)


mkosiarc (approved):

manish-jangra:
looks good to me

hack/preview.sh
```diff
@@ -141,6 +141,24 @@ if [ "$EC_DISABLE_DOWNLOAD_SERVICE" = "true" ]; then
...
+for node in $(kubectl get nodes -o name); do
+    echo "labeling $node..."
+    if kubectl label $node konflux-ci.dev/workload=konflux-tenants --overwrite; then
+        echo "successfully labeled $node"
+    else
+        echo "failed to label $node"
+    fi
+done
+
+kubectl get nodes --show-labels | grep konflux-ci.dev/workload
```
enarha, line 154: I think that the return code will have a 0 value even if one match (labeled node) is found. What you can do instead is take the count of the nodes and compare to the count of matched lines.

enarha: ... or do a loop on the nodes and check for each one. I like the above option more.

flacatus: i added a count for nodes.

kasemAlem (approved):

PR #4457, opened 2024-09-05
fix: kube_pod_container_status_restarts_total federate match

yftacherzog:
Fix a typo in the match string for the service monitor used for federating  metric kube_pod_container_status_restarts_total


theflockers (approved):

avi-biton (approved):
/lgtm

PR #4463, opened 2024-09-06
Grant cluster-monitoring-view role to all Konflux teams

hugares:
Until version 4.14 of OCP, all Konflux teams could access cluster monitoring, grant view cluster role was enough. Starting with 4.15, something changed and view clusterrole no longer allow to access monitoring and execute queries.

Granting cluster-monitoring-view cluster role to all teams resolve the issue.


amisstea (approved):

jhutar (approved):
/lgtm

PR #4465, opened 2024-09-06
Install minio-operator from upstream

amisstea:
The operator no longer exists in OperatorHub and cannot be installed on OCP 4.16+ without an enterprise license from MinIO.

See https://github.com/minio/operator/issues/2226

Version 5.0.15 is selected over 6.0.3 as the latter requires a minimum k8s version of 1.28 (OCP 4.15+). With 5.0.15 the minimum supported k8s version is 1.21 (OCP 4.8+).

### DIFF
--- a/components/pipeline-service/development/kustomization.yaml
+++ b/components/pipeline-service/development/kustomization.yaml
@@ -8,7 +8,55 @@ commonAnnotations:
   argocd.argoproj.io/sync-options: SkipDryRunOnMissingResource=true
 
 resources:
+  - github.com/minio/operator?ref=v5.0.15
   - main-pipeline-service-configuration.yaml
   - dev-only-pipeline-service-storage-configuration.yaml
   - ../base/rbac
 
+patches:
+  - target:
+      kind: Service
+      name: operator
+      namespace: minio-operator
+    patch: |
+      - op: add
+        path: /metadata/annotations/ignore-check.kube-linter.io~1dangling-service
+        value: This service is not dangling. The minio operator assigns the missing labels at runtime.
+  - target:
+      kind: Deployment
+      name: minio-operator
+      namespace: minio-operator
+    patch: |
+      - op: add
+        path: /metadata/annotations/ignore-check.kube-linter.io~1no-read-only-root-fs
+        value: The operator needs to be able to write to /tmp
+      - op: add
+        path: /spec/template/spec/containers/0/resources/limits
+        value:
+          cpu: 200m
+          memory: 256Mi
+      - op: remove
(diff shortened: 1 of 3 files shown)


amisstea:
An OCP 4.16 cluster can be successfully bootstrapped after this change.

gbenhaim:
@enarha  can you please review?

components/pipeline-service/development/kustomization.yaml
```diff
@@ -8,7 +8,55 @@ commonAnnotations:
   argocd.argoproj.io/sync-options: SkipDryRunOnMissingResource=true
 
 resources:
+  - github.com/minio/operator?ref=v5.0.15
```
enarha, line 11: It's mentioned in the commit message that the newer version requires k8s 1.28 (OCP 4.15) and that's already the case for the development (clusters used in CI) and the Konflux staging environment. Users also can't use earlier versions because the currently used version of OSP requires k8s 1.28. SO if there isn't other reason, you can use 6.0.3.
The rest is LGTM.

jhutar:
Hello @amisstea ! I'm deploying Konflux to OCP cluster in Hive and I'm hitting this issue:
```
FATAL Unable to initialize backend: parity validation returned an error: parity 2 should be less than or equal to 1 <- (2, 2), for pool(1st)
```
When I terminated ArgoCD sync and changed image tag `RELEASE.2024-08-26T15-33-07Z` back to `RELEASE.2022-09-17T00-09-45Z`, it started working.

Looks like for some reason change in the secret was not applied for me:
```
$ oc -n tekton-results get secrets/minio-storage-configuration -o json | jq -r '.data."config.env" | @base64d' | grep MINIO_STORAGE_CLASS_STANDARD
export MINIO_STORAGE_CLASS_STANDARD="EC:2"
```

Another way to "fix" the setup was when I fixed the secret and deleted StatefulSet and it's PVCs to allow Minio operator to recreate it.

Looking into the install log for my deployment, I can see that the related function ran, but returned here: https://github.com/amisstea/infra-deployments/blob/9774664ad202f1958829a191d1730f1b64269209/hack/secret-creator/create-plnsvc-secrets.sh#L35 :
```
[...]
Creating DB secret
DB secret already exists, skipping creation
Creating S3 secret
S3 secret already exists, skipping creation
Creating Postgres TLS certs
[...]
```

Am I missing something, or is there something we need to fix?

jhutar:
OK, I'm sorry. Looks like this was caused by unclean first installation I repeated few times without cleanup. With fresh cluster it all works. Please ignore me Alex.

PR #4466, opened 2024-09-08
chore(RHTAPWATCH-1266): update external secret location - staging

avi-biton:
Update external secret location in staging environment


yftacherzog (approved):
lgtm

PR #4474, opened 2024-09-09
Remove enable-tekton-oci-bundles from dev and staging TektonConfig

hugares:
This configuration options was removed in Tekton v0.61.0 which we now use staging since ef952976970f7dd17094483cb57b9bfb0e0f74ae


enarha (commented):
/lgtm

PR #4481, opened 2024-09-10
mintmaker: add ExternalSecret for registry.stage.redhat.io pull secret

qixiang:


qixiang:
This is secret is prepared by QE team, we would like to keep it in consistent format as the production one from build team `production/build/tekton-ci/registry-redhat-io-pull-secret`, however we don't have permission to check what's the exact format of the production one, before we verified the format is in the expected format, we won't consume it in mintmaker, that's why we don't add the label in this PR.

staticf0x (approved):
LGTM

PR #4380, opened 2024-08-21
fix: release controller oomkill on start up

theflockers:
this commit fixes the OOMKill happening during controller start up by increase the memory resources for the pod.

Signed-off-by: Leandro Mendes <lmendes@redhat.com>

PR #4483, opened 2024-09-10
fix(KFLUXBUGS-1607): increase memory limits for integration service

dirgim:
* The service manager pod is getting OOMKilled because of increased amout of resources on the production cluster(s)
* Bring the memory limits in line with similar Konflux services

### DIFF
--- a/components/integration/production/base/manager_resources_patch.yaml
+++ b/components/integration/production/base/manager_resources_patch.yaml
@@ -11,10 +11,10 @@ spec:
         resources:
           limits:
             cpu: 600m
-            memory: 1500Mi
+            memory: 2048Mi
           requests:
             cpu: 200m
-            memory: 600Mi
+            memory: 1024Mi
         env:
         - name: CONSOLE_NAME
           valueFrom:


Josh-Everett (approved):

components/integration/production/base/manager_resources_patch.yaml
```diff
@@ -11,10 +11,10 @@ spec:
         resources:
           limits:
             cpu: 600m
-            memory: 1500Mi
+            memory: 2048Mi
```
dheerajodha, line 14: +1, but how did you decide on these values?

dirgim: It's in line with what the other services have set (e.g. see #4380), and we've been seeing memory spikes of up to 1.7 G when starting or resyncing the service (which is the most memory demanding part of the lifecycle).

dheerajodha: thank you!

dheerajodha (approved):
/lgtm

PR #4484, opened 2024-09-10
mintmaker: add the new ExternalSecret to kustomization resources

qixiang:


qixiang:
I forgot to add the new ExternalSecret to kustomization resources in #4481.

PR #4487, opened 2024-09-10
Releasing Workspaces alpha6

dperaza4dustbit:
Adding only staging for now, prod will come after we verify that workspace availability metrics are showing correctly in our dashboards.

### DIFF
--- a/components/workspaces/staging/base/server/config/server/kustomization.yaml
+++ b/components/workspaces/staging/base/server/config/server/kustomization.yaml
@@ -1,0 +1,22 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- deployment.yaml
+- service.yaml
+configMapGenerator:
+- behavior: merge
+  files:
+  - traefik.yaml=./proxy-config/traefik.yaml
+  name: traefik-sidecar-static-config
+  options:
+    disableNameSuffixHash: true
+- behavior: merge
+  files:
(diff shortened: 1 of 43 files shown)


components/workspaces/staging/base/server/config/server/route.yaml
sadlerap, line 1: I think this needs to be removed, since we don't expose the route in internal staging.

components/workspaces/staging/base/server/config/server/kustomization.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- deployment.yaml
+- service.yaml
+- route.yaml
```
sadlerap, line 6: Take this out as well.

```suggestion
```

dperaza4dustbit: Got it

dperaza4dustbit: since we need to bring back when we move to base up a level, should I just remove from customize file reference?

sadlerap: Nope, we also have to do this for production environments, so let's just do it correctly here.  Any route we need to have lives in the environment folder underneath `staging/` or `production/`.

sadlerap (approved):
/lgtm

Let's merge this and see if anything breaks.

PR #4489, opened 2024-09-11
KFLUXINFRA-839: Cleaning up configs for stone-stg-m01

manish-jangra:
As we're decommissioning `stone-stg-m01` cluster, this requires cleaning up configs from infra-deployment repo, this is what this PR will do.


bamachrn (approved):
/lgtm

PR #4495, opened 2024-09-11
fix MPC provision script

mshaposhnik:


ifireball:
@mshaposhnik this kind of PR is usually created automatically when we merge stuff in MPC -is this PR a temporary change you intend to revert?

PR #4497, opened 2024-09-11
Promoting MintMaker from stage to prod

gnaponie:


components/mintmaker/production/base/kustomization.yaml
```diff
@@ -3,15 +3,15 @@ kind: Kustomization
 resources:
   - ../../base
   - ../../base/external-secrets
-  - https://github.com/konflux-ci/mintmaker/config/default?ref=3f8721be249d10e09cefea11a945dd5fe6714615
-  - https://github.com/konflux-ci/mintmaker/config/renovate?ref=3f8721be249d10e09cefea11a945dd5fe6714615
+  - https://github.com/konflux-ci/mintmaker/config/default?ref=3fa490ed3ca9180c408771c8a9731636ed669ea3
+  - https://github.com/konflux-ci/mintmaker/config/renovate?ref=3fa490ed3ca9180c408771c8a9731636ed669ea3
```
qixiang, line 7: This has not been deployed to stage yet, we need to test and verify it in stage first.
I'd suggest we release the changes for supporting secure registries together, this is still one PR for it under review.

components/mintmaker/production/base/manager_patches.yaml
```diff
@@ -17,4 +17,4 @@ spec:
             memory: 256Mi
         env:
         - name: RENOVATE_IMAGE
-          value: "quay.io/konflux-ci/mintmaker-renovate-image:333b7ba"
+          value: "quay.io/konflux-ci/mintmaker-renovate-image:e8ede3a"
```
qixiang, line 20: This image has not been related yet, and we need to test it in stage first. The release is still in-progress.

components/mintmaker/production/base/kustomization.yaml
staticf0x: Could we release an older commit: https://github.com/konflux-ci/mintmaker/commit/a7d0477e36c00266fd4ff32b96ea1c1024aa8f66 or is it better to postpone?

qixiang: yes, we can. Once we're ready to release the support for secure registries, we can include both the change for `inheritConfig` and the secure registries in the same announcement.

qixiang (approved):
LGTM

staticf0x:
The image is still the newest one, isn't it?

gnaponie:
> The image is still the newest one, isn't it?

correct.

PR #4508, opened 2024-09-12
mintmaker: update secret path for registry.state.redhat.io

qixiang:


qixiang:
The secret can't be created under the old path (because `production/qe/konflux-e2e-secrets` is already a secret), so QE team helped to create a new secret.

qixiang:
I'll add the label of `mintmaker.appstudio.redhat.com/secret-type: registry` after this is deployed and verified the secret is in good format, we don't have permission to check vault directly :( .

gnaponie (approved):

PR #4511, opened 2024-09-12
mintmaker: add label to stage registry pull secret

qixiang:


qixiang:
I've verified the secret is stored in good format, adding the label so mintmaker can consume it.

PR #4512, opened 2024-09-12
Remove of chain deep-inspection config option

enarha:
There is a change in the Tekton operator switching the supported value of Chains' enable-deep-inspection from bool to string ("true"/"false replace true/false). The update of OSP got blocked on some of the clusters because the TektonCOnfig value happened faster than the Catalog update and the actual upgrade. The upgrade was blocked because the old version of the operator does not support the new value in the tektonconfig. As a workaround, we remove that value, let the upgrade finish and in a following PR we switch to the new value.

### DIFF
--- a/components/pipeline-service/development/main-pipeline-service-configuration.yaml
+++ b/components/pipeline-service/development/main-pipeline-service-configuration.yaml
@@ -1485,7 +1485,6 @@ metadata:
 spec:
   chain:
     artifacts.oci.storage: oci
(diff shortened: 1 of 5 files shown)


lcarva (changes requested):
Nack on this. The impact here is that some images will go unsigned and unattested making them un-releasable.

enarha:
We need that workaround only for one of the staging clusters which I'm told is going to be decommisioned in the next few days, thus the change is not required and I'm closing it.

PR #4526, opened 2024-09-16
fix(KFLUXINFRA-836): STAGE: Align MPC image 2 src

ifireball:
Align the version of the MPC image we're using in STAGE and DEV to the latest MPC source.

### DIFF
--- a/components/multi-platform-controller/base/kustomization.yaml
+++ b/components/multi-platform-controller/base/kustomization.yaml
@@ -11,8 +11,8 @@ resources:
 
 images:
 - name: multi-platform-controller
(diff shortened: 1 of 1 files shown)


hugares:
the change LGTM, I will let @mshaposhnik vote if he is done testing on staging with temporary version he put there

ifireball:
> the change LGTM, I will let @mshaposhnik vote if he is done testing on staging with temporary version he put there

@hugares according to a thread in slack, the test version is still needed because it includes konflux-ci/multi-platform-controller#283. I think we need to go ahead and get that PR merged so we can stop having this kind of a test hack.

mshaposhnik (approved):

PR #4539, opened 2024-09-17
KFLUXINFRA-837: Changing AWS GPU Instance Type

manish-jangra:
Changing AWS GPU Instance Type to `g4dn.xlarge`


bamachrn (approved):
/lgtm

PR #4560, opened 2024-09-19
Set cpu/memory request and memory limit for tkn controllers in staging

hugares:
The tekton-pipelines-controller pod did not have request and limit set and this is causing the pods to be evicted from the nodes where they run in production. Not having request make those pods ideal candidates for eviction and when this happens, it cause a slight outage during restart of the controllers as they need to process the queue of pipeline runs.

Set cpu/memory request and memory limit in staging first, a follow up change will be done for production.

The cpu limit was not set intentionally as setting CPU limit in k8s is not a good idea and even an anti-pattern. Many articles are available out there explaining why but here is one of them[1].

The resources are different for stg-rh01 as we want them to be the same as production to be able to do load testing on that staging cluster.

[1]https://home.robusta.dev/blog/stop-using-cpu-limits

[SRVKP-6410](https://issues.redhat.com//browse/SRVKP-6410)

### DIFF
--- a/components/pipeline-service/staging/base/main-pipeline-service-configuration.yaml
+++ b/components/pipeline-service/staging/base/main-pipeline-service-configuration.yaml
@@ -1562,6 +1562,18 @@ spec:
                       requests:
                         cpu: 200m
                         memory: 200Mi
+        tekton-pipelines-controller:
+          spec:
+            template:
+              spec:
+                containers:
+                  - name: tekton-pipelines-controller
+                    resources:
+                      requests:
+                        memory: 1Gi
+                        cpu: 100m
+                      limits:
+                        memory: 2Gi
       disabled: false
       horizontalPodAutoscalers:
         tekton-operator-proxy-webhook:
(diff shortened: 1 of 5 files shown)


components/pipeline-service/staging/base/main-pipeline-service-configuration.yaml
```diff
@@ -1562,6 +1562,18 @@ spec:
...
                         memory: 200Mi
+        tekton-pipelines-controller:
+          spec:
+            template:
+              spec:
+                containers:
+                  - name: tekton-pipelines-controller
+                    resources:
+                      requests:
+                        memory: 500Mi
```
enarha, line 1573: You have to set this to the value you want and run the generate script to update the deploy.yaml files. It looks like you've edited them manually and the next time someone runs the generate script, it'll override those values.

hugares:
/hold
I will change values for rh01

enarha (commented):
/lgtm

PR #4562, opened 2024-09-19
Set cpu/memory request and memory limit for tkn controllers in prod

hugares:
The tekton-pipelines-controller pod did not have request and limit set and this is causing the pods to be evicted from the nodes where they run in production. Not having request make those pods ideal candidates for eviction and when this happens, it cause a slight outage during restart of the controllers as they need to process the queue of pipeline runs.

The cpu limit was not set intentionally as setting CPU limit in k8s is not a good idea and even an anti-pattern. Many articles are available out there explaining why but here is one of them[1].

This commit is a follow up of this commit[2] where the same was changed but for staging.

[1] https://home.robusta.dev/blog/stop-using-cpu-limits
[2] 90ee9cb7a3ce13e58e53a695f48c53597649c350

[SRVKP-6410](https://issues.redhat.com//browse/SRVKP-6410)

### DIFF
--- a/components/pipeline-service/production/base/main-pipeline-service-configuration.yaml
+++ b/components/pipeline-service/production/base/main-pipeline-service-configuration.yaml
@@ -1566,6 +1566,18 @@ spec:
                     requests:
                       cpu: 200m
                       memory: 200Mi
+        tekton-pipelines-controller:
+          spec:
+            template:
+              spec:
+                containers:
+                  - name: tekton-pipelines-controller
+                    resources:
+                      requests:
+                        memory: 6Gi
+                        cpu: "1"
+                      limits:
+                        memory: 8Gi
       disabled: false
       horizontalPodAutoscalers:
         tekton-operator-proxy-webhook:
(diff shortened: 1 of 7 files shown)


components/pipeline-service/production/base/main-pipeline-service-configuration.yaml
```diff
@@ -1566,6 +1566,18 @@ spec:
...
+        tekton-pipelines-controller:
+          spec:
+            template:
+              spec:
+                containers:
+                  - name: tekton-pipelines-controller
+                    resources:
+                      requests:
+                        memory: 6Gi
+                        cpu: 500m
```
enarha, line 1578: Is that value enough? I see on the rh01 cluster it's going above 500m (530m) and maybe we need to leave some extra limit.

hugares: I changed it to 1 CPU, I saw peaks around 800mi in the last 2 weeks

enarha (commented):
/lgtm

PR #4583, opened 2024-09-23
feat(KONFLUX-4508): Expose metrics for Pipelines leases distribution alerts

jhutar:
This is a dependency of https://github.com/redhat-appstudio/o11y/pull/388

Regarding how many labels combinations this mean, on `stone-stg-rh01` during last week I see 51. There are actually only 2 labels that change over time:

* `lease` - if we do not reconfigure Tekton Pipelines Controller to use different buckets count, there will be just 4 of these for every reconciler (so 12)
* `lease_holder` - this is a Pipelines Controller pod identification, so every time that pod restarts or so, new label value appears here (these pods are supposed to be stable, always running and restarts will be matter of upgrades or signs of cluster issues)

### DIFF
--- a/components/monitoring/prometheus/base/monitoringstack/monitoringstack.yaml
+++ b/components/monitoring/prometheus/base/monitoringstack/monitoringstack.yaml
@@ -49,7 +49,8 @@ spec:
           storageclass|volumename|release_reason|instance|result|deployment_reason|\
           validation_reason|strategy|succeeded|target|name|method|code|sp|le|\
           unexpected_status|failure|hostname|label_app_kubernetes_io_managed_by|status|\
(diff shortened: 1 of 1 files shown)


jhutar:
Hello @yftacherzog . Please could you review this PR as well?

yftacherzog:
> Hello @yftacherzog . Please could you review this PR as well?

I have limited capacity over the last few days as I'm off most of the time. @amisstea will you be able to find someone who could have a look?

amisstea:
I confirmed there's a limited set of time series (e.g. 12 on prod rh01) that would be remote written as a result of this change so it shouldn't cause issues with our RHOBS quota.

/lgtm

amisstea (approved):

hmariset (approved):

mftb (approved):
/lgtm
/approve

PR #4602, opened 2024-09-26
Update OSP version on stone-prd-m01

enarha:
Update already verified on the staging clusters. We start with updating a single prod clusters to verify the process and MPC update after initially encountering some problems with the staging clusters.


hugares:
@ifireball Can you update this PR to include the MPC  update so they both goes in at the same time

ifireball:
> @ifireball Can you update this PR to include the MPC update so they both goes in at the same time

done.

ifireball (approved):
/lgtm

PR #4603, opened 2024-09-26
Sending traces to SignalFX instead of Honeycomb

mftb:
Honeycomb is being shut down, sending traces to SignalFX instead.


hugares (approved):
/lgtm

PR #4606, opened 2024-09-27
Increase tekton results memory

hugares:
In production, on rh01, the container got OOMKilled. Increase from 6 to 8 and also set staging to same values. We want staging to be a mirror of prod.

### DIFF
--- a/components/pipeline-service/staging/stone-stage-p01/deploy.yaml
+++ b/components/pipeline-service/staging/stone-stage-p01/deploy.yaml
@@ -1646,10 +1646,10 @@ spec:
         resources:
           limits:
             cpu: 250m
-            memory: 4Gi
+            memory: 8Gi
           requests:
             cpu: 250m
-            memory: 4Gi
+            memory: 8Gi
         securityContext:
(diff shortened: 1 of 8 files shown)


components/pipeline-service/staging/stone-stage-p01/deploy.yaml
```diff
@@ -2063,12 +2063,12 @@ spec:
...
-                    whenUnsatisfiable: DoNotSchedule
-                    labelSelector:
-                      matchLabels:
-                        app: tekton-pipelines-controller
+                - labelSelector:
+                    matchLabels:
+                      app: tekton-pipelines-controller
+                  maxSkew: 1
+                  topologyKey: topology.kubernetes.io/zone
+                  whenUnsatisfiable: DoNotSchedule
```
hugares, line 2071: Regenerating the files using script change the order as when this was added, it was not done with the script but this is a no-op change, only order is changing

dperaza4dustbit (approved):
/lgtm

scoheb (approved):

PR #4612, opened 2024-09-27
kubesaw: add network policy for in-namespace communication

sadlerap:
Host operator needs to talk to registration-service to check health endpoints to set the status of the toolchainstatus resource.  However, since there are network policies defined on registration service, host-operator is disallowed from making network connections to registration-service, causing the toolchainstatus resource to report that registration service was unavailable.

To fix this, we need a network that allows pods within toolchain-host-operator to talk with each other.  This policy was already defined on the external clusters and the p01 internal cluster, but not on the others.


alexeykazakov (approved):

MatousJobanek (approved):

dperaza4dustbit (approved):
/lgtm

PR #4621, opened 2024-09-30
adding cloud-init configuration for gpu instances

manish-jangra:
This PR will take care of the followings --

- cloud-init config will take care of configuring nvme disk attached to the gpu instance type `g6.xlarge`
- this configures a new custom ami which is built from fedora 40.
- this will also generate nviDIa configuration to run gpu builds.


hugares (approved):
/lgtm

PR #4624, opened 2024-10-01
Change KubeArchive namespace because of ROSA

rh-hemartin:
This PR aims to change the namespace of KubeArchive because ROSA does not support namespaces for applications that are `kube.*`, which matches `kubearchive`. I changed it to be `product-kubearchive` given that we don't have any other name and I didn't want to use our shorthand (`k9e`). I tested it on CRC with the `prepare-crc.sh` script and then the `bootstrap-cluster.sh preview`.


rh-hemartin:
I added the `konflux-kubearchive` group to the list of groups that can see the clusters.

hugares (approved):
/lgtm

hugares (approved):
/lgtm

PR #4627, opened 2024-10-01
Increase maximum number of aarch64 nodes can spin up for public prod

bamachrn:
increasing the number nodes in the cluster


hugares (approved):
/lgtm

PR #4635, opened 2024-10-01
Fix kubearchive database secret extraction

hugares:
ExternalSecret are using go templates when using template to create the target secret. If the key to access contains a dot (e.g. db.user), we must use an index function to access it otherwise is is looking for an object(db) containing another object (user).


rh-hemartin (approved):
Looks good to me, thanks Hugo

PR #4637, opened 2024-10-02
feat(KONFLUX-4508): Run more Pipelines Controller replicas across more zones in Prod

jhutar:
This was deployed in Stage by c9f164e835f22ce89bed625a97f23097a0572668

### DIFF
--- a/components/pipeline-service/production/stone-prd-rh01/deploy.yaml
+++ b/components/pipeline-service/production/stone-prd-rh01/deploy.yaml
@@ -2055,6 +2055,13 @@ spec:
                     requests:
                       cpu: "1"
                       memory: 6Gi
+                topologySpreadConstraints:
+                - labelSelector:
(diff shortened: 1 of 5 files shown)


components/pipeline-service/production/stone-prd-rh01/deploy.yaml
hugares, line 1: These deploy file are generated from a script, I just ran the script and the order of the lines is not the same. Can you regenerate the deploy files by running the script:
```
./hack/generate-deploy-config.sh -c components/pipeline-service/production
```

jhutar: Oooh, there is a script for it! Always did it manually :))) Should be fixed now. Thank you!

hugares (approved):
/lgtm

PR #4638, opened 2024-10-02
Fix DB secret name for KubeArchive

rh-hemartin:
This PR changes the name of the Secret created by ExternalSecret to the correct one. It needs to be called `kubearchive-database-credentials` not `kubearchive-database-secret`.


hugares (approved):
/lgtm

PR #4640, opened 2024-10-02
Fix DB secret keys for KubeArchive

rh-hemartin:
A key for the KubeArchive's DB secret was wrong, sorry about the multiple PRs.


hugares (approved):
/lgtm

PR #4644, opened 2024-10-02
release workspaces v0.1.0-alpha7 to staging

sadlerap:
This release of workspaces implements support for two labels on workspaces.konflux-ci.dev resources.

- `internal.workspaces.konflux-ci.dev/is-owner` indicates whether the requesting user is or is not an owner of the workspace.  Contains the values `true` or `false`.
- `internal.workspaces.konflux-ci.dev/has-direct-access` indicates whether the requesting user has been given access directly or not. With Kubesaw as a backend, this means that there is at least one SpaceBinding granting a user access to the Space.  Contains the values `true` or `false`.

This also fixes an issue where the data field of the `workspaces-metrics-reader` secret was assigned ownership to ArgoCD, which was causing workspaces to read as out-of-sync.

Full release notes can be found [here](https://github.com/konflux-workspaces/workspaces/releases/tag/v0.1.0-alpha7).

### DIFF
--- a/components/workspaces/staging/base/kustomization.yaml
+++ b/components/workspaces/staging/base/kustomization.yaml
@@ -1,0 +1,5 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- operator/config/default
+- server/config/default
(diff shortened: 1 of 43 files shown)


filariow (approved):
/lgtm

PR #4645, opened 2024-10-02
Use MCE provided images for the hypershift CLI

amisstea:
The previously used image repo unexpectedly removed tags for OCP versions. The MCE operator provides its own HO image which should provide better guarantees of compatibility since the image tag corresponds to the version of MCE deployed.


hmariset (approved):
/lgtm

PR #4646, opened 2024-10-02
feat(RELEASE-1192): allow rhtap-releng to use es

scoheb:
- this allows ExternalSecrets to work in the rhtap-releng-tenant


hugares (approved):
/lgtm

PR #4668, opened 2024-10-08
add Workspace CRD in staging's member cluster

filariow:
To allow end users to retrieve the Custom Resource Definition for Workspaces  from the proxy, it needs to be deployed in the member clusters also.

A subsequent PR for production will follow


sadlerap (approved):

hugares (approved):
/lgtm

PR #4670, opened 2024-10-08
feat(KFLUXINFRA-889): Proactive Autoscaler Configuration for Staging Overlays

manish-jangra:
This Pull Request will make the following changes -

1. Create Priority Class
2. Dummy Deployment for Proactive Scaling

This will create two priority classes one that sets the highest priority to all the pods if not set already and another class that sets the least priority but doesn't apply that to the pods unless explicitely specified. 

This will also create 3 deployment pods for each node pool (m6a.2xlarge, m6a.4xlarge and c5d.4xlarge), these will be fake pods with least priority set to them, their job will be to influence the autoscaler to add more nodes (1 for node pool for now) and when actual workload is generated by konflux builds, these fake pods will be evicted making the way to schedule tenant workload and moving on to ask autoscaler to add more nodes to cluster.

This way, we will always have one node per node pool ready all the time.

### DIFF
--- a/argo-cd-apps/base/member/infra-deployments/proactive-scaler/proactive-scaler.yaml
+++ b/argo-cd-apps/base/member/infra-deployments/proactive-scaler/proactive-scaler.yaml
@@ -1,0 +1,40 @@
+apiVersion: argoproj.io/v1alpha1
+kind: ApplicationSet
+metadata:
+  name: proactive-scaler
+spec:
+  generators:
+    - merge:
+        mergeKeys:
+          - nameNormalized
+        generators:
+          - clusters:
+              values:
+                sourceRoot: configs/proactive-scaler
+                environment: staging
+          - list:
+              elements: []
+  template:
+    metadata:
+      name: proactive-scaler-{{nameNormalized}}
+    spec:
+      project: default
+      source:
+        path: '{{values.sourceRoot}}/{{values.environment}}'
+        repoURL: https://github.com/redhat-appstudio/infra-deployments.git
+        targetRevision: main
+      destination:
+        namespace: default
+        server: '{{server}}'
+      syncPolicy:
+        automated:
+          prune: true
+          selfHeal: true
+        syncOptions:
+        - CreateNamespace=true
+        retry:
+          limit: -1
+          backoff:
+            duration: 10s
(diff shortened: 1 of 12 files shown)


argo-cd-apps/base/member/infra-deployments/proactive-scaler/proactive-scaler.yaml
hugares, line 1: Let's delete this application from the development overlay too

hugares (changes requested):

configs/proactive-scaler/base/deployments.yaml
```diff
@@ -0,0 +1,96 @@
...
+      containers:
+        - name: reserve-resources
+          image: registry.k8s.io/pause
+          resources:
+            requests:
+              memory: "45Gi"
+              cpu: "10"
+---
+apiVersion: apps/v1
+kind: Deployment
```
jhutar, line 67: Hello @hugares ! I thought we will just create CronJob that will trigger a Job with 1 second pause every minute or so. I do not understand this too much, but this Deployment will just create pods that will be handing there forever, so they will not make the resources available. Or does it work differently?

hugares: This change is creating 2 priority classes, default one(for all pods) will have higher priority that the dummy deployments. Dummy deployments get schedule on node A, as soon as a pod need to be scheduled, that does not fit on any other node than node A, then dummy pods get evicted(takes less than a second) and pod get scheduled on node A. The dummy pod, will be pending and cause the autoscaler to add another node.

hugares: I know we talked about cron but this solution is more elegant and has the advantage to not having to wait until cron runs again to cause auto scaling

argo-cd-apps/base/member/infra-deployments/proactive-scaler/proactive-scaler.yaml
manish-jangra: removed from development overlays.

hugares (approved):
/lgtm

hugares:
@manish-jangra PR remove the hold when you are ready to merge it

PR #4673, opened 2024-10-09
KFLUXINFRA-812: Dynamic host for s390x in external stage

bamachrn:
This is for testing dynamic s390x in stage


ifireball (approved):
lgtm, this might solve some issues I've been having

PR #4677, opened 2024-10-09
KFLUXINFRA-799: Creating Cluster Secret Store for Insights

manish-jangra:
Creating Cluster Secret Store for Insights to retreive secrets in konflux build pipelines.

### DIFF
--- a/components/cluster-secret-store/base/insights-secret-store.yaml
+++ b/components/cluster-secret-store/base/insights-secret-store.yaml
@@ -1,0 +1,30 @@
+---
+apiVersion: external-secrets.io/v1beta1
+kind: ClusterSecretStore
+metadata:
+  name: insights-appsre-vault
+  annotations:
+    argocd.argoproj.io/sync-options: SkipDryRunOnMissingResource=true
+spec:
+  provider:
+    vault:
+      server: "https://vault.ci.ext.devshift.net"
+      path: insights
+      version: v1
+      auth:
+        # VaultAppRole authenticates with Vault using the
+        # App Role auth mechanism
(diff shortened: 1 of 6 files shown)


components/cluster-secret-store/production/insights-approle-id-patch.yaml
```diff
@@ -0,0 +1,4 @@
+---
+- op: add
+  path: /spec/provider/vault/auth/appRole/roleId
+  value: 573e6024-0c58-7e95-365b-d2ad6127e838
```
hugares, line 4: Is it intentional to have the same app-role in stating and production?

skateman: I'm not sure if the secret is needed in staging at all, but we don't have a separate staging one. I guess it's safe as we use the secrets only for RHEL subscription and it does not matter where the built containers get the subscription context from.

components/cluster-secret-store/base/insights-secret-store.yaml
```diff
@@ -0,0 +1,30 @@
...
+        # App Role auth mechanism
+        # https://www.vaultproject.io/docs/auth/approle
+        appRole:
+          # Path where the App Role authentication backend is mounted
+          path: approle
+          # RoleID configured in the App Role authentication backend
+          roleId: 
+          # Reference to a key in a K8 Secret that contains the App Role SecretId
+          secretRef:
+            name: insights-appsre-vault
```
gbenhaim, line 25: can we deploy this secret from our vault? (not a blocker for this pr)

PR #4618, opened 2024-09-30
Move cachi2 pipelines from tekton-ci to konflux-ci

slimreaper35:
Related PR: https://github.com/containerbuildsystem/cachi2/pull/665

PR #4695, opened 2024-10-11
Revert "Move cachi2 pipelines from tekton-ci to konflux-ci (#4618)"

eskultety:
This reverts commit 3076d38f801845b322d9d14bc01584659c531014.

We don't have all the pieces needed in place yet and the change in question would break cachi2's release process. Revert the change until cachi2 is truly ready to make the change. This unblocks release of cachi2 which is pending.


MartinBasti:
@psturc can you please force merge? You did the same with original PR. TY

PR #4714, opened 2024-10-15
increase concurrency limits for platforms RHEL team is using

brianwcook:
increase concurrency limits for platforms RHEL team is using on p02

### DIFF
--- a/components/multi-platform-controller/production-downstream/host-config.yaml
+++ b/components/multi-platform-controller/production-downstream/host-config.yaml
@@ -60,7 +60,7 @@ data:
   dynamic.linux-arm64.aws-secret: aws-account
   dynamic.linux-arm64.ssh-secret: aws-ssh-key
   dynamic.linux-arm64.security-group-id: sg-0903aedd465be979e
-  dynamic.linux-arm64.max-instances: "50"
+  dynamic.linux-arm64.max-instances: "250"
   dynamic.linux-arm64.subnet-id: subnet-0aa719a6c5b602b16
 
   dynamic.linux-mlarge-arm64.type: aws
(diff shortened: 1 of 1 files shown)


components/multi-platform-controller/production-downstream/host-config.yaml
```diff
@@ -60,7 +60,7 @@ data:
   dynamic.linux-arm64.aws-secret: aws-account
   dynamic.linux-arm64.ssh-secret: aws-ssh-key
   dynamic.linux-arm64.security-group-id: sg-0903aedd465be979e
-  dynamic.linux-arm64.max-instances: "50"
+  dynamic.linux-arm64.max-instances: "250"
```
hugares, line 63: we should increase amd64 too

arewm:
Can we support this many parallel instances in IBM Cloud? Were there discussions that were happening to try to prepare for increasing the max count?

components/multi-platform-controller/production-downstream/host-config.yaml
brianwcook: why? it is a local platform.

arewm: `amd64` isn't local, it is a provisioned VM. The local platforms are defined above. `x86_64` is an example of a local one.

hugares, line 63: We cannot increase that before we get a new CIDR block and increase the number of IPs on AWS side, we only have 128 ips for now. The actual config is already broken, if we sum up the number of VMs that can be spawned concurrently, it is already more than 128

brianwcook: I'm going to drop the PR then because I opened it based on bad info

PR #4718, opened 2024-10-15
Fix sync issues in the cluster-as-a-service component

amisstea:
- ArgoCD would get stuck trying to sync a resource whose CRD didn't yet exist.
- A sync wave was added to the multicluster-engine namespace to speed up the overall sync process.


hmariset (approved):
/lgtm

PR #4722, opened 2024-10-16
Increasing MPC VM node count to 100 for internal prod cluster

bamachrn:
Increasing this to 100 VMs in each arch.


manish-jangra (approved):
/lgtm

PR #4723, opened 2024-10-16
fix(KFLUXBUGS-1735): increase memory limits for integration service

dirgim:
* The service manager pod is getting OOMKilled because of increased amout of resources on the production cluster(s)
* Increase the limit to 4096Mi as agreed on architecture calls
* Increase the requests to 2048Mi as that's the new expected consumption


jencull (approved):
/lgtm

chipspeak (approved):

jsztuka (approved):
lgtm

PR #4736, opened 2024-10-16
Deploy insight vault only to RH member clusters

hugares:
We do not need this vault on host, m01 and eaas cluster in production neither we need this vault on any staging clusters.

Fix this by moving the clusterscretstore to cluster-secret-store-rh application, which is only deployed on production rh member clusters.

[KFLUXBUGS-1739](https://issues.redhat.com//browse/KFLUXBUGS-1739)


mshaposhnik (approved):

manish-jangra (approved):
/lgtm

PR #4743, opened 2024-10-17
fix(KFLUXBUGS-1600): internalrequests for timed out pipelineruns are not removed

theflockers:
this PR add a new CronJob to prune internalrequests that are left from timed out pipelineruns, 24 hours after their creation.


johnbieren (approved):
/lgtm
/approve

PR #4750, opened 2024-10-18
Enabling Proactive Autoscaling for the Production Clusters

manish-jangra:
This Pull Request will enable the proactive autoscaling for the production overlay and disable it for the stating overlay.

Related Pull Request that was implemented for the staging environment --

- https://github.com/redhat-appstudio/infra-deployments/pull/4670


hugares (approved):
/lgtm

configs/proactive-scaler/base/deployments.yaml
```diff
@@ -62,35 +62,3 @@ spec:
             requests:
               memory: "45Gi"
               cpu: "10"
----
```
hugares, line 65: @manish-jangra Let's remove 2x large from here since we remove that from p02 and we want to get rid of that from rh01 and p01 too

PR #4755, opened 2024-10-18
More replicas do not seem to improve the situation

arewm:
Reducing to one. Instead, we will see if multiple reconciliations will be any more successful konflux-ci/multi-platform-controller/pull/318

### DIFF
--- a/components/multi-platform-controller/production-downstream/replicas_patch.yaml
+++ b/components/multi-platform-controller/production-downstream/replicas_patch.yaml
@@ -4,7 +4,7 @@ metadata:
   name: multi-platform-controller
   namespace: multi-platform-controller
 spec:
-  replicas: 30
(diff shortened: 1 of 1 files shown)


components/multi-platform-controller/production-downstream/replicas_patch.yaml
```diff
@@ -4,7 +4,7 @@ metadata:
   name: multi-platform-controller
   namespace: multi-platform-controller
 spec:
-  replicas: 30
+  replicas: 1
```
hugares, line 7: if we reduce to 1, do we need to remove the pod affinity section, line 8 to 21?

mshaposhnik: we''ll remove it later after experiments

PR #4763, opened 2024-10-21
Fixup MPC container patch

mshaposhnik:
With adding the `rbac-proxy` container, the order is changed and patched one is now index `1`


psturc (approved):
/lgtm

PR #4764, opened 2024-10-21
Add Knative Event, KubeArchive dependency, as infra component

rh-hemartin:
I'm adding Knative Eventing because its a dependency for KubeArchive. There was an [ADR](https://github.com/konflux-ci/architecture/pull/202) to install it for everyone use, but my intention here is that is used by KubeArchive.

cc @ralphbean @hugares

### DIFF
--- a/components/knative-eventing/base/kustomization.yaml
+++ b/components/knative-eventing/base/kustomization.yaml
@@ -1,0 +1,35 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - https://github.com/knative/eventing/releases/download/knative-v1.15.3/eventing-core.yaml?timeout=90s
+
+patches:
+# kube-lint fixes
+- patch: |-
+    apiVersion: apps/v1
+    kind: Deployment
+    metadata:
+      name: eventing-controller
+      namespace: knative-eventing
+    spec:
+      template:
+        spec:
+          containers:
+            - name: eventing-controller
+              resources:
+                requests:
+                  cpu: 35m
+                  memory: 100Mi
+                limits:
+                  cpu: 100m
+                  memory: 200Mi
+# This was causing issues with kube-linter and I didn't want
+# to increase its replicas to 2, so I deleted it instead
+- patch: |-
+    $patch: delete
+    apiVersion: policy/v1
+    kind: PodDisruptionBudget
+    metadata:
+      name: eventing-webhook
+      namespace: knative-eventing
--- a/argo-cd-apps/base/member/infra-deployments/knative-eventing/knative-eventing.yaml
+++ b/argo-cd-apps/base/member/infra-deployments/knative-eventing/knative-eventing.yaml
@@ -1,0 +1,40 @@
+---
+apiVersion: argoproj.io/v1alpha1
+kind: ApplicationSet
+metadata:
+  name: knative-eventing
+spec:
+  generators:
+    - merge:
+        mergeKeys:
+          - nameNormalized
+        generators:
+          - clusters:
+              values:
+                sourceRoot: components/knative-eventing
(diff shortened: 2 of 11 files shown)


rh-hemartin:
I will submit corrections for the kube-lint problems.

argo-cd-apps/base/all-clusters/infra-deployments/knative-eventing/knative-eventing.yaml
```diff
@@ -0,0 +1,40 @@
+---
```
ralphbean, line 1: I think, rather than `argo-cd-apps/base/all-clusters/` you want this in `argo-cd-apps/base/member/`.

While we still have the kubesaw architecture, there is a notion of `host` cluster and `member` cluster. The host clusters just provide proxying and user management. All the workload and user namespaces are on the member clusters. I think you only want knative eventing on the member clusters, just like you want kubearchive only on the member clusters.

components/knative-eventing/base/kustomization.yaml
```diff
@@ -0,0 +1,35 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - https://github.com/knative/eventing/releases/download/knative-v1.15.3/eventing-core.yaml?timeout=90s
```
ralphbean, line 5: For the RH instances here, do we want to use the RH product version of knative eventing instead of the upstream build? I would assume we want to unless we've had deliberate conversations that concluded to the contrary.

ralphbean:
I think you may be missing important bits in `argo-cd-apps/overlays/*/delete-applications.yaml` and possibly also `argo-cd-apps/overlays/development/kustomization.yaml`

argo-cd-apps/base/all-clusters/infra-deployments/knative-eventing/knative-eventing.yaml
ralphbean: ... eventually, we're going to consolidate and drop the `host` clusters at which point the only kind of cluster will be `member` clusters, but it's just not in place yet.

components/knative-eventing/base/kustomization.yaml
rh-hemartin: KubeArchive needs Knative Eventing v1.15 and the latest OpenShift Serverless version is not there yet. It may be when we reach internal instances, so we may decide to use the OpenShift Serverless variant. For now we should stick to the upstream project.

argo-cd-apps/base/all-clusters/infra-deployments/knative-eventing/knative-eventing.yaml
rh-hemartin: Moved!

components/knative-eventing/base/kustomization.yaml
ralphbean: Ack. @hugares any opinion on this from your team's pov?

ralphbean:
@rh-hemartin this looks good to me, but I'd like @hugares to ack it as well to minimize any surprises.

hugares:
> @rh-hemartin this looks good to me, but I'd like @hugares to ack it as well to minimize any surprises.

I have a packed morning, I will check this later today

components/knative-eventing/base/kustomization.yaml
hugares: for  Tekton, we use Openshift Pipelines so following that, we should use OpenShift Serverless  but if serveless do not have the knative version we need, we have no other choice to go with upstream

hugares:
@rh-hemartin The knative application deployed but there a sync loop, some cluster role are out of sync, then sync, then out of sync and so on.

I did not have time to look into this, please have a look

PR #4767, opened 2024-10-21
remove workspaces from production clusters

sadlerap:
With the decision from upstream Konflux to not pursue a workspaces abstraction, it doesn't make sense to have workspaces deployed to the production clusters.  Staging will be removed once all of the dependencies on workspaces that exist there (UI) have been changed to no longer depend on workspaces.


hugares:
We are sure that the current UI, both for external cluster(hosted on consoledot) and for internal cluster(hosted in the konflux cluster) do not have any dependency on the workspaces?

sadlerap:
I don't believe so, but it'd be good to get confirmation from the UI team just to be sure.

hugares:
> We are sure that the current UI, both for external cluster(hosted on consoledot) and for internal cluster(hosted in the konflux cluster) do not have any dependency on the workspaces?

@sahil143 Can you confirm?

hugares (approved):

sahil143:
> > We are sure that the current UI, both for external cluster(hosted on consoledot) and for internal cluster(hosted in the konflux cluster) do not have any dependency on the workspaces?
> 
> @sahil143 Can you confirm?

Yes, We do not use the workspaces api.

PR #4768, opened 2024-10-21
Fix priority class of the proactive scaler pods

hugares:
This priority class is meant to be lower that priority class used by the tenant pods. The original change created a default class with priority value of 0 and scaler pods used a class with priority of -1.

The problem is all the tenants pods get assigned to sandbox-users-pods priority class which has a value of -3, causing the scaler pods to not be evicted from the node when resource are needed.

We are planning to remove sandbox so this sandbox-users-pods priority class will be deleted eventually but in the meantime, set scaler pods to -10.

Also rename the 2 priority classes so their name are more meaningful.

[KFLUXINFRA-889](https://issues.redhat.com//browse/KFLUXINFRA-889)

### DIFF
--- a/configs/proactive-scaler/base/deployments.yaml
+++ b/configs/proactive-scaler/base/deployments.yaml
@@ -22,7 +22,7 @@ spec:
           operator: "Equal"
           value: "konflux-tenants"
           effect: "NoSchedule"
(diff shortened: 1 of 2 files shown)


manish-jangra (approved):
/lgtm

PR #4774, opened 2024-10-23
Update production to the latest MPC with all recent fixes

mshaposhnik:
This change include:
 - Apply version with better concurrency on public cluster (showed good results on private);
 - Apply `s390x` 100Gb VM disc size.


hugares (approved):
/lgtm

PR #4779, opened 2024-10-24
Ignore Aggregated Roles in Knative Eventing

rh-hemartin:
#4764 Knative Eventing was introduced, but the deployment was in and out of sync constantly. This is because Knative Eventing uses aggregated roles. The following PR makes the Knative Eventing ArgoCD application ignore the roles that were being aggregated.

PD: I found no way to tell the `ignoreDifferences` option to ignore "all clusterroles that have `aggregateRules`". So I had to exclude them one by one.

cc @hugares


hugares (approved):
/lgtm

rh-hemartin:
Tide is "pending" but the test has already completed as far as I know, any way to reevaluate tide?

PR #4787, opened 2024-10-24
Make sure we have 2 worker nodes in advance

hugares:
Looking at the tekton task throttled by node, we can see that one node in advance is helping but not enough to get rid of task pending due to lack of nodes.


manish-jangra (approved):
/lgtm

PR #4791, opened 2024-10-25
Configure Pipelines as Code to require /ok-to-test on every new commit

mmorhun:


hugares:
I am ok with the change but is there a jira issue to get the context of why changing this or maybe explain in the commit/PR description

arewm:
I am okay with this change as well. We need to announce this change before we merge it as it will affect how users interact with the services. I have drafted up an email that we can use to make the announcement.

PR #4809, opened 2024-10-29
feat(KONFLUX-5124): Test with more VMs available

jhutar:
For testing purposes.

https://redhat-internal.slack.com/archives/C02CTEB3MMF/p1730181905884509


mshaposhnik:
It;s looks little too intense to me, let's sick at 20 for IBM ?

jhutar:
Hello @mshaposhnik . Shouldn't we test it at least at scale customers are going to use it? Of course we will put it back to 2 once we are happy with the test results.

mshaposhnik:
AFAIK prod now set at 30.. but yes, if it is temporary - let's try with 50 and see if we can increase prod up to that.

mshaposhnik (approved):

PR #4811, opened 2024-10-29
Increase tekton remote resolver resources to prevent OOMKilled events

jkhelil:
After increasing to 16Gi then decreasing to 4Gi remote resolver memory, we could see OOMKilled events for resolver pod
Increasing memory to 8Gi


hugares (approved):
/lgtm

hugares:
afecting prod, merging right away

PR #4812, opened 2024-10-29
let RememberOKToTest to be true

jkhelil:


enarha:
> Should be done through /hack/generate-deploy-config.sh

Indeed. Change https://github.com/redhat-appstudio/infra-deployments/blob/main/components/pipeline-service/production/base/update-tekton-config-pac.yaml#L10 . I also see this overriden for particular cluster https://github.com/redhat-appstudio/infra-deployments/blob/main/components/pipeline-service/production/stone-prod-p01/resources/update-tekton-config-pac.yaml#L10 . Not sure if the intention is to have that value different on the stone-prod-p01 cluster.

PR #4819, opened 2024-10-31
Bump hypershift-aws-template chart version

amisstea:
The imageContentSources chart value is made available as a result of this change.


hmariset (approved):
/lgtm

PR #4822, opened 2024-11-01
Update CatalogSource mem target to avoid eviction

dperaza4dustbit:
When nodes are overcommited pods that are over their requested memory can be evicted. Increasing the target memory will reduce this risk.


hugares (approved):
/lgtm

PR #4832, opened 2024-11-04
add additional params in build pipelines config for build-service

rcerven:
build-platforms for multiarch pipeline


arewm (approved):
Thanks for getting around to this before me. :)

PR #4839, opened 2024-11-05
Update OSP prod-m01

jkhelil:
Dev and Stage already verified.
Bumping prod-m01 in preparation for bumping all prod


enarha (commented):
You can do that (change the deploy.yaml directly), but if anyone adds another change and regenerate the yamls as they should, your change will be gone, which is not ideal. Instead you should add kustomize patch here https://github.com/redhat-appstudio/infra-deployments/tree/main/components/pipeline-service/production/stone-prd-m01/resources and regenerate the yaml, which will only change the deploy.yaml for that specific cluster.

hugares (approved):
/lgtm

PR #4842, opened 2024-11-05
Set number of buckets for OSP resolver controllers

hugares:
Increase the number of buckets so the 2 controllers can acquire leases.


hugares:
We are trying to get a work around together for prod issue, merging this one right away .

PR #4843, opened 2024-11-05
Increase OSP remote resolver replicas to 2

hugares:
This change is an attempt to improve the resolutionRequest time, we see a lot of resolution timing out. We cannot configure the number of threads per controller yet so increase the number of controllers.

[KFLUXBUGS-1784](https://issues.redhat.com//browse/KFLUXBUGS-1784)


hugares:
This is affecting prod, merging right away

PR #4850, opened 2024-11-06
fix(KFLUXBUGS-1764): update notification-controller production

avi-biton:
update notification-controller production


yftacherzog (approved):
/approve
/lgtm

PR #4866, opened 2024-11-07
fix(KFLUXINFRA-777): Adding a readonly variable in install.sh script

manish-jangra:
Adding a readonly variable for a lock file


bamachrn (approved):
LGTM

PR #4869, opened 2024-11-08
feat(KONFLUX-5352): Workaround frequent restarts on stone-prd-rh01

jhutar:


components/pipeline-service/production/base/main-pipeline-service-configuration.yaml
```diff
@@ -1576,10 +1576,10 @@ spec:
                   - name: tekton-pipelines-controller
                     resources:
                       requests:
-                        memory: 6Gi
+                        memory: 10Gi
                         cpu: "1"
                       limits:
-                        memory: 8Gi
+                        memory: 12Gi
```
hugares, line 1582: Let's start using new pattern of memory request=limits. The reason is  that the pod will be allowed to go higher than requests but will be considered in eviction process if node is under pressure.

I suggest to make both requests and limits 12Gi, WDYT?

hugares:
@jhutar We should change stating to be the same, this will help when we run performance test on staging to reflect production

components/pipeline-service/production/base/main-pipeline-service-configuration.yaml
jhutar: Yeah, I was aiming at smallest possible change, but you are right. When if not now? :-)

hugares (approved):
/lgtm

PR #4870, opened 2024-11-08
feat(tekton-pipelines-webhook): Long term usage is on par with requests, so give more

jhutar:


components/pipeline-service/production/base/main-pipeline-service-configuration.yaml
```diff
@@ -1566,8 +1566,8 @@ spec:
                       cpu: "1"
                       memory: 1Gi
                     requests:
-                      cpu: 200m
-                      memory: 200Mi
+                      cpu: 400m
+                      memory: 400Mi
```
hugares, line 1570: Same comment as in the other PR, let's start using new pattern of memory request=limits. The reason is that the pod will be allowed to go higher than requests but will be considered in eviction process if node is under pressure.

We also need to do the change for staging so staging is a perfect replica of production for our performance tests

jhutar: Right. Please check now.

hugares (approved):
/lgtm

PR #4872, opened 2024-11-08
fix: increase memory limit

theflockers:
This PR increases the memory limit to 8Gb to decrease the Pod restarts over the week.

### DIFF
--- a/components/release/k-components/manager-resources-patch/manager_resources_patch.yaml
+++ b/components/release/k-components/manager-resources-patch/manager_resources_patch.yaml
@@ -11,7 +11,7 @@ spec:
         resources:
           limits:
             cpu: 500m
-            memory: 4096Mi
+            memory: 8192Mi
           requests:
             cpu: 100m
-            memory: 3072Mi
+            memory: 8192Mi


mmalina (approved):

components/release/k-components/manager-resources-patch/manager_resources_patch.yaml
```diff
@@ -11,7 +11,7 @@ spec:
         resources:
           limits:
             cpu: 500m
-            memory: 4096Mi
+            memory: 8192Mi
```
hugares, line 14: I suggest to set memory request=limits. The reason is that the pod will be allowed to go higher than requests but will be considered in eviction process if node is under pressure. To really reduce the number of restart, set the request as well to 8192Mi

theflockers: thanks for the hint. Changed.

happybhati (approved):

components/release/k-components/manager-resources-patch/manager_resources_patch.yaml
theflockers: Although thinking now, would this not be risky in case the cluster can not give me the requests and the Pod would not spawn?

seanconroy2021 (approved):
LGTM

theflockers:
@hugares can you LGTM this one? Not sure why the ones given are not valid.

PR #4875, opened 2024-11-11
feat(KONFLUX-5297): Expose metrics for Pipelines leases distribution alerts

jhutar:


hugares:
@jhutar there was a problematic change at some point causing the e2e to fail, can you rebase this PR, failure might be caused by that problem which is fixed on latest main

jhutar:
Cool, it helped, thank you @hugares !

hugares (approved):
/lgtm

PR #4900, opened 2024-11-13
Configure MPC memory for staging and production

hugares:
Align MPC controller memory to be the same on all staging and production environments. Also reduce memory from 16Gi to 10Gi, rh01 is the most used cluster and memory is never going past 6Gi.


ifireball (approved):
/lgtm

PR #4903, opened 2024-11-13
patch to staging cluster-provisioner pod read

ascerra:
This PR should patch the clusterRole cluster-provisioner to also allow for getting pod logs.

### DIFF
--- a/components/cluster-as-a-service/staging/kustomization.yaml
+++ b/components/cluster-as-a-service/staging/kustomization.yaml
@@ -5,6 +5,8 @@ resources:
   - ../base
   - ../../openshift-gitops
   - external-secrets.yaml
+  - namespace-manager-pod-reader-role.yaml
+  - namespace-manager-pod-reader-binding.yaml
 patches:
   - path: add-hypershift-params.yaml
     target:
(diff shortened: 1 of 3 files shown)


components/cluster-as-a-service/staging/cluster-provisioner-read-pod-logs.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
```
amisstea, line 2: Patching the resource like this will replace it entirely. Use `kustomize build components/cluster-as-a-service/staging` to verify the resulting ClusterRole contains the rules from the base rbac plus your changes.

Take a look at this patch for an example of how to append to an array: https://github.com/redhat-appstudio/infra-deployments/blob/b22e629801630a927a1b0f5aed8d6dfea8319c9c/components/cluster-as-a-service/staging/add-hypershift-params.yaml

ascerra: Alright based on the newest changes I'm adding new resources instead of editing existing ones so I shouldn't be replacing anything.

PR #4914, opened 2024-11-15
Tune Build Service and Image Controller resources to prevent often eviction

mmorhun:


components/build-service/production/base/manager_resources_patch.yaml
```diff
@@ -0,0 +1,17 @@
...
+    spec:
+      containers:
+      - name: manager
+        resources:
+          limits:
+            cpu: 500m
+            memory: 2Gi
+          requests:
+            cpu: 250m
+            memory: 768Mi
```
hugares, line 17: if you want to prevent eviction, you must set memory requests=limits. If the memory usage is above requests, e.g. >=768Mi and the node is running low on memory(pod scheduling is done based on requests value, not limit), the pod will be considered for eviction.

That said. it the current proposed memory request is above typical usage, we can leave the change as is.

components/build-service/staging/base/manager_resources_patch.yaml
```diff
@@ -0,0 +1,17 @@
...
+    spec:
+      containers:
+      - name: manager
+        resources:
+          limits:
+            cpu: 500m
+            memory: 1Gi
+          requests:
+            cpu: 150m
+            memory: 512Mi
```
hugares, line 17: We are running performance test on staging, we should have identical values between staging and prod. Please set same values for mem/cpu requests/limits as for prod

components/image-controller/staging/base/manager_resources_patch.yaml
```diff
@@ -13,6 +13,6 @@ spec:
             cpu: 500m
             memory: 2Gi
           requests:
-            cpu: 100m
-            memory: 20Mi
+            cpu: 150m
+            memory: 512Mi
```
hugares, line 17: same comment as above: we are running performance test on staging, we should have identical values between staging and prod. Please set same values for mem/cpu requests as for prod

components/build-service/production/base/manager_resources_patch.yaml
mmorhun: I looked at current usage and gave slightly higher value than typical, this should prevent eviction. I still don't want to consume more than actually needed for now.

components/build-service/staging/base/manager_resources_patch.yaml
mmorhun: ok, will do

PR #4933, opened 2024-11-19
chore: promote project-controller to production

yftacherzog:
This introduces into production a change in which the controller watches also over the project and template resources.
https://github.com/konflux-ci/project-controller/pull/124


yftacherzog:
@ifireball I confirmed in stage that template changes are reconciled and that creating the project after the dev stream and template results in owner ref being created on the dev stream.

ifireball (approved):
:+1:

PR #4940, opened 2024-11-20
Run MintMaker every 4 hours instead of 2 hours

gnaponie:
Due to the amount of components on-boarding to Konflux, the load of MintMaker runs is increasing quite a lot. Let's reduce the frequency, at least untile we figure out how to improve the performances.


gnaponie:
@qixiang @staticf0x @querti @FernandesMF @nvtnlucie PTAL
There is some discussion in slack you can refer to for the reason of this change.

PR #4941, opened 2024-11-20
APPSRE-11137 add app-sre SecretStore

fishi0x01:
**Intent**

Adding a `SecretStore` for the `app-sre-tenant` namespace in **public** Konflux instance.
secretID is already provisioned manually in our own workspace `app-sre-tenant`.

Open questions:

- how can I make sure this is only provisioned in the public konflux instance?
~~- how do I provision the secretID in `appsre-vault`?~~

### DIFF
--- a/components/cluster-secret-store-rh/base/app-sre-secret-store.yaml
+++ b/components/cluster-secret-store-rh/base/app-sre-secret-store.yaml
@@ -27,7 +27,7 @@ spec:
           roleId: "" # will be added by the overlay
           # Reference to a key in a K8 Secret that contains the App Role SecretId
           secretRef:
-            name: app-sre-vault
+            name: secret-store-credentials
             key: secret-id
-            namespace: appsre-vault
+            namespace: app-sre-tenant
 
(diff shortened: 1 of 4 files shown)


components/cluster-secret-store-rh/base/app-sre-secret-store.yaml
```diff
@@ -0,0 +1,33 @@
+---
+apiVersion: external-secrets.io/v1beta1
+kind: SecretStore
```
hugares, line 3: @gbenhaim Since this is a secretStore, should we create this in infra-deployments? Tenants can create secretstore in their NS either manually or using gitops in the tenants repo

fishi0x01: Having this in konflux-release-data would also allow me to specify which konflux instance I want this applied in. Would be fine for me too :)

hugares: ok, let's put this in konflux-release-data, it make more sense since infra-deployments is the GitOps repo for Konflux itself, not the user payloads

fishi0x01: Got you. Will do :+1:

PR #4944, opened 2024-11-20
Temporarily adjust Renovate jobs to run every 4 hours from Friday to Sunday

qixiang:
This change will be reverted after the new release of the customized schedule for managers is deployed, planned for tomorrow.

The change is not applied to p01 and p02 clusters.


arewm (approved):
@hugares , does this look like a correct kustomization?

hugares (approved):
/lgtm

hugares:
> @hugares , does this look like a correct kustomization?

it does, I ran the kustomize build for all cluster and they are all ok

arewm:
Is the next cron scheduled to kick off at 11 AM ET? If so, we should make sure that this change is merged before hand.

PR #4948, opened 2024-11-20
mintmaker: fix invalid k8s cronjob schedule

qixiang:
The k8s CronJob schedule uses a syntax similar to Linux crontab, but it has a key difference: the day of the week field only accepts values from 0 to 6.


hugares (approved):
/lgtm

PR #4951, opened 2024-11-21
Promote mintmaker from stage to prod

qixiang:
This also removes the temporary adjustment to the Renovate job schedule.


qixiang:
/lgtm

This probably doesn't work (because this is a PR opened by myself), but just give it a try.

PR #4952, opened 2024-11-21
Add Vector for forwarding Pipelines log

khrm:
This is needed to enable a newer version of Results which is shipped to other customers.

For now, we are using play.min.io for dev environment.


khrm (commented):
/assign @enarha

hack/secret-creator/create-plnsvc-secrets.sh
```diff
@@ -35,34 +45,12 @@ create_s3_secret() {
...
-type: Opaque
-stringData:
-  config.env: |-
-    export MINIO_ROOT_USER="$USER"
-    export MINIO_ROOT_PASSWORD="$PASS"
-    export MINIO_STORAGE_CLASS_STANDARD="EC:1"
-    export MINIO_BROWSER="on"
-EOF
+      --from-literal=aws_access_key_id="Q3AM3UQ867SPQQA43P2F" \
+      --from-literal=aws_secret_access_key="zuf+tfteSlswRu7BJ86wekitnifILbZam1KYY3TG" \
```
Roming22, line 50: I would not recommend exposing a token. If it is removed from the PR, it should be rotated.

khrm: This is a public instance of min.io for testing purpose. It's already expose by minio.

components/pipeline-service/development/kustomization.yaml
```diff
@@ -8,55 +8,6 @@ commonAnnotations:
   argocd.argoproj.io/sync-options: SkipDryRunOnMissingResource=true
 
 resources:
-  - github.com/minio/operator?ref=v5.0.15
```
enarha, line 11: I remember this was recently changed by @amisstea , so I wonder if his team has any use of it. @amisstea please confirm we can remove that.

hack/secret-creator/create-plnsvc-secrets.sh
enarha: I was skeptical as well, but those credentials are publicly available here https://min.io/docs/minio/linux/administration/minio-console.html#logging-in , and that's considered a temporary solution, so I'm OK with it.

PR #4971, opened 2024-11-24
KFLUXINFRA-1058: Deploy the new UI to public stage

gbenhaim:
Deploy the new UI to the public staging cluster so it can be tested. The deployment supports connecting from a frontend which run on a local machine. This is useful when developing the frontend.

Authentication to the new UI is done using dex and the OCP connector.


yftacherzog (approved):
/lgtm

PR #5004, opened 2024-11-28
Fix Vector secret name

enarha:
This fixes the secret name used by Vector. It also unifies the names used in staging and development (CI).

### DIFF
--- a/components/pipeline-service/staging/base/main-pipeline-service-configuration.yaml
+++ b/components/pipeline-service/staging/base/main-pipeline-service-configuration.yaml
@@ -1957,27 +1957,27 @@ spec:
           - name: AWS_ACCESS_KEY_ID
             valueFrom:
               secretKeyRef:
-                name: s3-conf
+                name: tekton-results-s3
                 key: aws_access_key_id
           - name: AWS_SECRET_ACCESS_KEY
             valueFrom:
               secretKeyRef:
-                name: s3-conf
+                name: tekton-results-s3
                 key: aws_secret_access_key
           - name: AWS_DEFAULT_REGION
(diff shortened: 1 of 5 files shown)


khrm:
Vector one uses `s3-conf` in the `tekton-logging`  NS.  `tekton-results-s3` exists in the `tekton-results` NS.

enarha:
> Vector one uses `s3-conf` in the `tekton-logging` NS. `tekton-results-s3` exists in the `tekton-results` NS.

I only changed the name to match, it is the same secret, 2 different namespaces.

khrm (commented):
/lgtm

/approve

components/pipeline-service/staging/base/main-pipeline-service-configuration.yaml
```diff
@@ -1957,27 +1957,27 @@ spec:
           - name: AWS_ACCESS_KEY_ID
             valueFrom:
               secretKeyRef:
-                name: s3-conf
+                name: tekton-results-s3
```
hugares, line 1960: You need to change the ExternalSecret to create `tekton-results-s3` instead of `s3-conf` in tekton-logging NS: https://github.com/redhat-appstudio/infra-deployments/blob/main/components/pipeline-service/staging/base/main-pipeline-service-configuration.yaml#L1855

enarha: It does that https://github.com/redhat-appstudio/infra-deployments/blob/main/components/pipeline-service/staging/base/main-pipeline-service-configuration.yaml#L1868 ... unless I'm missing something

PR #5017, opened 2024-11-29
Fix Environment Variables for tekton results api and s3

khrm:
Blob SDK Go use by results uses standard enviornment. This changes to use use that.


khrm (commented):
/assign @mshaposhnik @enarha @hugares

enarha (approved):
/lgtm

PR #5027, opened 2024-12-02
Schedule Tekton Logging Forwarder to konflux-ci.dev/workload

khrm:
Schedule Tekton Logging Forwarder to konflux-ci.dev/workload.
We don't require a forwarder in master.

### DIFF
--- a/components/pipeline-service/development/main-pipeline-service-configuration.yaml
+++ b/components/pipeline-service/development/main-pipeline-service-configuration.yaml
@@ -2029,7 +2029,7 @@ spec:
                 key: endpoint
         tolerations:
           - effect: NoSchedule
-            key: node-role.kubernetes.io/master
+            key: konflux-ci.dev/workload
             operator: Exists
         securityContext:
           allowPrivilegeEscalation: false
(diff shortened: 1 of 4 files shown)


components/pipeline-service/development/main-pipeline-service-configuration.yaml
```diff
@@ -2029,7 +2029,7 @@ spec:
                 key: endpoint
         tolerations:
           - effect: NoSchedule
-            key: node-role.kubernetes.io/master
+            key: konflux-ci.dev/workload
```
hugares, line 2032: The pipelines will only run on the nodes with label "konflux-ci.dev/workload: konflux-tenants". The current change will setup log forwarder on both the nodes where we run konflux services "konflux-ci.dev/workload: konflux-services"  and where we run pipelines "konflux-ci.dev/workload: konflux-tenants".

I am just letting you know in case you want to deploy the log forwarder only on the nodes where we run tekton pipelines, then you will need to update this toleration to check value of label

khrm: Yes, we are interested only in `konflux-tenants`.

khrm: I have updated with that value.

hugares (approved):
/lgtm

PR #5035, opened 2024-12-03
chore(KFLUXVNGD-89): Add service account and related config

hmariset:
Add service account, cluster role, role binding and related ArgoCD configuration for deployment.

### DIFF
--- a/argo-cd-apps/base/eaas/crossplane-config/crossplane-config.yaml
+++ b/argo-cd-apps/base/eaas/crossplane-config/crossplane-config.yaml
@@ -1,0 +1,41 @@
+apiVersion: argoproj.io/v1alpha1
+kind: ApplicationSet
+metadata:
+  name: crossplane-config
+spec:
+  generators:
+    - merge:
+        mergeKeys:
+          - nameNormalized
+        generators:
+          - clusters:
+              values:
+                sourceRoot: components/crossplane-config
+                environment: staging
+                clusterDir: ""
+          - list:
+              elements: []
+  template:
+    metadata:
+      name: crossplane-config-{{nameNormalized}}
+    spec:
+      project: default
+      source:
+        path: "{{values.sourceRoot}}/{{values.environment}}/{{values.clusterDir}}"
+        repoURL: https://github.com/redhat-appstudio/infra-deployments.git
+        targetRevision: main
+      destination:
+        namespace: crossplane-config
+        server: "{{server}}"
+      syncPolicy:
+        automated:
+          prune: true
+          selfHeal: true
+        syncOptions:
+          - CreateNamespace=true
+        retry:
+          limit: -1
+          backoff:
+            duration: 10s
+            factor: 2
+            maxDuration: 3m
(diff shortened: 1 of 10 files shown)


argo-cd-apps/base/eaas/crossplane-config/crossplane-config.yaml
```diff
@@ -0,0 +1,46 @@
...
+        repoURL: https://github.com/redhat-appstudio/infra-deployments.git
+        targetRevision: main
+      destination:
+        namespace: crossplane-config
+        server: "{{server}}"
+      ignoreDifferences:
+        - group: argoproj.io
+          kind: ApplicationSet
+          jqPathExpressions:
+            - .spec.generators
```
amisstea, line 34: This appset isn't deploying any appsets itself. I don't think this is needed.

components/crossplane-config/base/rbac.yaml
```diff
@@ -0,0 +1,35 @@
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: crossplane-sa
+  namespace: crossplane-system
```
amisstea, line 6: The `crossplane-system` namespace won't exist on the EaaS clusters since we don't deploy crossplane there. I think you can put this SA and its RBAC in the `crossplane-config` namespace.

components/crossplane-config/base/rbac.yaml
```diff
@@ -0,0 +1,35 @@
...
+  name: crossplane-rb
+  namespace: crossplane-system
+subjects:
+- kind: ServiceAccount
+  name: crossplane-sa
+  namespace: crossplane-system
+roleRef:
+  kind: ClusterRole
+  name: crossplane-clusterrole
+  apiGroup: rbac.authorization.k8s.io
```
amisstea, line 35: This only grants `crossplane-sa` the ability to manage secrets within the `crossplane-system` namespace. Instead, this SA needs to be able to create, at a minimum, the following resources across the entire cluster:

- namespaces
- serviceaccounts
- secrets
- roles
- rolebindings
- resourcequotas
- clusterresourcequotas
- networkpolicies

List adapted from the manifests in https://github.com/redhat-appstudio/infra-deployments/tree/main/components/sandbox/tiers/src/appstudio-env.

argo-cd-apps/base/eaas/crossplane-config/kustomization.yaml
amisstea, line 1: Don't you also need to reference this kustomization dir in https://github.com/redhat-appstudio/infra-deployments/blob/main/argo-cd-apps/base/eaas/kustomization.yaml?

amisstea (changes requested):

components/crossplane-config/base/rbac.yaml
```diff
@@ -0,0 +1,56 @@
...
+  name: crossplane-clusterrole
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - namespaces
+  - secrets
+  - serviceaccounts
+  verbs:
+  - create
```
amisstea, line 21: I believe we need to grant all verbs (`"*"`) for each rule listed here. Crossplane will create but then eventually need to access and delete these resources upon removal of a claim.

hmariset: updated

amisstea (approved):
/lgtm

amisstea (approved):
/lgtm

manish-jangra (approved):
looks good to me

manish-jangra:
@hmariset remove the secret from there. We shouldn't keep secrets there, argocd might override the token generated.

PR #5044, opened 2024-12-03
Increase disk size of ART arm64 instance to 80GB

hugares:
We are getting "no space left on device" when building on arm VMs, increase the size to see if VM disk size is the issue.


ifireball (commented):
/lgtm

PR #5046, opened 2024-12-04
chore: Remove crossplane-config secret file

hmariset:
Remove crossplane-config secret file since argocd
might override the token generated.


amisstea (approved):
/lgtm

PR #5049, opened 2024-12-04
KFLUXBUGS-1610 - Update logging operator to 6.0.2. to pick up changes…

mike-kingsbury:
… to better troubleshoot long log lines being dropped.

Documentation used: 
https://docs.openshift.com/container-platform/4.16/observability/logging/logging-6.0/log6x-upgrading-to-6.html
https://github.com/openshift/cluster-logging-operator/blob/master/docs/administration/upgrade/v6.0_changes.adoc
... 
and many `oc explain clusterlogforwarders.observability.openshift.io.spec...` commands. 

TL;DR: `clusterlogging.logging.openshift.io` and `clusterlogforwarder.logging.openshift.io` were combined into `clusterlogforwarder.observability.openshift.io`.    Filters got pulled out into their own section, details about the outputs also pulled into an object named after the output type.


components/monitoring/logging/base/configure-logging/configure-log-collectors.yaml
```diff
@@ -1,10 +1,9 @@
 ---
-apiVersion: "logging.openshift.io/v1"
-kind: "ClusterLogging"
+apiVersion: "observability.openshift.io/v1"
+kind: "ClusterLogForwarder"
```
hugares, line 3: with this change, we now have 2 ClusterLogForwarder named "instance"

mftb: This also looks like the reason why kube-linter is failing

PR #5050, opened 2024-12-04
Create new ARM flavor with larger disk

hugares:
Change back all ARM flavors to default disk size and create a new flavor, same as default one with larger disk. Create a new one to not increase default thus increasing running cost for all builds that do not need more space.

[KFLUXINFRA-1141](https://issues.redhat.com//browse/KFLUXINFRA-1141)

### DIFF
--- a/components/multi-platform-controller/production-downstream/kflux-ocp-p01/host-config.yaml
+++ b/components/multi-platform-controller/production-downstream/kflux-ocp-p01/host-config.yaml
@@ -14,6 +14,7 @@ data:
   dynamic-platforms: "\
     linux/arm64,\
     linux/amd64,\
+    linux-d160/arm64,\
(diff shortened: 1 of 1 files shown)


arewm (changes requested):
I think we need to add the platform to line 14
```
dynamic-platforms: "\
```

hugares:
> I think we need to add the platform to line 14
> 
> ```
> dynamic-platforms: "\
> ```

Good point, I know this but somehow forgot..

hugares:
> > I think we need to add the platform to line 14
> > ```
> > dynamic-platforms: "\
> > ```
> 
> Good point, I know this but somehow forgot..

Done

hugares:
As usual, merging this change as e2e are not covering the files change in this PR

PR #5055, opened 2024-12-05
Pass QPS, Burst and Thread Flag to Results Watcher

khrm:


components/pipeline-service/production/kflux-ocp-p01/deploy.yaml
```diff
@@ -1999,6 +2003,9 @@ spec:
                   value: "konflux-tenants"
                   effect: "NoSchedule"
             default-timeout-minutes: "120"
+        config-leader-election-resolvers:
```
khrm, line 2006: Why didn't we update the manifest for this cluster earlier?

hugares: We might have forgot to do it, this is a new cluster that was just added, it was probably added by copy pasting another one and we forgot to regenerate the deploy.yaml

hugares (approved):
/lgtm

components/pipeline-service/development/main-pipeline-service-configuration.yaml
```diff
@@ -1278,6 +1278,9 @@ spec:
             - -completed_run_grace_period=2h
             - -store_deadline=1m
             - -forward_buffer=1m
+            - -kube-api-qps=50
```
hugares, line 1281: Looks like this kube-api-qps is not a valid option:
```
flag provided but not defined: -kube-api-qps
Usage of watcher:
-api_addr string
Address of API server to report to (default "localhost:8080")
-auth_mode string
Authentication mode to use when making requests. If not set, no additional credentials will be used in the request. Valid values: [google]
-check_owner
If enabled, owner references will be checked while deleting objects (default true)
-completed_run_grace_period duration
Grace period duration before Runs should be deleted. If 0, Runs will not be deleted. If < 0, Runs will be deleted immediately.
-disable_crd_update
Disables Tekton CRD annotation update on reconcile.
-label_selector string
Selector (label query) to filter objects to be deleted. Matching objects must satisfy all labels requirements to be eligible for deletion
-logs_api
Disable sending logs. If not set, the logs will be sent only if server support API for it (default true)
-namespace string
Should the Watcher only watch a single namespace, then this value needs to be set to the namespace name otherwise leave it empty.
-requeue_interval duration
How long the Watcher waits to reprocess keys on certain events (e.g. an object doesn't match the provided selectors) (default 10m0s)
-threadiness int
Number of threads (Go routines) allocated to each controller (default 2)
-token string
Authentication token to use in requests. If not specified, on-cluster configuration is assumed.
```

PR #5058, opened 2024-12-05
fix: Ingress Controller Configuration

manish-jangra:
This Pull Request will limit the Ingress Controler Configuration to  `stone-stg-host` && `stone-prd-host1`  clusters only


hugares (approved):
/lgtm

argo-cd-apps/base/host/ingresscontroller/ingresscontroller.yaml
```diff
@@ -10,17 +10,22 @@ spec:
         generators:
           - clusters:
               values:
-                environment: ""
-                clusterDir: ""
+                sourceRoot: configs/ingresscontroller
+                environment: staging
+                clusterDir: base
```
hugares, line 15: By defaulting all `clusters to configs/ingresscontroller/base`, this means that the time out patch will be applied to all clusters, not only dev, stg-host and prd-host1. Is this what you want?

hugares: ignore my comment, we detault to configs/ingresscontroller/staging/base which does not exists. I think we are good

argo-cd-apps/base/host/ingresscontroller/ingresscontroller.yaml
```diff
@@ -10,17 +10,22 @@ spec:
         generators:
           - clusters:
               values:
-                environment: ""
-                clusterDir: ""
+                sourceRoot: configs/ingresscontroller
+                environment: staging
```
hugares, line 14: we are missing somthing that will replace environment value from staging to production in the production overlay

manish-jangra: should be fixed now, added a patch for this in its overlay.

PR #5060, opened 2024-12-05
feat(KONFLUX-5480): Update pipeline references

arewm:
Includes https://github.com/konflux-ci/build-definitions/pull/1688 which bases the FBC pipeline off multi-arch, so expose the `build-platforms` additional parameter to the PipelineRun parameters.


chmeliik (approved):
/lgtm
/approve

PR #5066, opened 2024-12-05
Revert "Pass QPS, Burst and Thread Flag to Results Watcher (#5055)"

hugares:
This reverts commit 247d757c412a067eed746e753de94c202d59879a. Watcher is not starting due to invalid -kube-api-qps.

PR #5067, opened 2024-12-05
feat: Pass QPS, Burst and Thread Flag to Results Watcher, attempt 2

jhutar:
This was initially attempted here: https://github.com/redhat-appstudio/infra-deployments/pull/5055

And reverted here: https://github.com/redhat-appstudio/infra-deployments/pull/5066

Looking at this https://github.com/tektoncd/results/blob/main/cmd/watcher/main.go#L60-L80, looks like right options should not be `-kube-api-qps=50 -kube-api-burst=50 -threadiness=32`, but maybe `-qps=50 -burst=50 -threadiness=32`.

I have not tested this at all though :-/


khrm (commented):
/approve 
/lgtm

hugares (approved):
/lgtm

PR #5068, opened 2024-12-05
Allow the pipeline runner SA to list secrets

amisstea:
The SA can already get secrets but this [stepaction](https://github.com/konflux-ci/build-definitions/tree/main/stepactions/eaas-copy-secrets-to-ephemeral-cluster/0.1) requires list permissions.


sadlerap (approved):
/lgtm

PR #5079, opened 2024-12-08
KFLUXINFRA-1131: Allow tekton to use the appstudio-scc

gbenhaim:
In the kflux-prd-rh02 cluster we don't deploy kubesaw which configure the pipelines service account in each namespace to have permissions to access the appstudio scc. Instead, give this permission to tekton's pipeline controller.


enarha (commented):
/lgtm

PR #5080, opened 2024-12-09
Add PipelineAsCode labels selector to the forwarder log

khrm:
We had only Tekton Labels in vector, this adds PAC labels also.


enarha (commented):
/lgtm

PR #5087, opened 2024-12-09
Deploy crossplane XNamespace composition

amisstea:
The composition creates a LimitRange so `crossplane-sa` is granted permission to manage them.


yftacherzog (approved):
/approve
/lgtm

PR #5093, opened 2024-12-10
chore(KFLUXVNGD-89): Add stage kubernetes provider config

hmariset:
Crossplane kubernetes provider config for staging environment is added in crossplane-config.

### DIFF
--- a/components/crossplane-control-plane/staging/provider-config.yaml
+++ b/components/crossplane-control-plane/staging/provider-config.yaml
@@ -1,0 +1,35 @@
+---
+apiVersion: kubernetes.crossplane.io/v1alpha1
+kind: ProviderConfig
+metadata:
+  name: eaas-stage-kubernetes-provider-config
+  annotations:
+    argocd.argoproj.io/sync-options: SkipDryRunOnMissingResource=true
+spec:
+  credentials:
+    source: Secret
+    secretRef:
+      namespace: crossplane-system
+      name: eaas-cluster
+      key: kubeconfig
+---
+apiVersion: external-secrets.io/v1alpha1
+kind: ExternalSecret
+metadata:
+  name: eaas-cluster
+  namespace: crossplane-system
+  annotations:
+    argocd.argoproj.io/sync-options: SkipDryRunOnMissingResource=true
+    argocd.argoproj.io/sync-wave: "-1"
+spec:
+  dataFrom:
+    - extract:
+        key: staging/eaas/eaas-stage-kubeconfig
+  refreshInterval: 1h
+  secretStoreRef:
+    kind: ClusterSecretStore
+    name: appsre-stonesoup-vault
+  target:
+    creationPolicy: Owner
+    deletionPolicy: Delete
+    name: eaas-cluster
--- a/components/crossplane-control-plane/staging/kustomization.yaml
+++ b/components/crossplane-control-plane/staging/kustomization.yaml
@@ -1,5 +1,6 @@
 resources:
 - ../base
+- provider-config.yaml
 
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization


components/crossplane-config/staging/provider-config.yaml
```diff
@@ -0,0 +1,13 @@
+---
+apiVersion: kubernetes.crossplane.io/v1alpha1
+kind: ProviderConfig
+metadata:
+  name: eaas-stage-kubernetes-provider-config
+  api-server: https://api.kflux-stg-es01.21tc.p1.openshiftapps.com:6443
```
amisstea, line 6: This isn't a valid field and it's not needed anyway. The apiserver should be included as a key in the secret. Please remove this.

components/crossplane-config/staging/provider-config.yaml
```diff
@@ -0,0 +1,13 @@
...
+spec:
+  credentials:
+    source: Secret
+    secretRef:
+      namespace: crossplane-config
+      name: eaas-stage-kubeconfig
```
amisstea, line 12: Please include the `ExternalSecret` that generates this secret.

components/crossplane-config/staging/kustomization.yaml
```diff
@@ -3,3 +3,4 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
   - ../base
+  - provider-config.yaml
```
amisstea, line 6: The ProviderConfig shouldn't reside in the `crossplane-config` component since that's deployed to the target (EaaS) clusters. We need this on the "member" clusters where the control plane is running (the `crossplane-control-plane` component).

components/crossplane-config/staging/provider-config.yaml
```diff
@@ -0,0 +1,13 @@
...
+apiVersion: kubernetes.crossplane.io/v1alpha1
+kind: ProviderConfig
+metadata:
+  name: eaas-stage-kubernetes-provider-config
+  api-server: https://api.kflux-stg-es01.21tc.p1.openshiftapps.com:6443
+spec:
+  credentials:
+    source: Secret
+    secretRef:
+      namespace: crossplane-config
```
amisstea, line 11:
```suggestion
      namespace: crossplane-system
```

components/crossplane-config/staging/provider-config.yaml
amisstea: Nit: since the secret will include more than just a kubeconfig and the env info is redundant (we always know the cluster/env it's deployed to).
```suggestion
      name: eaas-cluster
```

components/crossplane-config/staging/provider-config.yaml
```diff
@@ -0,0 +1,13 @@
+---
+apiVersion: kubernetes.crossplane.io/v1alpha1
+kind: ProviderConfig
+metadata:
```
amisstea, line 4: We need to add the ArgoCD annotation to allow skipping when the CRD hasn't been registered yet.

amisstea (changes requested):

components/crossplane-control-plane/staging/provider-config.yaml
```diff
@@ -0,0 +1,38 @@
...
+  secretStoreRef:
+    kind: ClusterSecretStore
+    name: appsre-stonesoup-vault
+  target:
+    creationPolicy: Owner
+    deletionPolicy: Delete
+    name: eaas-cluster
+    template:
+      data:
+        apiserver: https://api.kflux-stg-es01.21tc.p1.openshiftapps.com:6443
```
amisstea, line 38: This should be stored in vault in the same secret as the kubeconfig.

hmariset: Updated

amisstea (approved):
/lgtm

PR #5111, opened 2024-12-10
chore: Update the External Secret version in crossplane

hmariset:
v1aplha1 doesn't have Extract option so updating it to v1beta1 in crossplane-control-plane ExternalSecret.


amisstea (approved):
/lgtm

PR #5114, opened 2024-12-11
KFLUXINFRA-1131: RBAC manifests for rh02

gbenhaim:
rh02 use our new architecture which doesn't include kubesaw. This requires creating ClusterRole that we be later assigned to Konflux users so they can use the system.


hugares (approved):
/lgtm

PR #5118, opened 2024-12-11
Fix the crossplane RBAC and staging ProviderConfig

amisstea:
Our XNamespace composition references the ProviderConfig by a different name. It also assigns the `edit` ClusterRole so `crossplane-sa` must also be granted these permissions.


hmariset (approved):
/lgtm

PR #5120, opened 2024-12-11
Roll out MPC update to prod

ralphbean:
From https://github.com/redhat-appstudio/infra-deployments/pull/5075


ralphbean:
Not urgent - but I would like to use the new logs and events to help with debugging.

hugares (approved):
/lgtm

PR #5123, opened 2024-12-11
Increase results-watcher number of threads

hugares:
The other tekton controllers processing pipelineRun have 4 replicas with 32 threads each. Results watcher only has one replica with 2 thread. Try to increase the processing capacity of watcher by configuring number of threads. The other options we use normally (qps and burst) cannot be configured yet as the version of results in production does not support them.

[KFLUXINFRA-1115](https://issues.redhat.com//browse/KFLUXINFRA-1115)


enarha (commented):
/lgtm

PR #5125, opened 2024-12-11
Disable OSP operator TektonConfig generation

enarha:
This prevent the operator from generating TektonConfig on start. Instead the TektonConfig stored in Git is used. This lead to race conditions in the past and we must have single source of truth (git).


hugares (approved):
/lgtm

PR #5128, opened 2024-12-12
Add two grades of s390x VM

mshaposhnik:
Single 4x16G type looks liltte too heavy, let's make 2 grades as it is done for internal cluster.


hugares (approved):
/lgtm

PR #5132, opened 2024-12-12
remove references to jvm-build-service

tnevrlka:
- jvm-build-service is currently not being used by anyone
- It is going to undergo major breaking changes and is going to be removed for now until the changes are done

Split into two commits:
- first is removing the references in most of the repository.
- second is removing  references to jbs in `components/sandbox/tiers` 

The second commit contains files created by `ksctl` as mentioned in https://github.com/redhat-appstudio/infra-deployments/blob/main/components/sandbox/tiers/README.adoc leading to thousands of newly added lines


chmeliik (commented):
The manual changes LGTM, just not sure why something ended up generating 5000 lines of new stuff :confused:

components/sandbox/tiers/production/appstudio/kustomization.yaml
```diff
@@ -13,16 +13,19 @@ resources:
 - tiertemplate-appstudio-clusterresources-593233715-593233715.yaml
 - tiertemplate-appstudio-clusterresources-809836689-809836689.yaml
 - tiertemplate-appstudio-contributor-1817914940-1817914940.yaml
+- tiertemplate-appstudio-contributor-3265408548-3265408548.yaml
```
mmorhun, line 16: Why do we need such changes?

tnevrlka: Sorry, it looks really messy, but I'm not aware of anything I can do about it. 

The source of truth and the only thing I'm changing is in [components/sandbox/tiers/src/appstudio](https://github.com/redhat-appstudio/infra-deployments/tree/main/components/sandbox/tiers/src/appstudio). 
We need such changes, because there are references to jvm-build-service and I'm removing them, see an example [here](https://github.com/redhat-appstudio/infra-deployments/pull/5132/commits/6c6aed66ba1035c64fb86d42d9894b501fab424e#diff-94de3b39ce770e418903c21543b146f7dca0e63b16db7af023aec3c63c9e6ed5L107-L115)

After changes are made in the directory, a command is ran as described in the [README](https://github.com/redhat-appstudio/infra-deployments/blob/main/components/sandbox/tiers/README.adoc) which autogenerates majority of the changes in https://github.com/redhat-appstudio/infra-deployments/pull/5132/commits/6c6aed66ba1035c64fb86d42d9894b501fab424e.

This seems to me to indeed be the intended workflow, see https://github.com/redhat-appstudio/infra-deployments/pull/5137 as an example

tnevrlka:
> The manual changes LGTM, just not sure why something ended up generating 5000 lines of new stuff 😕

It scares me a little bit, but it kinda makes sense to me when you consider [this PR](https://github.com/redhat-appstudio/infra-deployments/pull/5137/files#diff-9112ab363ffcf6b17411f197c164b062f8e5635f62f774eca74b8fde4134cdc9) which only changes the maintainer file and has 1225 newly added lines.

I've changed four files which could result in 4 times as many lines.
Sanity check: 1225*4=4900, I have around 4855. Q.E.D. 

It seems like it might be a good idea to get someone from infra to take a look at this though

components/sandbox/tiers/production/appstudio/kustomization.yaml
hugares: This is normal. Changing one access in a role will case a generation of a new role file, thus explaining all those new lines. This is how kubesaw works to deploy the new tier.

chmeliik (approved):
/lgtm
/approve

hugares:
/hold
We did a few changes to the tiers since this change was originally written. We might need to rebase and re-generate the tiers. Blocking this change, I will verify it later today

hugares:
> @hugares Any updates? I have regenerated the tiers during the last rebase 2 days ago

Thanks, I wanted to make sure this was done.

/unhold

hugares (approved):
/lgtm

PR #5133, opened 2024-12-12
Add Vector and update Results stone-prod-p01

enarha:
Add Vector for forwarding Pipelines log.
Update Tekton Results to version 0.13.2-2.
The base file for prod is updated and the deploy.yaml file updated for a single cluster stone-prod-p01 to allow gradual update.


khrm (commented):
/lgtm
/approve

PR #5136, opened 2024-12-13
Move cachi2 pipelines from tekton-ci to konflux-ci

slimreaper35:


MartinBasti:
Force merged, tekton didn't want to accept ok-to-test and it's just namesapce unrelated directly to konflux

PR #5141, opened 2024-12-13
Kyverno Policies used during demo

dperaza4dustbit:
These are the files we used for demo, wanted to share for early review and to allow other team members to use in other demos or in final pr for infra deployment


gbenhaim:
thsnk @dperaza4dustbit. 
can you please also share the manifests for install kyverno?

PR #5163, opened 2024-12-16
add namespace-lister deployment to staging

sadlerap:
This is largely for testing purposes, to see if this deployment configuration is functional.  As such, we're only going to deploy this to one of the member clusters for the time being, and we're not going to integrate this with the UI's nginx config until we have a stable deployment.

### DIFF
--- a/argo-cd-apps/overlays/konflux-public-staging/delete-applications.yaml
+++ b/argo-cd-apps/overlays/konflux-public-staging/delete-applications.yaml
@@ -11,9 +11,3 @@ kind: ApplicationSet
 metadata:
   name: konflux-rbac
 $patch: delete
----
-apiVersion: argoproj.io/v1alpha1
-kind: ApplicationSet
-metadata:
-  name: namespace-lister
-$patch: delete
--- a/argo-cd-apps/overlays/konflux-public-staging/delete-applications.yaml
+++ b/argo-cd-apps/overlays/konflux-public-staging/delete-applications.yaml
@@ -11,3 +11,9 @@ kind: ApplicationSet
 metadata:
   name: konflux-rbac
 $patch: delete
+---
+apiVersion: argoproj.io/v1alpha1
+kind: ApplicationSet
+metadata:
+  name: namespace-lister
+$patch: delete
--- a/components/namespace-lister/base/deployment.yaml
+++ b/components/namespace-lister/base/deployment.yaml
@@ -1,0 +1,47 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: namespace-lister
+  namespace: namespace-lister
+  labels:
+    apps: namespace-lister
+spec:
+  selector:
+    matchLabels:
+      apps: namespace-lister
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        apps: namespace-lister
+    spec:
+      # securityContext:
+      #   runAsNonRoot: true
+      serviceAccountName: namespace-lister
+      containers:
+      - image: namespace-lister:foo
+        name: namespace-lister
+        imagePullPolicy: IfNotPresent
+        env:
+        - name: LOG_LEVEL
+          value: "0"
+        # - name: AUTH_USERNAME_HEADER
+        #   value: "X-User"
+        resources:
+          limits:
+            cpu: 500m
+            memory: 128Mi
+          requests:
+            cpu: 10m
+            memory: 64Mi
+        ports:
+        - containerPort: 8080
+          name: http
+        securityContext:
+          allowPrivilegeEscalation: false
+          readOnlyRootFilesystem: true
+          runAsNonRoot: true
+          capabilities:
+            drop:
+            - "ALL"
+      terminationGracePeriodSeconds: 60
--- a/components/namespace-lister/base/deployment.yaml
+++ b/components/namespace-lister/base/deployment.yaml
@@ -9,14 +9,12 @@ spec:
   selector:
     matchLabels:
       apps: namespace-lister
-  replicas: 1
+  replicas: 3
   template:
     metadata:
       labels:
         apps: namespace-lister
     spec:
-      # securityContext:
-      #   runAsNonRoot: true
       serviceAccountName: namespace-lister
       containers:
       - image: namespace-lister:foo
--- a/components/namespace-lister/base/deployment.yaml
+++ b/components/namespace-lister/base/deployment.yaml
@@ -5,6 +5,8 @@ metadata:
   namespace: namespace-lister
   labels:
     apps: namespace-lister
+  annotations:
+    ignore-check.kube-linter.io/no-anti-affinity: "Using topologySpreadConstraints"
 spec:
   selector:
     matchLabels:
@@ -16,6 +18,13 @@ spec:
         apps: namespace-lister
     spec:
       serviceAccountName: namespace-lister
+      topologySpreadConstraints:
+      - maxSkew: 1
+        topologyKey: topology.kubernetes.io/zone
+        whenUnsatisfiable: ScheduleAnyway
+        labelSelector:
+          matchLabels:
+            apps: namespace-lister
       containers:
       - image: namespace-lister:foo
         name: namespace-lister
@@ -23,8 +32,6 @@ spec:
         env:
         - name: LOG_LEVEL
           value: "0"
-        # - name: AUTH_USERNAME_HEADER
-        #   value: "X-User"
         resources:
           limits:
             cpu: 500m
--- a/components/namespace-lister/base/nginx.conf
+++ b/components/namespace-lister/base/nginx.conf
@@ -1,0 +1,59 @@
+worker_processes auto;
+error_log /var/log/nginx/error.log;
+pid /run/nginx.pid;
+
+# Load dynamic modules. See /usr/share/doc/nginx/README.dynamic.
+include /usr/share/nginx/modules/*.conf;
+
+events {
+    worker_connections 1024;
+}
+
+http {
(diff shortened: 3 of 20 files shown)


sadlerap:
I think it may not be a bad idea to hold merging this until after the holidays, since I'd rather avoid situations where we accidentally cause an outage with the addition of this component.

gbenhaim (commented):
I suggest that it would be included as a side car in the proxy deployment https://github.com/redhat-appstudio/infra-deployments/tree/main/components/konflux-ui/production/base/proxy it would be much simpler (you can see how it was done for `workspace-manager`). I don't see a benefit of having it in its own deployment. In addition, since we used impersonation today, token review can't be turned on in namespace-lister, so it means anyone would be able to list namespaces.

filariow:
@gbenhaim I'm mostly fine with any setup we chose, but I'd prefer to have separate deployments for the following reasons.

> I suggest that it would be included as a side car in the proxy deployment https://github.com/redhat-appstudio/infra-deployments/tree/main/components/konflux-ui/production/base/proxy it would be much simpler (you can see how it was done for `workspace-manager`). I don't see a benefit of having it in its own deployment. 

If we chose separate deployments we'll have better isolation and scaling.

**Isolation**: A bug on `namespace-lister` won't affect the proxy, but just the `GET namespace` endpoint. If the `namespace-lister` starts crashing and it is in the same pod of the proxy, the whole pod will crash and -from a user perspective- nothing will work. If we have separate deployments, only the `GET namespace` endpoint won't work.

**Scaling**: we can scale the proxy and the `namespace-lister` independently.

> In addition, since we used impersonation today, token review can't be turned on in namespace-lister, so it means anyone would be able to list namespaces.

I'm not sure I got it correctly, but if it is a concern about external users (1) or other deployments (2) invoking the namespace-lister, we could avoid exposing the namespace-lister outside (1) and use a [NetworkPolicy](https://kubernetes.io/docs/concepts/services-networking/network-policies/) to only grant requests from the proxy deployed in the same namespace (1,2); right?

sadlerap:
I think I'll leave it as a separate deployment for now.

To me, the biggest reason to do so is a separation of concerns: in particular, we don't need nearly as many permissions on the namespace-lister deployment.  The list of permissions that the proxy's service account would grant is far beyond what namespace-lister needs.  Why should we grant namespace-lister the extra permissions that the proxy's service account would grant?

Also, I'm not certain it should be moved under the konflux-ui component.  Unless that component is also going to be deployed on the internal clusters, we'd be duplicating the resources between the `konflux-ui` and some other components, which doesn't make sense to me.  It seems better to me to unify this underneath a single component.

@gbenhaim thoughts?

components/namespace-lister/base/proxy.yaml
filariow, line 1: We may simplify by not deploying a Proxy at all and by using `port-forward` to test namespace-lister is working properly.

gbenhaim:
> I think I'll leave it as a separate deployment for now.
> 
> To me, the biggest reason to do so is a separation of concerns: in particular, we don't need nearly as many permissions on the namespace-lister deployment. The list of permissions that the proxy's service account would grant is far beyond what namespace-lister needs. Why should we grant namespace-lister the extra permissions that the proxy's service account would grant?

ok, you convinced me with this argument.
> 
> Also, I'm not certain it should be moved under the konflux-ui component. Unless that component is also going to be deployed on the internal clusters, we'd be duplicating the resources between the `konflux-ui` and some other components, which doesn't make sense to me. It seems better to me to unify this underneath a single component.
> 

The konflux-ui component is going to be deployed on all clusters (external and internal).

> @gbenhaim thoughts?

You would still need to do the following changes:

1. Remove the Route, namespace lister shouldn't bu publicly exposed.
2. Remove the nginx deployment, instead edit the nginx config in the konflux-ui component to route traffic to namespace-lister.
3. Create a network policy that allows only the proxy deployment from the konflux-ui namespace to access namespace-lister (since it doesn't use any authentication).

components/namespace-lister/base/deployment.yaml
```diff
@@ -0,0 +1,47 @@
...
+  selector:
+    matchLabels:
+      apps: namespace-lister
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        apps: namespace-lister
+    spec:
+      # securityContext:
```
gbenhaim, line 18: can you please remove the comments?

components/namespace-lister/base/deployment.yaml
```diff
@@ -0,0 +1,47 @@
...
+metadata:
+  name: namespace-lister
+  namespace: namespace-lister
+  labels:
+    apps: namespace-lister
+spec:
+  selector:
+    matchLabels:
+      apps: namespace-lister
+  replicas: 1
```
gbenhaim, line 12: please use 3 replicas and spread them across AZ.
You can see an example of how to do it in the deployment of nginx in the konflux-ui component.

components/namespace-lister/base/proxy.yaml
sadlerap: I'd like to be able to test API access alongside namespace-lister, so if possible, I'd like at least a minimal proxy in-place.

components/namespace-lister/base/route.yaml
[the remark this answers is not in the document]
sadlerap: Adjusted in a8adc9f4.

components/namespace-lister/base/deployment.yaml
sadlerap: Done in 7c94977f.

components/namespace-lister/base/deployment.yaml
sadlerap: Fixed in 7c94977f.

sadlerap:
@gbenhaim FWIW my primary intention with this PR is to test namespace-lister and ensure it functions as intended on the staging cluster.  Long-term, I will integrate it with konflux-ui's proxy, but until I've tested it in a production-like environment, I would like to keep things separate.  This will let me iterate faster in the event I do run into issues during testing.

gbenhaim:
> @gbenhaim FWIW my primary intention with this PR is to test namespace-lister and ensure it functions as intended on the staging cluster. Long-term, I will integrate it with konflux-ui's proxy, but until I've tested it in a production-like environment, I would like to keep things separate. This will let me iterate faster in the event I do run into issues during testing.

@sadlerap when do you plan to integrat it into Konflux' UI proxy? how many days do you need? we really need to get it integrated soon so the UI team can use it while developing the required changes for dropping worksapces.

why do you think that integrating it into the Konflux ui will make you iterates slower?

components/namespace-lister/base/nginx.conf
```diff
@@ -0,0 +1,59 @@
...
+
+# Load dynamic modules. See /usr/share/doc/nginx/README.dynamic.
+include /usr/share/nginx/modules/*.conf;
+
+events {
+    worker_connections 1024;
+}
+
+http {
+    log_format upstreamlog '[$time_local] $remote_addr - $remote_user - $server_name $host to: $proxy_host  $upstream_addr: $request $status upstream_response_time $upstream_response_time msec $msec request_time $request_time';
```
filariow, line 13: Minor: This is logging a lot of info. We can simplify it. I'd suggest using the same format we have in `ui`

filariow:
> @gbenhaim  we really need to get it integrated soon so the UI team can use it while developing the required changes for dropping worksapces.

If this is blocking UI's development, we should be able to unblock them just by giving them the right to `list` namespaces. They'll receive much more results than expected, but it should be fine for development.

sadlerap:
> @sadlerap when do you plan to integrat it into Konflux' UI proxy? how many days do you need?

As soon as I can verify that our deployment is stable and won't cause any issues.  I don't expect this process to take a long time - less than a day after this merges?

I've been bitten by adding components before - something usually goes wrong, and we have to follow up with a few extra PRs to fix things.  For an example of this, see our recent PRs adding Kyverno to this repository (#5203, #5272, #5274, #5275, and #5276).

By taking this route, I'm hoping to reduce the blast radius of any instability.

gbenhaim (approved):

argo-cd-apps/overlays/konflux-public-staging/delete-applications.yaml
```diff
@@ -11,3 +11,9 @@ kind: ApplicationSet
 metadata:
   name: konflux-rbac
 $patch: delete
+---
```
gbenhaim, line 14: you should remove this change in order to deploy to staging.

filariow: great catch!

components/namespace-lister/base/nginx.conf
sadlerap: This is the same as what we're using in `components/ui/base/proxy/nginx.conf`.

argo-cd-apps/overlays/konflux-public-staging/delete-applications.yaml
sadlerap: Fixed in 1cd4b0af.

components/namespace-lister/base/rbac.yaml
```diff
@@ -0,0 +1,50 @@
...
+  name: namespace-lister-auth-delegator
+subjects:
+- apiGroup: ""
+  kind: ServiceAccount
+  name: namespace-lister
+  namespace: namespace-lister
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: system:auth-delegator
```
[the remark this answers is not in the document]
sadlerap: Fixed in 09102540.

PR #5168, opened 2024-12-17
KFLUXINFRA-1183: Public Konflux information

gbenhaim:
Expose a configmap which is readable by all authenticated users that contains public information about the Konflux instance.

This information, for example, can be used by the UI, instead of hardcoding it in the UI codebase.

### DIFF
--- a/argo-cd-apps/base/member/infra-deployments/konflux-info/konflux-info.yaml
+++ b/argo-cd-apps/base/member/infra-deployments/konflux-info/konflux-info.yaml
@@ -1,0 +1,58 @@
+apiVersion: argoproj.io/v1alpha1
+kind: ApplicationSet
+metadata:
+  name: konflux-info
+spec:
+  generators:
+    - merge:
+        mergeKeys:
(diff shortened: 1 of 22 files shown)


argo-cd-apps/base/member/infra-deployments/konflux-public/konflux-public.yaml
```diff
@@ -0,0 +1,58 @@
+apiVersion: argoproj.io/v1alpha1
+kind: ApplicationSet
+metadata:
+  name: konflux-public
```
hugares, line 4: Disclaimer, this is a subjective comment. "Konflux-public" name might be confused with internal vs public instances. I would have named it something more like "konflux-info"

Feel free to ignore this comment :)

gbenhaim: I changed it to `konflux-info`

hugares (approved):
/lgtm

PR #5170, opened 2024-12-17
Reduce mumber of IBM VM's for Ocp-Art cluster account

mshaposhnik:
Looks like IBM cannot handle 50 (and even more so 100)  VM's in the single DC. The empiric value which work is about 30 VMs, if continue to start more VM they're dropping into `failed` state.


hugares (approved):
/lgtm

PR #5174, opened 2024-12-17
fix: moving monitor configs to named clusters

theflockers:


seanconroy2021 (approved):
LGTM

theflockers:
@hugares could you take a look in this PR? I am not sure if the way done here it would load the resources as expected, so loading the configs from the "staging" directories for different clusters.

PR #5175, opened 2024-12-17
KFLUXBUGS-1610 - Move tuning component into Splunk object for the Clu…

mike-kingsbury:
…sterLogForwarder object.

This setting was initially missed as it was applied in a patch.


mftb (approved):
/lgtm
/approve

PR #5178, opened 2024-12-18
Install Tekton Logs Collector using Kustomize generator

khrm:
ArgoCD CR's Application/ApplicationSet CR isn't available in every cluster. So we switched to using kustomize generator for vector installation.


khrm (commented):
/assign @enarha

components/pipeline-service/production/base/main-pipeline-service-configuration.yaml
```diff
@@ -601,7 +608,7 @@ data:
     LOGS_API=false
     LOGS_TYPE=File
     LOGS_BUFFER_SIZE=5242880
-    LOGS_PATH=//logs
+    LOGS_PATH=/logs
```
enarha, line 611: Why was that value changed? I see further below we also downgrade (reference older image) of Results, why?

khrm (commented):
@enarha I have removed the prod changes.

khrm:
Vector got deployed to Dev env.
image

/hold cancel

PR #5188, opened 2024-12-19
KFLUXINFRA-1129: View Only Cluster Role for in-cluster Grafana

manish-jangra:
This Pull Request will do the following -
- Create a Cluster Role to provide view only access to in-cluster grafana
- Create a ClusterRoleBinding to allow users (part of konflux-grafana-viewers rover group) to use the Cluster Role


hugares (approved):
/lgtm

PR #5191, opened 2024-12-19
Disable Creation of Tekton-Logging NS in Pipeline Service

khrm:
This avoids sync issue due to race condition between vector and pipeline service.


khrm (commented):
/assign @enarha

hugares (approved):
/lgtm

enarha:
@khrm The namespace was explicitly added to avoid the race condition with the ExternalSecret which is synced quite early by ArgoCD https://github.com/redhat-appstudio/infra-deployments/blob/main/components/pipeline-service/base/external-secrets/tekton-logging/tekton-results-s3.yaml#L7 . Was there a real issue between vector and pipeline service? I would expect that if the namespace exists (created by pipeline-service), the Vector deployment continues as usual.

PR #5196, opened 2024-12-20
KFLUXINFRA-1154: Configure Backup for kflux-prd-rh02

manish-jangra:
This Pull Request will configure the backup for newly provisioned cluster `kflux-prd-rh02`.

This Cluster is hosting `toolchain-member-operator` as well as `toolchain-host-operator`, this backup targets both of them.

### DIFF
--- a/components/backup/production/kflux-prd-rh02/kustomization.yaml
+++ b/components/backup/production/kflux-prd-rh02/kustomization.yaml
@@ -1,0 +1,24 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - ../../base/all-clusters
+  - backup-tenants-schedule.yaml
+patches:
+  - target:
+      group: external-secrets.io
+      version: v1beta1
+      kind: ExternalSecret
+      name: backup-s3-credentials
+    path: backup-s3-credentials-patch.yaml
+  - target:
+      group: oadp.openshift.io
+      version: v1alpha1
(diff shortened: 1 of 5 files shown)


components/backup/production/kflux-prd-rh02/kustomization.yaml
```diff
@@ -0,0 +1,25 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - ../../base/all-clusters
+  - ../../base/host/schedules
+  - ../../base/member/schedules
```
hugares, line 6: as mentioned above, we are not using toolchain host/member on this cluster so we need to include member schedules but we only need one of the two from that folder, we only need the tenant one

components/backup/production/kflux-prd-rh02/kustomization.yaml
```diff
@@ -0,0 +1,25 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - ../../base/all-clusters
+  - ../../base/host/schedules
```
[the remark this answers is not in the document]
manish-jangra: noted, changed made

components/backup/production/kflux-prd-rh02/kustomization.yaml
manish-jangra: noted, changed made

gbenhaim (approved):

PR #5202, opened 2024-12-20
fix(KFLUXINFRA-1129): Fix in-cluster grafana permissions

manish-jangra:
Fix in-cluster grafana permissions to allow users of rover group `konflux-grafana-viewers` to have view-only access to the in-cluster grafana.

Related Pull Request -
- https://github.com/redhat-appstudio/infra-deployments/pull/5188


avi-biton (approved):
lgtm

PR #5215, opened 2024-12-23
Remove enable-tekton-oci-bundles TektonConfig

enarha:
Remove enable-tekton-oci-bundles from production clusters. This configuration options was removed in Tekton v0.61.0.


hugares (approved):
/lgtm

PR #5218, opened 2024-12-29
Allow pipeline-service to delete logging pods

enarha:
The way the Vector pods are deployed today it happened that when a new version was deployed, we ended up with 2 DaemonSets, with the pods of the old DaemonSets running and blocking the pods of the new DaemonSet. Deleting the DaemonSet/pods allowed the new DaemonSet to spawn its own pods. We need that flexibility as managers of the service.


khrm (commented):
/approve
/lgtm

PR #5228, opened 2025-01-06
KFLUXINFRA-1129: In-Cluster Grafana Dashboard Access

manish-jangra:
This pull request will create a ClusterRole and a Namespace specific RoleBinding to allow the users of konflux-grafana-viewers rover group to view the grafana dashboards. This will be a read-only access to the in-cluster grafana dashboards.


hugares (approved):
/lgtm

PR #5229, opened 2025-01-06
chore(KFLUXVNGD-136): export label deployment

yftacherzog:
The deployment Prometheus label is needed for monitoring project-controller's availability in RHOBS.


ifireball:
@yftacherzog how come we didn't need this before? If de "deployment" label is uniqe to the project-controller, perhaps we don't need it?

yftacherzog:
> @yftacherzog how come we didn't need this before? If de "deployment" label is uniqe to the project-controller, perhaps we don't need it?

It's not unique. We need it now because we want the `deployment` label to be translated to the `service` label in the `konflux_up` time series we're adding. Others did not do that specific transformation.

avi-biton (approved):
/lgtm

PR #5243, opened 2025-01-09
Switch back to local Minio in development

enarha:
The Minio testing service play.min.io is down for almost a day blocking the CI, so switching back to local deployment in development.


khrm (commented):
/approve
/lgtm

PR #5270, opened 2025-01-14
revert production patch for notification-service

avi-biton:
The new definitions only apply for dev and stage, the production environment remains the same for now and therefor should not been updated.


yftacherzog (approved):
/lgtm
/approve

PR #5274, opened 2025-01-14
kyverno: specify values inline

sadlerap:
Argocd doesn't do template substitution when determining the location of helm's values files, which we were relying on to pick up the values file for the helm chart.  To fix this, specify the values in the ApplicationSet itself.


filariow (approved):

hugares (approved):
/lgtm

PR #5282, opened 2025-01-15
Increase Build Service memory request to be bigger than actual average usage

mmorhun:
The investigation was done by perf team and the result is [this issue](https://issues.redhat.com/browse/KONFLUX-5911) which this PR fixes.


hugares (approved):
/lgtm

PR #5288, opened 2025-01-15
namespace-lister: disable configmap name suffix hashing

sadlerap:
Currently, when we generate the configmap for the proxy, kustomize will automatically append a hash to its name.  This breaks the proxy deployment, since it looks for a configmap with a specific name.

We can fix this either by disabling name suffix hashing or by injecting the name via a replacement.  The former is easier to implement, so opt to implement that method.

### DIFF
--- a/components/namespace-lister/base/kustomization.yaml
+++ b/components/namespace-lister/base/kustomization.yaml
@@ -12,6 +12,8 @@ configMapGenerator:
 - files:
   - nginx.conf=nginx.conf
   name: nginx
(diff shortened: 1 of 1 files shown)


sadlerap:
FWIW, I'm opting to fix proxy here since I don't have permissions to do port-forwarding on the staging clusters.

```
$ oc auth can-i create pods/port-forward -n namespace-lister
no
```

filariow (approved):

PR #5297, opened 2025-01-16
integrate konflux-ui with namespace-lister

filariow:


components/konflux-ui/staging/base/proxy/nginx.conf
```diff
@@ -161,6 +161,24 @@ http {
...
+            # authenticate request
+            auth_request_set $email  $upstream_http_x_auth_request_email;
+            auth_request /oauth2/auth;
+
+            # Impersonate requesting user
+            proxy_set_header Impersonate-User $email;
+            include /mnt/nginx-generated-config/bearer.conf;
+
+            # namespace-lister endpoint
+            rewrite ^/(.*)/$ /$1 permanent;
```
gbenhaim, line 174: we should use `break` instead of `permanent`. we want to reverse proxy not redirect.

gbenhaim (changes requested):

components/konflux-ui/staging/base/proxy/nginx.conf
```diff
@@ -161,6 +161,24 @@ http {
             include /mnt/nginx-generated-config/bearer.conf;
         }
 
+        location ~* /api/v1/namespaces(/?)$ {
+            # authenticate request
+            auth_request_set $email  $upstream_http_x_auth_request_email;
+            auth_request /oauth2/auth;
+
+            # Impersonate requesting user
+            proxy_set_header Impersonate-User $email;
```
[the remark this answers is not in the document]
filariow: the namespace-lister is expecting authentication details somewhere. In the proposed configuration they are in the `Impersonate-User` header. We can use something different, like `X-Email` either here and [in the with-header-auth.yaml](https://github.com/redhat-appstudio/infra-deployments/blob/main/components/namespace-lister/base/patches/with-header-auth.yaml). I'll change that to use `X-Email` as it seems cleaner to me.

sadlerap (commented):
lgtm, please fix merge conflicts

filariow:
@sadlerap thanks for merging the conflicts

sadlerap (approved):

components/konflux-ui/staging/base/proxy/nginx.conf
```diff
@@ -161,6 +161,26 @@ http {
...
         }
 
+        location ~* /api/k8s/api/v1/namespaces(/?)$ {
+            # authenticate request
+            auth_request_set $email  $upstream_http_x_auth_request_email;
+            auth_request /oauth2/auth;
+
+            # Impersonate requesting user
+            proxy_set_header X-Email $email;
+            include /mnt/nginx-generated-config/bearer.conf;
```
filariow, line 171:
```suggestion
```

components/namespace-lister/base/service.yaml
```diff
@@ -3,11 +3,13 @@ kind: Service
 metadata:
   name: namespace-lister
   namespace: namespace-lister
+  annotations:
+    service.beta.openshift.io/serving-cert-secret-name: namespace-lister-tls
```
gbenhaim, line 7: I would expect to see a change to namespace-lister deployment as well for mounting the tls certificates.

components/konflux-ui/staging/base/proxy/nginx.conf
```diff
@@ -161,6 +161,25 @@ http {
             include /mnt/nginx-generated-config/bearer.conf;
         }
 
+        location ~* /api/k8s/api/v1/namespaces(/?)$ {
```
gbenhaim, line 164: I think this can be simplified by dropping the regex expression. wdyt?

filariow: IIUC, If we remove this regex this will become the path every namespaced request will take, e.g. `api/v1/namespaces/my-ns/pods` will be processed by this rule

sadlerap: Correct.  We only want to match on requests for namespace resources, i.e. when the path is `/api/k8s/api/v1/namespaces` or `/api/k8s/api/v1/namespaces/`.

If I'm understanding [nginx's docs](https://nginx.org/en/docs/http/ngx_http_core_module.html#location) correctly, we could set it up like this and have it work:
```suggestion
        location = /api/k8s/api/v1/namespaces/ {
```

This would match the path we want, both with and without the trailing slash.  However, this has slightly different behavior than what we currently have: nginx will issue a 301 permanent redirect from `/api/k8s/api/v1/namespaces` to `/api/k8s/api/v1/namespaces/`, which I'm not sure I'm a fan of.  To me, the regex is clearer in its intent.

Also, if we _do_ decide to use regex matching here, we should probably do case-sensitive matching:
```suggestion
        location ~ /api/k8s/api/v1/namespaces(/?)$ {
```

components/namespace-lister/base/service.yaml
filariow: namespace-lister is not supporting TLS at the moment. let's move this to a future PR.

sadlerap: I've filed https://github.com/konflux-ci/namespace-lister/pull/29 to enable it in namespace-lister.

components/konflux-ui/staging/base/proxy/nginx.conf
```diff
@@ -161,6 +166,39 @@ http {
...
+            try_files $uri @$ns_target;
+        }
+
+        location @namespacelister {
+            auth_request_set $email  $upstream_http_x_auth_request_email;
+            auth_request /oauth2/auth;
+            proxy_read_timeout 30m;
+            proxy_set_header X-Email $email;
+
+            rewrite ^.*$ /api/v1/namespaces break;
```
gbenhaim, line 184: instead of re writing the entire path to a constnat, you can add add the path to the url in the proxy_pass directive.

filariow: right, let me apply this change

sadlerap (commented):
With the suggestion by @gbenhaim, lgtm

PR #5300, opened 2025-01-16
stage: Add RBAC details to konflux-public-info cm

gbenhaim:
The information will be used by the UI when
rendering the user access page and when creating
role bindings for users.


hugares (approved):
/lgtm

PR #5313, opened 2025-01-17
Push refactored version of MintMaker to stage

gnaponie:
This is a temporary release to stage, as we need to test some new refactoring in MintMaker.

### DIFF
--- a/components/mintmaker/staging/base/kustomization.yaml
+++ b/components/mintmaker/staging/base/kustomization.yaml
@@ -10,8 +10,8 @@ namespace: mintmaker
 
 images:
 - name: quay.io/konflux-ci/mintmaker
-  newName: quay.io/konflux-ci/mintmaker
-  newTag: 7f0cd9e10e0ede8e2dcce8e5cbaa3647b4548cc4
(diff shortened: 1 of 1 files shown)


gnaponie:
@qixiang @staticf0x @FernandesMF PTAL

components/mintmaker/staging/base/kustomization.yaml
```diff
@@ -9,9 +9,9 @@ resources:
 namespace: mintmaker
 
 images:
-- name: quay.io/konflux-ci/mintmaker
-  newName: quay.io/konflux-ci/mintmaker
-  newTag: 7f0cd9e10e0ede8e2dcce8e5cbaa3647b4548cc4
+- name: quay.io/repository/redhat-user-workloads/konflux-mintmaker-tenant/mintmaker-refactoring-pipelinerun
```
qixiang, line 12: `quay.io/konflux-ci/mintmaker` is our controller image.
You can't change this, the `name` is the image kustomize should transform for you, if you use this new image, kustomize won't find anything to be transformed.

PR #5318, opened 2025-01-19
Update refs to Minio images

enarha:
By default the Minio images are pulled from docker.io, which makes the CI fail when the pull limit is reached. Use quay.io instead.


khrm (commented):
/test appstudio-upgrade-tests

PR #5322, opened 2025-01-20
Deploy konflux-rbac to public staging

gbenhaim:
The cluster roles will be used for granting
permissions to users after we stop using kubesaw.

Right now those roles are needed by the UI team
for testing changes they make to the UI.


hugares (approved):
/lgtm

PR #5324, opened 2025-01-20
[Staging] Optimizing Vector Configuration to avoid missing logs

khrm:
Optimized the vector configuration to avoid missing logs.

### DIFF
--- a/components/vector-tekton-logs-collector/development/vector-helm-values.yaml
+++ b/components/vector-tekton-logs-collector/development/vector-helm-values.yaml
@@ -17,7 +17,8 @@ customConfig:
     kubernetes_logs:
       type: kubernetes_logs
       rotate_wait_secs: 5
-      glob_minimum_cooldown_ms: 15000
+      glob_minimum_cooldown_ms: 500
+      max_read_bytes: 3145728
       auto_partial_merge: true
       extra_label_selector: "app.kubernetes.io/managed-by in (tekton-pipelines,pipelinesascode.tekton.dev)"
     internal_metrics:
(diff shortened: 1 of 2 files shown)


khrm (commented):
/assign @enarha

hugares (approved):

enarha:
@khrm you also need to regenerate the deploy.yaml files for the staging clusters.

components/vector-tekton-logs-collector/development/vector-helm-values.yaml
```diff
@@ -17,7 +17,8 @@ customConfig:
     kubernetes_logs:
       type: kubernetes_logs
       rotate_wait_secs: 5
-      glob_minimum_cooldown_ms: 15000
+      glob_minimum_cooldown_ms: 500
```
enarha, line 20: Maybe do that 1s (1000)? IMO that'll be enough to detect the new logs, without stressing the system too much. I've already heard complains of vector (not our deployment) being too resource hungry.

khrm: I use this reference while setting this: https://github.com/vectordotdev/vector/issues/7934#issuecomment-864972060

enarha: OK. I was thinking most short lived taskruns I've seen are taking ~3s, so setting that value to 1s will be enough to not miss logs without over stressing the system. Anyway, we can start with that lower value and monitor the resource utilization and increase the value if needed.

PR #5327, opened 2025-01-20
kyverno: rollout to all staging clusters

sadlerap:
Kyverno's deployments in the existing staging environments have stabilized at this point, so it's safe to continue the rollout to the rest of the staging clusters.

### DIFF
--- a/argo-cd-apps/overlays/production-downstream/delete-applications.yaml
+++ b/argo-cd-apps/overlays/production-downstream/delete-applications.yaml
@@ -75,3 +75,9 @@ kind: ApplicationSet
 metadata:
   name: namespace-lister
 $patch: delete
+---
+apiVersion: argoproj.io/v1alpha1
+kind: ApplicationSet
+metadata:
+  name: kyverno
+$patch: delete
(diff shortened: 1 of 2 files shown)


filariow (approved):
/lgtm

argo-cd-apps/overlays/development/delete-applications.yaml
```diff
@@ -121,12 +121,6 @@ $patch: delete
 ---
 apiVersion: argoproj.io/v1alpha1
 kind: ApplicationSet
-metadata:
-  name: kyverno
-$patch: delete
----
-apiVersion: argoproj.io/v1alpha1
```
hugares, line 128: this change is causing kyverno to be deployed in dev overlay. Is this want you want? I am ok with it, just that commit message/PR title is mentioning staging clusters

sadlerap: Good catch - development overlays should be a separate PR I think.

sadlerap: Fixed in b9234b04.

hugares (approved):
/lgtm

PR #5340, opened 2025-01-21
Fixing application sync complaint about workspaces-system

dperaza4dustbit:
Taking care of this sync issue:

One or more objects failed to apply, reason: namespaces "workspaces-system" not found.


filariow (approved):
/lgtm

argo-cd-apps/base/member/infra-deployments/workspaces/workspaces.yaml
```diff
@@ -27,7 +27,6 @@ spec:
         repoURL: https://github.com/redhat-appstudio/infra-deployments.git
         targetRevision: main
       destination:
-        namespace: workspaces-system
```
hugares, line 30: Removing this is the right thing to do but we need to make sure none of the NS resources are missing NS name in the metadata.

I checked and only the 2 following resources are relying on Argo to set NS. Please set then NS explicitly on them as part of this change:

- https://github.com/redhat-appstudio/infra-deployments/blob/main/components/workspaces/team/base/workspaces-team-role.yaml#L1
- https://github.com/redhat-appstudio/infra-deployments/blob/main/components/workspaces/team/base/workspaces-team-rolebinding.yaml#L1

dperaza4dustbit: Got it

dperaza4dustbit: those resources are not needed any longer, removing them

components/workspaces/team/base/kustomization.yaml
hugares, line 1: kubelinter is failing at kustomize build because you did not remove usage of  "../../team/base" in:
- components/workspaces/staging/stone-stage-p01/kustomization.yaml
- components/workspaces/staging/stone-stg-host/kustomization.yaml

dperaza4dustbit: done

hugares (approved):
/lgtm

PR #5343, opened 2025-01-22
fix: add pod reader role to appstudio-pipeline SA

tisutisu:
In konflux-ci namespace, appstudio-pipeline SA needs pod reader role to archive e2e-tests logs.

### DIFF
--- a/components/konflux-ci/base/kustomization.yaml
+++ b/components/konflux-ci/base/kustomization.yaml
@@ -4,6 +4,8 @@ resources:
 - repository.yaml
 - konflux-ci-maintainers-rb.yaml
 - appstudio-pipelines-runner-rolebinding.yaml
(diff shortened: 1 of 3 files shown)


components/konflux-ci/base/pod-reader-role.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: pod-reader
```
mmorhun, line 4: I think it should be `pod-logs-reader`

mmorhun (approved):

components/konflux-ci/base/pod-reader-role.yaml
tisutisu: Updated the names now.

PR #5350, opened 2025-01-22
Remove limit for one knative eventing component

rh-hemartin:
I talked with the Eventing team and they have a fix, meanwhile the solution is to provide more memory. I don't want to keep trying values, so I opted to remove the limit for memory instead. What do you think? This will be removed and a limit will be established again when the fix is merged and then its released. I will keep an eye on this.


hugares (approved):
/lgtm

maruiz93 (approved):

PR #5351, opened 2025-01-22
Update Mintmaker image

nvtnlucie:
Updates the version of RPM lockfile.


staticf0x:
I think we can close this?

PR #5357, opened 2025-01-22
deploy namespace-lister to production rh02

sadlerap:
The deployments for namespace-lister appear to be stable on the relevant staging cluster (`stone-stg-rh01`), so we can begin to roll this out to the relevant production clusters.

For now, the only cluster we wish to deploy to is `kflux-prd-rh02`, since the component `konflux-ui` has not been deployed to any other production clusters yet.  All other clusters affected by the `konflux-public-production` overlay will see no new resources added to the cluster.

Adding namespace-lister to konflux-ui's production proxy config will come in a separate PR.

### DIFF
--- a/components/namespace-lister/empty-base/kustomization.yaml
+++ b/components/namespace-lister/empty-base/kustomization.yaml
@@ -1,0 +1,3 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: namespace-lister
--- a/components/namespace-lister/production/base/kustomization.yaml
+++ b/components/namespace-lister/production/base/kustomization.yaml
@@ -1,0 +1,27 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- deployment.yaml
+- namespace.yaml
+- rbac.yaml
+- service.yaml
+- network_policy.yaml
+- metrics/
+namespace: namespace-lister
+images:
+- name: namespace-lister
+  newName: quay.io/konflux-ci/namespace-lister
+  newTag: 1181ecb4ba10d6fbe5ea7bb94393b04cb2a932f8
+patches:
+- path: ./patches/with_cachenamespacelabelselector.yaml
+  target:
+    group: apps
+    kind: Deployment
+    name: namespace-lister
+    namespace: namespace-lister
+- path: ./patches/with_header_auth_user.yaml
+  target:
+    group: apps
+    kind: Deployment
+    name: namespace-lister
+    namespace: namespace-lister
--- a/components/namespace-lister/production/base/patches/with_cachenamespacelabelselector.yaml
+++ b/components/namespace-lister/production/base/patches/with_cachenamespacelabelselector.yaml
@@ -1,0 +1,5 @@
+- op: add
+  path: /spec/template/spec/containers/0/env/-
+  value:
+    name: CACHE_NAMESPACE_LABELSELECTOR
+    value: 'konflux.ci/type=user'
--- a/components/namespace-lister/production/kflux-prd-rh02/kustomization.yaml
+++ b/components/namespace-lister/production/kflux-prd-rh02/kustomization.yaml
@@ -1,0 +1,4 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- ../base/
--- a/argo-cd-apps/base/member/infra-deployments/namespace-lister/namespace-lister.yaml
+++ b/argo-cd-apps/base/member/infra-deployments/namespace-lister/namespace-lister.yaml
@@ -10,15 +10,21 @@ spec:
         generators:
           - clusters:
               values:
+                # This app must have cluster specific configurations, so we
+                # need to provide a config for clusters we do not wish to deploy
+                # to.  The `empty-base` clusterDir is an overlay that deploys no
+                # resources to non-matching clusters.
+                clusterDir: empty-base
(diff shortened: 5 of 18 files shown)


components/namespace-lister/production/kflux-prd-rh02/kustomization.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- ../../base/
```
filariow, line 4: I'd suggest to use a separate `base` under production so we don't share manifests between environments

filariow (changes requested):

components/namespace-lister/empty-base/kustomization.yaml
```diff
@@ -0,0 +1,3 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: namespace-lister
```
filariow, line 3:
```suggestion
```

gbenhaim:
I agree with @filariow comments, other than that, lgtm.

components/namespace-lister/empty-base/kustomization.yaml
sadlerap: Kustomize needs something other than an empty config, otherwise it will exit with an error (and the corresponding application in argo will always be reporting an error).  Applying this change yields the following:
```
$ kustomize build components/namespace-lister/empty-base
Error: kustomization.yaml is empty
```

Note that this configuration won't actually make any resources, since the namespace generator does nothing without any resources to act on.

components/namespace-lister/production/kflux-prd-rh02/kustomization.yaml
sadlerap: Fixed in d84b890f.

components/namespace-lister/production/base/kustomization.yaml
```diff
@@ -0,0 +1,20 @@
...
+- deployment.yaml
+- namespace.yaml
+- rbac.yaml
+- service.yaml
+- network_policy.yaml
+namespace: namespace-lister
+images:
+- name: namespace-lister
+  newName: quay.io/konflux-ci/namespace-lister
+  newTag: fd195c941b3151c165ddf376ce5f44d57db3f071
```
filariow, line 13: This image is quite old, can we bump it?

components/namespace-lister/production/base/kustomization.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
```
filariow, line 3: What about metrics?

filariow (changes requested):

components/namespace-lister/production/base/kustomization.yaml
sadlerap: Fixed in 28bd12fe.

components/namespace-lister/production/base/kustomization.yaml
sadlerap: Added in 28bd12fe.

gbenhaim (approved):

components/namespace-lister/production/base/patches/with_header_auth_email.yaml
```diff
@@ -0,0 +1,5 @@
+- op: add
+  path: /spec/template/spec/containers/0/env/-
+  value:
+    name: AUTH_USERNAME_HEADER
+    value: X-Email
```
filariow, line 5: production is already explicitly using usernames, let's use `X-User` here
* https://github.com/redhat-appstudio/infra-deployments/blob/265a37d5659ad07ae0951ecf7033cad2a4ad9e00/components/konflux-ui/production/base/proxy/proxy.yaml#L71
* https://github.com/redhat-appstudio/infra-deployments/blob/265a37d5659ad07ae0951ecf7033cad2a4ad9e00/components/konflux-ui/production/base/proxy/nginx.conf#L69

```suggestion
    value: X-User
```

We also need to update the filename to `with_header_auth_user.yaml`

filariow (changes requested):

components/namespace-lister/production/base/patches/with_header_auth_email.yaml
sadlerap: Fixed in 192f5512.

filariow (changes requested):

components/namespace-lister/production/base/patches/with_cachenamespacelabelselector.yaml
```diff
@@ -0,0 +1,5 @@
+- op: add
+  path: /spec/template/spec/containers/0/env/-
+  value:
+    name: CACHE_NAMESPACE_LABELSELECTOR
+    value: 'toolchain.dev.openshift.com/type=tenant'
```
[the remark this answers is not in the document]
sadlerap: Do we also need to change this on the staging clusters?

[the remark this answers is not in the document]
filariow: no, I'd say just this one. I'd prefer to have staging rh01 to be aligned to prod rh01 to test KubeSaw migration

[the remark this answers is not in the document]
sadlerap: Fixed in 96375032.

filariow (approved):
/lgtm

PR #5244, opened 2025-01-09
KONFLUX-5670: Enable NVMe Disk Configuration

manish-jangra:
This Pull Request will enable NVMe Disk Configuration for stone-stage-p01

PR #5358, opened 2025-01-23
Revert "KONFLUX-5670: Enable NVMe Disk Configuration (#5244)"

jhutar:
This reverts commit df73235785aa650efbd8afd975cf4141afa847b6.

Yesterday we discussed on "Konflux Performance WG weekly" that this NVMe permission problem is not up to Perf team to solve, so I'm going to stop looking into it, so we do not need it deployed on Stage.


hugares (approved):
/lgtm

PR #5362, opened 2025-01-23
Increase application-service requests nearer to current usage

ralphbean:
Long term, application-service-controller-manager memory usage is above its requests which makes it very fragile when there is a memory pressure on the cluster.

Fixes [KONFLUX-5891](https://issues.redhat.com//browse/KONFLUX-5891)


ralphbean:
Related, I'm proposing a change to the OWNERS file at https://github.com/redhat-appstudio/infra-deployments/pull/5363

hugares (approved):
/lgtm

PR #5370, opened 2025-01-23
namespace-lister: disable https temporarily

sadlerap:
Until TLS certificates have been configured, disable serving over https.


filariow (approved):
/lgtm

PR #5379, opened 2025-01-24
Mintmaker: create a cronjob for generating the OSV database

querti:
The database will be in a persistent volume which will be mounted by pods running renovate. The cronjob downloads the upstream OSV database and generates custom RPM and docker DB files on each run. The job uses the same image as MintMaker controller, so no modifications to the release process are necessary.

The reason why the cronjob was duplicated as a job is that the schedule "every X hours" does not create a job upon cronjob creation. This can cause issues when bootstrapping, since no DB will exist for several hours. The job is implemented as a hook, and the resource will self-delete after its run.

### DIFF
--- a/components/mintmaker/base/osv-database-builder/create-database-pvc.yaml
+++ b/components/mintmaker/base/osv-database-builder/create-database-pvc.yaml
@@ -1,0 +1,12 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: osv-offline-db
+  namespace: mintmaker
+spec:
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 10Gi
+  volumeMode: Filesystem
--- a/components/mintmaker/base/kustomization.yaml
+++ b/components/mintmaker/base/kustomization.yaml
@@ -1,6 +1,7 @@
 resources:
 - cronjobs/
 - rbac/
+- osv-database-builder/
 
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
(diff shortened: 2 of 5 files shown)


components/mintmaker/base/osv-database-builder/create-database-pvc.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
```
gbenhaim, line 1: The default storage class provides EBS volumes. The pvc is going to be bounded to a single node, it means that all the jobs that will use it would need to be scheduled on the same node, this can cause a bottleneck.

gbenhaim:
what is going to prevent the job to update the files in the pvc while other pods are using it?

components/mintmaker/base/osv-database-builder/create-database-pvc.yaml
gbenhaim: not at this moment.

gbenhaim:
I suggest to build a container image with the data base and then pull in the job that need it.
You can also upload the files as an OCI artifact and then pull them using tool like oras (https://oras.land/).
This approach will overcome the two issues I mentioned above.

gbenhaim:
> > what is going to prevent the from updating the files in the pvc while other pods are using it?
> 
> I haven't considered this possibility, but checking the docs of the package which will use these files (https://github.com/seald/nedb), it claims "A copy of the whole database is kept in memory". I assume it will load the file to memory and perform the queries on the in-memory version. So maybe this won't be much of an issue.
> 
> Regardless, is there a simple way to overcome this problem?

what is the size of that db?

gbenhaim:
> > I suggest to build a container image with the data base and then pull in the job that need it. You can also upload the files as an OCI artifact and then pull them using tool like oras (https://oras.land/). This approach will overcome the two issues I mentioned above.
> 
> The issue is that we want to refresh the data every few hours. I'm not sure we'd want to rebuild our image periodically like that.

why not? what is the issue with that?

qixiang:
/hold

Adding the do-not-merge label to prevent accidental merging, as this change is not ready for merging.

PR #5380, opened 2025-01-24
Remove invalid fields from monitoring kustomize transformation

hugares:
This transformation does have invalid fields which are ignored by kustomize version < 5.4.0. Starting with 5.4.0, kustomize uses strict unmarsheling so kustomize build command is failing. This change is needed to upgrade from gitops 1.12 to 1.15 which include ArgoCD version which uses kustomize version 5.4.3

[KFLUXINFRA-1211](https://issues.redhat.com//browse/KFLUXINFRA-1211)


pacho-rh (approved):
/lgtm
/approve

PR #5389, opened 2025-01-26
feat(KFLUXDP-140): Prepare all Konflux components to integrate with Sealights

flacatus:
This Pull Requests prepare all the konflux development stages to integrate with Sealights. The sealights images will require to have a token in the environment to work properly to send results to Sealights for tests statistics


components/build-service/development/sealights-token-patch.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  name: build-service-controller-manager
+spec:
+  template:
+    spec:
+      containers:
+      - name: manager
+        env:
+          - name: SEALIGHTS_TOKEN
+            value: ""
```
psturc, line 12: since we push the argocd changes during bootstrap to a public repo, wouldn't the token be exposed?

should we consider loading the token from the secret instead?

flacatus: do we want to make users to have a token locally?

psturc (approved):

PR #5036, opened 2024-12-03
move pipelines and resolvers controller from leader election to state…

jkhelil:
…fulsets

PR #5398, opened 2025-01-27
switch pipelines and remote resolver controllers on prod m01 to state…

jkhelil:
…fulSets

linked to https://github.com/redhat-appstudio/o11y/pull/462

### DIFF
--- a/components/pipeline-service/production/stone-prd-m01/deploy.yaml
+++ b/components/pipeline-service/production/stone-prd-m01/deploy.yaml
@@ -2090,40 +2090,6 @@ spec:
                     requests:
                       cpu: 100m
                       memory: 100Mi
-        tekton-pipelines-controller:
-          spec:
-            template:
-              spec:
-                containers:
-                - name: tekton-pipelines-controller
-                  resources:
-                    limits:
-                      memory: 12Gi
-                    requests:
-                      cpu: "1"
-                      memory: 12Gi
-                topologySpreadConstraints:
-                - labelSelector:
-                    matchLabels:
-                      app: tekton-pipelines-controller
-                  maxSkew: 1
-                  topologyKey: topology.kubernetes.io/zone
-                  whenUnsatisfiable: DoNotSchedule
-        tekton-pipelines-remote-resolvers:
-          spec:
-            replicas: 4
-            template:
-              spec:
-                containers:
-                - args:
-                  - --threads-per-controller=32
-                  name: controller
-                  resources:
-                    limits:
-                      memory: 8Gi
-                    requests:
-                      cpu: 500m
-                      memory: 8Gi
         tekton-pipelines-webhook:
           spec:
             template:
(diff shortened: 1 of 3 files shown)


jkhelil:
@hugares @enarha can you have a look ?

components/pipeline-service/production/stone-prd-m01/deploy.yaml
```diff
@@ -2173,12 +2139,48 @@ spec:
...
+                  topologyKey: topology.kubernetes.io/zone
+                  whenUnsatisfiable: DoNotSchedule
+        tekton-pipelines-remote-resolvers:
+          spec:
+            replicas: 4
+            template:
+              spec:
+                containers:
+                - args:
+                  - --threads-per-controller=32
```
enarha, line 2169: Is that really needed? I'm looking at the existing deployment on m01 and I see the following parameters passed to the container `-kube-api-burst=50 -kube-api-qps=50 -threads-per-controller=32 --threads-per-controller=32` . Isn't enough we have threads-per-controller in the performance field?

enarha (commented):
Strictly speaking that configuration slightly differs from the one in staging where `tekton-pipelines-remote-resolvers` is missing from the list of statefulSets. It is still working though as I see the statefulSet exist on the cluster. Ideally we'll have matching configurations in staging and prod. I do not mind which changes as far as they match. (I know that because I'm working on the same change, but you beat me to it).

components/pipeline-service/production/stone-prd-m01/deploy.yaml
enarha: > The threads-per-controller=32 setting should now be superseded by the performance flags defined under spec/pipeline/performance.

That was my point, given that threads-per-controller=32 is already set through spec/pipeline/performance and it is indeed passed to the container, we do not need it here.

> On the staging environment, we already have the same configuration for tekton-pipelines-controller. However, at the time I made this change, there was nothing related to remote-resolvers under spec/pipeline/deployments, as this section had already been moved to spec/pipeline/statefulSets

I see here https://github.com/redhat-appstudio/infra-deployments/pull/5036/files#diff-e961bdc3a405dfda1b39ed17ee82576bb68e2a35e9288e7de208e695e4ee5944L2226-L2228 is removed from deployments and not added to statefulSets. This change includes it. If we indeed need it, then we should update the staging overlay accordingly. I do not want to mix prod and staging changes, so I can open a separate PR to update staging. Once we deploy and validate it, we can proceed with this change.

> 
> if you are trying to align stage with prod, lets do your PR before this then, we can hold this for a bit

jkhelil: @enarha sound good to me, I wil wait for your PR on stage, once merged, we proceed with this one

enarha: OK, I'll open one in a min. But what about the `--threads-per-controller=32` (my original comment). Do you agree it should be removed?

jkhelil: Yes I will removed it

jkhelil:
/hold 
we need to discuss monitoring updates

enarha (approved):
/lgtm
/approve

PR #5400, opened 2025-01-27
Improve MintMaker memory resources

gnaponie:
We are experiences memory peaks, it might happen to reach the limit in high cluster loads period. Let's increase the memory a bit to prevent the pod from crashing.


gnaponie:
@staticf0x @nvtnlucie @FernandesMF PTAL

PR #5401, opened 2025-01-27
Fix namespace-claim-cleaner cronjob

amisstea:
The jobs would fail when multiple claims were found. The frequency of the cronjob has also been increased to once per hour. Otherwise users could have access to ephemeral namespaces for up to 24h.


hmariset (approved):
/lgtm

PR #5402, opened 2025-01-27
expose metrics for namespace lister

sadlerap:
See https://github.com/konflux-ci/namespace-lister/pull/39 for details on the metrics added.

### DIFF
--- a/components/namespace-lister/base/kustomization.yaml
+++ b/components/namespace-lister/base/kustomization.yaml
@@ -6,7 +6,7 @@ resources:
 - rbac.yaml
 - service.yaml
 - network_policy.yaml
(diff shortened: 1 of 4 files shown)


components/namespace-lister/base/kustomization.yaml
```diff
@@ -6,11 +6,12 @@ resources:
 - rbac.yaml
 - service.yaml
 - network_policy.yaml
+- ./metrics/
```
filariow, line 9: very minor, should we remove the `./` here?
```suggestion
- metrics/
```

filariow (approved):
/lgtm

PR #5415, opened 2025-01-28
PVO11Y-4603 Deploy cluster observability operator

kubasikus:
As detailed in this [KCS article](https://access.redhat.com/articles/7103797), the Observability Operator (OBO) will no longer be deployed automatically (by Hive) on ROSA clusters as of Jan 30th.

This means that the new Cluster Observability Operator (aka COO, the productized version of OBO), should be deployed on all Konflux clusters. 

Before rolling this out to production, we will test this only in the stone-stage-p01 cluster and validate. There will be a follow up PR where the production deployments will be added.


raks-tt (approved):
/approve
/lgtm

PR #5416, opened 2025-01-28
kvyerno: move helm values to file

sadlerap:
Keeping the helm values inline to the applicationset is getting in our way, so we should move away from this if we can.

This copies the approach that the `tracing-workload-otel-collector` applicationset uses.


hugares (approved):
/lgtm

filariow (approved):
/lgtm

PR #5418, opened 2025-01-28
namespace-lister: label service

sadlerap:
The servicemonitor will search for endpoints by label, and we're not setting any label on services, which will create endpoints.  This means that the servicemonitor isn't finding any metrics to use.

To fix this, apply the label that the servicemonitor is listening for.


filariow (approved):
/lgtm

PR #5377, opened 2025-01-24
feat: Add quay secret to push test artifacts

tisutisu:
This PR adding a quay secret needed to push the artifacts to `quay.io/konflux-test-storage/konflux-team/build-definitions` repo.
Part of the story: https://issues.redhat.com/browse/STONEBLD-3025

PR #5424, opened 2025-01-29
fix: the target name of the test-artifacts-push-secret

tisutisu:
- when created this secret in the [last PR](https://github.com/redhat-appstudio/infra-deployments/pull/5377), target name was wrong
- added github-bot-token key in the secret, needed for interacting with github


mmorhun (approved):
/lgtm

PR #5429, opened 2025-01-29
PVO11Y-4603 Install Cluster Observability Operator in stage and prod

kubasikus:
Observability Operator will no longer be installed by default on ROSA clusters starting Jan 30th 2025. Deploy Cluster Observability Operator, a productized version of OBO, in both stage and prod to keep the related Resources synchronized.


pacho-rh (approved):
/lgtm
/approve

PR #5430, opened 2025-01-29
namespace-lister: provide a tls config to ServiceMonitor

sadlerap:
The service monitor doesn't know anything about the service it's talking to in terms of TLS, so we need to provide the following information:

- A Certificate Authority for requests to authenticate against.
- A server name to verify that we're talking to the correct service.


filariow (approved):
/lgtm

PR #5433, opened 2025-01-30
fix: the secret key name and use `_` instead

tisutisu:
during argo sync of this ExternalSecret..  resource is degraded with this error 
```
could not apply template: could not execute template: could not execute template: unable to execute template at key github-bot-commenter-token: unable to parse template at key github-bot-commenter-token: template: github-bot-commenter-token:1: bad character U+002D '-'
```
`-` character in the key name is causing the issue, so changed key name to use `_` instead.


chmeliik (approved):
/lgtm
/approve

PR #5435, opened 2025-01-30
fix: remove template part from the external secret

tisutisu:
Still we are getting this error
```
could not apply template: could not execute template: could not execute template: unable to execute template at key github-bot-commenter-token: unable to parse template at key github-bot-commenter-token: template: github-bot-commenter-token:1: bad character U+002D '-'
```
Removing the template part and use the secret as it is.


chmeliik (approved):
/lgtm
/approve

dirgim (approved):
/lgtm

PR #5436, opened 2025-01-30
[Staging] Change Tekton Vector collection to in-memory and ack enabled

khrm:
This is to resolve missing logs issue.


khrm (commented):
/assign @enarha

enarha:
/lgtm
/approve
/hold

Adding the hold label as we probably want to bump the memory limit to 3GB (.resources.requests.limits.memory). I was checking a prod instance and at regular load the memory usage was standing around 1 GB. Now with the in memory buffer, the usage will increase and it's better to avoid the OOM and pod restarts.

khrm:
@enarha I have updated the memory to 3GB.

PR #5443, opened 2025-01-31
increase namespace-lister resources

filariow:
namespace-lister is getting OOMKilled. Let's give it a bit more resources.

### DIFF
--- a/components/namespace-lister/base/deployment.yaml
+++ b/components/namespace-lister/base/deployment.yaml
@@ -40,7 +40,7 @@ spec:
         resources:
           limits:
             cpu: 500m
-            memory: 256Mi
+            memory: 1Gi
           requests:
             cpu: 500m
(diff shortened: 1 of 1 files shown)


components/namespace-lister/base/deployment.yaml
```diff
@@ -40,10 +40,10 @@ spec:
         resources:
           limits:
             cpu: 500m
-            memory: 128Mi
+            memory: 256Mi
```
sadlerap, line 43: As long as we're increasing things, maybe let's set the memory limit to something much higher than what we're requesting?

```suggestion
            memory: 1Gi
```

filariow: makes sense. commited

sadlerap (approved):
/lgtm

PR #5482, opened 2025-02-06
release new version of namespace-lister

filariow:
* https://github.com/konflux-ci/namespace-lister/pull/51
* https://github.com/konflux-ci/namespace-lister/pull/48

### DIFF
--- a/components/namespace-lister/base/metrics/monitor.yaml
+++ b/components/namespace-lister/base/metrics/monitor.yaml
@@ -9,7 +9,7 @@ spec:
     - interval: 15s
       scheme: https
       path: /metrics
-      port: http
+      port: metrics
       authorization:
         credentials:
           key: token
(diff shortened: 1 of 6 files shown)


dperaza4dustbit (approved):
/lgtm

components/namespace-lister/base/metrics/monitor.yaml
```diff
@@ -18,9 +18,9 @@ spec:
         ca:
           secret:
             key: tls.crt
-            name: namespace-lister-tls
+            name: namespace-lister-metrics-tls
```
sadlerap, line 21: hmmm, we don't have a method for using different keys for both the api server and the metrics server.  Why change this?

sadlerap (changes requested):

components/namespace-lister/base/metrics/monitor.yaml
filariow: right, so we need to rely on the cert generated for the other service... that's a bit ugly

filariow: should be fixed by 3f25ba1e01d24fddd928f26a1e4a330035050787

sadlerap (approved):
/lgtm

PR #5483, opened 2025-02-06
PVO11Y-4624 Increase COO resource limits

kubasikus:
This should prevent out-of-memory Pod killing of the Obervability Operator deployed by the Cluster observability operator.

This has a side effect of increasing limits of all pods managed by this operator.


hugares (approved):
/lgtm

PR #5486, opened 2025-02-06
Add per-platform instance tags

enkeefe00:
When determining if a particular VM instance can be created, the VM flavor's maximum number of instances is compared to the current number of instances with the same `instance-tag` to ensure no more than the maximum number of VMs are created. Since the current configurations have all flavors using the same instance tag, the count of unique instances (flavors, in this case) is wrong. Adding per-platform/flavor instance tags should fix this.

### DIFF
--- a/components/multi-platform-controller/staging-downstream/host-config.yaml
+++ b/components/multi-platform-controller/staging-downstream/host-config.yaml
@@ -55,6 +55,7 @@ data:
   dynamic.linux-arm64.region: us-east-1
   dynamic.linux-arm64.ami: ami-03d6a5256a46c9feb
   dynamic.linux-arm64.instance-type: m6g.large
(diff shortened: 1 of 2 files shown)


mshaposhnik:
In general, this approach is OK for me. I have the following tag format proposal though:

```
[prod|stage]-[platform]-[optional_extra_quantifier]
```
 i..e `prod-amd64`, `stage-arm64-2mxlarge`, `prod-ppc64-d200`

The reason for that is that some IBM accounts hosts both staging and prod VMs, and it has to have different tags both for the amount control as well as for us "adminitrators" is always helpful to see in the console where those machines come from.

And I would like ask to add more explicit examples into the documentation, so that fuctionality become more trasparent.

enkeefe00:
@hugares or @mshaposhnik can you guys give this an LGTM so I can try to test it in the staging clusters?

PR #5497, opened 2025-02-10
[prod] switch pipelines and remote resolver controllers to statefulSets

jkhelil:
After few days checking on prod m01, this PR shanges tekton pipelines controller and tekton resolvers controller to statefulSet on all prod environement


enarha (approved):
/lgtm
/approve

PR #5500, opened 2025-02-10
Add read perms of kac resources

maruiz93:
KubeArchive maintainers need permissions to
read the kubearchive resources clusterwide.


rh-hemartin:
I'm not sure about the clusterwide permission in relation to security practices, @hugares can you comment on that? Otherwise looks good to me.

hugares:
> I'm not sure about the clusterwide permission in relation to security practices, @hugares can you comment on that? Otherwise looks good to me.

Clusterwide permissions are ok if the permission does not grant access to secrets, password, tokens,... Other Konflux teams do the same to be able to see the resource their controller is reconciling.

hugares (approved):
/lgtm

PR #5508, opened 2025-02-10
fix(RELEASE-1441): allow rsc in development

scoheb:
- the rsc needs to managed outside of the release
      service so that clusters can have a specific version
      for testing.
- A version can be deployed and managed by ArgoCD
      via infra-deployments for a specific environment
- This PR only affects the development overlay
- includes contents of PR https://github.com/redhat-appstudio/infra-deployments/pull/5541

### DIFF
--- a/components/release/development/release_service_config.yaml
+++ b/components/release/development/release_service_config.yaml
@@ -1,0 +1,6 @@
+apiVersion: appstudio.redhat.com/v1alpha1
+kind: ReleaseServiceConfig
+metadata:
+  name: release-service-config
+spec:
+  debug: true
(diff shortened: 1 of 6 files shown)


johnbieren:
What about the konflux-ci installer thing? If we manage the RSC here instead of release-service, won't we break them?

components/release/development/release_service_config.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: appstudio.redhat.com/v1alpha1
+kind: ReleaseServiceConfig
+metadata:
+  name: release-service-config
```
johnbieren, line 4: I am fine with either name, but it should be the same across envs

scoheb: either name? This is the name that the controller expects

scoheb: I see the issue. thanks

scoheb:
> If we manage the RSC here instead of release-service, won't we break them

I guess we will. I'll have to supply a patch to them

johnbieren (approved):

happybhati (approved):
/lgtm
/approve

PR #5515, opened 2025-02-11
KubeArchive: increase operator memory limit

rh-hemartin:
KubeArchive's operator is being OOMKilled, it needs more memory. We don't know why this is the case, in theory nothing points to increased memory usage with increased cluster size, but looks like that is the problem.

### DIFF
--- a/components/kubearchive/base/kustomization.yaml
+++ b/components/kubearchive/base/kustomization.yaml
@@ -132,6 +132,12 @@ patches:
                 runAsNonRoot: true
               ports:
                 - containerPort: 8081
+              resources:
+                requests:
+                  memory: 256Mi
+                limits:
+                  memory: 256Mi
+                
 - patch: |-
     apiVersion: apps/v1
     kind: Deployment


components/kubearchive/base/kustomization.yaml
```diff
@@ -132,6 +132,9 @@ patches:
                 runAsNonRoot: true
               ports:
                 - containerPort: 8081
+              resources:
+                limits:
+                  memory: 256Mi
```
ggallen, line 137: I think we should increase the request as well.

hugares: this is correct, requests is the most important one in fact. Scheduling is done base on requests and only request is guaranteed. I suggest to use the pattern of setting memory requests=memory limits

rh-hemartin: Changed to make it guaranteed at 256Mi

rh-hemartin:
> I get an error when I run locally. What's different?

I can't see the KAC on the cluster, also the error here is different. In theory it should be populated by changes made in https://github.com/redhat-appstudio/infra-deployments/pull/5479. The error can be seen at https://console-openshift-console.apps.stone-stg-rh01.l2vh.p1.openshiftapps.com/k8s/ns/product-kubearchive/pods/kubearchive-operator-98dfc7d5d-vnd6m/logs

ggallen:
> I can't see the KAC on the cluster

I was able to see it via argoCD using the link you sent.

rh-hemartin:
> And I don't see any error in that log file.

Yep, sorry about that. The error is not on the logs, but you can see where it "stopped". The error I'm referring to is on ArgoCD as well: https://argocd-server-konflux-public-staging.apps.rosa.appsres09ue1.24ep.p3.openshiftapps.com/applications/konflux-public-staging/kubearchive-stone-stg-rh01?view=tree&resource=&node=%2FPod%2Fproduct-kubearchive%2Fkubearchive-operator-98dfc7d5d-vnd6m%2F0

```
Container is waiting because of CrashLoopBackOff. It is not started and not ready.
The container last terminated with exit code 137 because of OOMKilled.
```

rh-hemartin:
> I get an error when I run locally. What's different?

Also the error from the issue you mention occurs after the scenario that happens related to this PR. The operator on Konflux staging stops at `in reconcileFilterConfigMap` but with the issue you mention there are 7 or 8 more lines of logs.

components/kubearchive/base/kustomization.yaml
```diff
@@ -132,6 +132,12 @@ patches:
                 runAsNonRoot: true
               ports:
                 - containerPort: 8081
+              resources:
+                resources:
```
skoved, line 136: shouldn't this second `resources` be `requests`

```suggestion
                requests:
```

rh-hemartin: You are totally right, thanks!

rh-hemartin: Weirdly, kubelinter does not flag it as an error.

hugares (approved):
/lgtm

skoved (approved):

PR #5518, opened 2025-02-11
KONFLUX-5876 Deploy logging collectors on tainted nodes in stage clus…

kubasikus:
…ters

Enables user log collection from tenant workloads.

Adds tolerations to pods deployed by the Logging Operator. This should allow said pods to schedule on Nodes with this taint.

Also add resource requests and limits as starting point. They will likely have to be tweaked in the future based on Pod load.


FaisalAl-Rayes (approved):

raks-tt (approved):
/lgtm

PR #5525, opened 2025-02-11
Allow namespace admins to list and watch pods

gbenhaim:
This permissions is required in order tail logs
using opc.


hugares (approved):
/lgtm

PR #5529, opened 2025-02-11
namespace-lister: unify metrics and api service

sadlerap:
Metrics is getting served on a different service from the api service that namespace-lister provides, but they use the same tls certificate. This is causing some issues, so let's unify these behind the same service.

TODO: do we need to update the network policies for this service to allow traffic from openshift user workload monitoring?


filariow (approved):
/lgtm

PR #5534, opened 2025-02-12
Fix typo on kubearchive-maintainer cr

maruiz93:
The resource name needs to be in plural, the same way it appears in `kubectl api-resources`.

This should fix the current issue of not being able to see this resources.


hugares (approved):
/lgtm

PR #5560, opened 2025-02-14
Make large IBM disk flavour of large machine

mshaposhnik:


components/multi-platform-controller/production-downstream/stone-prod-p02/host-config.yaml
```diff
@@ -463,6 +447,23 @@ data:
   dynamic.linux-large-s390x.allocation-timeout: "1800"
 
 
+  # Same as linux-s390x-large but with 200GB disk instead of default 100GB
+  dynamic.linux-d200-s390x.type: ibmz
```
hugares, line 451: should we put "large" in the name? so we now these are same as "linux-large-s390x" but with large disk

mshaposhnik: Fixed

hugares (approved):
/lgtm

PR #5561, opened 2025-02-14
use correct label selector for metrics

sadlerap:
We don't label any of our services with the `apps: namespace-lister-metrics` label, so we shouldn't be using it in our service monitor to watch for metrics.


filariow (approved):
/lgtm nice catch!

filariow (approved):

PR #5568, opened 2025-02-17
Revert memory settings for MPC on RH01 to default (10Gb)

mshaposhnik:
After recent changes with rh02 detachment and update of kube-runtime version, memory consumption seems stablilized on the level below 10Gb. 
The problem with 16Gb is that is is usualy requires new node to start a new pod, which takes a lot of time during updates.

image


hugares (approved):
/lgtm

PR #5578, opened 2025-02-17
Remove finalizer workaround for cluster templates

amisstea:
This is currently causing issues with cleanup in production.

The recently upgraded ArgoCD no longer requires the workaround. The apps are getting stuck in the deleting state while the argocd controllers endlessly append duplicate entries to the finalizers list.


yftacherzog (approved):
/lgtm
/approve

PR #5579, opened 2025-02-17
Remove patch for memory limit.

ggallen:


hugares:
@ggallen I forgot that we required memory limit in the kubelinter check. Either you wave this kubelinter check for the operator or set limit way higher.

What is the issue we are trying to solve here, container is getting OOMKilled?

ggallen:
@hugares, yeah, just trying to confirm that it is OOM actually. I can't believe it is. It's happening on a Kubernetes API `Get` call to retrieve the sink `ConfigMap`.

I have no idea how to waive the test.  Do I even have permissions to do that?

skoved (approved):

hugares:
> @hugares, yeah, just trying to confirm that it is OOM actually. I can't believe it is. It's happening on a Kubernetes API `Get` call to retrieve the sink `ConfigMap`.
> 
> I have no idea how to waive the test. Do I even have permissions to do that?

here is how to ignore this lint:

```
diff --git a/components/kubearchive/base/kustomization.yaml b/components/kubearchive/base/kustomization.yaml
index 342276b6..403d0062 100644
--- a/components/kubearchive/base/kustomization.yaml
+++ b/components/kubearchive/base/kustomization.yaml
@@ -122,6 +122,8 @@ patches:
     metadata:
       name: kubearchive-operator
       namespace: kubearchive
+      annotations: ignore-check.kube-linter.io/unset-memory-requirements: >
+        "This is temporary to troubleshoot an issue"
     spec:
       template:
         spec:
```

skoved (approved):

components/kubearchive/base/kustomization.yaml
```diff
@@ -122,6 +122,9 @@ patches:
     metadata:
       name: kubearchive-operator
       namespace: kubearchive
+      annotations: ignore-check.kube-linter.io/unset-memory-requirements: >
+        "This is temporary to troubleshoot an issue"
+
```
hugares, line 127: sorry, suggested code was not right, it should be:
```suggestion
      annotations:
        ignore-check.kube-linter.io/unset-memory-requirements: >
          "This is temporary to troubleshoot an issue"
```

ggallen:
@hugares, I added it but the test is still listed as failing?

hugares:
> @hugares, I added it but the test is still listed as failing?

yeah, my bad, see my in file comment

skoved (approved):

hugares (approved):
/lgtm

PR #5582, opened 2025-02-17
Fix split Results API service staging

enarha:
A previous PR introduced that change, which was properly configured in the development overlay (CI), but not in the staging overlay.


hugares (approved):
/lgtm

PR #5583, opened 2025-02-17
Fix Results API cert after service split

enarha:
The Results API service was split into two services. This fixes the certificate of the new service.


hugares (approved):
/lgtm

PR #5584, opened 2025-02-17
Revert max instances decrease on internal

enkeefe00:
Testing has been completed regarding the maximum number of instances for a particular VM flavor, so this reverts the maximum number of instances back to their original values on the internal cluster.


hugares (approved):
/lgtm

PR #5585, opened 2025-02-17
Add per-platform instance tags to production

enkeefe00:
When determining if a particular VM instance can be created, the VM flavor's maximum number of instances is compared to the current number of instances with the same instance-tag to ensure no more than the maximum number of VMs are created. Since the current configurations have all flavors using the same instance tag, the count of unique instances (flavors, in this case) is wrong. Adding per-platform/flavor instance tags has been shown to fix this in the staging environment.

### DIFF
--- a/components/multi-platform-controller/production-downstream/kflux-ocp-p01/host-config.yaml
+++ b/components/multi-platform-controller/production-downstream/kflux-ocp-p01/host-config.yaml
@@ -66,6 +66,7 @@ data:
   dynamic.linux-arm64.region: us-east-1
   dynamic.linux-arm64.ami: ami-03d6a5256a46c9feb
   dynamic.linux-arm64.instance-type: m6g.large
+  dynamic.linux-arm64.instance-tag: prod-arm64
   dynamic.linux-arm64.key-name: kflux-ocp-p01-key-pair
   dynamic.linux-arm64.aws-secret: aws-account
   dynamic.linux-arm64.ssh-secret: aws-ssh-key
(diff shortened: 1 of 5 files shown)


components/multi-platform-controller/production-downstream/kflux-ocp-p01/host-config.yaml
```diff
@@ -583,22 +615,24 @@ data:
...
-  dynamic.linux-d200-large-ppc64le.crn: "crn:v1:bluemix:public:power-iaas:dal12:a/eebc7153d0c844f9915fb6f3b9850a3b:d6faf373-c0b6-4ba9-b0bc-4f7e0a74f9d5::"
-  dynamic.linux-d200-large-ppc64le.url: "https://us-south.power-iaas.cloud.ibm.com"
-  dynamic.linux-d200-large-ppc64le.network: "4d164943-b043-4c6e-bba5-d233abf66c6a"
-  dynamic.linux-d200-large-ppc64le.system: "e980"
-  dynamic.linux-d200-large-ppc64le.cores: "4"
-  dynamic.linux-d200-large-ppc64le.memory: "16"
-  dynamic.linux-d200-large-ppc64le.max-instances: "30"
-  dynamic.linux-d200-large-ppc64le.allocation-timeout: "1800"
-  dynamic.linux-d200-large-ppc64le.disk: "200"
+  dynamic.linux-d400-large-ppc64le.type: ibmp
```
mshaposhnik, line 621: this should be `d200` both tag and name

mshaposhnik (approved):

PR #5587, opened 2025-02-18
Temporarily disable sync of kubearchive application

hugares:
Kuebearchive is only deployed on staging clusters and there is an issue so disable the auto sync to allow modifying the deployment in staging to troubleshoot the issue. This change will be reverted once troubleshooting is done.

### DIFF
--- a/argo-cd-apps/overlays/konflux-public-staging/delete-sync-policy.yaml
+++ b/argo-cd-apps/overlays/konflux-public-staging/delete-sync-policy.yaml
@@ -1,0 +1,2 @@
+- op: remove
+  path: /spec/template/spec/syncPolicy
--- a/argo-cd-apps/overlays/konflux-public-staging/kustomization.yaml
+++ b/argo-cd-apps/overlays/konflux-public-staging/kustomization.yaml
@@ -12,3 +12,9 @@ resources:
 namespace: konflux-public-staging
 patchesStrategicMerge:
   - delete-applications.yaml
+patches:
+ - path: delete-sync-policy.yaml
+   target:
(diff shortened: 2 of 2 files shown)


ggallen (approved):

ggallen:
> @ggallen @rh-hemartin My change will need to be reworked. Removing sync policy is causing e2e to fail because kubearchive application stays out of sync...
> 
> Conceptually, we would need to disable sync only on staging

Can we just skip this test for now?

I have a fix that I believe works, but argoCD keeps overwriting it.

hugares:
> > @ggallen @rh-hemartin My change will need to be reworked. Removing sync policy is causing e2e to fail because kubearchive application stays out of sync...
> > Conceptually, we would need to disable sync only on staging
> 
> Can we just skip this test for now? Can you force the merge?
> 
> I have a fix that I believe works, but argoCD keeps overwriting it.

we cannot because every PR in infra-deployment will fail e2e until we revert that change

ggallen:
That's unfortunate. So how do you propose I test my change?  I guess I'll have to go through all the hoops to create a PR that points to my gist that in turn points to my own image?

hugares:
> That's unfortunate. So how do you propose I test my change? I guess I'll have to go through all the hoops to create a PR that points to my gist that in turn points to my own image?

I meant we can still disable sync but we should do it only on staging. Updating my PR.

hugares:
@ggallen is external staging OK? Or you needed to disable it on internal staging as well?

hugares:
This PR is only changing staging overlay, which is not tested by e2e, merging now, no point in waiting for the tests

PR #5592, opened 2025-02-18
Fix Results API cert after service split - development

enarha:
The issue did not actually affect development, but we've seen it in staging. This brings the development overlay in sync with staging.


hugares (approved):
/lgtm

PR #5595, opened 2025-02-18
Read Results API service metrics from api-for-watcher

enarha:
Splitting the load of the Results API service, switched the watcher/controller related request  to a new pod which handles most of the load. With the original service now used only by users, it gets much less load (that was the purpose), but it triggers some false positives from the monitoring service. Here we switch the ServiceMonitor to the new pods running most of the load.

### DIFF
--- a/components/pipeline-service/development/main-pipeline-service-configuration.yaml
+++ b/components/pipeline-service/development/main-pipeline-service-configuration.yaml
@@ -1810,7 +1810,7 @@ spec:
   jobLabel: app.kubernetes.io/name
   selector:
     matchLabels:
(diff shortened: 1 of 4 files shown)


aThorp96 (commented):
@enarha I'm worried that this could be a moving target, and could potentially cause the service monitor to falsely alert on request volume for the regular results-api service. 

What do you think about instead doing the following, like we do with Tekton Chains:
- on the Results services, add the labels `app.kubernetes.io/component:` with `watcher` and `api` respectively
- switch the label selector to use `app.kubernetes.io/part-of: tekton-results` and `app.kubernetes.io/component: api`.

PR #5597, opened 2025-02-18
Refactor results api labels and service monitor

aThorp96:
The Service Monitor for Tekton Results needs to label with `job: tekton-results-api` since pagerduty alerts and grafana dashboards both depend on the `job` label. However the monitor needs to match both the normal `tekton-results-api` service as well as the new "writer" service `tekton-results-api-for-watcher`.

This PR adds the `app` and `app.kubernetes.io/component` labels to all `tekton-results` services, similar to what is done for tekton chains, in order for the ServiceMonitor to correctly match both API services and not the Watcher service.


enarha (approved):
/lgtm

PR #5603, opened 2025-02-19
Pull Results API metrics from both Services

enarha:
After splitting the workload betwen two services, this updates the ServiceMonitor to monitor both.


divyansh42 (commented):
/lgtm

PR #5613, opened 2025-02-20
KFLUXSPRT-1907: Adding New ARM Instance Type for Multi-Platform Build

manish-jangra:
This pull request will add `c6gd.2xlarge` instance type for ARM64 Platform Builds, this instance type has a local NVMe Disk of 400GB attached. This disk will be configured and mounted during runtime to `/vat/tmp` and other directories and this should us enough enough room in `/var/tmp` directory.


meyrevived:
@manish-jangra do you only want to add this type of machine to that one cluster? Why not update them all to have the option for it?

mshaposhnik (approved):

PR #5634, opened 2025-02-24
don't filter namespace-lister from internal prod overlay

sadlerap:
The application set for namespace-lister is ready to go - don't filter it from the overlay.

PR #5635, opened 2025-02-24
ns-lister: apply prod overlay patch for prod deployments

sadlerap:
Without this patch, argocd tries to pull the production manifests from the staging environment, which is not correct.

Fixes #5634.


filariow (approved):
/lgtm

hugares (approved):
/lgtm

PR #5670, opened 2025-02-27
KFLUXINFRA-1269: Adding Configuration to etcd defrag periodically

manish-jangra:
This Pull Request will create following resources -
- Namespace
- ClusterRole
- ClusterRoleBinding
- ServiceAccount
- CronJob that will run every 5 minutes

### DIFF
--- a/configs/etcd-defrag/base/cronjob.yaml
+++ b/configs/etcd-defrag/base/cronjob.yaml
@@ -1,0 +1,24 @@
+---
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: etcd-maintenance
+  namespace: etcd-maintenance
+spec:
+  schedule: "*/5 * * * *"
+  jobTemplate:
+    spec:
+      template:
+        spec:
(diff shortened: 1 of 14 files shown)


configs/etcd-defrag/base/cronjob.yaml
```diff
@@ -0,0 +1,24 @@
+---
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: etcd-maintenance
+  namespace: etcd-maintenance
+spec:
+  schedule: "*/15 * * * *"
```
hugares, line 8: in you PR description, you say cron that run every 5 minutes and here it is 15. Which one it wrong?

IMHO: I think 5 minutes is better

manish-jangra: okay, changed to 5 minutes

hugares (approved):
/lgtm

/hold @manish-jangra Remove hold when you are ready to merge and monitor deployment

PR #5681, opened 2025-02-27
PVO11Y-4650 - Add costmanagement-metrics-operator ns to cluster-secret-store

raks-tt:
This change is required for cost-management metrics operator to access the secret stored in vault.


hugares (approved):
/lgtm

PR #5683, opened 2025-02-27
Allow MPC sample multi-arch repo on staging

enkeefe00:
This repo is private so it's been added to the allow list.


hugares (approved):
/lgtm

PR #5685, opened 2025-02-27
Adding temp access to workspace team to assist with migration

dperaza4dustbit:
Missing prod-p01 and ocp-p01 from the list that needs to be migrated off KubeSaw


hugares (approved):
/lgtm

PR #5686, opened 2025-02-27
[OCP] linux-d250-largecpu variants

ashwindasr:
Since 250 GB is the max limit for now. This VM to be used for ose-installer-artifacts


hugares (approved):
/lgtm

PR #5691, opened 2025-02-28
[OCP] rename linux 250d variant

ashwindasr:
Since `linux-d200-large-ppc64le` is working, we should be good with this new name `linux-d250-8x16-ppc64le`


mshaposhnik:
you need shorter tag, profile name doesn't matter

ashwindasr:
Thanks @mshaposhnik, updated PR.

mshaposhnik (approved):

PR #5692, opened 2025-02-28
[ocp] increase default linux-large max-instances to 30

ashwindasr:
10 instances might be too low. Requesting the number to be at least the same as the default VM flavor


hugares (approved):
/lgtm

PR #5703, opened 2025-03-03
Bump Vector (Results) batch max size

enarha:
The batch max size is cutting logs short and the default value (which is also wrongly documented) cuts the logs at around 140KB size. With the value 524288000 I was able to store log files of size close to 10MB, which should be enough given the biggest step/container logs files I've encounted in Konflux are ~4.7MB.

Removed kubernetes_logs.max_read_bytes as it does not play a role in the logs storing and it was confusing.


hugares (approved):
/lgtm

PR #5704, opened 2025-03-03
chore: Remove Pat from integration-service's OWNERS file

dheerajodha:
This is because Pat has left the integration service team, and I would like to make his life a little bit better.

Signed-off-from: Dheeraj\<djodha@redhat.com>


dirgim (approved):
/lgtm
/approve

PR #5706, opened 2025-03-03
Increase memory request and limit for appstudio-federate-ms

pacho-rh:
Resolves: PVO11Y-4622

### DIFF
--- a/components/monitoring/prometheus/base/monitoringstack/monitoringstack.yaml
+++ b/components/monitoring/prometheus/base/monitoringstack/monitoringstack.yaml
@@ -23,7 +23,9 @@ spec:
   resources: # ensure that you provide sufficient amount of resources
     requests:
       cpu: 500m
-      memory: 1Gi
+      memory: 16Gi
+    limits:
+      memory: 24Gi
   prometheusConfig:
     externalLabels: {} # added by overlays
     replicas: 2  # ensures that at least one prometheus is running during upgrade
--- a/components/monitoring/prometheus/base/monitoringstack/monitoringstack.yaml
+++ b/components/monitoring/prometheus/base/monitoringstack/monitoringstack.yaml
@@ -25,7 +25,7 @@ spec:
       cpu: 500m
       memory: 16Gi
     limits:
-      memory: 24Gi
+      memory: 16Gi
   prometheusConfig:
     externalLabels: {} # added by overlays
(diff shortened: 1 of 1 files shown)


components/monitoring/prometheus/base/monitoringstack/monitoringstack.yaml
```diff
@@ -23,7 +23,9 @@ spec:
   resources: # ensure that you provide sufficient amount of resources
     requests:
       cpu: 500m
-      memory: 1Gi
+      memory: 16Gi
+    limits:
+      memory: 24Gi
```
hugares, line 28: I suggest we set the limit the same as requests, i.e. 16Gi in current version of the change. I find 16Gi quite high., this means almost one node reserved only for that pod. Are we sure we need that much memory, is it possible there is an issue with this pod, i.e. memory leak?

hugares:
> Current configuration is preventing the MonitoringStack from being deployed in the public production Konflux cluster as the requested 1Gi of memory is above the memory limit of 512Mi (some new default in the cluster?)

Since when this pod is not starting? Is something broken because of it, do we need that pod?

pacho-rh:
> > Current configuration is preventing the MonitoringStack from being deployed in the public production Konflux cluster as the requested 1Gi of memory is above the memory limit of 512Mi (some new default in the cluster?)
> 
> Since when this pod is not starting? Is something broken because of it, do we need that pod?

Seems like the pod has not started since Feb 6th 2025 @ 10am UTC looking at AppSRE Grafana (using PromQL such as `container_memory_usage_bytes{source_cluster=~"stone-prd-rh01" on the RHOBS datasource}`

This actually affects two pods. Both of which are the prometheus instances I believe are responsible for collecting and forwarding metrics from the data plane to RHOBS.

hugares (approved):
/lgtm

components/monitoring/prometheus/base/monitoringstack/monitoringstack.yaml
pacho-rh: Updated the resource limit to match the request (16Gi).

I felt the request is pretty high myself. We'll need to watch the pods' memory resources carefully once we get it back up and running. Identifying and seeing if we can fix any memory leak issues. But for now, thinking we can set it like to get the pod back up and potentially prevent it from being evicted from memory pressure.

PR #5710, opened 2025-03-04
Correct c6gd2xlarge platform host configurations

enkeefe00:
An issue provisioning a c6gd2xlarge VM instance for the private staging cluster prompted a look at the host configuration for this platform. The AMI field for the private staging cluster in addition to instance tags for this platform are fixed in this commit.


manish-jangra (approved):
/lgtm

PR #5726, opened 2025-03-05
Tune Vector Results config

enarha:
- Using bigger batches requires more memory
- Update batch max_bytes
- Remove internal_metrics source, not used by any sinc
- Remove S3 acknowledgements, not supported by the kubernetes_logs source


hugares (approved):
/lgtm

PR #5728, opened 2025-03-05
add konflux-ui to kflux-ocp-p01

sadlerap:
Adds konflux-ui to the internal production cluster `kflux-ocp-p01`.


filariow (approved):
/lgtm

PR #5733, opened 2025-03-06
Deploy latest crossplane configs

amisstea:
This change should fix an issue where ephemeral namespace names containing illegal dot characters couldn't be created.


yftacherzog (approved):
/approve
/lgtm

PR #5735, opened 2025-03-07
Reduce minio replicas to 1 for development

rh-hemartin:
This PR aims to reduce the number of replicas specified for development related to the minio operator. Currently the replicas is set to 2, but the pods have a antiaffinity constraint in which they can't be in the same node as another pod called minio-operator. This results in the second replica not being deployed, leaving the deployment in a degraded state, so the `preview.sh` script can't progress.


hugares (approved):
/lgtm

PR #5743, opened 2025-03-07
fix(KFLUXSPRT-2070): fix taskrun PaC url template for rh02

ralphbean:
There is no application-pipeline/ prefix here.


hugares (approved):
/lgtm

PR #5751, opened 2025-03-10
add kyverno to stone-prd-rh01

sadlerap:
Configuration is a copy of kflux-prd-rh02, since there's nothing specific to that cluster in its configuration.


filariow (approved):
/lgtm

PR #5755, opened 2025-03-10
nslister: add egress NetworkPolicy

filariow:
This change allows the namespace-lister to only reach-out to the APIServer


components/namespace-lister/base/network_policy_allow_to_apiserver.yaml
```diff
@@ -0,0 +1,22 @@
...
+  policyTypes:
+  - Egress
+  egress:
+  - to:
+    - podSelector:
+        matchLabels:
+          app: openshift-kube-apiserver
+      namespaceSelector:
+        matchLabels:
+          kubernetes.io/metadata.name: openshift-kube-system
```
sadlerap, line 19: On `kflux-prd-rh02` I see the pods with the `app: openshift-kube-apiserver` label in the `openshift-kube-apiserver` namespace, not `openshift-kube-system`.

filariow: great catch, that was a typo. fixed in 11777aca2bdb17c34904c0a45dd831d2b81ad62d

sadlerap (approved):
/lgtm

PR #5757, opened 2025-03-10
Fix view nodes on EaaS cluster

hugares:
On EaaS clusters, where Cluster as a service operator is installed, this permission is needed to list the nodes.


amisstea (approved):
/lgtm

PR #5758, opened 2025-03-10
namespace-lister: add healthz and readyz checks to staging

sadlerap:
Add healthz and readyz checks, so that we have hooks to inform the pod controller if we're alive and if we're ready to serve requests.

Production clusters will follow in a separate PR.


filariow (approved):
/lgtm

PR #5767, opened 2025-03-11
Optimize Vector Results

enarha:
To reduce the memory consumption and avoid OOM pod removal:
- Reduce the number of events by combining the log lines for multiple events into single event, avoiding storing the same event metadata for each event.
- Reduce the metadata we store for each event to only what's actually needed. Previously for each log line (~70 bytes), we were storing 3.5Kb and more of metadata.


hugares (approved):
/lgtm

khrm (commented):
/approve

PR #5775, opened 2025-03-11
Only enable pause pods in production

amisstea:
This is to experiment without them in staging.

They should be avoided in development due to typical resource constraints.


yftacherzog (approved):
/lgtm
/approve

yftacherzog:
@amisstea I put the hold label just so you can merge it at the time that suits you.

PR #5790, opened 2025-03-12
konfluxui: promote prod manifests to stg

filariow:


components/konflux-ui/staging/base/proxy/proxy.yaml
```diff
@@ -97,10 +84,10 @@ spec:
             cpu: 10m
             memory: 64Mi
       containers:
-      - image: registry.access.redhat.com/ubi9/nginx-124@sha256:b924363ff07ee0f8fd4f680497da774ac0721722a119665998ff5b2111098ad1
-        name: nginx
+      - image: openresty/openresty:1.25.3.1-0-jammy
```
sadlerap, line 87: Why are we swapping out for openresty?  I thought we were intentionally moving toward our own build of nginx.

filariow: great catch, that was not supposed to happen. wrong copy paste I guess

sadlerap (approved):
/lgtm

PR #5797, opened 2025-03-12
 fix: add ServerSideApply to sync release-service

theflockers:
this PR sets the ServerSideApply flag to argocd
syncOptions, to avoid errors syncing big chuncks
of text present in Release CRDs, that are added
to the object annotations.

### DIFF
--- a/argo-cd-apps/base/member/infra-deployments/release/release.yaml
+++ b/argo-cd-apps/base/member/infra-deployments/release/release.yaml
@@ -33,6 +33,7 @@ spec:
           selfHeal: true
         syncOptions:
         - CreateNamespace=true
+        - ServerSideApply=true
         retry:
           limit: 50
           backoff:


theflockers:
Can you take a look too @johnbieren ?

johnbieren:
> Can you take a look too @johnbieren ? This PR fixes the situation where the annotation is too big. Using server side apply, it allows the change to happen. It is tested, btw.

I have never even heard of this option, but you say it is tested and I see David already approved, so I certainly won't get in the way of this being merged

davidmogar:
> > Can you take a look too @johnbieren ? This PR fixes the situation where the annotation is too big. Using server side apply, it allows the change to happen. It is tested, btw.
> 
> I have never even heard of this option, but you say it is tested and I see David already approved, so I certainly won't get in the way of this being merged

Just so you have some more context https://argo-cd.readthedocs.io/en/latest/proposals/server-side-apply/. It's frequently mentioned as a solution to avoid the problem we are having right now with the metadata annotations becoming too big.

davidmogar (approved):

dirgim (approved):
/lgtm
/approve

PR #5804, opened 2025-03-13
konfluxui: add replace and force argocd annotation to job

filariow:
ArgoCD errors synching the Job because the PodSpec is an immutable field.
This annotation will force the deletion of the job before creating the new version


sadlerap (approved):
/lgtm

PR #5805, opened 2025-03-13
Run Vector pods only on nodes running pipelines prod

enarha:
This limits the Vector pods only to nodes used to run user pipelines.


hugares (approved):
/lgtm

PR #5816, opened 2025-03-14
Disable pause pods in production for ephemeral cluster provisioning

amisstea:
Testing in staging demonstrated that these pause pods didn't make much difference in overall performance and stability.


hugares (approved):
/lgtm

PR #5818, opened 2025-03-14
Add Role For Monitoring Release Pipelines

rhartman93:
We want to create a service account that can be used by users to automate monitoring the status of their managed release pipelines. The thought for now is it will only require:
* get
* list
* watch

permissions for:
* pipelineruns
* pipelines
* pods
* pods/logs

### DIFF
--- a/components/release/base/release-pipeline-monitor-role.yaml
+++ b/components/release/base/release-pipeline-monitor-role.yaml
@@ -8,5 +8,5 @@ rules:
     resources: ["pods", "pods/log"]
     verbs: ["get", "watch", "list"]
   - apiGroups: ["tekton.dev"]
-    resources: ["pods", "pods/log"]
+    resources: ["pipelineruns", "pipelines"]
     verbs: ["get", "watch", "list"]
--- a/components/release/base/release-pipeline-monitor-role.yaml
+++ b/components/release/base/release-pipeline-monitor-role.yaml
@@ -1,0 +1,18 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: monitor-release-pipeline-role
+rules:
+  - apiGroups:
+      - tekton.dev
+      - ""
+    resources:
+      - pipelineruns
+      - pipelines
+      - pods
+      - pods/logs
+    verbs:
+      - get
+      - list
+      - watch
--- a/components/release/base/release-pipeline-monitor-role.yaml
+++ b/components/release/base/release-pipeline-monitor-role.yaml
@@ -4,15 +4,9 @@ kind: ClusterRole
 metadata:
   name: monitor-release-pipeline-role
 rules:
-  - apiGroups:
-      - tekton.dev
-      - ""
-    resources:
-      - pipelineruns
-      - pipelines
-      - pods
-      - pods/logs
-    verbs:
-      - get
-      - list
-      - watch
+  - apiGroups: [""]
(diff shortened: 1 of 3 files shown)


components/release/base/release-pipeline-monitor-role.yaml
```diff
@@ -0,0 +1,18 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: monitor-release-pipeline-role
```
scoheb, line 5: this can be used for any type of pipeline though

components/release/base/release-pipeline-monitor-role.yaml
```diff
@@ -0,0 +1,18 @@
...
+rules:
+  - apiGroups:
+      - tekton.dev
```
scoheb, line 8: you may have to separate the tekton and the core ones?

components/release/base/release-pipeline-monitor-role.yaml
```diff
@@ -0,0 +1,18 @@
...
+      - ""
+    resources:
+      - pipelineruns
+      - pipelines
+      - pods
+      - pods/logs
+    verbs:
+      - get
+      - list
+      - watch
```
dchourasia, line 18:
```suggestion
- apiGroups: [""]
  resources: ["pods", "pods/log"]
  verbs: ["get", "watch", "list"]
- apiGroups: ["tekton.dev"]
  resources: ["pipelineruns", "pipelines"]
  verbs: ["get", "watch", "list"]
```

components/release/base/release-pipeline-monitor-role.yaml
rhartman93: Are you saying I should rename it to be more generic? or suggesting that there's a way I could make the role more tailored to release pipelines?

components/release/base/release-pipeline-monitor-role.yaml
```diff
@@ -0,0 +1,12 @@
...
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: monitor-release-pipeline-role
+rules:
+  - apiGroups: [""]
+    resources: ["pods", "pods/log"]
+    verbs: ["get", "watch", "list"]
+  - apiGroups: ["tekton.dev"]
+    resources: ["pods", "pods/log"]
```
dchourasia, line 11:
```suggestion
    resources: ["pipelineruns", "pipelines"]
```

components/release/base/release-pipeline-monitor-role.yaml
rhartman93: done

components/release/base/release-pipeline-monitor-role.yaml
rhartman93: Done

gbenhaim:
| We want to create a service account that can be used by users to automate monitoring the status of their managed
is it going to be a single service account that will be used by multiple users?

rhartman93:
hey folks, after my updates can someone re-check to see if this is ok?

rhartman93:
> | We want to create a service account that can be used by users to automate monitoring the status of their managed is it going to be a single service account that will be used by multiple users?
It could be, for now there is just one team requesting this, so i was going to make a dedicated account for their purposes, but we could create a "watcher" account to be used by multiple people? I'm not sure which is preferrable

scoheb (approved):

PR #5819, opened 2025-03-14
chore: remove konflux-test repository from tekton-ci namespace

14rcole:
The konflux-test repository was left over from the migration to konflux-ci.  The repository is no longer needed and is preventing the integration team from onboarding the konflux-test repo to konflux.

### DIFF
--- a/components/tekton-ci/base/repository.yaml
+++ b/components/tekton-ci/base/repository.yaml
@@ -16,13 +16,6 @@ spec:
 apiVersion: pipelinesascode.tekton.dev/v1alpha1
 kind: Repository
 metadata:
-  name: konflux-test
-spec:
-  url: "https://github.com/konflux-ci/konflux-test"
----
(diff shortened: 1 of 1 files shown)


components/tekton-ci/base/kustomization.yaml
```diff
@@ -1,7 +1,6 @@
 resources:
 - namespace.yaml
 - serviceaccount.yaml
-- repository.yaml
```
MartinBasti, line 4: Why did you removed all repositories and not only konflux-test ?

MartinBasti (changes requested):
I don;t think that all repos should be removed

components/tekton-ci/base/kustomization.yaml
14rcole: Because I was rushing and didn't pay attention to what else was in that file.  Let's try that again

MartinBasti (approved):

rcerven (approved):

PR #5820, opened 2025-03-14
Add VMs with larger disks in RH01

hugares:
In order to simplify the documentation, add the flavors of VMs with larger disk on external prod so we no longer need to make distinction between internal and external in user documentation.


mshaposhnik (approved):

mshaposhnik:
Missing tags for arm64/amd64, the rest is fine

hugares:
> Missing tags for arm64/amd64, the rest is fine
Done

PR #5832, opened 2025-03-17
Align MPC VMs between p02 and rh01

hugares:
We should have the same spec between environments for documentation simplicity but also for user experience, large in one environment should be the same on another one.


meyrevived (approved):
/lgtm

PR #5845, opened 2025-03-18
kyverno: more resources to background controller in prod

filariow:
Kyverno's background controller is getting OOMKilled. This change gives it more resources


components/kyverno/production/kflux-ocp-p01/kyverno-helm-values.yaml
```diff
@@ -24,8 +24,12 @@ admissionController:
 backgroundController:
   replicas: 3
   resources:
+    requests:
+      cpu: 200m
+      memory: 256Mi
```
hugares, line 29: Changes look good but we should aim at request=limits, otherwise pod will get evicted if pod uses more than 256 and the node is packed

PR #5848, opened 2025-03-18
KubeArchive: pull release file to remove the dependency

rh-hemartin:
This PR pulls the manifests being used by KubeArchive (a patched version due to some problems with memory consumption) so the process is independent from GitHub. Soon KubeArchive will be renamed due to restrictions of the CNCF, so when the KubeArchive organization gets renamed the KubeArchive deployment here will fail. To avoid that scenario we copy the manifest in this PR.

### DIFF
--- a/components/kubearchive/base/kubearchive.yaml
+++ b/components/kubearchive/base/kubearchive.yaml
@@ -1,0 +1,1093 @@
+apiVersion: v1
+kind: Namespace
+metadata:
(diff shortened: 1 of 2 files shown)


maruiz93:
After the renaming of kubeArchive, the idea is to go back to the link of the released resources? or to keep updating this file?

maruiz93 (approved):

PR #5855, opened 2025-03-18
nslister-ui: add group support in staging

filariow:
- **bump namespace-lister version**
- **configure the proxy to set the impersonate-group header**


sadlerap (approved):
/lgtm

PR #5859, opened 2025-03-18
update components/has/production/kustomization.yaml

ralphbean:


ralphbean:
I verified this in stage. I can create applications and components and the logs of the kube-rbac-proxy container look sane.

14rcole (approved):

PR #5861, opened 2025-03-18
kyverno: bump memory limits in stone-stg-rh01 to 1Gi

sadlerap:
Like what we were seeing in stone-prd-rh01, the equivalent staging cluster doesn't have enough memory, so increase it.


filariow (approved):
/lgtm

PR #5868, opened 2025-03-19
Run Mintmaker DependencyUpdateCheck cronjobs as non-root

querti:
Mintmaker SA has been granted the anyuid SCC becuase it needs to run its subscription-manager pod as root[1]. A side effect is that now it tries to run all of its pods as root by default. This is not secure and also conflicts with the setting runAsNonRoot: true, resulting in failing jobs. Explicitly run these cronjobs as a non-root user.

[1] https://github.com/konflux-ci/mintmaker/pull/222

### DIFF
--- a/components/mintmaker/base/cronjobs/create-dependency-update-check.yaml
+++ b/components/mintmaker/base/cronjobs/create-dependency-update-check.yaml
@@ -38,5 +38,6 @@ spec:
                   cpu: 100m
                   memory: 100Mi
               securityContext:
+                runAsUser: 1001120000
                 runAsNonRoot: true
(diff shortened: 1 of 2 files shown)


staticf0x (approved):

staticf0x:
What about the PLR pods that run renovate?

staticf0x:
> > What about the PLR pods that run renovate?
> 
> I have explicitly defined these pods to use the nonroot user in the related MintMaker PR: https://github.com/konflux-ci/mintmaker/pull/222/files#diff-a642dd1dc8889d9d6bc3cd5137d23a4309c170ad03cb91012dd9034286d26477R141

My bad thank you!

PR #5887, opened 2025-03-20
fix(STONEINTG-1137): fix kubearchiveconfig kustomization

kasemAlem:
Fix the integration /stage kustomization path and namespace patch


hongweiliu17:
we don't need to maintain https://github.com/redhat-appstudio/infra-deployments/tree/main/components/integration/staging/stone-stage-p01, right?

kasemAlem:
> we don't need to maintain https://github.com/redhat-appstudio/infra-deployments/tree/main/components/integration/staging/stone-stage-p01, right?

you are right,  for now the KubeArchive was made on the Public Stage (stage-rh01)

dirgim (approved):
/lgtm
/approve

PR #5889, opened 2025-03-20
Add RBAC data to Konflux info for all clusters

gbenhaim:
This data is used by the new UI to list the
available role users can use.


sahil143 (approved):

filariow (approved):
/lgtm

PR #5897, opened 2025-03-20
fix(STONEINTG-1137-kubearchive): apply archiving snapshots with kubeArch

kasemAlem:


components/kubearchive/base/kubearchive-config.yaml
```diff
@@ -6,3 +6,8 @@ metadata:
   namespace: product-kubearchive
 spec:
   resources:
+    - selector:
+        apiVersion: appstudio.redhat.com/v1alpha1 
+        kind: Snapshot
+      archiveOnDelete: true
```
rh-hemartin, line 12: The type is `string` because its a CEL expression, so it should be quoted.

```suggestion
      archiveOnDelete: 'true'
```

rh-hemartin: I can't resolve the comment, not sure if you can.

PR #5898, opened 2025-03-20
fix: set the integration staging memory requests equal to limits

dirgim:
* Set the integration service staging memory memory requests equal to limits
* This helps avoid a scenario where the pod runs on a node that is already experiencing high memory usage, which could lead to the pod being killed


hugares (approved):
/lgtm

PR #5900, opened 2025-03-20
fix(KONFLUX-7357): increase integration service memory to 8gb

dirgim:
* Since the increased load and added functionalities are causing frequent restarts on staging and production clusters, increase the memory requirements and limits to 8192Mi


hugares (approved):
/lgtm

jencull (approved):
lgtm

PR #5907, opened 2025-03-20
Limit memory usage Pipeline Service components

enarha:
- Limit memory usage for controllers running multiple replicas staging.
- Limit Vector pods memory usage development and staging.


divyansh42 (approved):
/lgtm

PR #5916, opened 2025-03-21
Remove power monitoring from production

hugares:
We got a request from ROSA SRE team to remove that from the master nodes to help with an ongoing issue on stg-rh01. Kepler must run on all the nodes so instead of trying to fix the issue, remove the whole component.

This power-monitoring was deployed even though could VMs are not well supported, and this is what we use. In addition, we do not want to run workloads that are not necessary which is the case for power monitoring, it is deployed but we do not absolutely need this, at least not for now.


raks-tt (approved):
/lgtm

scoheb (approved):

PR #5917, opened 2025-03-21
Update Konflux Openshift Pipelines nightly release

aThorp96:
For deploy on Monday

### DIFF
--- a/components/pipeline-service/production/base/main-pipeline-service-configuration.yaml
+++ b/components/pipeline-service/production/base/main-pipeline-service-configuration.yaml
@@ -1861,6 +1861,8 @@ spec:
     resources:
       - pipelinerun
     schedule: '*/10 * * * *'
+  result:
+    disabled: true
   targetNamespace: openshift-pipelines
   trigger:
     options:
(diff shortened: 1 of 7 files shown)


components/pipeline-service/production/stone-prod-p02/deploy.yaml
```diff
@@ -2453,6 +2453,8 @@ spec:
     resources:
     - pipelinerun
     schedule: '*/10 * * * *'
+  results:
+    disabled: true
```
zakisk, line 2457: @aThorp96 why results disabled?

enarha: Because we still use it from our downstream fork. Now with Results being enabled by default in the operator, we must explicitly disable it. That's until we switch to deploying through the operator.

components/pipeline-service/production/base/main-pipeline-service-configuration.yaml
```diff
@@ -1861,6 +1861,8 @@ spec:
     resources:
       - pipelinerun
     schedule: '*/10 * * * *'
+  results:
```
enarha, line 1864:
```suggestion
  result:
```
see here: https://github.com/redhat-appstudio/infra-deployments/blob/main/components/pipeline-service/development/main-pipeline-service-configuration.yaml#L2053

PR #5919, opened 2025-03-21
Temporarily disable sync of kubearchive application

hugares:
Disable Kubearchive sync on external staging cluster to allow to investigate issues where something is deleting secret from ServiceAccount causing the secret to be recreated in a loop, which filled up etcd.

The change will be reverted once troubleshooting is done.


scoheb (commented):
/approve 
/lgtm

scoheb (approved):

hugares:
/hold
We are not sure who is updating the SA, it it's the kubearchive controller then we do not need this PR

PR #5932, opened 2025-03-25
Use Vector mirrored image prod

enarha:
The Helm chart used to deploy Vector reference the image stored in the Docker registry. With the image being pulled for each node running users' pipelines and CI jobs also pulling that image, we hit the rate limit on the Docker repo, fail to deploy Vector in a timely manner and miss logs from the users' pipelines. Here we switch to using image mirrored to Quay.


divyansh42 (commented):
/lgtm

PR #5937, opened 2025-03-26
PVO11Y-4689 Update the Konflux Splunk logging endpoint

raks-tt:


raks-tt:
> I recommend:
> 
> 1. Doing it gradually, first for staging to make sure that nothing breaks with the new Splunk endpoint.
> 2. Verify with Splunk IT that the HEC tokens haven't changed.

Just confirmed with Eran in the email that since the ITML team has been using the endpoint for some time now, we can make this change directly to both the clusters

amisstea (approved):
/lgtm

PR #5975, opened 2025-03-28
Add more PPc VM flavors in prod-p02

hugares:
Standardize the VM flavors in PPC. Provide 2x8, 4x16, 8x32 and 16x64 VMs with both 100GB or 200GB.

CPU allocation on PowerPC does not work the same way as other architectures. Number of core is specified at the VM level, this maps to virtual core which maps to vCPU as each core/virtual core has 8 threads. Any fractional core gets rounded up to up to the nearest whole number. So 0.25 core gets rounded up to 1 virtual core/8vCPU but it will get 15 seconds per minute of CPU time making this the equivalent of a 2vCPU even of OS list 8 CPUs.


hugares:
/hold

I want to control when this change is deployed, please review and I will remove hold myself

hugares:
this is not covered by e2e, merging now

PR #5999, opened 2025-04-01
KubeArchive: store Releases

rh-hemartin:
We want to store Releases as well as Snapshots.

cc @kasemAlem

### DIFF
--- a/components/kubearchive/base/kubearchive-config.yaml
+++ b/components/kubearchive/base/kubearchive-config.yaml
@@ -11,6 +11,10 @@ spec:
         kind: Snapshot
       archiveOnDelete: 'true'
     - selector:
+        apiVersion: appstudio.redhat.com/v1alpha1
+        kind: Release
+      archiveOnDelete: 'true'
+    - selector:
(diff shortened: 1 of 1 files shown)


skoved (approved):
/lgtm

mafh314 (approved):

ggallen (commented):
It seems like everything is using `archiveOnDelete`.  

Why don't we use `archiveWhen` to force things to happen  multiple times or at least earlier.  For example, we could archive `PipelineRuns` when they complete.

maruiz93:
> It seems like everything is using `archiveOnDelete`.
> 
> Why don't we use `archiveWhen` to force things to happen multiple times or at least earlier. For example, we could archive `PipelineRuns` when they complete.

I agree with Greg's point, in fact the initial proposed configuration of the Release team was this:

```yaml
   - selector:
     apiVersion: appstudio.redhat.com/v1alpha1
     kind: Release
     archiveWhen: has(status.completionTime)
 ```
 
 Check it out [here](https://github.com/redhat-appstudio/infra-deployments/pull/5479/files)

PR #6012, opened 2025-04-02
KubeArchive: change when resources are archived

rh-hemartin:
We want to archive resources at different points in time. Also Releases were meant to be archived on completion, not deletion (see https://github.com/redhat-appstudio/infra-deployments/pull/5479/files).

### DIFF
--- a/components/kubearchive/base/kubearchive-config.yaml
+++ b/components/kubearchive/base/kubearchive-config.yaml
@@ -13,16 +13,20 @@ spec:
     - selector:
         apiVersion: appstudio.redhat.com/v1alpha1
         kind: Release
+      archiveWhen: has(status.completionTime)
       archiveOnDelete: 'true'
     - selector:
         apiVersion: tekton.dev/v1
         kind: PipelineRun
+      archiveWhen: has(status.completionTime)
       archiveOnDelete: 'true'
     - selector:
         apiVersion: tekton.dev/v1
         kind: TaskRun
+      archiveWhen: has(status.completionTime)
       archiveOnDelete: 'true'
     - selector:
         apiVersion: v1
         kind: Pod
+      archiveWhen: has(metadata.labels) && "tekton.dev/taskRunUID" in metadata.labels && status.phase in ['Succeeded', 'Failed', 'Unknown']
       archiveOnDelete: has(metadata.labels) && "tekton.dev/taskRunUID" in metadata.labels


rh-hemartin:
@ggallen @maruiz93

components/kubearchive/base/kubearchive-config.yaml
```diff
@@ -13,16 +13,16 @@ spec:
     - selector:
         apiVersion: appstudio.redhat.com/v1alpha1
         kind: Release
-      archiveOnDelete: 'true'
+      archiveWhen: has(status.completionTime)
     - selector:
```
ggallen, line 17: I think we should have both here.

components/kubearchive/base/kubearchive-config.yaml
```diff
@@ -13,16 +13,16 @@ spec:
...
         apiVersion: tekton.dev/v1
         kind: PipelineRun
-      archiveOnDelete: 'true'
+      archiveWhen: has(status.completionTime)
     - selector:
```
ggallen, line 21: I think we should have both here.

components/kubearchive/base/kubearchive-config.yaml
```diff
@@ -13,16 +13,16 @@ spec:
...
         apiVersion: tekton.dev/v1
         kind: TaskRun
-      archiveOnDelete: 'true'
+      archiveOnDelete: has(status.completionTime)
     - selector:
         apiVersion: v1
         kind: Pod
-      archiveOnDelete: has(metadata.labels) && "tekton.dev/taskRunUID" in metadata.labels
+      archiveOnDelete: has(metadata.labels) && "tekton.dev/taskRunUID" in metadata.labels && status.phase == 'Succeeded'
```
ggallen, line 28: I think we want something like this here:
```
      archiveWhen: has(metadata.labels) && "tekton.dev/taskRunUID" in metadata.labels && status.phase in ('Succeeded', 'Failed', 'Unknown')
      archiveOnDelete: has(metadata.labels) && "tekton.dev/taskRunUID" in metadata.labels
```
I'm not sure if the syntax is correct on the `archiveWhen`.

components/kubearchive/base/kubearchive-config.yaml
```diff
@@ -13,16 +13,16 @@ spec:
...
     - selector:
         apiVersion: tekton.dev/v1
         kind: PipelineRun
-      archiveOnDelete: 'true'
+      archiveWhen: has(status.completionTime)
     - selector:
         apiVersion: tekton.dev/v1
         kind: TaskRun
-      archiveOnDelete: 'true'
+      archiveOnDelete: has(status.completionTime)
```
ggallen, line 24: I think we should have both here.

components/kubearchive/base/kubearchive-config.yaml
rh-hemartin: Added.

components/kubearchive/base/kubearchive-config.yaml
rh-hemartin: Added

components/kubearchive/base/kubearchive-config.yaml
rh-hemartin: The syntax is wrong, I just needed to change `(` by `[`

components/kubearchive/base/kubearchive-config.yaml
rh-hemartin: Added

skoved (approved):
/lgtm

PR #6014, opened 2025-04-02
KubeArchive: monitoring setup

rh-hemartin:
This PR aims to enable metrics scraping for KubeArchive. It introduces an OpenTelemetry Collector because KubeArchive does not expose Prometheus endpoints. This Collector is configured to expose that endpoint so it can be scrapped.

cc @redhat-appstudio/konflux-o11y so they can check as this is related to their domain


components/kubearchive/base/monitoring-otel-collector.yaml
```diff
@@ -0,0 +1,102 @@
...
+      app: otel-collector
+  template:
+    metadata:
+      labels:
+        app: otel-collector
+    spec:
+      serviceAccount: kubearchive-otel-collector
+      containers:
+        - name: kube-rbac-proxy
+          image: quay.io/brancz/kube-rbac-proxy:v0.19.0
```
yftacherzog, line 56: As far as I understood it, the program decided to move away from using kube-rbac-proxy altogether (rather than find alternative sources for the image).
Recent kubebuilder releases also scaffold services in a way that will not require it. Is it an option here?

rh-hemartin: One of the alternatives listed in the official discussion is to use `quay.io/brancz/kube-rbac-proxy` I'm using (https://github.com/kubernetes-sigs/kubebuilder/discussions/3907). Also I'm following the README.md from the monitoring component on this very same repo, which uses `quay.io/brancz/kube-rbac-proxy` as well.

yftacherzog: I was not aware the monitoring component readme is referencing this image, but have a look [here](https://docs.google.com/document/d/1B_CBBXvjUXHqTUunB1D7cy7nd6fI0PzRxd3szgEAaSA/edit?tab=t.0) under requirements.

rh-hemartin: I checked on the developers channel and it should be fine, as we can't avoid using the image (because its better than just not using it).

mafh314 (approved):

PR #6018, opened 2025-04-02
fix(itn-2025-00072): Moving PPC Builds Back to Dallas IBM DC

manish-jangra:
We are having intermittant connectivity issues with PowerVS Workspace in Toronto region while connecting to Red Hat Subscription Manager Service. This Pull Request will add a test config to trigger a build with new workspace settings


hugares (approved):
/lgtm

PR #6022, opened 2025-04-02
smee-client - livenessProbe creation

rubenmgx:
Based on https://issues.redhat.com/browse/SPRE-1255, as well as for smee server, we need to set up a livenessProbe for the client.


amisstea (approved):
/ok-to-test
/lgtm

PR #6025, opened 2025-04-02
Add missing env patch to konflux-rbac

gbenhaim:
The patch which sets the env for konflux-rbac
was missing in production-downstream.


filariow (approved):
/lgtm

PR #6027, opened 2025-04-02
Move prd-rh01 PowerPC back to Dallas

hugares:
We have an ongoing issue with Toronto where PPC VMs cannot reach subscription manager. Dallas does not have the issue to move to that region until IBM fix the issue


hugares:
/hold

Waiting for user confirmation that new WS is OK before merging

enkeefe00 (approved):

PR #6041, opened 2025-04-03
Update production RDS vault secret paths

enkeefe00:
AppSRE has changed the integration that manages the Konflux RDS instances from terraform-resources to external-resources. This, in turn, requires a change to the vault secret paths for RDS instances in order to ensure the instances have the most up-to-date passwords.


hugares (approved):
/lgtm

PR #6042, opened 2025-04-03
fix-up(PVO11Y-4652): Export missing etcd metrics

ci-operator:
Additional etcd metrics exported: mvcc db size and leader changes seen


raks-tt (approved):
LGTM

PR #6048, opened 2025-04-04
removing livenessprobe

rubenmgx:
pod on crashloop state, removing livenessprobe


hugares:
affecting prod, merging now

PR #6058, opened 2025-04-07
PVO11Y-4717 Remove observability operator memory limit

kubasikus:
Increase the limit to 3Gi. This should allow observability-operator pod to survive the initial memory spike before the usage stabilises.

Sideffect of this is that this applies limits to ALL Operator pods, including prometheus-operator, perses-operator and webhook-admission.
This should not impact scheduling as limits are not considered during scheduling.

Existing proposal for configuring operator resource usage: https://github.com/operator-framework/operator-controller/issues/1507


kubasikus:
Changing this to actually remove the memory limit entirely. Once this is merged, I will be able to test and find the right threshold based on empirical data.

TominoFTW (approved):
/lgtm
/approve

raks-tt (approved):

PR #6063, opened 2025-04-07
Draft - Update Namespaces with labels using policies

raks-tt:


components/cost-management/policies/kustomization.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- update_namespace_labels.yaml
+- kyverno_rbac.yaml
```
mike-kingsbury, line 5: Any chance of a race condition here where Kyverno will attempt to start applying the policy defined in `uodate_namespace_labels.yaml` prior to the RBAC being applied?  I know it appears to be a tiny window.

components/cost-management/policies/.chainsaw-test/chainsaw-test.yaml
```diff
@@ -0,0 +1,136 @@
+apiVersion: chainsaw.kyverno.io/v1alpha1
+kind: Test
+metadata:
+  name: label-propagation-valid-cost-center
+spec:
+  concurrent: true
```
filariow, line 6: as the the ClusterPolicy is installed as part of each test, we can not run them concurrently. This change applies also the other tests.

```suggestion
  concurrent: false
```

components/cost-management/policies/.chainsaw-test/chainsaw-test.yaml
```diff
@@ -0,0 +1,136 @@
...
+        - create:
+            file: ./resources/namespace-no-cost-center.yaml
+            template: true
+            bindings:
+              - name: namespace
+                value: demo-tenant-no-cost-center
+    # - name: Apply permission on namespaces
+    #   try:
+    #     - apply:
+    #         file: ../kyverno_rbac.yaml
```
filariow, line 112: this shouldn't be disabled

```suggestion
     - name: Apply permission on namespaces
       try:
         - apply:
             file: ../kyverno_rbac.yaml
```

components/cost-management/policies/.chainsaw-test/chainsaw-test.yaml
```diff
@@ -0,0 +1,136 @@
...
+  description: |
+    tests that the label is not applied to pods in a namespace
+    that does not match the label selector
+    in a tenant namespace
+  steps:
+    - name: Create namespaces for testing
+      try:
+        - create:
+            file: ./resources/namespace-nonmatching.yaml
+    - name: Apply kyverno Cluster Policy and assert it exists
```
filariow, line 65:
```suggestion
    - name: Apply RBAC
      try:
        - apply:
            file: ../kyverno_rbac.yaml
    - name: Apply kyverno Cluster Policy and assert it exists
```

components/cost-management/policies/kustomization.yaml
filariow: IIRC, kustomize should sort them such that (Cluster)RoleBindings are applied before the ClusterPolicies. Also, if permissions are missing a Kyverno's webhook won't allow ArgoCD to create the Policy. In a few iterations everything should work.

components/cost-management/policies/.chainsaw-test/chainsaw-test.yaml
filariow, line 1: we are missing the test case in which the label has an empty value

```yaml
labels:
    cost-center: ""
```

components/cost-management/policies/update_namespace_labels.yaml
```diff
@@ -0,0 +1,72 @@
...
+            value: ""
+      mutate:
+        targets:
+          - apiVersion: v1
+            kind: Pod
+        patchStrategicMerge:
+          metadata:
+            labels:
+              cost-center: "760"
+              insights_cost_management_optimizations: "true"
```
filariow, line 46: I think we should we add the Konflux domain to the label. WDYT?
```suggestion
              konflux-ci.dev/cost-center: "760"
              konflux-ci.dev/insights-cost-management-optimizations: "true"
```

filariow, line 46: I'd suggest using hyphens instead of underscores here:

```suggestion
              cost-center: "760"
              insights-cost-management-optimizations: "true"
```

filariow (changes requested):

components/cost-management/policies/update_namespace_labels.yaml
raks-tt: I think this is going to be a problem for Kyverno as it only accepts the following pattern ->^(?:\w+|\(.+\))$
(it doesn't accept -)

mftb (approved):
/lgtm
/approve

components/cost-management/policies/.chainsaw-test/chainsaw-test.yaml
raks-tt: We discussed that we won't be needing this use-case. I think we can skip this use-case as the namespace.sh script will make sure that it is never empty

components/cost-management/policies/update_namespace_labels.yaml
raks-tt: For the cost management operator, the label insights-cost-management-optimizations is required in its current form, adding a domain prefix would prevent it from being recognized correctly. Additionally, in some cases, I believe the same cost-center label is applied from AWS. Including a domain would result in an extra label, introducing complexity in mapping.
Given these factors, I prefer to opt out of a domain-prefixed label for either of these cases would not be practical or beneficial.

components/cost-management/policies/.chainsaw-test/chainsaw-test.yaml
raks-tt: I have an interesting observation, changing this to false results in a failure for the third use case, `label-no-cost-center` . Do you know why that could be the case?
It ends up having an empty label: {}

components/cost-management/policies/update_namespace_labels.yaml
filariow: thanks for the explanation

components/cost-management/policies/.chainsaw-test/chainsaw-test.yaml
filariow: try applying this change to all the `concurrent: true` occurrences in the file. Probably a task from a concurrent test is changing something shared, like deleting the ClusterPolicy.

raks-tt: You are right, but I never figured out what was causing. But tested locally after making few changes and it works

components/cost-management/policies/.chainsaw-test/chainsaw-test.yaml
raks-tt: Add this test case and works locally for me now

components/cost-management/policies/kyverno_rbac.yaml
```diff
@@ -0,0 +1,21 @@
...
+metadata:
+  name: kyverno-background-controller-pod-update
+  labels:
+    rbac.kyverno.io/aggregate-to-background-controller: "true"
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - pods
+  - namespaces
```
filariow, line 13: we don't need write operations on `namespaces`

components/cost-management/policies/kyverno_rbac.yaml
```diff
@@ -0,0 +1,21 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: kyverno-background-controller-pod-update
```
filariow, line 5: we may need to change the name as it is also providing permission on namespaces

components/cost-management/policies/kyverno_rbac.yaml
raks-tt: No i need to add the insights* label to namespaces (it is a requirement for ros)

mftb (approved):
/lgtm
/approve
/ok-to-test

components/cost-management/policies/kyverno_rbac.yaml
raks-tt: I added a small test case here to check if namespace gets updated. https://github.com/redhat-appstudio/infra-deployments/pull/6063/files#diff-de4b0e2c1cae9694ab647f0081681c4e17fce944b26607087ae136af1407f73eR174

PR #6064, opened 2025-04-07
PVO11Y-4717 Increase the limit to 4Gi

kubasikus:
This should allow observability-operator pod to survive the initial memory spike before the usage stabilises. The largest memory spike was noted on the stone-prod-p02 cluster as 1.5Gi.

Sideffect of this is that this applies limits to ALL Operator pods, including prometheus-operator, perses-operator and webhook-admission.
This should not impact scheduling as limits are not considered during scheduling.

Existing proposal for configuring operator resource usage: https://github.com/operator-framework/operator-controller/issues/1507


TominoFTW (approved):
/lgtm
/approve

raks-tt (approved):

PR #6062, opened 2025-04-07
add policy to create appstudio-pipeline rolebinding in staging

filariow:
this policy ensures that a rolebinding for the appstudio-pipeline-runner exists in every tenant namespace

Requires:
* [x] https://github.com/redhat-appstudio/infra-deployments/pull/6067

Signed-off-by: Francesco Ilario <filario@redhat.com>

PR #6067, opened 2025-04-07
actions: run chainsaw tests concurrently

sadlerap:
Chainsaw tests may leave artifacts on the cluster after executing tests, which might cause issues with later tests.  To address this, run each test suite independently and concurrently.

Fixes issues uncovered in #6062.


hugares (approved):
/lgtm

PR #6075, opened 2025-04-08
Fix KubeArchive metrics collection

rh-hemartin:
This PR adds `honorLabels: true` which will make [Prometheus to respect labels when they collide](https://docs.redhat.com/en/documentation/openshift_container_platform/4.18/html/monitoring_apis/servicemonitor-monitoring-coreos-com-v1#spec-endpoints-2) and adds a couple more options on the OpenTelemetry side: don't add units to the metrics names and receive traces, even if they are just printed to debug.

### DIFF
--- a/components/kubearchive/base/otel-collector-config.yaml
+++ b/components/kubearchive/base/otel-collector-config.yaml
@@ -1,5 +1,3 @@
-# Copyright KubeArchive Authors
-# SPDX-License-Identifier: Apache-2.0
 ---
 receivers:
   otlp:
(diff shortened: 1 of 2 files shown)


components/kubearchive/base/otel-collector-config.yaml
```diff
@@ -13,10 +11,17 @@ processors:
 exporters:
   prometheus:
     endpoint: '0.0.0.0:9090'
+    send_timestamps: true
+    add_metric_suffixes: false
+  debug:
```
mafh314, line 16: is this correct @rh-hemartin ? I mean to have the `debug`section defined and nothing under?

rh-hemartin: Yes, its defining the `debug` exporter where the traces are sent to.

PR #6076, opened 2025-04-08
move create pipeline-runner sa policy to konflux-rbac

filariow:
konflux-ci application doesn't seem to be present in relevant clusters.
Moving the ClusterPolicy to konflux-rbac.

In future we could want to create a dedicated component.


hugares (approved):
/lgtm

PR #6085, opened 2025-04-09
MintMaker: Deploy Redis for Renovate cache

querti:
Redis will be used by renovate PLRs to store and retrieve cache in order to improve performance and decrease the number of API calls. Connection is not password protected, but a network policy exists to restrict access only to pods in the "mintmaker" namespace. I think that a password is unnecessary since it would complicate the deployment and no sensitive or important data will be stored in the database. The persistent volume exists to back up the data so that Redis can reload it after restart.

The memory limit was chosen somewhat arbitrarily, but we can increase it if high use is observed over time. OOM errors should not happen since Redis limit is configured lower than Openshift limit.

### DIFF
--- a/components/mintmaker/base/kustomization.yaml
+++ b/components/mintmaker/base/kustomization.yaml
@@ -1,6 +1,7 @@
 resources:
 - cronjobs/
 - rbac/
+- redis-cache/
 
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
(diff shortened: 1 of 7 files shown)


staticf0x:
This looks great! My only question is whether you tested it on some repos (especially Go based) and if you have any numbers/observations if this cache reduces the amount of HTTP requests and how much.

staticf0x:
Great, thanks for the comment. There might be many factors, Renovate uses GraphQL for some GitHub API operations, but REST for some other (I think), and AFAIK we also don't know how the rate limit is reset on GitHub's side, but if there are hits on the cache, that's a good sign imo.

staticf0x (approved):

PR #6095, opened 2025-04-10
bump kyverno resources in production

filariow:


mshaposhnik (approved):

mshaposhnik:
/lgtm
We're also using practic of removing the CPU  upper cap, but it's up2you.

PR #6098, opened 2025-04-10
fix chainsaw tests for bootstrap tenant ns policy

filariow:
[`error`](https://kyverno.github.io/chainsaw/0.2.3/operations/error/) is NOT checking that the resource doesn't exist. Let's [`delete`](https://kyverno.github.io/chainsaw/0.2.3/operations/delete/) the resource and check that the operation failed.


sadlerap (approved):
/lgtm

PR #6099, opened 2025-04-10
set generateExisting to false in bootstrap-tenant-namespace policy

filariow:
We don't need to set `generateExisting` to `true` in the `bootstrap-tenant-namespace` ClusterPolicy.

As the [doc](https://kyverno.io/docs/writing-policies/generate/#generate-for-existing-resources) says:
> Note that the benefits of using a “generate existing” rule is only the moment the policy is installed. Once the initial generation effects have been produced, the rule functions like a “standard” generate rule from that point forward. Generate existing rules are therefore primarily useful for one-time use cases when retroactive policy should be applied.

Requires:
* [x] #6098


sadlerap (approved):
/lgtm

sadlerap (approved):
/lgtm

PR #6101, opened 2025-04-10
Adding liveness Probe for sme server

rubenmgx:
Adding livenessProbe for smee server, based on jira request https://issues.redhat.com/browse/SPRE-1255 
Solution has been tested locally and it works correctly.


amisstea (commented):
/ok-to-test

PR #6107, opened 2025-04-10
rename tenant namespace boostrap policy

sadlerap:
Some changes recently made to this policy are causing issues with kyverno, since the fields that were changed are immutable.  Renaming the policy should provide us with two benefits:
- Works around this issue by making a new resource entirely
- Make the policy's purpose clearer from its name.


filariow (approved):
Great! 🚀

/lgtm

PR #6127, opened 2025-04-14
[dev+stage] Add SCC for nested containerization

chmeliik:
Add the appstudio-pipelines-nested-containerization-scc. It is a copy of the regular appstudio-pipelines-scc, the only difference being the seLinuxContext.seLinuxOptions setting that allows (forces, in fact) the usage of the container_runtime_t label.

The container_runtime_t label is needed to enable buildah-in-kubernetes workflows with a more robust isolation mechanism (rather than chroot). See https://github.com/containers/buildah/issues/5818 for details.

Allow the appstudio-pipelines-runner ClusterRole - and by extension, the appstudio-pipeline ServiceAccount - to use the new SCC.

All in all, this should allow the appstudio-pipeline SA to set the container_runtime_t label on Pods (but not force, because the SA can still use the regular appstudio-pipelines-scc), making it possible to use more robust mechanisms for buildah isolation.

### DIFF
--- a/components/pipeline-service/development/main-pipeline-service-configuration.yaml
+++ b/components/pipeline-service/development/main-pipeline-service-configuration.yaml
@@ -2187,3 +2187,45 @@ volumes:
 - persistentVolumeClaim
 - projected
 - secret
+---
+allowHostDirVolumePlugin: false
+allowHostIPC: false
+allowHostNetwork: false
+allowHostPID: false
+allowHostPorts: false
+allowPrivilegeEscalation: false
+allowPrivilegedContainer: false
+allowedCapabilities:
+- SETFCAP
+apiVersion: security.openshift.io/v1
+defaultAddCapabilities: null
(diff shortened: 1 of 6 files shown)


chmeliik:
It is not yet clear if this will achieve what I need. Unfortunately, I wasn't able to test this change in a dev cluster, because in dev clusters the `container_runtime_t` is just allowed without any change to the SCCs.

Proposing this change for staging clusters first so that I can test the behavior.

enarha (commented):
/lgtm

chmeliik:
@Roming22 @adambkaplan PTAL as the approvers

enarha:
> @Roming22 @adambkaplan PTAL as the approvers

I can approve this change. Since I'm the only reviewer so far, do you prefer someone to review as well, before I set the approved label?

filariow (approved):
/lgtm

PR #6113, opened 2025-04-11
make chainsaw tests more robust

filariow:
This PR
* changes the order of steps for the chainsaw tests pipeline to make it faster
* changes default namespace of chainsaw tests to `konflux-kyverno` to match the other environments
* enforce Helm manifests are used also in the `chainsaw` overlay

PR #6130, opened 2025-04-14
improve tests for generate-tenant-resources-admission ClusterPolicy

filariow:
requires
* [x] #6113

This PR adds some content to the ClusterRole `appstudio-pipeline-runner` in chainsaw tests to make it more adherent to the real use case.

This is to make sure the policy's manifests are providing kyverno with the permission required to create the RoleBinding, i.e. the same permission the ClusterRole in the RoleBinding will provide to the assignee.

For this purpose permissions on `horizontalpodautoscaler` are a good fit as they aren't something shared globally and -if missed- an error would be raised.


sadlerap (approved):
/lgtm

hugares:
Closing stale/outdated PR, please re-open if it is still needed

PR #6145, opened 2025-04-15
Reduce priority for nested containerization SCC

chmeliik:
Previously, appstudio-pipelines-scc and
appstudio-pipelines-nested-containerization-scc had the same priority. The nested-containerization SCC is considered more restrictive (it forces a specific label while the normal SCC does not). Because of that, OpenShift prioritized it over the normal SCC and all Tekton Task Pods were defaulting to the container_runtime_t label.

Reduce the appstudio-pipelines-nested-containerization-scc priority to make sure Pods default to appstudio-pipelines-scc.

Ref: https://docs.redhat.com/en/documentation/openshift_container_platform/4.18/html/authentication_and_authorization/managing-pod-security-policies#scc-prioritization_configuring-internal-oauth

### DIFF
--- a/components/pipeline-service/development/main-pipeline-service-configuration.yaml
+++ b/components/pipeline-service/development/main-pipeline-service-configuration.yaml
@@ -2209,7 +2209,7 @@ metadata:
     argocd.argoproj.io/sync-options: SkipDryRunOnMissingResource=true
     argocd.argoproj.io/sync-wave: "0"
   name: appstudio-pipelines-nested-containerization-scc
-priority: 10
+priority: 0
 readOnlyRootFilesystem: false
 requiredDropCapabilities:
 - MKNOD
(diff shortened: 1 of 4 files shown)


enarha:
@chmeliik Do I understand this correctly and this is just an intermediate step until we test this and then switch the priorities the other way around or is this a long term configuration?

chmeliik:
> @chmeliik Do I understand this correctly and this is just an intermediate step until we test this and then switch the priorities the other way around or is this a long term configuration?

This would be long term. The only Task in our Pipeline that will currently need the `container_runtime_t` label is the buildah task. It calls buildah inside an OpenShift Pod => needs elevated SELinux permissions to enable better containerization methods (currently it just uses chroot). For all other tasks, giving them the more powerful label is not necessary and we should avoid unnecessarily loosening the security.

This change aims to allow Tasks (buildah, specifically) to specify the `container_runtime_t` label, but still keep defaulting to the more restricted label for all Tasks that don't specify one explicitly

enarha (commented):
/lgtm

PR #6147, opened 2025-04-15
feat(RELEASE-1532): use trusted artifacts

scoheb:
- enable emptyDirs for rh-advisory and rh-push-to-registry-redhat-io


johnbieren:
Looks good to me, but IIUC we don't want to merge this until https://github.com/konflux-ci/release-service-catalog/pull/898 is approved

PR #6160, opened 2025-04-17
Change s390x cobfiguration to static for ocp-art-p01

mshaposhnik:
For the upcoming release, we change the provisioning way for s390x from dynamic to static.
All machines are `bz2-16x64 ` with a 1TB volume mount as `/home`


manish-jangra (approved):
/approve
/lgtm

PR #6169, opened 2025-04-17
allow authenticated metrics for namespace-lister

sadlerap:
We were allowing anyone who could access our metrics port to scrape our metrics, even though our servicemonitor was configured to use a bearer token to authenticate.  The latest build of namespace-lister addresses this issue.

PR #6170, opened 2025-04-17
prod: allow authenticated metrics for namespace-lister

sadlerap:
Rolls out #6169 to production clusters.


sadlerap:
I've checked both `stone-stg-rh01` and `stone-stage-p01`, and namespace-lister is serving metrics correctly on both of these clusters.

filariow (approved):
/lgtm

PR #6181, opened 2025-04-21
Configure OCP cluster to use static PPC64le VMs

enkeefe00:
Multi-architectural builds in the OCP cluster are timing out due to waiting for PPC VMs to be provisioned. To decrease the time needed to complete the build, a set of static VMs is provisioned so that the only wait time is for if a VM is at its current concurrency limit.

### DIFF
--- a/components/multi-platform-controller/production-downstream/kflux-ocp-p01/host-config.yaml
+++ b/components/multi-platform-controller/production-downstream/kflux-ocp-p01/host-config.yaml
@@ -825,104 +825,153 @@ data:
 #  dynamic.linux-d250-largecpu-s390x.instance-tag: prod-s390x-d250-largecpu
 #  dynamic.linux-d250-largecpu-s390x.disk: "250"
 
-  #PPC64LE dynamic nodes
-  dynamic.linux-ppc64le.type: ibmp
-  dynamic.linux-ppc64le.ssh-secret: "internal-prod-ibm-ssh-key"
-  dynamic.linux-ppc64le.secret: "internal-prod-ibm-api-key"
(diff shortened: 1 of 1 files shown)


components/multi-platform-controller/production-downstream/kflux-ocp-p01/host-config.yaml
```diff
@@ -825,104 +825,153 @@ data:
...
+  host.ppc64le-static-7.address: "10.130.84.165"
+  host.ppc64le-static-7.platform: "linux/ppc64le"
+  host.ppc64le-static-7.user: "root"
+  host.ppc64le-static-7.secret: "internal-prod-key"
+  host.ppc64le-static-7.concurrency: "8"
+
+  host.ppc64le-static-8.address: "10.130.84.210"
+  host.ppc64le-static-8.platform: "linux/ppc64le"
+  host.ppc64le-static-8.user: "root"
+  host.ppc64le-static-8.secret: "internal-prod-key"
```
mshaposhnik, line 874: I cannot see the secret, is there some plan for it?

enkeefe00: I was able to just use the regular IBM internal production key. I may have named it wrong here.

enkeefe00: Just fixed the name

mshaposhnik:
Great, could you confirm you've done al necessary operations with the disk mount on avary machine, so they're have the 2TB volumes actually accessible and used?

mshaposhnik (approved):

hugares (approved):
/lgtm

PR #6188, opened 2025-04-22
Set up RBAC to kubearchive group

maruiz93:
Allow access to kubearchive API through proxy

This configuration allows to access the KubeArchive API the following way:

```bash
curl -k -H "Authorization: Bearer $(oc whoami --show-token)" "https://api.stone-stg-rh01.l2vh.p1.openshiftapps.com:6443/api/v1/namespaces/kubearchive/services/https:kubearchive-api-server:server/proxy/livez"
```

This is needed for the cli, and any other KubeArchive API client (like Konflux UI)

We will need to extend this to every authenticated user within the cluster through a ClusterRoleBinding to allow Konflux users to use this API.
I will create a ticket for this.

### DIFF
--- a/components/kubearchive/base/rbac.yaml
+++ b/components/kubearchive/base/rbac.yaml
@@ -1,4 +1,32 @@
 ---
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: kubearchive-api-proxy
+rules:
+  - apiGroups:
+      - ""
+    resources:
+      - services/proxy
+    verbs:
+      - get
+    resourceNames:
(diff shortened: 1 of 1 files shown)


gbenhaim (changes requested):
thanks @Marta Añón. The correct way will be to expose the kubearchive service through the nginx proxy (like we do for tekton results), for example https://github.com/redhat-appstudio/infra-deployments/blob/main/components/konflux-ui/staging/base/proxy/nginx.conf#L90

maruiz93:
Closed in favor of https://github.com/redhat-appstudio/infra-deployments/pull/6206

PR #6190, opened 2025-04-22
Remove ppc64le from dynamic platforms for kflxu-ocp-p01

hugares:


hugares:
this is affecting prod and not tested by e2e, merging now

PR #6203, opened 2025-04-23
Mintmaker stage release - metrics

nvtnlucie:


qixiang:
Addressed by https://github.com/redhat-appstudio/infra-deployments/pull/6198

PR #6206, opened 2025-04-23
Expose KubeArchive API 

maruiz93:
Configure nginx proxy in Konflux UI so it can access it from staging.

The KubeArchive API is exposed with an Openshift Route, the same way it's done for Tekton Results.

I've replicated the NGINX configuration for Tekton Results in Konflux UI, but I haven't tested it. Please, let me know if it needs any changes.

### DIFF
--- a/components/konflux-ui/staging/stone-stg-rh01/kustomization.yaml
+++ b/components/konflux-ui/staging/stone-stg-rh01/kustomization.yaml
@@ -13,6 +13,7 @@ configMapGenerator:
   literals:
     - IMPERSONATE=true
     - TEKTON_RESULTS_URL=https://tekton-results-api-service.tekton-results.svc.cluster.local:8080
+    - KUBEARCHIVE_URL=https://kubearchive-api-server.product-kubearchive.svc.cluster.local:8081
 
 patches:
 - path: add-service-certs-patch.yaml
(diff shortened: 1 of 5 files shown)


components/konflux-ui/staging/stone-stg-rh01/kustomization.yaml
```diff
@@ -13,6 +13,7 @@ configMapGenerator:
   literals:
     - IMPERSONATE=true
     - TEKTON_RESULTS_URL=https://tekton-results-api-service.tekton-results.svc.cluster.local:8080
+    - KUBEARCHIVE_URL=https://kubearchive-api-server.product-kubearchive.svc.cluster.local:8080
```
gbenhaim, line 16: can you please update the internal staging cluster as well? since the changes you made to the nginx config will take place in both staging clusters.

gbenhaim, line 16: I see on the cluster that the service port is 8081

gbenhaim (changes requested):

components/konflux-ui/staging/stone-stg-rh01/kustomization.yaml
maruiz93: included

rh-hemartin (approved):

gbenhaim (approved):

PR #6216, opened 2025-04-24
fix: remove failing checks

theflockers:
this PR remove the failing checks github and pyxis. github is fails due to throttling (needs the secret) and pyxis, the release account don't have access to the new secrets, so the check from the release
namespace needs to be ajusted.

### DIFF
--- a/components/release/base/monitor/production/configmap.yaml
+++ b/components/release/base/monitor/production/configmap.yaml
@@ -7,20 +7,11 @@ data:
       pool_interval: 120
       metrics_prefix: release_service
     checks:
-      git:
-        - name: github
-          url: https://github.com/konflux-ci/release-service-catalog
-          revision: production
-          path: pipelines/fbc-release/fbc-release.yaml
       quay:
         - name: quayio
           tags:
             - latest
           pullspec: quay.io/konflux-ci/release-service-utils
-      http:
-        - name: pyxis
-          url: https://pyxis.api.redhat.com/v1/ping
-          insecure: true
 kind: ConfigMap
 metadata:
   name: release-service-monitor-config


components/release/base/monitor/production/configmap.yaml
```diff
@@ -7,20 +7,11 @@ data:
...
-          revision: production
-          path: pipelines/fbc-release/fbc-release.yaml
       quay:
         - name: quayio
           tags:
             - latest
           pullspec: quay.io/konflux-ci/release-service-utils
-      http:
-        - name: pyxis
-          url: https://pyxis.api.redhat.com/v1/ping
```
mmalina, line 22: Does this really require credentials? Isn't the issue rather the fact that some clusters are on the private network and this url won't work there?

johnbieren: that would make sense to me

johnbieren:
Looks good to me, but I am not putting the proper command for it because idk if the bot will merge with outstanding threads. If it is resolved and you need me to add the approval, just ping me

components/release/base/monitor/production/configmap.yaml
theflockers: Yes, that the issue, so some cluster requires auth and other don't. My latest configuration worked (with the credentials) in some places and others don´t - the ones that require internal URL. But I did not have the proper time to test the infra-deployments configuration and make it work per cluster.

I have done it in a old PR (https://github.com/redhat-appstudio/infra-deployments/pull/5174/files), but I could not make it work because deploying Konflux in quickcluster was not working. Later I could make konflux work in quickcluster (with the minimal for the release service) but then I could not test the changes due to other priorities.

So, I think for now the better is disabling it until I can work on this.

mmalina (approved):

components/release/base/monitor/production/configmap.yaml
theflockers: The base in the PR I pasted, it basically having one different configuration per cluster. I tried to add the conf only for stage and see if it would work when I would add others. It did not as the hierarchy of it is a bit (at least) confusing.

PR #6221, opened 2025-04-24
Increase memory limit for crossplane controller in prod and staging

amisstea:
OOMKilled errors are occurring with the default 1Gi limit.


yftacherzog (approved):
/lgtm
/approve

PR #6226, opened 2025-04-24
Application-layer new-cluster automation

ralphbean:
~One thing that's up in the air here is dealing with the valero backup key. I assume that is waiting on some new work to be done in the cluster-level automation and can be followed up on later.~

This is ready for review.

### DIFF
--- a/hack/new-cluster/playbook.yaml
+++ b/hack/new-cluster/playbook.yaml
@@ -98,7 +98,7 @@
 
   pre_tasks:
     - name: Validate and initialize variables
-      tags: [github]
+      tags: [always]
       include_tasks: tasks/variables.yaml
 
   tasks:
--- a/hack/new-cluster/playbook.yaml
+++ b/hack/new-cluster/playbook.yaml
@@ -1,0 +1,136 @@
+---
+# This ansible playbook will prepare the application-layer for a new cluster.
+#
+# Major steps:
+# * Provision new github application and insert credentials in vault
+# * Provision new chains signing secret and insert credentials in vault
+# * Locally write out configuration overlays for a new cluster based on templates.
+#
+
+- name: Create and patch YAML files
+  hosts: localhost
+  gather_facts: no
+
+  vars_prompt:
+    - name: cutename
+      prompt: "Cute name of the cluster (like rh09)"
+      private: false
+    - name: shortname
+      prompt: "Short name of the cluster (like kflux-prd-rh09)"
+      private: false
+    - name: longname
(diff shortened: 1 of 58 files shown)


manish-jangra:
@ralphbean If you're referring to kms key for velero backup, we will not use that anymore for new environments as we're creating s3 bucket (for velero backup) using the default s3 encryption managed by aws itself. We can always change that if we need to and change terraform module to create kms key for the velero backup.

hack/new-cluster/playbook.yaml
```diff
@@ -0,0 +1,136 @@
...
+      private: false
+    - name: env
+      prompt: "What environment (one of {{allowed_envs}})"
+      private: false
+    - name: network
+      prompt: "What network (one of {{allowed_networks}})"
+      private: false
+
+  vars:
+    vault_addr: "https://vault.devshift.net"
```
manish-jangra, line 32: This is an internal vault instance only accessible from Red Hat internal Network. We might want to specify this in REDME that we should be connected to the vpn before running this ansible playbook.

ralphbean:
@manish-jangra ah, thanks for the tip about the valero key - that _is_ what I was referring to. I'll drop those comments in the yaml and update the description here.

hack/new-cluster/templates/pipeline-service/resources/update-tekton-config-pac.yaml
```diff
@@ -0,0 +1,12 @@
+---
+- op: add
+  path: /spec/platforms/openshift/pipelinesAsCode/settings
+  value:
+    application-name: Konflux {{ shortname }}
+    custom-console-name: Konflux {{ shortname }}
+    custom-console-url: https://konflux-ui.apps.{{ longname }}.p1.openshiftapps.com
```
manish-jangra, line 7: @ralphbean I have seen `p1` in `p1.openshiftapps.com` changing to `p2` and `p3`. However, I have verified all of our konflux clusters, they all have `p1.openshiftapps.com`.

ralphbean: Ack - let's change this then so that ".p1" is always expected to be a part of `longname`.

ralphbean: Revised in 1d74b4711db446541be6465d0be8a945f2540c3a.

manish-jangra (approved):
/approve
/lgtm

PR #6242, opened 2025-04-28
Disable kyverno WebhookCleanup job in dev and stg

filariow:
This will prevent the scale-to-zero job from being created.
This is not required in a GitOps environment, as the helm's `pre-delete` hook is not supported by ArgoCD and the effect of the scale-to-zero job will be reverted by ArgoCD itself.

cf:
* https://github.com/kyverno/kyverno/blob/77e2a860c8d0b66c65e2203d97348d19d2cb0136/charts/kyverno/templates/hooks/pre-delete-scale-to-zero.yaml#L11
* https://argo-cd.readthedocs.io/en/latest/user-guide/helm/#helm-hooks

### DIFF
--- a/components/kyverno/development/kyverno-helm-values.yaml
+++ b/components/kyverno/development/kyverno-helm-values.yaml
@@ -69,6 +69,7 @@ policyReportsCleanup:
       drop:
       - "ALL"
 webhooksCleanup:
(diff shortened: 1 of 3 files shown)


sadlerap (approved):
/lgtm

PR #6244, opened 2025-04-28
fix: running ir shold not be removed

theflockers:
this PR fixes the case where a running IR is
wrongly removed by the cronjob due to the missing
completionTime field, that is only added after
finishing the IR run.

### DIFF
--- a/components/release/base/cronjobs/remove-internal-requests.yaml
+++ b/components/release/base/cronjobs/remove-internal-requests.yaml
@@ -31,7 +31,8 @@ spec:
                   YD=$(date -d 'yesterday' +%s)
                   kubectl get internalrequests --all-namespaces \
                   --sort-by=.status.completionTime \
-                  --template '{{range .items}}{{.metadata.name}}{{"\t"}}{{.metadata.namespace}}{{"\t"}}{{.status.completionTime}}{{"\n"}}{{end}}' > $KUBECTL_OUTPUT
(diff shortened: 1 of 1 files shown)


components/release/base/cronjobs/remove-internal-requests.yaml
```diff
@@ -31,7 +31,8 @@ spec:
                   YD=$(date -d 'yesterday' +%s)
                   kubectl get internalrequests --all-namespaces \
                   --sort-by=.status.completionTime \
-                  --template '{{range .items}}{{.metadata.name}}{{"\t"}}{{.metadata.namespace}}{{"\t"}}{{.status.completionTime}}{{"\n"}}{{end}}' > $KUBECTL_OUTPUT
+                  --template '{{range .items}}{{.metadata.name}}{{"\t"}}{{.metadata.namespace}}{{"\t"}}{{.status.completionTime}}{{"\n"}}{{end}}' \
+                    | grep -E "[0-9]{4}((-?)[0-9]{2})+T.*Z$" > $KUBECTL_OUTPUT
```
mmalina, line 35: I'm just a little worried that if the date format changes a little, it will not work, whereas deleting lines with `<no value>` would be more straightforward. But maybe it's acceptable - worst case we won't be deleting stuff and we hit the quota again.

theflockers: I had the very same thought but the opposite way. Eventually they change the "no value" string to something else. I guess both options have some risk.

mmalina: I guess the completion time is probably generated by us (the controller) and so the date format should stay like this, so hopefully it's ok.

mmalina (approved):

PR #6245, opened 2025-04-28
deploy knative-eventing to stone-prod-p02

sadlerap:
The ApplicationSet manifest for knative-eventing references deploying to stone-prod-p02, but no resources are actually deployed there.  Deploy knative-eventing's resources there just like the other clusters.


hugares (approved):
/lgtm

PR #6246, opened 2025-04-28
deploy knative-eventing to stone-stg-rh01

sadlerap:
It's on the production clusters and internal staging, no reason not to have it on external staging as well.


hugares (approved):
/lgtm

PR #6256, opened 2025-04-29
build-service update

fmudry:
Updating build-service refs.

Included PR: https://github.com/konflux-ci/build-service/pull/427

### DIFF
--- a/components/build-service/development/kustomization.yaml
+++ b/components/build-service/development/kustomization.yaml
@@ -2,14 +2,14 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
 - ../base
-- https://github.com/konflux-ci/build-service/config/default?ref=259cef0dd53e60e5ab7a996bef96a32d7b8d492d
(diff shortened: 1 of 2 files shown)


chmeliik:
FYI this also includes the update from https://github.com/redhat-appstudio/infra-deployments/pull/6252.

If the CI passes here, we should merge this and close https://github.com/redhat-appstudio/infra-deployments/pull/6252

tisutisu:
Looking at the failed test, we need to change [the value](https://github.com/konflux-ci/e2e-tests/blob/6ade58b7683db862b2a39582d4a44025e7d98b72/tests/konflux-demo/konflux-demo.go#L108) here as well

chmeliik:
> Looking at the failed test, we need to change [the value](https://github.com/konflux-ci/e2e-tests/blob/6ade58b7683db862b2a39582d4a44025e7d98b72/tests/konflux-demo/konflux-demo.go#L108) here as well

There is a PR opened to fix the e2e-tests: https://github.com/konflux-ci/e2e-tests/pull/1576

But it looks like it isn't updating the value you mentioned

mmorhun (approved):
/lgtm

PR #6258, opened 2025-04-29
removing livenessProbe

rubenmgx:
Based on the conversation on https://issues.redhat.com/browse/SPRE-1255 we will be removing the livenessProbe since is not meeting the expectations


rubenmgx:
hey @amisstea could you please check this PR? Thanks a lot!

PR #6271, opened 2025-04-30
add konflux-viewer-user-actions role to stg and prod

filariow:
Add the `konflux-viewer-user-actions` role. It provides the same rights the `konflux-contributor-user-actions `role provides.
It is required for the implementation of Public view feature.


sadlerap (approved):
/lgtm

gbenhaim:
@filariow if the role need to appear in the ui, you should add it in this configmap as well https://github.com/redhat-appstudio/infra-deployments/blob/main/components/konflux-info/staging/stone-stage-p01/info.json#L24 
(each cluster has its own configmap)

PR #6262, opened 2025-04-29
update console URLs for staging clusters

filariow:
follows up to #6187, #6125, #6089

Signed-off-by: Francesco Ilario <filario@redhat.com>

PR #6272, opened 2025-04-30
Align pipeline-service base with cluster configs

enarha:
Some changes were done without running the generate-deploy-config.sh script which keeps the deploy.yaml files aligned with the base. This PR does not introduce new changes, but just aligns the configs with the base.

### DIFF
--- a/components/pipeline-service/staging/stone-stg-rh01/deploy.yaml
+++ b/components/pipeline-service/staging/stone-stg-rh01/deploy.yaml
@@ -2462,10 +2462,10 @@ spec:
         settings:
           application-name: Konflux Staging
           custom-console-name: Konflux Staging
-          custom-console-url: https://konflux-ui.apps.stone-stg-rh01.l2vh.p1.openshiftapps.com/
-          custom-console-url-pr-details: https://konflux-ui.apps.stone-stg-rh01.l2vh.p1.openshiftapps.com/ns/{{
+          custom-console-url: https://console.dev.redhat.com/application-pipeline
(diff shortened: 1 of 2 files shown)


components/pipeline-service/staging/stone-stg-rh01/deploy.yaml
```diff
@@ -2462,10 +2462,10 @@ spec:
         settings:
           application-name: Konflux Staging
           custom-console-name: Konflux Staging
-          custom-console-url: https://konflux-ui.apps.stone-stg-rh01.l2vh.p1.openshiftapps.com/
-          custom-console-url-pr-details: https://konflux-ui.apps.stone-stg-rh01.l2vh.p1.openshiftapps.com/ns/{{
+          custom-console-url: https://console.dev.redhat.com/application-pipeline
+          custom-console-url-pr-details: https://console.dev.redhat.com/application-pipeline/ns/{{
```
aThorp96, line 2466: This URL seems to be diverging from the other staging deployment. Do you know if that's intentional?

aThorp96 (approved):
/lgtm

One question regarding some of the values

components/pipeline-service/staging/stone-stg-rh01/deploy.yaml
enarha: I do not know that for sure. Sometimes people are only updating a single cluster when they want to test a change at smaller scale. Having said that, I'm not worried about this one as #6262 updates these again and BTW this is how we found out about the inconsistency.

PR #6279, opened 2025-04-30
Configure prod-p02 cluster to test static PVMs

enkeefe00:
Multi-architectural builds in the prod-p02 cluster are timing out due to waiting for PPC VMs to be provisioned. To decrease the time needed to complete the build, a set of static VMs is provisioned so that the only wait time is for if a VM is at its current concurrency limit. These static VMs are currently prefaced with "test" so testing them will not interfere with current builds.

### DIFF
--- a/components/multi-platform-controller/production-downstream/stone-prod-p02/host-config.yaml
+++ b/components/multi-platform-controller/production-downstream/stone-prod-p02/host-config.yaml
@@ -649,6 +649,54 @@ data:
   dynamic.linux-d200-large-s390x.disk: "200"
   dynamic.linux-d200-large-s390x.instance-tag: prod-d200-s390x-large
 
+  host.testppc64le-static-1.address: "10.130.74.90"
+  host.testppc64le-static-1.platform: "linux/testppc64le"
+  host.testppc64le-static-1.user: "root"
+  host.testppc64le-static-1.secret: "internal-prod-ibm-ssh-key"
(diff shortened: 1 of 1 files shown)


components/multi-platform-controller/production-downstream/stone-prod-p02/host-config.yaml
```diff
@@ -649,6 +649,54 @@ data:
   dynamic.linux-d200-large-s390x.disk: "200"
   dynamic.linux-d200-large-s390x.instance-tag: prod-d200-s390x-large
 
+  host.testppc64le-static-1.address: "10.130.74.90"
+  host.testppc64le-static-1.platform: "linux/ppc64le"
```
hugares, line 653: what is defining the platform, the key or the value? should we set value to linux/testppc64le?

hugares: just to be on the safe side, should we have both key and value set to testppc64le?

enkeefe00: Sure, why not.

hugares: I just want to be 100% we are not messing with production, not yet :)

hugares (approved):
/lgtm

hugares:
this change is not tested by e2e, merging now

PR #6282, opened 2025-04-30
Add static s390x hosts for stone-prod-p02

hugares:
Add all the static host using a temporary platform name to be able to test them out without affecting prod. A follow up change will replace dynamic by static ones.

KFLUXINFRA-1538


hugares:
The VPC is not yet connected to RH network, PR is ready to review but we cannot merge until VPC is connected

manish-jangra (approved):

hugares:
this change is not tested by e2e, merging now

PR #6284, opened 2025-05-01
Change s390x and ppc64le to static VMs on P02

hugares:
Only change the default platforms to not do a breaking change, then an announcement will be sent to ask users to stop using larger platform flavors in favor of default ones. The default platforms are running on static VMs that are large enough to accommodate larger build needs.

A follow up change will remove the larger flavors once users had time to migrate.

[KFLUXINFRA-1538](https://issues.redhat.com//browse/KFLUXINFRA-1538)
[KFLUXINFRA-1540](https://issues.redhat.com//browse/KFLUXINFRA-1540)


enkeefe00 (approved):

arewm (approved):

hugares:
This change is not tested by e2e, merging now

PR #6290, opened 2025-05-02
Configure RH02 cluster to test static PVMs

enkeefe00:
[KFLUXINFRA-1532](https://issues.redhat.com/browse/KFLUXINFRA-1532): Multi-architectural builds in the RH02 cluster are timing out due to waiting for PPC VMs to be provisioned. To decrease the time needed to complete the build, a set of static VMs is provisioned so that the only wait time is for if a VM is at its current concurrency limit. These static VMs are currently using a test platform so testing them will not interfere with current builds.

### DIFF
--- a/components/multi-platform-controller/production/kflux-prd-rh02/host-config.yaml
+++ b/components/multi-platform-controller/production/kflux-prd-rh02/host-config.yaml
@@ -653,6 +653,54 @@ data:
   dynamic.linux-d200-large-s390x.disk: "200"
   dynamic.linux-d200-large-s390x.instance-tag: prod-d200-s390x-large
 
+  host.testppc64le-static-1.address: "10.244.0.139"
+  host.testppc64le-static-1.platform: "linux/testppc64le"
+  host.testppc64le-static-1.user: "root"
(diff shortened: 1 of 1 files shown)


components/multi-platform-controller/production/kflux-prd-rh02/host-config.yaml
```diff
@@ -653,6 +653,54 @@ data:
   dynamic.linux-d200-large-s390x.disk: "200"
   dynamic.linux-d200-large-s390x.instance-tag: prod-d200-s390x-large
 
+  host.testppc64le-static-1.address: "10.244.0.139"
+  host.testppc64le-static-1.platform: "linux/testppc64le"
+  host.testppc64le-static-1.user: "root"
+  host.testppc64le-static-1.secret: "ibm-production-testppc64le-ssh-key"
```
hugares, line 659: is the secret really named like this or this is a mass rename issue when changing the platform name?

enkeefe00: The latter; I just pushed a fix.

hugares (approved):
/lgtm

PR #6295, opened 2025-05-02
etcd-shield: deploy to staging

sadlerap:
Overview
========

etcd-shield is a load-shedding tool designed to provide a more consistent experience to users when an outage is occurring.  When a configured alert is firing, it denies the admission of pipelineruns into the cluster.

Currently, it's configured to watch for etcd's on-disk capacity (hence the name).  The database capacity limits are not stored as prometheus time-series entries, so we're forced to hard-code the current 8GiB limit in the PrometheusRule resource.

In the future, we could add more alerts to match on and more resources types for this service to deny.

Alert configuration
===================

The alert is configured to trigger at 95% capacity, and hold that trigger until capacity drops below 80%.  This is done to provide a buffer between when we stop allowing new resources and when we start providing them again.  We'd like to ensure that when we allow resource admission again, it isn't going to immediately begin denying them; having a gap of 15% will buy the cluster some breathing room.

Of course, these capacity percentages can be changed if we need to.

### DIFF
--- a/components/etcd-shield/base/deployment.yaml
+++ b/components/etcd-shield/base/deployment.yaml
@@ -1,0 +1,79 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: etcd-shield
+  labels:
+    app: etcd-shield
+spec:
+  selector:
+    matchLabels:
+      app: etcd-shield
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        app: etcd-shield
+    spec:
+      serviceAccountName: etcd-shield
+      containers:
+      - args:
+        - -leader-elect
+        - -health-probe-bind-address=:8081
+        - -config=/etc/etcd-shield/config.yaml
+        - -port=8443
+        env:
+        - name: "NAMESPACE"
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.namespace
+        image: etcd-shield:latest
+        name: etcd-shield
+        imagePullPolicy: IfNotPresent
+        livenessProbe:
+          initialDelaySeconds: 1
+          httpGet:
+            path: /healthz
+            port: 8081
+            scheme: HTTP
+        readinessProbe:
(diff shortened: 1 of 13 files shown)


components/etcd-shield/base/deployment.yaml
```diff
@@ -0,0 +1,79 @@
...
+            path: /readyz
+            port: 8081
+            scheme: HTTP
+        resources:
+          limits:
+            cpu: 500m
+            memory: 128Mi
+          requests:
+            cpu: 10m
+            memory: 64Mi
```
filariow, line 50: I'd suggest enforcing Guaranteed QoS to this service

```suggestion
        resources:
          limits:
            cpu: 500m
            memory: 128Mi
          requests:
            cpu: 500m
            memory: 128Mi
```

components/etcd-shield/base/kustomization.yaml
```diff
@@ -0,0 +1,21 @@
...
+  newName: quay.io/konflux-ci/etcd-shield
+  newTag: 3de52a71e086e152b3d2b33bfc897ce7bf8ea0dd
+namespace: etcd-shield
+resources:
+- deployment.yaml
+- ns.yaml
+- rbac.yaml
+- service.yaml
+- webhook.yaml
+- etcd_shield_alerts.yaml
```
filariow, line 21: a NetworkPolicy would be desirable too, we can add that in a subsequent PR though

filariow (commented):
mostly, lgtm

components/etcd-shield/base/kustomization.yaml
sadlerap: I think I'll punt that for a future PR.

components/etcd-shield/base/deployment.yaml
sadlerap: Added in b17bb498.

hugares (approved):
/lgtm

PR #6303, opened 2025-05-05
Configure RH01 cluster to test static PVMs (#6290)

enkeefe00:
[KFLUXINFRA-1536](https://issues.redhat.com/browse/KFLUXINFRA-1536): Multi-architectural builds in the RH01 cluster are timing out due to waiting for PPC VMs to be provisioned. To decrease the time needed to complete the build, a set of static VMs is provisioned so that the only wait time is for if a VM is at its current concurrency limit. These static VMs are currently using a test platform so testing them will not interfere with current builds.


hugares (approved):
/lgtm

PR #6316, opened 2025-05-07
Add temporary permissions to the nginx proxy sa

maruiz93:
While KubeArchive does not support impersonation
the Konflux UI team need to have unblocked the
access to the Snapshots and Releases in staging
for KubeArchive.

### DIFF
--- a/components/konflux-ui/staging/base/proxy/proxy.yaml
+++ b/components/konflux-ui/staging/base/proxy/proxy.yaml
@@ -272,6 +272,14 @@ rules:
 - apiGroups: ["authorization.k8s.io"]
   resources: ["localsubjectaccessreviews"]
   verbs: ["create"]
+- apiGroups:  # TEMPORARY SOLUTION TO UNBLOCK KONFLUX-166 WHILE KAR-510 IS NOT SOLVED
+    - "appstudio.redhat.com"
+  resources:
+    - "snapshots"
+    - "releases"
+  verbs:
+    - "list"
+    - "get"
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding


components/konflux-ui/staging/base/proxy/proxy.yaml
```diff
@@ -272,6 +272,14 @@ rules:
 - apiGroups: ["authorization.k8s.io"]
   resources: ["localsubjectaccessreviews"]
   verbs: ["create"]
+- apiGroups:  # TEMPORARY SOLUTION TO UNBLOCK KONFLUX-166 WHILE KAR-510 IS NOT SOLVED
```
gbenhaim, line 275: how this change is going to solve the issue?
are you going to use the proxie's identity when querying for snapshots and releases? if so, I don't think we should allow it.

gbenhaim (changes requested):

components/konflux-ui/staging/base/proxy/proxy.yaml
maruiz93: > are you going to use the proxie's identity when querying for snapshots and releases? if so, I don't think we should allow it.

Yes, that's what I'm trying to do here and I'm aware that this is not how this should work.
I'm asking to do this *temporarilly* while kubearchive doesn't support impersonation just for snapshots and releases and just in staging so the UI team is unblocked for their development.

That's why I added the comment, so it's clear that this piece shouldn't be there, this should be considered a temporary exception.

If I'm not mistaken, this will impact only for the access to the archived snapshots and releases, not the live ones so I think that the impact is quite delimited.

gbenhaim: no, it will give all the users access to all the snapshots and releases on the cluster.

gbenhaim: ok I see what you mean. Why not investing few more hours an configure https://github.com/openshift/kube-rbac-proxy?tab=readme-ov-file  that will do the token/access review for kubearchive.

gbenhaim (approved):

testcara (commented):
lgtm

PR #6329, opened 2025-05-07
rbac: disallow certain groups from rolebindings in prod

sadlerap:
Disallow the following groups from being subjects in a rolebinding in a tenant namespace:

- `system:anonymous`
- `system:masters`
- `system:unauthenticated`

Staging appears to be stable, so it's likely safe to rollout to production.


hugares (approved):
/lgtm

filariow (approved):
/lgtm

PR #6334, opened 2025-05-07
etcd-shield: fix ApplicationSet name

sadlerap:
Overwriting existing applicationsets is a bad idea, don't do it.


hugares (approved):
/lgtm

PR #6336, opened 2025-05-08
Fix Vector connection to Minio S3 storage in development environments

infernus01:
In development environments, Vector logs were showing "Healthcheck failed" errors 
when trying to connect to Minio (S3 storage).

We're using HTTPS for the MinIO endpoint. Based on [Vector's](https://vector.dev/docs/reference/configuration/sinks/aws_s3/#tls.verify_certificate) docs, this error could be due to TLS certificate issues.


enarha:
It looks like the CI failure is Vector related as the service never gets fully synced.

aThorp96:
@infernus01 Have you tested this? The healthcheck here is failing to dispatch the request to AWS, but this error has a number of causes, [authorization errors](https://docs.aws.amazon.com/sdk-for-rust/latest/dg/error-handling.html).

PR #6337, opened 2025-05-08
KubeArchive: bump to v1, breaking changes

rh-hemartin:
This PR bumps KubeArchive to v1.0.0. This release includes breaking changes in the `apiVersion` of the KubeArchive CRDs: the group changes from `kubearchive.kubearchive.org` to `kubearchive.org` and the version changes from `v1alpha` to `v1`.

I tested it locally and Kyverno should create the new objects without any problems and I tried the deletion too. All existing `KubeArchiveConfig`s that are `kubearchive.kubearchive.org/v1alpha` need to have their `finalizer` removed and then deleted. I'm using:

```
kubectl patch -n <namespace> kubearchiveconfigs.kubearchive.kubearchive.org kubearchive -p '{"metadata": {"finalizers": []}}' --type=merge
kubectl delete -n <namespace> kubearchiveconfigs.kubearchive.kubearchive.org kubearchive
```

The CRD also needs to be deleted as far as I know, even if ArgoCD will try to remove it:

```
kubectl delete crd kubearchiveconfigs.kubearchive.kubearchive.org
```

There are no changes no the database, so it can merged mostly unattended.

**Questions**

* I don't know if I have wide-cluster permissions, I guess I should not, so I'm not sure if I will be able to delete the old resources and the CRD. Someone could  jump in for that?

### DIFF
--- a/components/authentication/base/everyone-can-view.yaml
+++ b/components/authentication/base/everyone-can-view.yaml
@@ -4,7 +4,7 @@ metadata:
   name: view-appstudio
 rules:
 - apiGroups:
-  - kubearchive.kubearchive.org
+  - kubearchive.org
   resources:
   - kubearchiveconfigs
   verbs:
--- a/components/kubearchive/base/kubearchive-maintainer.yaml
+++ b/components/kubearchive/base/kubearchive-maintainer.yaml
@@ -4,7 +4,7 @@ metadata:
   name: kubearchive-maintainer
 rules:
   - apiGroups:
-      - kubearchive.kubearchive.org
+      - kubearchive.org
     resources:
       - kubearchiveconfigs
     verbs:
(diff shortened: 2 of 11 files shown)


components/authentication/base/everyone-can-view.yaml
filariow, line 1: IIUIC, this is shared among all the environments, but the change here is only for dev and stage. This change will likely remove everyone's view access on `kubearchiveconfigs` in prod.

I'd suggest to add the `kubearchive.org` entry in the `apiGroups` list without removing the `kubearchive.kubearchive.org` one until we deploy v1.0.0 in prod.

components/kubearchive/base/kubearchive-maintainer.yaml
```diff
@@ -4,7 +4,7 @@ metadata:
   name: kubearchive-maintainer
 rules:
   - apiGroups:
-      - kubearchive.kubearchive.org
```
filariow, line 7: ditto

filariow (changes requested):
mostly lgtm

rh-hemartin:
It did not catch the lgtm, could you go again @filariow ?

hugares (approved):
/lgtm

PR #6340, opened 2025-05-08
Add static s390x hosts for stone-prd-rh01

hugares:
Add all the static host using a temporary platform name to be able to test them out without affecting prod. A follow up change will replace dynamic by static ones.

[KFLUXINFRA-1541](https://issues.redhat.com//browse/KFLUXINFRA-1541)


sadlerap (approved):
/lgtm

PR #6353, opened 2025-05-09
test correct default pipelines security group in kflux-rh02

aThorp96:
The current default appears to be set as `pipelines-scc`, which does not exist....

More context in [this slack thread](https://redhat-internal.slack.com/archives/C04PZ7H0VA8/p1746731894794529?thread_ts=1746425126.665539&cid=C04PZ7H0VA8)


scoheb:
should this be applied to ***all*** clusters?

hugares:
You should not change deploy.yaml, this file is generated, see: https://github.com/redhat-appstudio/infra-deployments/tree/main/components/pipeline-service#update-procedure

You need to do patch in the right src folder for the cluster you want to change and then re-generate the deploy files

aThorp96:
@scoheb we're testing on this cluster and if it works as expected we will propagate to to all clusters. The issue in question is not reproducible in staging, hence testing on the cluster we know is affected.

hugares (approved):

PR #6355, opened 2025-05-09
Create ClusterRoles appropriate for service accounts

ralphbean:
Here, I'm thinking to add something similar like what is used by the art tenant. There, there's an in-namespace Role. But, general, consistent clusterroles across clusters will be helpful to other users who want to do similar things.

### DIFF
--- a/components/konflux-rbac/production/base/konflux-releaser-bot-actions.yaml
+++ b/components/konflux-rbac/production/base/konflux-releaser-bot-actions.yaml
@@ -1,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: konflux-builder-bot-actions
+rules:
+- apiGroups:
+  - appstudio.redhat.com
+  resources:
+  - releases
+  verbs:
+  - watch
+  - create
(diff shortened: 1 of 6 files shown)


components/konflux-rbac/production/base/konflux-releaser-bot-actions.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: konflux-builder-bot-actions
```
filariow, line 4: Is this a typo?

```suggestion
  name: konflux-releaser-bot-actions
```

components/konflux-rbac/staging/base/konflux-releaser-bot-actions.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: konflux-builder-bot-actions
```
filariow, line 4: ditto

gbenhaim:
lgtm (just need to fix the CR names)

gbenhaim:
how users will discover those roles? are we going to mention them in the documentation?

components/konflux-rbac/production/base/konflux-releaser-bot-actions.yaml
ralphbean: It is!

ralphbean: Resolved

components/konflux-rbac/staging/base/konflux-releaser-bot-actions.yaml
ralphbean: Good catch - resolved.

PR #6357, opened 2025-05-10
Revert "test correct default pipelines security group in kflux-rh02"

aThorp96:
Reverts redhat-appstudio/infra-deployments#6353

This change did not resolve the issue, so the change is being reverted


savitaashture (approved):
/lgtm

simonbaird:
See also https://github.com/enterprise-contract/ec-cli/pull/2513 which is a different workaround aiming to solve the same problem.

PR #6366, opened 2025-05-12
bootstrap tenant policy creates missing resources

filariow:
where missing, resources will be created.

### DIFF
--- a/components/konflux-rbac/policies/bootstrap-tenant-namespace/kyverno_rbac.yaml
+++ b/components/konflux-rbac/policies/bootstrap-tenant-namespace/kyverno_rbac.yaml
@@ -2,7 +2,7 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-  name: kyverno-admission:read-rolebindings
+  name: kyverno-admission:bootstrap-tenant-namespace
   labels:
     rbac.kyverno.io/aggregate-to-admission-controller: "true"
 rules:


sadlerap (approved):
/lgtm

components/konflux-rbac/policies/bootstrap-tenant-namespace/bootstrap-tenant-namespace-clusterpolicy.yaml
```diff
@@ -34,3 +34,151 @@ spec:
...
+    match:
+      any:
+      - resources:
+          kinds:
+          - Namespace
+          selector:
+            matchLabels:
+              konflux-ci.dev/type: tenant
+    generate:
+      generateExisting: true
```
gbenhaim, line 47: what does `generateExisting` mean? (I didn't find it in the documentation)

gbenhaim (commented):
lgtm. Can you please also create a ticket for revisiting the network policies? I think we don't need all of them.

components/konflux-rbac/policies/bootstrap-tenant-namespace/bootstrap-tenant-namespace-clusterpolicy.yaml
filariow: this way we'll create the resources in all the already existing tenant namespaces that don't have them. Nothing will happen if they already exist. After some time, when we are sure every namespace has been fixed, we can change this value to `false`

components/konflux-rbac/policies/bootstrap-tenant-namespace/bootstrap-tenant-namespace-clusterpolicy.yaml
```diff
@@ -34,3 +34,151 @@ spec:
...
+      synchronize: false
```
gbenhaim, line 48: why synchronize is false? don't we want to update the NetworkPolicy if the Kyverno ClusterPolicy is changed?

filariow: for the current implementation it could collide with KubeSaw's managed ones. We can modify the rules to take in consideration KubeSaw's labels or just wait for KubeSaw to be removed. I'd propose to do that in a dedicated PR if/when we need to update the resources.

PR #6332, opened 2025-05-07
fix(PVO11Y-4653): Include es01 cluster labels

gcpsoares:
- Include source cluster labels.

PR #6378, opened 2025-05-13
fix(PVO11Y-4653): Add es01 clusterDir for monitoring-workload-prometheus

pacho-rh:
#6332 adds some external labeling for es01 that is not being applied due to es01 clusters defaulting to the base kustomize configuration for monitoring-workload-prometheus.
Added es01 clusterDir so those clusters can use their own kustomize configuration instead.


pacho-rh:
cc @gcpsoares @ci-operator

raks-tt (approved):
LGTM

raks-tt (approved):
/lgtm

PR #6380, opened 2025-05-13
Builder serviceaccounts should be able to inspect events

ralphbean:
This is useful for summarizing why a pipelinerun failed.


filariow (approved):
/lgtm

PR #6387, opened 2025-05-14
Increase releaseplan and releaseplanadmission quotas

hugares:
For larger tiers, increase both. We are hitting the limit in managed namespace.


filariow (approved):
/lgtm

PR #6389, opened 2025-05-14
Bump kubelinter

sadlerap:
We're running a fairly outdated version of kube-linter; let's update it to the latest version.

Unfortunately, kube-linter has introduced some new lints, and some of the components in our repository aren't in compliance with them.  We'll need to disable them for now until these components are fixed.

As long as we're at it, let's also run kube-linter checks on push to `main`.  It should get sarif reports working on PRs, and we should check that commits that wind up in `main` are linted appropriately.

### DIFF
--- a/.github/workflows/kube-linter.yaml
+++ b/.github/workflows/kube-linter.yaml
@@ -5,6 +5,8 @@ on:
   # SARIF reports.
   pull_request:
     branches: [ main ]
+  push:
+    branches: [ main ]
 jobs:
   scan:
     permissions: write-all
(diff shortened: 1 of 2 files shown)


meyrevived (commented):
Are the excluded checks are necessary for anything? Are there plans to fix what linting issues these components run into, or did you chose an exclusion method because it's up to another team to fix what the linter finds?

sadlerap:
Largely the latter; if I have time to do so, I can try to fix the components that are causing issues with the lints here, but no guarantees.  The list of failing components is fairly small:
- build-service
- mintmaker
- notification-controller
- pipeline-service
- project-controller
- release
- repository-validator
- ui
At the very least, I'm planning on filing some bugs tomorrow.  Ideally, I wouldn't need to have these linter exclusions in place.

PR #6390, opened 2025-05-15
dev: Increase Argo's application controller resource limits

gbenhaim:
During deployment the application controller struggles with processing all the Konflux applications because of missing resources. This cause it to crash (OOM) and because of that the deployment takes more time.

Increase resource limits.


hugares (approved):
/lgtm

PR #6394, opened 2025-05-15
Increase Kyverno UpdateRequests threshold

filariow:
This PR increases the UpdateRequests Threshold for clusters where we are hitting the limit.
The old (default) value was 1000, so this PR doubles it.


hugares (approved):
/lgtm

PR #6404, opened 2025-05-16
etcd-shield: remove cluster-monitoring label

sadlerap:
It was used for debugging purposes, and has no business being set on the production nor staging clusters.


filariow (approved):
/lgtm

PR #6405, opened 2025-05-18
Disable Tekton Pruner

enarha:
Tekton Pruner was initially enabled to help the Results Watcher workqueue from growing too big. With the recent optimizations that probably won't be required anymore. Furthermore, we suspect the pruner removing forcefully the pipelineruns could be the reason some taskruns are not properly stored by Results.


khrm (commented):
/lgtm

PR #6407, opened 2025-05-19
KubeArchive: install on production

rh-hemartin:
* Added production clusters, but only enabled one to prevent problems on multiple production clusters at once.
* Patches are duplicated because we need them to be this way for clusters to be independent.


rh-hemartin:
@hugares @manish-jangra do you know which is the correct path for the database secrets? I just assumed the same path as the staging secret.

components/kubearchive/production/base/database-secret.yaml
```diff
@@ -0,0 +1,28 @@
...
+kind: ExternalSecret
+metadata:
+  name: database-secret
+  annotations:
+    argocd.argoproj.io/sync-options: SkipDryRunOnMissingResource=true
+    argocd.argoproj.io/sync-wave: '-1'
+spec:
+  dataFrom:
+    - extract:
+        key: production/kubearchive/rds
```
maruiz93, line 12: Could you please note with a comment or something that this is just a placeholder being modified on each cluster installation?

maruiz93:
Could you also add the configuration needed under `konflux-ui` component (based on [this](https://github.com/redhat-appstudio/infra-deployments/pull/6206/files#diff-66c19a34c45a748154d0bef240fa1da1f130742a980aa3c87fffde6b5f7949af)) so the kubearchive API is available for Konflux UI through its nginx proxy.

NOTE: There is no need to add the Route as the service doesn't need to be exposed for that. That is something we need to deal with when it comes to the kubearchive CLI usage, but not for the API per se.

components/kubearchive/production/base/database-secret.yaml
rh-hemartin: Added

rh-hemartin:
> Could you also add the configuration needed under `konflux-ui` component (based on [this](https://github.com/redhat-appstudio/infra-deployments/pull/6206/files#diff-66c19a34c45a748154d0bef240fa1da1f130742a980aa3c87fffde6b5f7949af)) so the kubearchive API is available for Konflux UI through its nginx proxy.

Done!

mafh314 (approved):

argo-cd-apps/base/member/infra-deployments/kubearchive/kubearchive.yaml
```diff
@@ -13,9 +13,27 @@ spec:
...
+                # Private
+                - nameNormalized: kflux-ocp-p01
+                  values.clusterDir: kflux-ocp-p01
+                - nameNormalized: stone-prod-p01
+                  values.clusterDir: stone-prod-p01
+                - nameNormalized: stone-prod-p02
+                  values.clusterDir: stone-prod-p02
+                # Public
+                - nameNormalized: stone-prd-m01
+                  values.clusterDir: stone-prd-m01
```
filariow, line 30:
```suggestion
```

m01 is deprecated and it will be removed soon. @hugares should skip it in this rollout already?

components/kubearchive/development/kustomization.yaml
```diff
@@ -14,10 +15,168 @@ secretGenerator:
...
+                securityContext:
+                  readOnlyRootFilesystem: true
+                  runAsNonRoot: true
+                resources:
+                  limits:
+                    cpu: 200m
+                    memory: 1Gi
+                  requests:
+                    cpu: 200m
+                    memory: 900Mi
```
filariow, line 137: the development overlay is used in e2e tests. Do you think we can reduce deployments' `resources` in this overlay so we'll require smaller clusters during e2e tests?

components/kubearchive/production/kflux-ocp-p01/kustomization.yaml
```diff
@@ -0,0 +1,176 @@
...
+                  runAsNonRoot: true
+                ports:
+                  - containerPort: 8081
+                resources:
+                  $patch: replace
+                  limits:
+                    cpu: 500m
+                  requests:
+                    cpu: 10m
+                    memory: 256Mi
```
filariow, line 104: to prevent eviction of this service in production, we'd like to ensure the deployment has Guaranteed QoS (i.e. `limits==requests`). This suggestion applies also to the other KubeArchive's deployments.

argo-cd-apps/base/member/infra-deployments/kubearchive/kubearchive.yaml
rh-hemartin: I removed it, I can add it back if needed.

components/kubearchive/development/kustomization.yaml
rh-hemartin: I reduced the memory for all deployments, currently we are sitting at ~30Mi on staging because we introduced a fix that solved the high memory usage.

components/kubearchive/production/kflux-ocp-p01/kustomization.yaml
rh-hemartin: It applies for both, I checked the docs.

argo-cd-apps/base/member/infra-deployments/kubearchive/kubearchive.yaml
```diff
@@ -13,23 +13,40 @@ spec:
               values:
                 sourceRoot: components/kubearchive
                 environment: staging
-                clusterDir: ""
+                clusterDir: base
```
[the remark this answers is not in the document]
rh-hemartin: I changed it, let me know if that will work. I'm trying to run:

```
argocd appset generate argo-cd-apps/base/member/infra-deployments/kubearchive/kubearchive.yaml
```

To see how the file will look like, but I'm getting a deadline error when connecting to the private control plane with argocd.

hugares (approved):
/lgtm

PR #6421, opened 2025-05-20
MintMaker weekly release: promote stage to prod

FernandesMF:
This commit promotes the latest images tested in stage to prod, as part of the weekly release for
MintMaker.

### Testing remarks
I monitored the deployment in stage and there were no errors. I also double-checked quay and the images were built without problem there.

I believe it is safe to promote this to prod.


qixiang:
There is automated PR for this: https://github.com/redhat-appstudio/infra-deployments/pull/6419

PR #6422, opened 2025-05-21
Etcd shield updates

sadlerap:
The highlights:
- 3 replicas instead of 1, so etcd-shield is more resiliant to pods falling over
- disable leader election, since we don't need it for anything.

### DIFF
--- a/components/etcd-shield/base/deployment.yaml
+++ b/components/etcd-shield/base/deployment.yaml
@@ -17,7 +17,7 @@ spec:
       serviceAccountName: etcd-shield
       containers:
       - args:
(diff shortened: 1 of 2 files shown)


gbenhaim:
@sadlerap need to add topology spread constraints.

here is an example:

https://github.com/redhat-appstudio/infra-deployments/blob/main/components/konflux-ui/production/base/proxy/proxy.yaml#L9

https://github.com/redhat-appstudio/infra-deployments/blob/main/components/konflux-ui/production/base/proxy/proxy.yaml#L26

gbenhaim (approved):

PR #6427, opened 2025-05-21
fix(PVO11Y-4776): Include new cpu label

gcpsoares:
- Include new cpu label to improve cluster cpu usage overview.


mftb (approved):
/lgtm
/approve
/ok-to-test

PR #6428, opened 2025-05-21
update resource configurations for Tekton components

ab-ghosh:
Added CPU and memory limits for following Tekton resources:

-    pipelines-as-code-controller

-     pipelines-as-code-watcher

-     pipelines-as-code-webhook

-     pipelines-console-plugin

-     tekton-chains-controller

-     tekton-events-controller

-     tekton-triggers-controller

-     tekton-triggers-core-interceptors

-     tekton-triggers-webhook

-     tkn-cli-serve

### DIFF
--- a/components/pipeline-service/development/main-pipeline-service-configuration.yaml
+++ b/components/pipeline-service/development/main-pipeline-service-configuration.yaml
@@ -1960,6 +1960,136 @@ spec:
                       requests:
                         cpu: 200m
                         memory: 200Mi
+        pipelines-as-code-controller:
+          spec:
+            template:
+              spec:
+                containers:
+                  - name: pac-controller
+                    resources:
+                      limits:
+                        cpu: 200m
+                        memory: 300Mi
+                      requests:
+                        cpu: 500m
+                        memory: 500Mi
+        pipelines-as-code-watcher:
+          spec:
+            template:
+              spec:
+                containers:
+                  - name: pac-watcher
+                    resources:
+                      limits:
+                        cpu: 500m
+                        memory: 3Gi
+                      requests:
+                        cpu: "1"
+                        memory: 4.5Gi
+        pipelines-as-code-webhook:
+          spec:
+            template:
+              spec:
+                containers:
+                  - name: pac-webhook
+                    resources:
+                      limits:
+                        cpu: 200m
+                        memory: 250Mi
+                      requests:
+                        cpu: 500m
+                        memory: 400Mi
+        pipelines-console-plugin:
+          spec:
+            template:
+              spec:
+                containers:
+                  - name: pipelines-console-plugin
+                    resources:
+                      limits:
+                        cpu: 50m
+                        memory: 25Mi
+                      requests:
+                        cpu: 100m
+                        memory: 35Mi
+        tekton-chains-controller:
+          spec:
+            template:
+              spec:
+                containers:
+                  - name: tekton-chains-controller
+                    resources:
+                      limits:
+                        cpu: 450m
+                        memory: 4Gi
+                      requests:
+                        cpu: "1500m"
+                        memory: 7Gi
+        tekton-events-controller:
+          spec:
+            template:
+              spec:
+                containers:
+                  - name: tekton-events-controller
+                    resources:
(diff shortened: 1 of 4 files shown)


components/pipeline-service/development/main-pipeline-service-configuration.yaml
```diff
@@ -1960,6 +1960,136 @@ spec:
...
+              spec:
+                containers:
+                  - name: tekton-chains-controller
+                    resources:
+                      limits:
+                        cpu: 450m
+                        memory: 4Gi
+                      requests:
+                        cpu: "1500m"
+                        memory: 7Gi
```
aThorp96, line 2027: I think the `limits` and `requests` might be backwards here. Also, should the limits and requests be the same?

aThorp96: I think that's the case on all of these actually.

ab-ghosh: Updated the limits

components/pipeline-service/development/main-pipeline-service-configuration.yaml
```diff
@@ -1960,6 +1960,136 @@ spec:
...
+                      requests:
+                        cpu: 500m
+                        memory: 500Mi
+        pipelines-as-code-watcher:
+          spec:
+            template:
+              spec:
+                containers:
+                  - name: pac-watcher
+                    resources:
```
enarha, line 1982: I think those are way too high for the development overlay, I'll suggest cpu: 500, memory: 1 or 2 GB. Basically for the development overlay I think we should use values as found on the staging clusters which usually do not see any load. The thing here is to pass the CI tests, without being too greedy for resources.
This applies for all development overlay values.

components/pipeline-service/development/main-pipeline-service-configuration.yaml
```diff
@@ -1960,6 +1960,136 @@ spec:
...
+        pipelines-console-plugin:
+          spec:
+            template:
+              spec:
+                containers:
+                  - name: pipelines-console-plugin
+                    resources:
+                      limits:
+                        cpu: 100m
+                        memory: 35Mi
```
enarha, line 2011: 35Mi? Isn't that too low? Even if it is enough, I'll suggest using 100 or 200 MB, which won't really make huge difference on the resources usage.

components/pipeline-service/development/main-pipeline-service-configuration.yaml
```diff
@@ -1960,6 +1960,136 @@ spec:
...
+                      requests:
+                        cpu: 100m
+                        memory: 35Mi
+        tekton-chains-controller:
+          spec:
+            template:
+              spec:
+                containers:
+                  - name: tekton-chains-controller
+                    resources:
```
enarha, line 2021: Those limits seem way too much IMO. As mention above, check the actual usage on the staging cluster and set accordingly. So for this one, I just checked and the average mem usage is around 250Mi with peaks up to 350, so maybe a value of 1GB should be enough here. For the cpu, the average stands to 40m with peaks up to 700m, so anything between 500m to 1 should be OK IMO.

components/pipeline-service/development/main-pipeline-service-configuration.yaml
```diff
@@ -1960,6 +1960,136 @@ spec:
...
+                      requests:
+                        cpu: 250m
+                        memory: 100Mi
+        tekton-triggers-core-interceptors:
+          spec:
+            template:
+              spec:
+                containers:
+                  - name: tekton-triggers-core-interceptors
+                    resources:
```
enarha, line 2073: These are also too high based on the actual user I observe.

enarha (commented):
My comments on the development overlay also apply to the staging overlay. When we update production in another PR, then the values will be much higher, again based on the observed values on the clusters.

aThorp96:
If the e2e tests continue to fail, we may want to lower the requests for the development configuration. It's possible that the e2e tests' cluster doesn't have sufficient resources for the requests, leading to controllers just never getting scheduled.

PR #6435, opened 2025-05-26
update resource configurations for Tekton components

ab-ghosh:
Added CPU and memory limits for following Tekton resources:
* pipelines-as-code-controller
* pipelines-as-code-watcher
* pipelines-as-code-webhook
* pipelines-console-plugin
* tekton-chains-controller
* tekton-events-controller
* tekton-triggers-controller
* tekton-triggers-core-interceptors
* tekton-triggers-webhook
* tkn-cli-serve


enarha (approved):
/lgtm

PR #6432, opened 2025-05-22
update o11y dsexporter in stage

FaisalAl-Rayes:
This is to test out the [changes made to the building of o11y/exporters](https://github.com/redhat-appstudio/o11y/pull/576)

PR #6437, opened 2025-05-26
Rollout o11y dsexporter in prod

FaisalAl-Rayes:
This is a follow up to https://github.com/redhat-appstudio/infra-deployments/pull/6432


martysp21 (approved):
lgtm

PR #6443, opened 2025-05-26
Update prod clusters with more recent MPC with improved cleanup task

mshaposhnik:
New version,  tested on RH02 for more than week and showed reduce of users leftovers number significantly.


hugares (approved):
/lgtm

PR #6446, opened 2025-05-27
Add testplatform ProviderConfig

danilo-gemoli:
Add a new `ProviderConfig` for crossplane. This is going to be needed in order to access the TestPlatform `appci` cluster via a `kubeconfig` stored into the Vault.

### DIFF
--- a/components/crossplane-control-plane/production/kustomization.yaml
+++ b/components/crossplane-control-plane/production/kustomization.yaml
@@ -4,6 +4,7 @@ kind: Kustomization
 resources:
 - ../base
 - provider-config.yaml
+- testplatform-provider-config.yaml
 
 patches:
   - patch: |-
--- a/components/crossplane-control-plane/production/testplatform-provider-config.yaml
+++ b/components/crossplane-control-plane/production/testplatform-provider-config.yaml
@@ -1,0 +1,35 @@
+---
+apiVersion: kubernetes.crossplane.io/v1alpha1
+kind: ProviderConfig
+metadata:
+  name: testplatform-kubernetes-provider-config
+  annotations:
+    argocd.argoproj.io/sync-options: SkipDryRunOnMissingResource=true
+spec:
+  credentials:
+    source: Secret
+    secretRef:
+      namespace: crossplane-system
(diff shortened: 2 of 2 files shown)


components/crossplane-control-plane/production/testplatform-provider-config.yaml
```diff
@@ -0,0 +1,35 @@
...
+metadata:
+  name: testplatform-cluster
+  namespace: crossplane-system
+  annotations:
+    argocd.argoproj.io/sync-options: SkipDryRunOnMissingResource=true
+    argocd.argoproj.io/sync-wave: "-1"
+spec:
+  dataFrom:
+    - extract:
+        key: production/openshift-ci/ephemeral-cluster
```
amisstea, line 27: It seems odd for this key to have `ephemeral` in the name. Is it expected to be short lived?

components/crossplane-control-plane/production/kustomization.yaml
amisstea, line 1: It's typical to deploy and test with the staging env first before deploying to prod. Is there a staging test platform cluster we can use?

danilo-gemoli: Unfortunately in TestPlatform we don't have any staging environment, but we rather test in production.

components/crossplane-control-plane/production/testplatform-provider-config.yaml
danilo-gemoli: `ephemeral-cluster` relates to the work I'm carrying on. The `kubeconfig` I'm extracting here is related to our main cluster `appci`. I can rename it into `production/openshift-ci/appci` if this sounds better to you.

amisstea: Yeah, or `appci-cluster` just to be extra clear.

components/crossplane-control-plane/production/kustomization.yaml
amisstea: Ok. In that case, let's deploy the same configs in the staging env so it can at least be tested from there.

PR #6450, opened 2025-05-27
fix(RELEASE-1645): allow serviceaccounts to see cm

scoheb:
- Grant access to authenticated users to be able to view openshift-pipelines configMaps. 
- The use case involves service accounts running automated tests being able to discover the custom URL for the Konflux UI

### DIFF
--- a/components/pipeline-service/base/rbac/openshift-pipelines/configmap-viewer.yaml
+++ b/components/pipeline-service/base/rbac/openshift-pipelines/configmap-viewer.yaml
@@ -1,0 +1,33 @@
+---
+# Grant access to authenticated users to be able to view
+# openshift-pipelines configMaps. The user case involves
+# service accounts running automated tests being able to
+# discover the custom URL for the Konflux UI
+kind: Role
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: configmap-viewer
+  namespace: openshift-pipelines
+rules:
+  - verbs:
+      - get
+      - watch
+      - list
+    apiGroups:
+      - ''
+    resources:
+      - configmaps
+---
+kind: RoleBinding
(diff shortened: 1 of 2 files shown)


components/pipeline-service/base/rbac/openshift-pipelines/configmap-viewer.yaml
```diff
@@ -0,0 +1,33 @@
+---
+# Grant access to authenticated users to be able to view
+# openshift-pipelines configMaps. The user case involves
+# service accounts running automated tests being able to
+# discover the custom URL for the Konflux UI
+kind: Role
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: configmap-viewer-role
```
aThorp96, line 9: Nitpick: is it necessary to include the resource kind in the resource name? Any objection to naming the role `configmap-viewer`? It's not the end of the world to leave the kind in the name, just a little verbose, like `type FooStruct struct {}` 😄

aThorp96 (approved):
/lgtm

One optional nitpick

components/pipeline-service/base/rbac/openshift-pipelines/configmap-viewer.yaml
```diff
@@ -0,0 +1,33 @@
+---
+# Grant access to authenticated users to be able to view
+# openshift-pipelines configMaps. The user case involves
+# service accounts running automated tests being able to
+# discover the custom URL for the Konflux UI
```
aThorp96, line 5: I really appreciate the "why" comment here. Since this is going to apply to all users, is there any risk in all users having read access to configmaps? I'm wondering if the the automated testing serviceaccounts have a shared group which this rolebinding can use as its subject instead of all authenticated users

scoheb:
no longer needed!

PR #6453, opened 2025-05-27
pulp access controller

YasenT:
PR for the pulp access controller. It's meant to provide secrets for each user who needs access to Pulp.


ralphbean:
Good start! Some first feedback:

* I think we want to try to use a kustomize *reference* to your pulp-access-controller git repo for the CRD definition, SA, etc. Most teams put this in their repos in a `config/manifests/` dir, like this: https://github.com/konflux-ci/build-service/blob/main/config/manifests/kustomization.yaml and then they refer to that from infra-deployments like this: https://github.com/redhat-appstudio/infra-deployments/blob/main/components/build-service/development/kustomization.yaml#L5 ; Other teams do this reference using a github release, for an extra official link between them as an upstream and this repo as a downstream deployment, like this: https://github.com/redhat-appstudio/infra-deployments/blob/main/components/kubearchive/base/kustomization.yaml#L5
* You need to add a reference to your component from `argo-cd-apps/` otherwise argo won't know to pick it up and deploy it. See here an example of `kubearchive` which is currently deployed only to staging, not to production:

```
❯ git grep kubearchive argo-cd-apps 
argo-cd-apps/base/member/infra-deployments/kubearchive/kubearchive.yaml:  name: kubearchive
argo-cd-apps/base/member/infra-deployments/kubearchive/kubearchive.yaml:                sourceRoot: components/kubearchive
argo-cd-apps/base/member/infra-deployments/kubearchive/kubearchive.yaml:      name: kubearchive-{{nameNormalized}}
argo-cd-apps/base/member/infra-deployments/kubearchive/kubearchive.yaml:        namespace: product-kubearchive
argo-cd-apps/base/member/infra-deployments/kubearchive/kustomization.yaml:- kubearchive.yaml
argo-cd-apps/base/member/infra-deployments/kustomization.yaml:  - kubearchive
argo-cd-apps/overlays/development/kustomization.yaml:      name: kubearchive
argo-cd-apps/overlays/konflux-public-production/delete-applications.yaml:  name: kubearchive
argo-cd-apps/overlays/production-downstream/delete-applications.yaml:  name: kubearchive
argo-cd-apps/overlays/staging-downstream/delete-applications.yaml:  name: kubearchive
```

There, see that the *overlays* explicitly _delete_ kubearchive from the production instances. We want to do that with pulp controller here now as well.

* You should add yourself and some team members to `components/pulp-access-controller/OWNERS` so we know who is responsible.
* Please fill out a `components/pulp-access-controller/README.md` file explaining your controller and its purpose.
* We need to use an `ExternalSecret` to get your pulp cert and key in place. There are lots of examples in the other components. Your vault path is `stonesoup/staging/pulp/pulp-access-controller`.

components/pulp-access-controller/base/combined.yaml
```diff
@@ -0,0 +1,107 @@
...
+  name: pulpaccess-operator-role
+rules:
+  - apiGroups: ["konflux.io"]
+    resources: ["pulpaccessrequests"]
+    verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
+  - apiGroups: ["apiextensions.k8s.io"]
+    resources: ["customresourcedefinitions"]
+    verbs: ["get", "list", "watch"]
+  - apiGroups: [""]
+    resources: ["secrets", "events", "pods"]
```
ralphbean, line 50: Do you really need to be able to create/update/patch pods?

components/pulp-access-controller/base/combined.yaml
```diff
@@ -0,0 +1,107 @@
...
+metadata:
+  name: pulpaccess-authenticated-users-binding
+subjects:
+  - kind: Group
+    name: system:authenticated
+    apiGroup: rbac.authorization.k8s.io
+roleRef:
+  kind: ClusterRole
+  name: pulpaccess-user-role
+  apiGroup: rbac.authorization.k8s.io
```
ralphbean, line 86: Heh, I think this one needs to be removed, or at least it should be here only temporarily.

It means "every user can create/update/delete pulp access requests in **every namespace**; in other peoples' namespaces"

components/pulp-access-controller/README.md
```diff
@@ -0,0 +1,3 @@
+# Pulp Access Controller
+
+The `pulp-access-controller` is a CRD/Controller which creates namespace secrets that provide access to Pulp over TLS.
```
ralphbean, line 3:
```suggestion
The `pulp-access-controller` is a CRD/Controller which creates namespace secrets that provide access to Pulp over TLS.

It is similar in concept to `image-controller`. Image Controller provisions quay repos and creates secrets for user pipelines to use. Pulp Access Controller provisions pulp tenants and provides secrets for user pipelines to access those.
```

ralphbean: My hope is that this sentence helps Konflux contributors understand the role of this thing.

ralphbean:
> And can I skip the "delete from prod" part? There seems to be a bit of a push to test this by the SP RHEL team, early next week. So can we use them as a quick test? With the controller being basic, I am targeting to get it to production fast, if possible.

No, I don't want to risk you knocking over production clusters. I don't have reason to think that would happen, but let's get you to stage first and confirm that it starts up. We can follow up later to get you to prod.

> Also, initially can I use a reference just to my main branch, and when this deploys and works i will move to a tag? Just opening the door for easier testing in a way.

Yes, I think that's ok.

We really want you on a digest git ref so that we can be sure we know what is deployed at any given time. It gives us a way to `git revert` a deployment here.

ralphbean:
This still needs some work, but I'm going to merge it to iterate with Yasen on it.

PR #6459, opened 2025-05-28
mintmaker: update redis deployment to reduce CPU and memory footprint

qixiang:
Seems we don't need so much cpu/memory resource for redis-deployment.

To prevent the Redis deployment from consuming excessive resources while deploying Konflux, I reduced the resource requests.


qixiang:
@querti could you help to confirm whether this is safe to be changed?

PR #6469, opened 2025-05-28
Update kyverno policy to validate the cost-center label

raks-tt:
Separated out the policies for stage and prod clusters. 
Prod will be handled in https://github.com/redhat-appstudio/infra-deployments/pull/6649

Update Kyverno policy to validate the cost-center label for staging clusters:
      (1) New tenant namespaces have the `cost-center` label.
      (2) The `cost-center` label value contains only numeric characters.
      (3) The `cost-center` label is propagated from non-managed tenant namespaces to pods.
      (4) The `insights_cost_management_optimizations` label is set to `true` on Pods to which the `cost-center` label was propagated.

- Organized policies to differentiate between staging and production environments for better cluster-specific management.


components/cost-management/policies/update_namespace_labels.yaml
```diff
@@ -27,68 +24,12 @@ spec:
...
-      mutate:
-        targets:
-          - apiVersion: v1
-            kind: Pod
-        patchStrategicMerge:
-          metadata:
-            labels:
-              cost-center: "760"
-              insights_cost_management_optimizations: "true"
+              cost-center: "?*"
```
filariow, line 32: I think we should we validate the value of the label too -not only the existence and the not-emptiness. It should be a string composed by just numbers, right?

In this case, something like this should do the trick:

```suggestion
              cost-center: "?*"
    - name: validate-cost-center-label-value
      match:
        any:
          - resources:
              kinds:
              - Namespace
              operations:
              - CREATE
              - UPDATE
              selector:
                matchLabels:
                  konflux-ci.dev/type: tenant
                  cost-center: "?*"
      validate:
        message: "Tenant namespaces 'cost-center' label can only contain numbers."
        cel:
          expressions:
          - expression: "object.metadata.labels['cost-center'].matches('^[0-9]+$')"
```

here the test

```yaml
---
apiVersion: chainsaw.kyverno.io/v1alpha1
kind: Test
metadata:
  name: deny-namespace-with-invalid-cost-center
spec:
  concurrent: false
  description: |
    Tests that a namespace labeled as `konflux-ci.dev/type: tenant`
    is denied if the `cost-center` label is invalid.
  steps:
    - name: Apply RBAC
      try:
        - apply:
            file: ../kyverno_rbac.yaml
    - name: Apply Kyverno Cluster Policy and assert it exists
      try:
        - apply:
            file: ../update_namespace_labels.yaml
        - assert:
            file: chainsaw-assert-clusterpolicy.yaml
    - name: Attempt to create a tenant namespace with invalid cost-center label
      try:
        - create:
            file: ./resources/namespace-cost-center.yaml
            template: true
            bindings:
              - name: namespace
                value: tenant
              - name: cost_center
                value: "invalid-should-be-all-numbers"
            expect:
            - match:
                kind: Namespace
              check:
                ($error != null): true
```


cc: @hugares

filariow (changes requested):
As it is right now, this PR will publish the new version of the policy in staging and production at the same time. That's something we don't want to do. Please change it to only target the staging cluster. We'll promote it to production in a second PR.

components/cost-management/policies/update_namespace_labels.yaml
raks-tt: Thank you, I have added the new rule as well as the test case and tested this locally.

raks-tt: Also moved the policies folder under staging.

components/cost-management/staging/policies/kyverno_rbac.yaml
```diff
@@ -0,0 +1,21 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: kyverno-background-controller-update
+  labels:
+    rbac.kyverno.io/aggregate-to-background-controller: "true"
```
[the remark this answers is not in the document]
raks-tt: updated

components/cost-management/staging/policies/.chainsaw-test/chainsaw-test.yaml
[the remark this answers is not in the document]
raks-tt: added

components/cost-management/staging/policies/kyverno_rbac.yaml
```diff
@@ -0,0 +1,21 @@
...
+  - pods
+  - namespaces
+  verbs:
+  - list
+  - get
+  - create
+  - delete
+  - update
+  - patch
+  - watch
```
filariow, line 21: definitely we don't need `create` and `delete` 

```suggestion
  - list
  - get
  - update
  - patch
  - watch
```

If I'm not mistaken, this is the minimum we should require
```suggestion
  - list
  - get
```

components/cost-management/staging/policies/update_namespace_labels.yaml
```diff
@@ -0,0 +1,93 @@
...
+    policies.kyverno.io/category: Cost Management
+    policies.kyverno.io/severity: low
+    policies.kyverno.io/subject: Pod
+    policies.kyverno.io/description: >-
+      This policy ensures that for Namespaces labeled as tenant type:
+      (1) Validates that new tenant namespaces have a `cost-center` label.
+      (2) Ensures the `cost-center` label value contains only numeric characters during namespace creation or update.
+      (2) Propagates the `cost-center` label from tenant namespaces to pods within the namespace.
+      (3) The `insights_cost_management_optimizations` label is set to `true` on Pods if the cost-center label was propagated.
+      (4) Excludes managed tenant namespaces from the policy using the `NotIn` condition.
```
filariow, line 17:
```suggestion
      This policy ensures that:
      (1) New tenant namespaces have the `cost-center` label.
      (2) The `cost-center` label value contains only numeric characters.
      (3) The `cost-center` label is propagated from non-managed tenant namespaces to pods.
      (4) The `insights_cost_management_optimizations` label is set to `true` on Pods to which the `cost-center` label was propagated.
```

components/cost-management/staging/policies/update_namespace_labels.yaml
```diff
@@ -0,0 +1,93 @@
...
+            kinds:
+            - Pod
+            namespaceSelector:
+              matchLabels:
+                konflux-ci.dev/type: tenant
+              matchExpressions:
+                - key: kubernetes.io/metadata.name
+                  operator: NotIn
+                  values:
+                      - rhtap-releng-tenant
```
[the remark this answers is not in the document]
raks-tt: I have added this tenant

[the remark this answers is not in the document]
raks-tt: Wondering why managed-tenant-namespace should be committed? Should we remove it ?

components/cost-management/production/base/kustomization.yaml
```diff
@@ -2,6 +2,7 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
   - ../../base
+  - ../policies/
```
filariow, line 5: as we disabled that in production with #6486, we need to revert this change

```suggestion
```

components/cost-management/staging/policies/update_namespace_labels.yaml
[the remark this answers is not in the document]
raks-tt: So the one that can fail should have `Enforce` and the one that should fail(propogating) should set it as "Audit" is that correct to assume. I made the changes with this assumption (I will have to update the tests).

components/cost-management/staging/policies/propogate_namespace_labels.yaml
```diff
@@ -0,0 +1,55 @@
...
+    policies.kyverno.io/title: Propagate Cost-Center from Namespace
+    policies.kyverno.io/category: Cost Management
+    policies.kyverno.io/severity: low
+    policies.kyverno.io/subject: Pod
+    policies.kyverno.io/description: >-
+      This policy ensures that: 
+      (1) The `cost-center` label is propagated from non-managed tenant namespaces to pods.
+      (2) The `insights_cost_management_optimizations` label is set to `true` on Pods to which the `cost-center` label was propagated.
+spec:
+  validationFailureAction: Audit
```
filariow, line 16: this is not needed here as the policy only contains a `mutate` rule. we need to use `spec.webhookConfiguration.failurePolicy: Ignore` instead

```suggestion
  webhookConfiguration:
    failurePolicy: Ignore
```

components/cost-management/staging/policies/propogate_namespace_labels.yaml
```diff
@@ -0,0 +1,55 @@
...
+  name: propagate-cost-management-labels
+  annotations:
+    policies.kyverno.io/title: Propagate Cost-Center from Namespace
+    policies.kyverno.io/category: Cost Management
+    policies.kyverno.io/severity: low
+    policies.kyverno.io/subject: Pod
+    policies.kyverno.io/description: >-
+      This policy ensures that: 
+      (1) The `cost-center` label is propagated from non-managed tenant namespaces to pods.
+      (2) The `insights_cost_management_optimizations` label is set to `true` on Pods to which the `cost-center` label was propagated.
```
filariow, line 14: as we are allowing the policy to fail without preventing the pod from being created, I'd suggest to change this `ensure` with something like `tries to`. 
```suggestion
      This policy tries to that: 
      (1) Propagate the `cost-center` label from non-managed tenant namespaces to pods.
      (2) Set the `insights_cost_management_optimizations` label is set to `true` on Pods to which the `cost-center` label was propagated.
      If the webhook fails, the pod will be created without the labels.
```

components/cost-management/staging/policies/kustomization.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- validate_namespace_labels.yaml
+- propogate_namespace_labels.yaml
```
filariow, line 5: typo 
```suggestion
- propagate_namespace_labels.yaml
```

components/cost-management/staging/policies/propogate_namespace_labels.yaml
raks-tt: Sure. Changed this

components/cost-management/staging/policies/.chainsaw-test/chainsaw-test.yaml
```diff
@@ -0,0 +1,512 @@
...
+              spec:
+                replicas: 0
+
+    - name: Wait for Kyverno admission controller to be down
+      try:
+        - script:
+            content: |
+              echo "Waiting for Kyverno admission controller to scale down..."
+              sleep 15
+              kubectl wait --for=delete pod -l app.kubernetes.io/name=kyverno -n konflux-kyverno --timeout=60s || true
```
filariow, line 446: maybe we can use the [wait helper](https://kyverno.github.io/chainsaw/0.2.3/operations/helpers/wait/) here.

Also
* the `|| true` at the end is very strange to me
* is the `sleep 15` needed?

components/cost-management/staging/policies/.chainsaw-test/chainsaw-test.yaml
```diff
@@ -0,0 +1,512 @@
...
+              INSIGHTS_LABEL=$(kubectl get pod demo-pod -n tenant-with-cost-center -o jsonpath='{.metadata.labels.insights_cost_management_optimizations}' 2>/dev/null || echo "")
+              if [ "$INSIGHTS_LABEL" = "true" ]; then
+                echo "FAIL: Pod should not have insights_cost_management_optimizations label when webhooks are disabled. Found: $INSIGHTS_LABEL"
+                kubectl get pod demo-pod -n tenant-with-cost-center -o yaml
+                exit 1
+              fi
+
+              echo "SUCCESS: Pod created without cost management labels as expected"
+              echo "Cost-center label: '$COST_CENTER_LABEL'"
+              echo "Insights label: '$INSIGHTS_LABEL'"
```
filariow, line 494: can't we just use an `assert` here with the manifest of the pod?

components/cost-management/staging/policies/.chainsaw-test/chainsaw-test.yaml
```diff
@@ -0,0 +1,512 @@
...
+                namespace: konflux-kyverno
+              spec:
+                replicas: 1
+    - name: Verify Kyverno is working again (optional validation)
+      try:
+        - script:
+            content: |
+              echo "Waiting for Kyverno to recover..."
+              sleep 5
+              echo "Kyverno should be operational again"
```
filariow, line 512: this is not verifying that kyverno is back up

filariow (changes requested):

components/cost-management/production/base/kustomization.yaml
```diff
@@ -2,6 +2,7 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
   - ../../base
+  # - ../policies/
```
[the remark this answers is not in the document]
raks-tt: https://github.com/redhat-appstudio/infra-deployments/pull/6649

filariow (approved):
/lgtm
/approve

mftb (approved):
/lgtm
/approve

PR #6476, opened 2025-05-29
Ensure directory is present in new-cluster playbook

ralphbean:
This should fix an error reported by Manish.


ralphbean:
This only affected MPC because it is the only one with a subdir that looks like "production-downstream" instead of just "production".

PR #6479, opened 2025-05-29
Grant builder and releaser SAs rights to create secrets

ralphbean:
The need here is for an external CI provider that wants to initiate a build or release and needs to supply a short-term secret to that run. We want this SA to be able to create or update a previously created secret, but not get or read that secrets or any other secrets in the namespace.

### DIFF
--- a/components/konflux-rbac/production/base/konflux-builder-bot-actions.yaml
+++ b/components/konflux-rbac/production/base/konflux-builder-bot-actions.yaml
@@ -15,4 +15,9 @@ rules:
   - update
   - patch
   - delete
-
+- apiGroups:
+  - ""
+  resources:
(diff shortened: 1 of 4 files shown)


filariow:
Some considerations on the current proposal:

1. only admins will be able to assign this role as they are the only ones with these permissions.
2. `list` and `get` provides the same level of detail, so we can't give `list` neither (cf. https://kubernetes.io/docs/reference/access-authn-authz/authorization/#request-verb-resource)
3. we shouldn't give resource-unscoped `delete`, `update`, and `patch` as users modify or delete well-known or guessed secrets.
4. Kubernetes RBAC can't scope `create` to a resource name, so users will be able to create secrets with any name.

PR #6480, opened 2025-05-29
feat(KONFLUX-4930) Add service namespaces to k8s metrics rules

jdcasey:
This adds metric export match rules for kubernetes pod/replica counts for known service namespaces. It also clears out some regex aggregations and simplifies expressions and adds comments to make it easier to find service-related metrics.

TODO: I have been minimally invasive with regard to existing match rules. We need to refine those rules further and look at places where we can group service-related things together. Then, we should consider which of those are more generally useful, and apply the same approach as we have here for those other metrics.


jdcasey:
Closing. I'll open a new PR to try to get the workflows unjammed.

PR #6484, opened 2025-05-29
Revert "leverage etcd-shield to reject PipelineRuns in rh01"

filariow:
Reverts redhat-appstudio/infra-deployments#6483

As the issue with rh01 is gone, we want to resume the normal operativity of the cluster, allowing again creation of PipelineRuns.


filariow:
/hold

⚠️ Please merge this when the rh01 is stable again.

gbenhaim (approved):

PR #6486, opened 2025-05-29
disable cost-management policies

filariow:
To reduce pressure on rh01, we want to remove this non-core policy

### DIFF
--- a/components/cost-management/base/kustomization.yaml
+++ b/components/cost-management/base/kustomization.yaml
@@ -5,7 +5,7 @@ resources:
   - costmanagement-metrics-config.yaml
   - external-service-account-secret.yaml
   - rbac
-  - ../policies/
+  # - ../policies/
 
 commonAnnotations:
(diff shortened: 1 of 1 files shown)


filariow:
no more needed

gbenhaim:
Kyverno's admission controller use leader election - https://kyverno.io/docs/high-availability/#admission-controller-1

We recently see the controller restart because leader election fails while etcd is doing fragmentation (this is expected https://docs.redhat.com/en/documentation/openshift_container_platform/4.15/html/scalability_and_performance/recommended-performance-and-scalability-practices-2#etcd-defrag_recommended-etcd-practices)

Users reports that pods for taskruns are failed to get created because the Kyverno mutating webhook is not responding.

I'm going to disable the policy which mutates pods until we will sort out this issue.

gbenhaim:
no need to wait for the openshift-ci job since the cost-management app isn't deployed as part of it https://github.com/redhat-appstudio/infra-deployments/blob/main/argo-cd-apps/overlays/development/delete-applications.yaml#L125C9-L125C24

PR #6491, opened 2025-05-30
kyverno disable scale-to-zero job in stg

filariow:
The scale-to-zero job is not adding value in a ArgoCD managed environment


hugares (approved):
/lgtm

PR #6499, opened 2025-05-30
chore: update integration service in all environments

14rcole:
Pushes a patch fixing a race condition in snapshot creation that is causing a CrashLoopBackoff


jencull (approved):
lgtm

PR #6519, opened 2025-06-02
Add cost management operator to kflux-prd-rh03 cluster

raks-tt:


mike-kingsbury (approved):
LGTM.  Wish there was a way to inherit a list & iterate over it and not have to make these sorts of changes

PR #6522, opened 2025-06-03
fix: update development pipelines sha

flacatus:


mshaposhnik (approved):

components/pipeline-service/development/main-pipeline-service-configuration.yaml
```diff
@@ -2101,7 +2101,7 @@ metadata:
   namespace: openshift-marketplace
 spec:
   displayName: custom-operators
-  image: quay.io/openshift-pipeline/pipelines-index-4.15@sha256:bf032558eeb696ed351cb9c00503f6d415e4764639ef23295abd51bad566a966
+  image: quay.io/openshift-pipeline/pipelines-index-4.15:next
```
gbenhaim, line 2104: I prefer to not use a floating tag since it makes it harder to understand why errors happen (we don't know what the content it points to changes).

gbenhaim, line 2104: I prefer to not use a floating tag since it makes it harder to understand why errors happen (we don't know what the content it points to changes).

pramodbindal: may be an alternative approach we can have it to add some custom tag on this image once nightly is qualified by QA 
something like  `quay.io/openshift-pipeline/pipelines-index-4.15:next-tested`

flacatus:
old image was restored from quay

PR #6535, opened 2025-06-03
disable `mutate-namespace-enforce-label` policy in staging

filariow:
At this point we migrated all the namespace to the new label. We are no more expecting any Namespace to be created with the `toolchain.dev.openshift.com` or the `konflux.ci` labels.


sadlerap (approved):
/lgtm

PR #6536, opened 2025-06-03
give more cpu to Kyverno background on rh01

filariow:
Kyverno background is using all the CPU time we provided.
Let's give it more compute power


sadlerap (approved):
/lgtm

sadlerap (changes requested):
This is the admission controller

sadlerap (approved):
/lgtm

PR #6551, opened 2025-06-04
changes to bootstrap to accomodate the pulpaccessrequets

YasenT:
The current problem with pulpaccessrequets is that when you try to use the new controller from insdie a pipeline, the automatically generated user which is namespace based, doesn't have access to the crd
And looking at the rbac, I believe that a change to the namespace bootstrap will help with this


ralphbean:
No, this isn't the way to go for two reasons.

First, we're retiring the `appstudio-pipeline` service account. It will be decommissioned by the end of June and gone.

Second, I don't think that creating PARs from inside the pipeline is the right pattern. Instead, users create PARs when they onboard and are setting up their Components and ImageRepositories, which provisions the secret. Their pipeline should mount that secret.

PR #6555, opened 2025-06-04
fix group propagation for nslister in stg

filariow:
namespace-lister is not showing the correct results for group `system:authenticated`.


sadlerap (approved):
/lgtm

PR #6579, opened 2025-06-05
Set number of OSP buckets to 4

hugares:
Config is set to 8 but ArgoCD is showing TektonConfig out of sync in all the production cluster, the live TektonConfig only has 4 buckets


aThorp96 (approved):
/lgtm

PR #6581, opened 2025-06-05
Fix MPC deployment on kflux-rhel-p01 cluster

hugares:
The manifest was overwriting the version of MPC for a version that no longer exists. Remove all customization and use base folder, same as other internal clusters

[KFLUXINFRA-1636](https://issues.redhat.com//browse/KFLUXINFRA-1636)


sadlerap (approved):
/lgtm

PR #6582, opened 2025-06-05
Move secrets file to each internal MPC cluster folder

hugares:
Having the secrets in the base is making things more complicated as some clusters do not need all the secrets and some others, secrets must be patched to have right path in vault.

As first step, duplicate the secrets in each cluster and adjust accordingly. We have a task to standardize the MPC secrets name, one this is done we will be able to remove duplication and created them in the base overlay.


sadlerap (approved):
/lgtm

hugares:
this is not covered by e2e, merging now

PR #6583, opened 2025-06-05
Revert "give unlimited cpu to Kyverno background on rh01 (#6537)"

sadlerap:
Kyverno no longer needs unlimited cpu resources, so add back the limits we had previously.

This reverts commit a5666c8d46d48b7dcbbbb6474ff87f29cac2b3c5.


filariow:
we are fine with no limits on cpu

PR #6577, opened 2025-06-05
increase QPS and Burst for Kyverno in prod rh01

filariow:
Kyverno is failing as it's throttled while renovating the lease. Bumping QPS and Burst could help. Default values were 300 each.

```yaml
      containers:
        - args:
            [ ... REDACTED ... ]
            - --enablePolicyException=false
            - --enableReporting=validate,mutate,mutateExisting,imageVerify,generate
            - --clientRateLimitQPS=1000
            - --clientRateLimitBurst=1000
          env:
            - name: KYVERNO_SERVICEACCOUNT_NAME
```

Useful links:
* https://kyverno.io/docs/installation/customization/#container-flags

Signed-off-by: Francesco Ilario <filario@redhat.com>

PR #6586, opened 2025-06-06
Revert "exclude kubesaw managed namespace from tenant bootstrapping in stg"

filariow:
Reverts redhat-appstudio/infra-deployments#6567

It seems KubeSaw and Kyverno are not competing over the same resources.
redhat-appstudio/infra-deployments#6577 fixed the issue in prod.


hugares (approved):
/lgtm

PR #6592, opened 2025-06-06
remove cpu limits for nslister

filariow:
we don't need to set limit for cpu, we just require accurate request.


hugares (approved):
/lgtm

sadlerap (approved):
/lgtm

PR #6595, opened 2025-06-06
kyverno remove cpu limits in prod

filariow:
removes cpu limits as they are not needed


sadlerap (approved):
/lgtm

PR #6606, opened 2025-06-06
Remove s390x and ppc VMs from p02

hugares:
The us-south datacenter is full for s390x. Remove 6 VMs from p02 to move them to the rhel cluster. Since 6 VMs can do 24 concurrent builds, reduce the power accordingly since there is no point of being able to do more ppc than s390x

[KFLUXINFRA-1639](https://issues.redhat.com//browse/KFLUXINFRA-1639)


hugares:
this is not covered in e2e test, merging now

PR #6614, opened 2025-06-09
Remove the deletion of nvme-storage app from downstream overlays

gbenhaim:
It's already getting deleted in the parent overlay so if we
    try to delete it again kustomize build fails.


manish-jangra (approved):
/lgtm

PR #6628, opened 2025-06-09
Configure kflux-rhel-p01 s390x and ppc VMs

hugares:
Add all 8 ppc machines and only 3 s390x machines. We are missing 13 s390x machines that will be added once we sort it out with IBM/IT, we might need to move to another datacenter.

A follow up PR will be done with remaining machines.

[KFLUXINFRA-1638](https://issues.redhat.com//browse/KFLUXINFRA-1638)
[KFLUXINFRA-1639](https://issues.redhat.com//browse/KFLUXINFRA-1639)


manish-jangra (approved):
/approve
/lgtm

PR #6630, opened 2025-06-10
feat(KFLUXUI-504): Add banner content yaml and its kustomization files

testcara:
[KFLUXUI-500](https://issues.redhat.com/browse/KFLUXUI-500) aims to provide konflux banner to users.

The PR does:
supporting the single manual banner:
- add the banner-content.yaml for each cluster under konflux-info
- update its kustomize to ensure the its confimap could be generated
- add README.md and github workflow validation support for banner content including schema & safe content
supporting the multiple auto generated alerts:
- add the auto-alerts directory and its fake kustomize 
- update genearl kustomize for auto-alerts
- update README.md to explain how we can enjoy the auto-alerts.

Here is one video about how UI integrate with them:
https://drive.google.com/file/d/17X3qkgLD7aXWq_wnuteAWUu3WE7_J2M5/view

### DIFF
--- a/components/konflux-info/production/kflux-ocp-p01/banner-content.yaml
+++ b/components/konflux-info/production/kflux-ocp-p01/banner-content.yaml
@@ -1,0 +1,5 @@
+enable: false
+type: warning # options: info, warning, danger
+summary: The Konflux platform will undergo scheduled maintenance tonight. Temporary service interruptions may occur.
+startTime: "2025-05-15T21:00:00Z"
+endTime: "2025-05-15T23:00:00Z"
--- a/components/konflux-info/production/kflux-prd-rh02/auto-alerts/auto-alert-1.yaml
+++ b/components/konflux-info/production/kflux-prd-rh02/auto-alerts/auto-alert-1.yaml
@@ -1,0 +1,12 @@
+apiVersion: v1
+data:
+  auto-alert-content.yaml: |-
+    enable: false
+    type: warning
+    summary: pipeline run cannot run
+kind: ConfigMap
+metadata:
+  labels:
+    konflux-auto-alert: "true"
+  name: konflux-autoalert-configmap-1
+  namespace: konflux-info
--- a/.github/workflows/validate-banner.yaml
+++ b/.github/workflows/validate-banner.yaml
@@ -1,0 +1,38 @@
+name: Validate Banners
+
+on:
+  pull_request:
+    paths:
+      - 'components/konflux-info/**/**/banner-content.yaml'
+      - 'components/konflux-info/banner-schema.json'
+
+env:
+  BANNER_DIR: components/konflux-info
+  SCHEMA_FILE: components/konflux-info/banner-schema.json
+  TMP_DIR: .tmp/banners
+
+jobs:
+  validate-banner:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+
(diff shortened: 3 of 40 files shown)


gbenhaim (changes requested):
Please move the content of https://github.com/testcara/konflux-banner into a directory in this repo (under the components directory)

Thanks !

gbenhaim:
I really liked the implementation because it will also be possible to generate those confimaps with the banner details dynamically based on alerts.

I left a comment above about the way it should be deployed.

Thank you for working on this.

gbenhaim:
I have a question, does each banner stored in its own configmap? I would like to also generate banners dynamically, it means that we would need the UI to read multiple configmaps with the banner details from the cluster. 
Is it possible?

cc @sahil143

testcara:
@gbenhaim hi, Gal, i refreshed the PR, could you help to review again? BTW, the description includes what the PR does or not covered, hope it makes the review a bit easier. Thank you for the help.

components/konflux-info/production/kflux-ocp-p01/banner-content.yaml
```diff
@@ -0,0 +1,13 @@
+enable: false
```
testcara, line 1: If UX is able to provide the multiple banner solution, how about i improve the single banner yaml to banner content list yaml, then we do not need to create multiple configmaps.

gbenhaim: we need multiple configmap because some will get created by humans via gitops and some via automation directly in the cluster. we don't want gitops and controllers to edit the same configmap.

gbenhaim (changes requested):

components/konflux-info/production/kflux-ocp-p01/banner-content.yaml
testcara: @gbenhaim , Hi. Gal.
Konflux UI has a new issue related to SLO dashboards and SOPs: [KFLUXUI-456](https://issues.redhat.com/browse/KFLUXUI-456). I’m considering whether auto-detected alerts should be included in its scope.

The banner appears at the top of the entire application and is visible to all users. Such banners need to be manageable, reliable, and clear for everyone.

Auto-detected alerts are unpredictable in timing, content, volume, and duration. This unpredictability risks causing unnecessary broad concerns and may undermine user trust.

Therefore, I believe manually curated (GitOps) banners and auto-detected alerts should be handled separately. In particular, GitOps banners work should not be blocked auto alerts ideas, as the UI would treat these two kinds of messages differently—regardless of whether it’s handled by KFLUXUI-456 or another issue.

What is your idea?

testcara:
@gbenhaim  i tried to support the two kinds of banners now, see the https://drive.google.com/file/d/17X3qkgLD7aXWq_wnuteAWUu3WE7_J2M5/view. looking forward to any feedback. thank you.

components/konflux-info/production/kflux-prd-rh02/auto-alerts/auto-alert-1.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
+data:
+  auto-alert-content.yaml: |-
+    enable: false
+    type: warning
+    summary: pipeline run cannot run
+kind: ConfigMap
+metadata:
+  labels:
+    konflux-auto-alert: "true"
```
testcara, line 10: i would change the label to recommended way as 'konflux.auto.alert'

PR #6652, opened 2025-06-10
kyverno: increase leader election retry period in prod rh01

filariow:
In prod rh01 kyverno is restarting because it can not renovate the lease.
We want to have leases that lasts at least `25 seconds`.

Kyverno accepts the flag `leaderElectionRetryPeriod` defaulted to `2s` (cf. https://kyverno.io/docs/installation/customization/#container-flags) and it's multiplying this factor for `6` to calculate the `LeaseDuration` value.

Bumping the `leaderElectionRetryPeriod` to `5s` will make the lease last at least `30s`.


filariow:
/hold

wait for #6650

enkeefe00 (approved):

PR #6664, opened 2025-06-11
cost-management validate policy as ValidatingAdmissionPolicy

filariow:
Starting from Kubernetes 1.30 / OCP 4.17 we can use the kubernetes native ValidatingAdmissionPolicy for some validating policy.

> Validating admission policies offer a declarative, in-process alternative to validating admission webhooks.

This commit rewrites the cost-management validate namespace policy.

* https://docs.okd.io/4.17/rest_api/extension_apis/validatingadmissionpolicy-admissionregistration-k8s-io-v1.html
* https://kubernetes.io/docs/reference/access-authn-authz/validating-admission-policy/


raks-tt:
@filariow - Does this mean we don't need a `ClusterPolicy` from Kyverno and you would recommend this should be used instead?

PR #6687, opened 2025-06-12
Increase updateRequestThreshold in stg-rh01

gbenhaim:
This is required in order to let Kyverno process the policy which create Kueue LocalQueue in each tenant namespace. Without it the update is blocked because of the threshold.


filariow:
@amisstea fyi

PR #6697, opened 2025-06-12
KubeArchive: fix always out of sync

rh-hemartin:
ArgoCD shows that KubeArchive is always out of sync because the DB secrets keeps being changed by the ExternalSecrets operator and needs to be reestablished to its original content. This is not the case, so we will ignore the differences from now on.

### DIFF
--- a/argo-cd-apps/base/member/infra-deployments/kubearchive/kubearchive.yaml
+++ b/argo-cd-apps/base/member/infra-deployments/kubearchive/kubearchive.yaml
@@ -49,6 +49,11 @@ spec:
         # not specifcy namespace
         namespace: product-kubearchive
         server: "{{server}}"
+      ignoreDifferences:
+        # Secret get modified by ExternalSecrets operator
+        - group: ""
+          kind: Secret
+          name: kubearchive-database-credentials
       syncPolicy:
         automated:
           prune: true


hugares:
> > Why is ArgoCD tracking the secret in the first place? We should only have ExternalSecret in ArgoCD and the ES will create the secret. Are you creating the secret in kubearchive manifest?
> 
> The problem is a secret named `kubearchive-database-credentials` is coming from the kubearchive manifest (https://github.com/kubearchive/kubearchive/releases/download/v1.0.1/kubearchive.yaml) but in this Konflux deployment, we want the secret to be created by ExternalSecret so all we have to do is delete this Secret using Kustomize.

here is what I suggest, instead of telling ArgoCD to ignore:
```
git diff
diff --git a/components/kubearchive/production/stone-prod-p02/kustomization.yaml b/components/kubearchive/production/stone-prod-p02/kustomization.yaml
index e6c20baf..8016cd83 100644
--- a/components/kubearchive/production/stone-prod-p02/kustomization.yaml
+++ b/components/kubearchive/production/stone-prod-p02/kustomization.yaml
@@ -183,3 +183,11 @@ patches:
       metadata:
         name: "kubearchive-operator-certificate"
         namespace: kubearchive
+  # We don't need this Secret as it will be create by an ExternalSecret
+  - patch: |-
+      $patch: delete
+      apiVersion: v1
+      kind: Secret
+      metadata:
+        name: kubearchive-database-credentials
+        namespace: kubearchive
```

hugares (approved):
/lgtm

PR #6717, opened 2025-06-13
add deployment visibility to konflux-info

marcin-michal:
related to [KONFLUX-8558](https://issues.redhat.com/browse/KONFLUX-8558), it is beneficial for the UI to be able to differentiate, if the current deployment is public or private

### DIFF
--- a/components/konflux-info/production/kflux-ocp-p01/info.json
+++ b/components/konflux-info/production/kflux-ocp-p01/info.json
@@ -49,5 +49,6 @@
                 "name": "konflux-contributor-user-actions"
             }
         }
-    ]
+    ],
+    "visibility": "private"
 }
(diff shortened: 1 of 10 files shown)


gbenhaim:
What is the reason th ui needs to know this detail?

marcin-michal:
> What is the reason th ui needs to know this detail?

On the namespace list page, there is a link to the internal documentation for instructions on how to create a new namespace. This should not be displayed on the public instances as it leads to possibly inaccessible places, as described in [KONFLUX-8558](https://issues.redhat.com/browse/KONFLUX-8558)

sahil143:
@gbenhaim For now we can proceed with just setting up the visibility in the config map, this is blocking a bug resolution.  We can create a feature request to have docs direct links as part of config maps. Currently, there is only link (namespace documentation) in UI that's pointing to internal documentation.

PR #6723, opened 2025-06-13
complete removal of mutate-namespace-enforce-label policy

filariow:
we don't need this anymore as onboarding is managed just by ArgoCD nowadays


amisstea (approved):
/lgtm

PR #6714, opened 2025-06-13
introducing the policies component in stg and dev

filariow:
we want to migrate all the Kyverno ClusterPolicies to a dedicated component

Signed-off-by: Francesco Ilario <filario@redhat.com>

PR #6724, opened 2025-06-13
migrate kubearchive policies to policies component in stg

filariow:
Requires:

* [ ] https://github.com/redhat-appstudio/infra-deployments/pull/6714


rh-hemartin:
Hey! Does this force KubeArchive to be installed in all staging clusters? We are not installed on p01 and we don't have plans to yet.

PR #6726, opened 2025-06-13
Automate adding NAT Gateway IPs to smee allow list

enkeefe00:
This adds a task to communicate with AWS and get the cluster VPC's NAT Gateway IPs and adds some tasks to the main playbook to add the retrieved IPs if they do not already exist in the the cluster's smee IP allow list.

Assisted-by: Gemini


hack/new-cluster/playbook.yaml
```diff
@@ -139,6 +140,36 @@
...
+      include_tasks: tasks/aws/smee-ips.yaml
+    
+    - name: Get current content of {{ env }} smee IP allow list
+      # If the path does not exist, yq will return an empty string to stdout.
+      command: yq '.[0].value' {{ smee_ip_allow_list_file }}
+      register: current_ip_allow_list
+      changed_when: false # This task only reads data, it doesn't change anything.
+
+    - name: Add the NAT Gateway Public IPs to the {{ env }} IP allow list
+      # This task executes only if the current content in the file
```
ralphbean, line 155:
```suggestion
      tags: [smee-ips]
      # This task executes only if the current content in the file
```

PR #6764, opened 2025-06-16
feat(PVO11Y-4814): Promote Kanary exporter to a production cluster

FaisalAl-Rayes:
This is meant for:
- Publishing metrics to `rhtap-observatorium-prod` datasource which feeds the SLO alerts routing to #konflux-slo-alerts on slack

related issue: [PVO11Y-4814](https://issues.redhat.com/browse/PVO11Y-4814)


kubasikus (approved):
/approve
/lgtm

filariow (commented):
usually we don't have components in production without having them also in staging. Is there a specific need for this one to be removed from staging and installed in production?

kubasikus (approved):

kubasikus:
@filariow We discussed some more with @FaisalAl-Rayes and @mftb and will go for stage and prod deployments. Please PTAL.

mftb (approved):
/lgtm
/approve

argo-cd-apps/base/monitoring-workload-kanary/monitoring-workload-kanary.yaml
filariow, line 1: I suspect this will match `stone-prod-p01`, `kflux-ocp-p01`, and `kflux-rhel-p01` too and install there what's under `components/monitoring/kanary/production/base`? 

@gbenhaim @hugares am I right?

FaisalAl-Rayes: @filariow thats a very good point! Thank you for noticing.
I will move the exporter from the base/kustomization.yaml to <environment>/<cluster-name>/kustomization.yaml since we are not aiming for multi cluster deployment (per environment) as of now.

filariow: This way we'll still install the RBAC and the ExternalSecret in the other production clusters.
Should we move also those resoruces to `components/monitoring/kanary/${environment}/stone-prod-p02/kustomization.yaml` and leave `production/base` with just a dummy resource (needed not to fail), like the Namespace?

FaisalAl-Rayes: @kubasikus what do you think?

I remember that before we were thinking that it should still not be an issue to have that deployed in clusters other than our target cluster.

Having it stay they way it is will make it easier for us in the future since we would need to still deploy whatever is in base to all the clusters in a given environment.

kubasikus: @FaisalAl-Rayes I think it makes sense to only deploy the resources to cluster where we actually need them. We can always move the resources to the base if we decide to deploy across multiple clusters in the future.

kubasikus: @filariow Can you take another look?

filariow: Looks way better to me. Thanks for addressing my comments.

dirgim (approved):
/lgtm
/approve

PR #6786, opened 2025-06-18
fix(STONEINTG-1259): disable integration leader election

dirgim:
* Disable the leader election for the integration service controller manager in staging/development
* The leader election is not needed for the integration service at this point
* The etcd defragmentation is causing pods to restart when failing to renew the lease


sonam1412 (approved):
lgtm

jsztuka (approved):
lgtm

PR #6788, opened 2025-06-18
Update the Tekton Results watcher timeouts

enarha:
More information about the change:
https://github.com/redhat-appstudio/infra-deployments/pull/6779 We start with single production cluster, before updating the rest.

The `enable-param-enum: true` change is unrelated, but since the kflux-rhel-p01 cluster was out of sync with the base, it was now added to the generated diff and all clusters brought to sync with base.


khrm (commented):
/approve
/lgtm

PR #6792, opened 2025-06-19
scrape kyverno metrics in dev and stage

filariow:
Kyverno doesn't seem to support HTTPS on the metrics server natively even though the Helm chart allows to configure the ServiceMonitor for that.
Further investigation is needed, deferring it to another PR.


hugares (approved):
/lgtm

PR #6794, opened 2025-06-19
KubeArchive: ignoreDifferences in ClusterRoleBinding

rh-hemartin:
This PR aims to ignoreDifferences of a CRB that KubeArchive modifies in its activity. ArgoCD is trying to turn it back to the original state, which will make KubeArchive malfunction. This will be applied to all installations, which makes sense.


rh-hemartin:
@hugares could you take a look?

hugares (approved):
/lgtm

PR #6808, opened 2025-06-20
KubeArchive: changing commonAnnotations

rh-hemartin:
This PR moves the SkipDryRunOnMissingResources to the application set so we can remove commonAnnotations and then use specific annotations for certain resources. The Job we use to update the database schema needs specific annotations.


maruiz93 (approved):

rh-hemartin:
@manish-jangra could you take a look?

PR #6812, opened 2025-06-20
enable multi-repo Github App token scoping

aThorp96:
Disabling this setting allows users to enumerate additional private repositories for which their Github Token will be scoped. Importantly, the additional private repositories can only be included in the token's scope if both repositories have the same Github App installed and the their respective Repository CRs exist in the same namespace, which is in line with Konflux's security model. 

Part of https://issues.redhat.com/browse/KONFLUX-8780


hugares (approved):
/lgtm

PR #6814, opened 2025-06-20
Update Pipelines Production to be in line with staging and devopment

aThorp96:
The `quay.io/openshift-pipeline/pipelines-index-4.15@sha256:a1f651150d5e361d64249d8b38ed21c5bb1e623d2d776ecec2a0ad2633fdaf63` index has been on staging since June 10


enarha (approved):
/lgtm

PR #6834, opened 2025-06-23
etcd-shield: fix metrics in staging

sadlerap:
Metrics for etcd-shield were misconfigured in two ways.  First, the labels selectors were incorrect and not applied to the relevant service. Applying the correct labels and label selectors allows the servicemonitor to pick up the correct metrics endpoints.  Second, because etcd-shield is run not as a user workload but as a core service, the prometheus-k8s service account in the namespace `openshift-monitoring` needs to be given appropriate permissions to look at resources in the `etcd-shield` namespace.  This is fixed by adding a role and a rolebinding.

### DIFF
--- a/components/etcd-shield/base/metrics/metrics-service.yaml
+++ b/components/etcd-shield/base/metrics/metrics-service.yaml
@@ -1,6 +1,8 @@
 apiVersion: v1
 kind: Service
 metadata:
+  labels:
+    app: etcd-shield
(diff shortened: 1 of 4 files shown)


sadlerap:
/retest

@filariow do you have any ideas why e2e tests might be failing?  None of the components that fail are related to `etcd-shield`.

hugares (approved):
/lgtm

PR #6839, opened 2025-06-24
feat(KFLUXVNGD-331): Smee server liveness probe

ifireball:
Add sidecar-based full-roundtrip liveness probe to the Smee server.


components/smee/staging/deployment.yaml
```diff
@@ -0,0 +1,118 @@
...
+              type: RuntimeDefault
+            readOnlyRootFilesystem: true
+            runAsNonRoot: true
+          resources:
+            limits:
+              cpu: 1
+              memory: 750Mi
+            requests:
+              cpu: 1
+              memory: 750Mi
```
yftacherzog, line 73: This client is going to serve 2 events once every 30 seconds. I think we can try using less resources.

components/smee/staging/deployment.yaml
```diff
@@ -0,0 +1,118 @@
...
+            readOnlyRootFilesystem: true
+            runAsNonRoot: true
+          resources:
+            limits:
+              cpu: 1
+              memory: 256Mi
+            requests:
+              cpu: 1
+              memory: 256Mi
+        - image: "ghcr.io/chmouel/gosmee:v0.26.1"
```
yftacherzog, line 51: We need this one downgraded as well. Otherwise, it will refuse working with the server.

components/smee/staging/deployment.yaml
yftacherzog: In konflux-ci/konflux-ci we have this
```
          resources:
            limits:
              cpu: 100m
              memory: 32Mi
            requests:
              cpu: 10m
              memory: 32Mi
```
Probably we can do the same, but make requests equal limits.

components/smee/staging/deployment.yaml
ifireball: will do

components/smee/staging/deployment.yaml
```diff
@@ -0,0 +1,118 @@
...
+  selector:
+    matchLabels:
+      app: gosmee
+  template:
+    metadata:
+      labels:
+        app: gosmee
+    spec:
+      containers:
+        - image: "ghcr.io/chmouel/gosmee:v0.20.2"
```
avi-biton, line 21: Don't we want to upgrade the server to the latest version?

PR #6844, opened 2025-06-24
fix: correct memory request format in Smee deployment configuration

ifireball:
Updated the memory request in the Smee server deployment from an incorrect value to the correct format, ensuring proper resource allocation.

Signed-off-by: Barak Korren <bkorren@redhat.com>


yftacherzog (approved):
/lgtm
/approve

PR #6853, opened 2025-06-25
chore: change liveness probe on smee sidecar

yftacherzog:
This should prevent a condition in which smee and its sidecar crashloop forever.


avi-biton (approved):
/lgtm

PR #6857, opened 2025-06-25
Update store deadline and pruner time values

enarha:
On the busiest production cluster, we see pipelineruns and more rarely taskruns are not reconciled timely resulting objects removed from the cluster before they are actually stored in the DB. With the updated values we wait longer before giving up and force remove the objects.
In addition we also scale up the number of replicas of the watcher controller. Once we see the effect of it, we can lower back the time values.


khrm (commented):
/approve
/lgtm

hugares (approved):
/lgtm

PR #6871, opened 2025-06-25
Enable kubearchive access from nginx in prod if configured

maruiz93:
I've included the same config we added in staging to make kubearchive accessible withink konflux-ui if the KUBEARCHIVE_URL parameter is set up.

Related: https://issues.redhat.com/browse/KFLUXINFRA-1629

Currently it's only set up for `stone-prod-p02`. Related: https://issues.redhat.com/browse/KFLUXINFRA-1778

### DIFF
--- a/components/konflux-ui/production/base/proxy/proxy.yaml
+++ b/components/konflux-ui/production/base/proxy/proxy.yaml
@@ -81,6 +81,18 @@ spec:
               "proxy_pass ${TEKTON_RESULTS_URL:?tekton results url must be provided};" \
               > /mnt/nginx-generated-config/tekton-results.conf
 
+            if [[ "$KUBEARCHIVE_URL" != "" ]]; then
+              echo "location /api/k8s/plugins/kubearchive/ {" > /mnt/nginx-generated-config/kubearchive.conf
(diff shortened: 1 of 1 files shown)


filariow:
/lgtm

/hold
for changes to be deployed in internal staging first

rh-hemartin:
> > This change is already on internal staging, without it the proxy would fail to start, as KubeArchive is not deployed there.
> 
> Not sure I got it correctly. Will we have internal stage and internal production in the same state after we merge this PR? If no, I'd suggest to align internal staging before merging this one

Please, don't block this PR until that happens, it could be a long time. We need a DB there. I created https://issues.redhat.com/browse/KFLUXINFRA-1842 for it.

filariow:
> > > This change is already on internal staging, without it the proxy would fail to start, as KubeArchive is not deployed there.
> > 
> > 
> > Not sure I got it correctly. Will we have internal stage and internal production in the same state after we merge this PR? If no, I'd suggest to align internal staging before merging this one
> 
> Please, don't block this PR until that happens, it could be a long time. We need a DB there. I created https://issues.redhat.com/browse/KFLUXINFRA-1842 for it.

got it, feel free to `unlock` the PR

PR #6874, opened 2025-06-25
Follow-up: Apply Resource Limits for Tekton Components in Production

ab-ghosh:
This PR mirrors the changes introduced in [#6435](https://github.com/redhat-appstudio/infra-deployments/pull/6435) by applying CPU and memory limits for Tekton components in the production environments.


aThorp96:
/lgtm
/hold

We want to make sure we're monitoring this as it deploys. Because these are all new resource limits/requests, and since a misconfiguration here could cause the components to become unschedulable or be repeatedly evicted, let's target merging this during the maintenance window this Friday. 

CC @enarha @hugares

PR #6875, opened 2025-06-25
etcd-shield: use in-memory IPC in staging

sadlerap:
Etcd-shield has two sets of threads:
- One set queries alertmanager for etcd's state.
- One set handles admission requests.

These threads communicated by reading/writing serialized state to a configmap.  However, etcd-shield now supports communicating this state over memory shared between threads.  Configure etcd-shield to use this new IPC mechanism.


filariow (approved):
/lgtm

PR #6892, opened 2025-06-27
migrate kubearchive prod policies to dedicated component

filariow:
The migration is performed in two phases. In this first phase the policies are copied to the `policies` component. In a subsequent PR they will be removed from the kubearchive component.


filariow:
> Shouldn't we disable the policies from the KubeArchive component folder?

#6893 will do this, we can merge it right after this one when we are sure that the policies have been correctly migrated in Argo

PR #6912, opened 2025-06-29
chore: move smee sidecar to file based probe

yftacherzog:
This eliminates the use of GET calls with side effects.


avi-biton (approved):
/lgtm

PR #6941, opened 2025-07-01
update the public part of prod clusters with recent MPC

mshaposhnik:
Apply  fix https://issues.redhat.com/browse/KONFLUX-8777 to public prod


filariow (approved):
/lgtm

PR #6947, opened 2025-07-01
Make appstudio-pipelines-runner role common for all overlays

mshaposhnik:
it is moved outside of sandbox


components/konflux-rbac/production/base/kustomization.yaml
```diff
@@ -1,6 +1,7 @@
 kind: Kustomization
 apiVersion: kustomize.config.k8s.io/v1beta1
 resources:
+  - ../../base
```
filariow, line 4: I'd suggest NOT to share manifests between dev+stage environments and production

```suggestion
```

components/konflux-rbac/production/base/kustomization.yaml
filariow, line 1: I'd suggest to revert this changes and target one environment for each PR. Let's scope this one to dev+stage only and a subsequent one to prod

components/sandbox/toolchain-member-operator/base/rbac/kustomization.yaml
filariow, line 1: I'd suggest doing the removal in a follow up PR so that we have time to double-check the role was correctly added to the `konflux-rbac` application.

filariow (changes requested):
great work, some suggestions on the procedure

PR #6962, opened 2025-07-02
Update PostgreSQL image to 17.5.0 in development overlay

ab-ghosh:
This PR updates the PostgreSQL image version used in the development overlay for the pipeline-service to 17.5.0. The update ensures alignment with the latest supported version in the Bitnami Helm chart (16.7.15)..


enarha:
> /lgtm
> 
> Are there any migrations or changes necessary, since we're updating the development overlay 4 major versions?

Given that's only used in CI, created for each run and removed after, I guess no migration is required. The same question applies to stage and prod though and I think the answer is negative, but I'll let @ab-ghosh to confirm.

PR #6967, opened 2025-07-02
etcd-shield: remove leader election permissions in prod

sadlerap:
Under no circumstances should etcd-shield use leader election.  When a cluster is having etcd capacity issues, the leader won't be able to update its lease, causing etcd-shield's pods to lose leadership.

The current build and configuration of etcd-shield is configured to never use leader election, so it's safe to remove etcd-shield's permissions to use it.


hugares (approved):
/lgtm

PR #6969, opened 2025-07-02
Update store deadline and pruner time values - 2

enarha:
While we are working on multiple optimization to increase the throughput of Tekton Results, to avoid missing PipelineRuns/TaskRuns we need to allow enough time for those objects to be stored before they are pruned from the cluster. At peak times we see reconciliation time taking 3+ hours.

### DIFF
--- a/components/pipeline-service/production/base/main-pipeline-service-configuration.yaml
+++ b/components/pipeline-service/production/base/main-pipeline-service-configuration.yaml
@@ -1504,7 +1504,7 @@ spec:
         - -check_owner=false
         - -completed_run_grace_period=5m
         - -requeue_interval=2m
-        - -store_deadline=90m
+        - -store_deadline=240m
         - -forward_buffer=1m
         - -threadiness=32
         - -logs_api=true
@@ -1863,10 +1863,10 @@ spec:
   profile: all
   pruner:
     disabled: false
-    keep-since: 120
+    keep-since: 240
     resources:
       - pipelinerun
-    schedule: '*/10 * * * *'
+    schedule: '*/30 * * * *'
   result:
     disabled: true
   targetNamespace: openshift-pipelines
--- a/components/pipeline-service/production/kflux-ocp-p01/deploy.yaml
+++ b/components/pipeline-service/production/kflux-ocp-p01/deploy.yaml
@@ -1937,7 +1937,7 @@ spec:
         - -check_owner=false
         - -completed_run_grace_period=5m
         - -requeue_interval=2m
-        - -store_deadline=90m
+        - -store_deadline=240m
         - -forward_buffer=1m
         - -threadiness=32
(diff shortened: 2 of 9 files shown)


components/pipeline-service/production/base/main-pipeline-service-configuration.yaml
```diff
@@ -1504,7 +1504,7 @@ spec:
         - -check_owner=false
         - -completed_run_grace_period=5m
         - -requeue_interval=2m
-        - -store_deadline=90m
+        - -store_deadline=240m
```
aThorp96, line 1507: For a paper trail on why this is necessary: since there is an order of magnitude more TaskRuns than PipelineRuns, TaskRuns may take 60-120 minutes to be reconciled after they're complete, whereas their corresponding PipelineRun may be reconciled in 20-30 minutes after completion. When this happens, the PipelineRun reconciler sees that the PipelineRun has been complete for 30, 60, or 90 minutes but its TaskRuns have not yet been stored. It then deletes the pipelinerun without storing because its passed the "store_deadline". This change allows the PipelineRun reconciler to wait longer for its TaskRuns to be stored/deleted before giving up on the PipelineRun.

aThorp96 (approved):
@enarha should we limit this to one or two clusters first to make sure keeping the PLRs for longer doesn't result in too much etc load?

aThorp96:
Marking as `hold` to prevent auto-merge, regarding my question of a partial rollout

components/pipeline-service/production/base/main-pipeline-service-configuration.yaml
enarha: The store_deadline applies to finished PipelineRuns, which means at that time all TaskRuns should be finished as well. From that point we wait 4 hours for everything to be stored before we delete the objects without a "stored" annotation. Obviously if objects are stored earlier, they will be also pruned earlier.

enarha:
> @enarha should we limit this to one or two clusters first to make sure keeping the PLRs for longer doesn't result in too much etc load?

There isn't much sense IMO to do that on a less busy cluster as we'll only see the positive and eventually negative effect on a busier cluster. So I can start with one of them. This will prolong the process with a few days, but it's possible. My thinking here was that with 2 clusters out of n bearing most of the load, updating all will potentially have extra negative effect on just 1 (given I plan to update a busy one). Of course we can make it even more gradual, 3 step process: less busy -> busy -> all clusters, but given how easy it is to adjust or revert, it should not really be a big issue.
Anyway, I'm OK with either of the three options.

aThorp96:
Makes sense. Is there a way we can monitor the etcd load while this is being rolled out? IIUC recovery from etcd getting too full is not trivial. CC @hugares 

Maybe we can deploy to all clusters during the maintenance window

PR #6974, opened 2025-07-02
Revert "konflux-ci: add dockerhub pull secret (#5421)"

MartinBasti:
This reverts commit be18ad5d4c60aa9e472dc1cfa839ba3b28e88a4b.

This was mainly for Hermeto builds and we don't need dockerhub pull secret anymore for that.


MartinBasti:
I deactivated pull secret token, so if somebody is still using it, we will see soon

rcerven (approved):

PR #6989, opened 2025-07-03
Increase kyverno UpdateRequests in p02

hugares:
Increase to 2000 to unblock things and allow us to investigate why we reached the max, which was 1000.


filariow (approved):
/lgtm

PR #6996, opened 2025-07-03
etcd-shield: add ClusterRole for metrics auth

sadlerap:
The metrics service for etcd-shield needs to be able to create TokenReviews and SubjectAccessReviews to ensure metrics clients have appropriate permissions to view metrics.  Since a RoleBinding is already configured to accept this, add the appropriate ClusterRole to give etcd-shield's service account permissions to create these resources.


meyrevived (approved):
/lgtm

PR #7030, opened 2025-07-07
chore: set GOMEMLIMIT for smee-sidecar in staging

yftacherzog:
It doesn't seem as the sidecar is leaking memory, but at the same time, its container does ask for more memory over time.

Setting GOMEMLIMIT should hopefully make its memory request stabilize.


amisstea (approved):
/lgtm

PR #7032, opened 2025-07-07
Add SA to each namespace for running cron jobs

brianwcook:
This PR supports adding a UI for https://konflux-ci.dev/docs/testing/integration/periodic-integration-tests/ In order for the UI to work an SA with correct permissions to mutate snapshots must exist in the namespace with a predictable name.

### DIFF
--- a/components/policies/staging/base/konflux-rbac/bootstrap-tenant-namespace/bootstrap-tenant-namespace-np-konflux-cron-sa.yaml
+++ b/components/policies/staging/base/konflux-rbac/bootstrap-tenant-namespace/bootstrap-tenant-namespace-np-konflux-cron-sa.yaml
@@ -1,0 +1,66 @@
+apiVersion: kyverno.io/v1
+kind: ClusterPolicy
+metadata:
+  name: init-ns-cron-sa
+spec:
+  generateExisting: true
+  rules:
+    - name: generate-serviceaccount
+      match:
+        any:
+        - resources:
+            kinds:
+            - Namespace
+            selector:
+              matchLabels:
+                konflux-ci.dev/type: tenant
+      generate:
+        kind: ServiceAccount
+        apiVersion: v1
+        name: konflux-cron-sa
+        namespace: '{{request.object.metadata.name}}'
+        synchronize: true
+    - name: generate-snapshot-role
+      match:
+        any:
+        - resources:
+            kinds:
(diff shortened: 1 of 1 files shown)


gbenhaim:
with this permission users can change the snapshot content. are we ok with it?

brianwcook:
Users (maintainers and admins) can already change snapshot content. This gives us a service account that can do it so that we can run ITS jobs with cron.

arewm:
but the ClusterRole for admins has:
https://github.com/redhat-appstudio/infra-deployments/blob/4e205ca7d9aa4cf164a4ce51a2b6d42a29a06163/components/konflux-rbac/staging/base/konflux-admin-user-actions.yaml#L24-L35

@gbenhaim , how do we protect against changing Snapshots today?

brianwcook:
the issue is that we need to be able to add a label to an existing snapshot as described [here](https://konflux-ci.dev/docs/testing/integration/periodic-integration-tests/) in order to kick off the ITS.  I am following the instructions given on creating an SA and what permissions to assign to it. I think that if we are worried about this PR, then we should also be worried about the advice given in those docs.

arewm:
This also appears to be consistent with https://github.com/konflux-ci/architecture/blob/main/ADR/0011-roles-and-permissions.md. It seems like we might need additional protections in place to protect against mutating the components.

brianwcook:
@gbenhaim we decided to add a new webhook to make snapshot components immutable. 
https://github.com/konflux-ci/integration-service/pull/1212

components/policies/staging/base/konflux-rbac/bootstrap-tenant-namespace/bootstrap-tenant-namespace-np-konflux-cron-sa.yaml
```diff
@@ -0,0 +1,66 @@
...
+                konflux-ci.dev/type: tenant
+      generate:
+        kind: Role
+        apiVersion: rbac.authorization.k8s.io/v1
+        name: snapshot-access
+        namespace: '{{request.object.metadata.name}}'
+        synchronize: true
+        data:
+          rules:
+            - apiGroups: ["snapshot.mycompany.io"]  # Replace with actual CRD group
```
gbenhaim, line 40: can you please use the correct group?

PR #7044, opened 2025-07-08
Kubearchive (stone-prod-p02): double operator memory

rh-hemartin:
* Looks like we crossed the line for the operator, before was working OK, now its restarting endlessly. Increasing its memory to prevent that. We will work to see if we can reproduce this and then lower its consumption on upstream.

### DIFF
--- a/components/kubearchive/production/stone-prod-p02/kustomization.yaml
+++ b/components/kubearchive/production/stone-prod-p02/kustomization.yaml
@@ -116,11 +116,11 @@ patches:
                   - containerPort: 8081
                 resources:
                   limits:
-                    cpu: 500m
-                    memory: 256Mi
+                    cpu: 100m
+                    memory: 512Mi
                   requests:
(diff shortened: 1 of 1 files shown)


components/kubearchive/production/stone-prod-p02/kustomization.yaml
```diff
@@ -116,11 +116,11 @@ patches:
                   - containerPort: 8081
                 resources:
                   limits:
-                    cpu: 500m
-                    memory: 256Mi
+                    cpu: 100m
```
filariow, line 119: Is this downscale intentional?

Usually we don't enforce a limit on CPU, so a requiring component can take all the available left free. If not needed, feel free to remove it.

rh-hemartin: I'm setting the limit because of the Guaranteed QoS, not that we ran into any eviction problem but still.

PR #7045, opened 2025-07-08
mintmaker: update brew.registry.redhat.io secret labels

qixiang:
Add the following lable to brew.registry.redhat.io secret:

   mintmaker.appstudio.redhat.com/secret-type: registry

mintmaker will combine secrets with this label, and mount it to Renovate container, then podman/skopeo can use the combined secret for image pulling.


qixiang:
I confirmed the brew.registry.redhat.io secret is created as expected, so I'm adding the label to secret, so podman/skopeo can use this secret.

qixiang:
@staticf0x @querti this PR is missing the lgtm label, could you help add it?

PR #7060, opened 2025-07-09
stage: Give the Vanguard team admin permission on the smee-client ns

gbenhaim:
The team needs it in order to maintain the smee client.


avi-biton (approved):
/lgtm

PR #7061, opened 2025-07-09
Set memory limit as Mi

enarha:
Initial value (1.5Gi) was set as Gi, but the controller is transforming it to Mi (1536Mi) and it's "fighting" with ArgoCD over that value, blocking the TektonConfig update. So no real change here, the actual value will remain the same, only the representation is changed.


hugares (approved):
/lgtm

PR #7075, opened 2025-07-09
Apply set-tenant-namespace-visibility policy

enkeefe00:
Applies the ClusterPolicy to all staging clusters.


filariow:
/hold 

we may not need this in the short term

PR #7076, opened 2025-07-09
Upgrade MCE operator to 2.8 in prod

amisstea:
This adds support for provisioning OCP 4.18 with EaaS.


yftacherzog:
@amisstea I added hold label just so you merge it when it's convenient to you.

PR #7077, opened 2025-07-09
ClusterRole for integration pipeline service account runner

amisstea:
Integration pipelines will eventually use new service accounts bound to this role.

Permissions for SpaceRequests are not included since toolchain is obsolete.

Assisted-by: cursor (claude-4-sonnet)

### DIFF
--- a/components/konflux-rbac/base/konflux-integration-runner.yaml
+++ b/components/konflux-rbac/base/konflux-integration-runner.yaml
@@ -1,0 +1,51 @@
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: konflux-integration-runner
+rules:
+  - verbs:
+      - get
+      - list
+    apiGroups:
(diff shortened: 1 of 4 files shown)


amisstea:
This currently deploys the ClusterRole to dev and staging clusters but not prod clusters since the `components/konflux-rbac/base` directory isn't included in the kustomization file. How is the `appstudio-pipelines-runner` clusterrole deployed to prod today?

filariow:
> @filariow I duplicated the ClusterRole in the `sandbox/toolchain-member` component for production only. This should now deploy to all envs.

as we are deprecating the `sandbox` component, may you move it under `konflux-rbac`?

amisstea:
@filariow I moved the ClusterRole for production under `konflux-rbac`. It should be safe to cover all envs together with this change IMO. #7108 is a different story. We'll definitely want to deploy staging and verify there before moving on to prod.

PR #7079, opened 2025-07-10
KONFLUX-134: Cancel in-progress PipelineRuns for updated pull requests (staging)

ralphbean:
This should reduce some load on the cluster when pull requests are updated over and over again. Earlier unnecessary PipelinesRuns will be cancelled and impose less load.


ralphbean:
If this works out well, it should be promoted to staging later (and added to `hack/new-cluster/`).

aThorp96 (approved):

PR #7097, opened 2025-07-10
KubeArchive: allow exported_job to reach Prometheus

rh-hemartin:
Due to how KubeArchive exposes metrics the `job` gets overwritten by a higher layer and it can't be changed, so we need to let `exported_job` to be passed to Prometheus. The reasoning behind that `honorLabels` is not honored is because users could potentially override labels as `namespace` leading to confusion and problems in metrics analysis. https://issues.redhat.com/browse/MON-793 and  PVO11Y-4845 for details.


mftb (approved):
/lgtm
/approve

PR #7102, opened 2025-07-10
Bump hypershift-aws-template chart to 0.1.8

amisstea:
This should fix an issue where fips is always enabled on EaaS ephemeral clusters when it should be disabled.


hmariset (approved):
/lgtm

PR #7104, opened 2025-07-10
Set the qps and burst for Results api

enarha:
We are hitting the limit with the default values and seeing `due to client-side throttling, not priority and fairness` errors.


khrm (commented):
/approve
/lgtm

PR #7114, opened 2025-07-11
ensure kyverno admission controller has Guaranteed QoS in stg

filariow:
we want every kyverno's pod to be assigned with Guaranteed QoS.


hugares (approved):
/lgtm

PR #7134, opened 2025-07-13
PVO11Y-4852 Fix: add new rolebinding to kustomization file

kubasikus:
The new rolebinding was not deployed because it was not added to the kustomization file resources.


kubasikus:
@martysp21 @TominoFTW PTAL

PR #7160, opened 2025-07-15
KubeArchive: disable leader election

rh-hemartin:
Our operator is getting restarted due to etcd defrag that makes the leases unavailable for more than 15s, the default timeout for leases. I'm disabling the leader election since patching the timeout will take some days.

### DIFF
--- a/components/kubearchive/development/kustomization.yaml
+++ b/components/kubearchive/development/kustomization.yaml
@@ -99,6 +99,7 @@ patches:
           spec:
             containers:
               - name: manager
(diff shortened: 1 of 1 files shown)


components/kubearchive/development/kustomization.yaml
```diff
@@ -99,6 +99,7 @@ patches:
           spec:
             containers:
               - name: manager
+                args: [--health-probe-bind-address=:8081]
```
maruiz93, line 102: how does this disable leader election?

PR #7171, opened 2025-07-15
WIP: configure kc oauth for dev/test clusters

psturc:


gbenhaim:
Why it's needed?

psturc:
> Why it's needed?

After toolchain is gone, we won't be able to [create users via sandbox anymore in e2e-tests](https://github.com/konflux-ci/e2e-tests/blob/6c92ed5bbd24631d0f3ea9baaf9942f9ec9fade5/pkg/clients/kubernetes/client.go#L137-L148), so I thought that one way to mimic the staging/prod clusters environment would be to set up SSO (since we're already deploying it in dev env) as oauth and update the e2e framework to do the folllowing for each test (that will have a dedicated namespace where the test resources are created):
1. create new namespace with required labels
2. create new keycloak user
3. assign proper [cluster role](https://github.com/redhat-appstudio/infra-deployments/blob/bd2d4a83dc7f2d324db56247c5334fd444392195/components/konflux-rbac/staging/base/konflux-admin-user-actions.yaml)
4. log in with that new user and use it [for interactions with test resources from a user's perspective](https://github.com/konflux-ci/e2e-tests/blob/6c92ed5bbd24631d0f3ea9baaf9942f9ec9fade5/pkg/framework/framework.go#L154)

I'm definitely open to suggestions, this is just an idea that I haven't fully tested yet

gbenhaim:
@psturc why do you need to create users as part of the test? you can just create namespace. With the new Konflux architecture there is no requirement to have an SSO user for creating a namespace.

PR #7176, opened 2025-07-15
KubeArchive: add labels for histograms

rh-hemartin:
Adding some labels so we can pass enough information so the histograms are well constructed.


kubasikus (changes requested):
See my note in slack - instead of adding more labels, I think we should rename metrics themselves. This way, there should be no collision with other metrics gathered from the cluster.

PR #7178, opened 2025-07-15
Deploy smee server only on host clusters

hugares:
Smee server was deployed to the clusters with host role which resulted in deploying it to clusters with both member and host role, such as internal clusters, rh02, rh03,... We only need smee server on stone-stg-host and stone-prd-host1 clusters.

To achieve that, change applicationSet to deploy empty application to all the clusters with host role and make sure the 2 where we need this actually deploy the smee server.

KFLUXINFRA-1950


yftacherzog (approved):
/lgtm
/approve
/hold
@hugares Looks good to me. Added hold label so you can merge it when the time is right for you.

PR #7184, opened 2025-07-15
Updated resources tekton-pipelines-controller in pipeline-service pro…

ab-ghosh:
This PR updates the resource requests and limits for the tekton-pipelines-controller in the pipeline-service production overlay to address frequent pod restarts due to OOMKilled (ExitCode 137).


aThorp96 (approved):
/lgtm
/approve

PR #7197, opened 2025-07-16
Deploy the integration kyverno policy to production

amisstea:


components/policies/production/base/integration/kustomization.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- ../../../development/integration/
```
amisstea, line 4: This feels like an anti-pattern but on the other hand I don't want to duplicate it all for production. Since there's no common base, how should we handle this?

filariow: we usually duplicate in production, so we are sure that changes to dev/stg are not propagated to production

filariow (approved):
/lgtm

PR #7198, opened 2025-07-16
Promote MintMaker stage to production

FernandesMF:
This commit promotes the current stage deployment of MintMaker to production (weekly MintMaker release).


qixiang:
This duplicates https://github.com/redhat-appstudio/infra-deployments/pull/7195, I approved that one.

PR #7204, opened 2025-07-17
fix: reduce-tekton-pr kyverno policy for e2e tests

psturc:
previous kyverno policy did not apply to all pods created by e2e test pipelines, because when the component name is long, the TaskRun's pod name no longer contains - `-on-pull-request-` or `-on-push-` in its name, but the suffix `-pod` always remains

### DIFF
--- a/components/policies/development/taskruns-cluster-policy.yaml
+++ b/components/policies/development/taskruns-cluster-policy.yaml
@@ -16,8 +16,7 @@ spec:
           kinds:
           - Pod
           names:
-          - '*-on-pull-request-*'
-          - '*-on-push-*'
+          - '*-pod'
           operations:
           - CREATE
           selector:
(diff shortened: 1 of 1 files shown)


filariow (approved):
/lgtm

Out of curiosity, isn't the label we are already matching enough? Do we have any other label we can match instead of the name?

/hold
To allow you to view my comment. Feel free to unhold when it's ready to merge.

psturc:
> /lgtm
> 
> Out of curiosity, isn't the label we are already matching enough? Do we have any other label we can match instead of the name?
> 
> /hold To allow you to view my comment. Feel free to unhold when it's ready to merge.

the main reason why I'm using a pod name matcher is because I've seen test failures caused by this kyverno policy applied to `affinity-assistant-*` pods, which have the same label as a related TR pod, which caused the TR pod to be pending due to issue with scheduling on a node.. basically this kind of error
```
0/x nodes are available: x node(s) didn't match pod affinity rules, x node(s) had volume node affinity conflict,...
```

I didn't look more into it, but updating the policy to match the pod name fixed this issue... 🤷

PR #7220, opened 2025-07-17
Fold MPC staging-downstream into staging folder

hugares:
Simplify the folder structure under MPC. We used to have a staging and staging-downstream. Instead of that, folder both folders into staging and create one folder per cluster, like we do in any components where we have specifics manifest depending on the cluster.


sadlerap (approved):
/lgtm

gbenhaim:
Can you please update the script which updates the cluster queues as well https://github.com/redhat-appstudio/infra-deployments/blob/0a70d48440841fac74913133cf7fcda859c7dc8a/hack/kueue-vm-quotas/generate-queue-config.sh#L19 ?

PR #7246, opened 2025-07-21
Support local platforms in CQ generation and verify using GH action

gbenhaim:
KFLUXINFRA-1982: Consider local platform when generating CQ
    
    - When generating the cluster queue take include the local platform.
    
    - Local platforms has a fixed quota of 1000 (for simplicity). It's
      possible to configure it in that way because the workload will be
      queued based on the quota for pipelineruns (which will likely be lower).
    
    - Update the CQ's in staging.
  
    KFLUXINFRA-1985: Validate queue definition is up to date
    
    - Add --verify-no-change flag to the generate-queue-config.sh script.
      With this flag the script will fail if there were any modifications
      to the files containing the queue definition.
    
    - Add a Github action for running the verification on PRs.

### DIFF
--- a/.github/workflows/verify-kueue-queue-configs.yaml
+++ b/.github/workflows/verify-kueue-queue-configs.yaml
@@ -1,0 +1,37 @@
+name: Verify Kueue Queue Configurations
+
+on:
+  pull_request:
+    branches: [ main ]
+    paths:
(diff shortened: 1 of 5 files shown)


gbenhaim:
@hugares can you PTAL?

PR #7259, opened 2025-07-21
add pipelines-as-code namespace URL

aThorp96:


enarha (commented):
Shouldn't we also update the development overlay?

enarha:
> @enarha the development overlay doesn't configure any of the URLs for PaC since it doesn't have a consistent hostname per deployment. They are only used for links in the git provider's comments so missing URLs aren't a functional issue for development

Will it break anything if we add it? Even if it does not have any effect, it's better to have it in dev, so the dev and stage overlays have less diffs. If it has negative effect though (breaks CI), then we can keep it as is.

aThorp96:
> Even if it does not have any effect, it's better to have it in dev, so the dev and stage overlays have less diffs. If it has negative effect though (breaks CI), then we can keep it as is.

It won't necessarily break anything (unless there is CI to check that links are not dead). Do you have any suggestions on what value to use? I'm not sure of a sensible default since every deployment will have a default value. Instead the CI updating each link to `some.setting.not.set/foo/bar` it will be a wrong URL 🤔 

I guess maybe the string `unknown` would work?

PR #7262, opened 2025-07-21
add resource limits and requests for pipelines staging

aThorp96:
Values calculated based on the resources consumed by pods in the past 7 days in stone-prd-rh01, stone-prod-p01, and stone-prod-p02, as seen in [this grafana query](https://grafana-access-appstudio-grafana.apps.stone-prd-rh01.pg1f.p1.openshiftapps.com/explore?schemaVersion=1&panes=%7B%22y99%22:%7B%22datasource%22:%22064f2b18-a714-4631-997b-1fe80bb695b5%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22max%20by%28deployment%29%20%28label_replace%28container_memory_working_set_bytes%7Bnamespace%3D%5C%22openshift-pipelines%5C%22,%20container%3D%5C%22%5C%22%7D,%20%5C%22deployment%5C%22,%20%5C%22$1%5C%22,%20%5C%22pod%5C%22,%20%5C%22%28%5B-a-z%5D%2B%29-.%2A%5C%22%29%29%5Cn%22,%22range%22:true,%22instant%22:true,%22datasource%22:%7B%22type%22:%22prometheus%22,%22uid%22:%22064f2b18-a714-4631-997b-1fe80bb695b5%22%7D,%22editorMode%22:%22code%22,%22legendFormat%22:%22__auto%22%7D%5D,%22range%22:%7B%22from%22:%22now-7d%22,%22to%22:%22now%22%7D%7D%7D&orgId=1): `max by(deployment) (label_replace(container_memory_working_set_bytes{namespace="openshift-pipelines", container=""}, "deployment", "$1", "pod", "([-a-z]+)-.*"))`

Screenshot of RH01:
image

### DIFF
--- a/components/pipeline-service/staging/base/main-pipeline-service-configuration.yaml
+++ b/components/pipeline-service/staging/base/main-pipeline-service-configuration.yaml
@@ -1781,6 +1781,136 @@ spec:
                       requests:
                         cpu: 400m
                         memory: 1Gi
+        pipelines-as-code-controller:
+          spec:
+            template:
+              spec:
+                containers:
+                  - name: pac-controller
+                    resources:
+                      limits:
+                        cpu: 500m
+                        memory: 2.5Gi
+                      requests:
+                        cpu: 500m
+                        memory: 2.5Gi
+        pipelines-as-code-watcher:
+          spec:
+            template:
+              spec:
+                containers:
+                  - name: pac-watcher
+                    resources:
+                      limits:
+                        cpu: 250m
+                        memory: 1Gi
+                      requests:
+                        cpu: 250m
+                        memory: 1Gi
+        pipelines-as-code-webhook:
+          spec:
+            template:
+              spec:
+                containers:
+                  - name: pac-webhook
+                    resources:
+                      limits:
+                        cpu: 400m
+                        memory: 100Mi
+                      requests:
+                        cpu: 400m
+                        memory: 100Mi
+        pipelines-console-plugin:
+          spec:
+            template:
+              spec:
+                containers:
+                  - name: pipelines-console-plugin
+                    resources:
+                      limits:
+                        cpu: 200m
+                        memory: 8Mi
+                      requests:
+                        cpu: 200m
+                        memory: 8Mi
+        tekton-chains-controller:
+          spec:
+            template:
+              spec:
+                containers:
+                  - name: tekton-chains-controller
+                    resources:
+                      limits:
+                        cpu: 1500m
+                        memory: 8Gi
+                      requests:
+                        cpu: 1500m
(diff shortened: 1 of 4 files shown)


components/pipeline-service/staging/base/main-pipeline-service-configuration.yaml
```diff
@@ -1781,6 +1781,136 @@ spec:
...
+              spec:
+                containers:
+                  - name: webhook
+                    resources:
+                      limits:
+                        cpu: 250m
+                        memory: GiMi
+                      requests:
+                        cpu: 250m
+                        memory: GiMi
```
mathur07, line 1887: GiMi?

components/pipeline-service/staging/base/main-pipeline-service-configuration.yaml
```diff
@@ -1781,6 +1781,136 @@ spec:
...
+                        memory: 8Mi
+        tekton-chains-controller:
+          spec:
+            template:
+              spec:
+                containers:
+                  - name: tekton-chains-controller
+                    resources:
+                      limits:
+                        cpu: 500m
```
enarha, line 1844: this seems low. I see some peaks on one of the clusters going above 800m, so probably 1000m will be a better value.
And in general, it feels most value a bit conservative and we should be able to withstand peaks x2, x3 or even x5 above the normal load.

enarha (commented):
As I mentioned in the comment, most of those values feel a bit low to me. I checked the actual usage in the last 2 days and they seem to be around x2 the normal load. Was that your target? I think we could be a bit more "generous", especially on the cpu side.
As a side note, there are some pods which are hardly or not at all used. For instance - pipelines-console-plugin, tekton-triggers-core-interceptors, tkn-cli-serve and there could be more. Maybe a better approach to conserve resources is to remove all unused pods.  This should be a separate change and it should go through the architecture committee to approve that.

components/pipeline-service/staging/base/main-pipeline-service-configuration.yaml
aThorp96: Updated. My reasoning here was generally ~1.5-2x highest spike I saw, rounded to a friendly number. I must have missed this when looking at CPU though. I've revised the numbers now

components/pipeline-service/staging/base/main-pipeline-service-configuration.yaml
aThorp96: Good catch. Thanks!

aThorp96:
D'oh. Thanks for catching. Fixed

PR #7270, opened 2025-07-22
Extract kite database secret for stone-stg-rh01

hugares:
Create ExternalSecret to extract database connection info. Create ES in base and patch the path in the cluster dir as each cluster will have a different secret path.

Also allow to get secret from appsre-vault in konflux-kite namespace.


enkeefe00 (changes requested):
Shouldn't the path be `konflux-kite/staging/stone-stg-rh01`?

CryptoRodeo (approved):
LGTM

PR #7287, opened 2025-07-22
Add larger disk variant to RH03 cluster

enkeefe00:
[KFLUXINFRA-1980](https://issues.redhat.com/browse/KFLUXINFRA-1980): Add a 160GB disk variant of the c8xlarge amd64 & arm64 AWS VM platforms to the kflux-prd-rh03 cluster host configuration file.


hugares:
@enkeefe00 I think we should align all the clusters to have this and add this to the documented VMs in https://konflux.pages.redhat.com/docs/users/getting-started/multi-platform-builds.html#cpu-optimized

At least all the non-dedicated clusters, p01, p02, rh01, rh02 and rh03

hugares (approved):
/lgtm

PR #7298, opened 2025-07-23
restrict update event for integration's ClusterPolicy

filariow:
We need to create the resources only when the namespace is promoted to tenant namespace. We can skip the other updates to the namespaces.

### DIFF
--- a/components/policies/development/integration/bootstrap-namespace/.chainsaw-test/chainsaw-test.yaml
+++ b/components/policies/development/integration/bootstrap-namespace/.chainsaw-test/chainsaw-test.yaml
@@ -188,3 +188,141 @@ spec:
         expect:
         - check:
             ($error != null): true
+---
+# yaml-language-server: $schema=https://raw.githubusercontent.com/kyverno/chainsaw/main/.schemas/json/test-chainsaw-v1alpha1.json
+apiVersion: chainsaw.kyverno.io/v1alpha1
+kind: Test
+metadata:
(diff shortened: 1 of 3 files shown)


components/policies/development/integration/bootstrap-namespace/bootstrap-namespace.yaml
```diff
@@ -19,6 +19,12 @@ spec:
           selector:
             matchLabels:
               konflux-ci.dev/type: tenant
+          operations:
+          - CREATE
+          - UPDATE
```
amisstea, line 24: I suspect the declaration of `operations` is related to the chainsaw test failure. These should be the default values according to the [docs](https://kyverno.io/docs/policy-types/cluster-policy/match-exclude/#match-statements) so I'm not sure why. It was something I encountered with #7108 but didn't investigate any further.

Does the policy still work if you omit these?

components/policies/development/integration/bootstrap-namespace/.chainsaw-test/chainsaw-test.yaml
```diff
@@ -128,6 +128,7 @@ spec:
     - assert:
         file: chainsaw-assert-clusterpolicy.yaml
   - name: then-serviceaccount-is-created
+    timeout: 180s
     try:
     - assert:
```
filariow, line 133:
```suggestion
    try:
    - assert:
        timeout: 180s
```

amisstea (approved):

PR #7216, opened 2025-07-17
Replace rhel-p01 AMD VMs by m7a

hugares:
The RHEL rpm builds are building x86_64 using MPC VMs, not in cluster so change the MPC EC2 instances from m6a to m7a.

[KONFLUX-8247](https://issues.redhat.com//browse/KONFLUX-8247)

PR #7301, opened 2025-07-23
Revert kflux-rhel-p01 AMD VMs upgrade

enkeefe00:
Reverted commit #7216 back to using m6a AWS AMD VMs for the kflux-rhel-p01 cluster.


hugares:
We cannot revert that, RHEL team actually need m7 which is supporting x86 v4

PR #7305, opened 2025-07-23
set GIT_SSL_CAINFO env var on resolvers in development

aThorp96:


mathur07:
@enarha PTAL

enarha:
I cherry-picked that commit in a separate PR #7311 so I can have a proper look on the CI cluster and the problem is one of the images (tkn-cli-serve) fails to deploy as the deployment seems to be updated once applied to the cluster and the image tag removed. That blocks the deployment of the other components, including PaC and then that further affects the build service deployment.

PR #7306, opened 2025-07-24
KFLUXINFRA-1970: Change Max Instances for MPC Platforms

manish-jangra:
This commit will change the max-instance for all the platforms specified in host-config making sure allowed number of instances do not exceed 250 instances.

### DIFF
--- a/components/multi-platform-controller/production/stone-prd-rh01/host-config.yaml
+++ b/components/multi-platform-controller/production/stone-prd-rh01/host-config.yaml
@@ -66,7 +66,7 @@ data:
   dynamic.linux-arm64.aws-secret: aws-account
   dynamic.linux-arm64.ssh-secret: aws-ssh-key
   dynamic.linux-arm64.security-group-id: sg-0fbf35ced0d59fd4a
-  dynamic.linux-arm64.max-instances: "50"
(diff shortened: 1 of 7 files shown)


components/multi-platform-controller/production/stone-prd-rh01/host-config.yaml
```diff
@@ -66,7 +66,7 @@ data:
   dynamic.linux-arm64.aws-secret: aws-account
   dynamic.linux-arm64.ssh-secret: aws-ssh-key
   dynamic.linux-arm64.security-group-id: sg-0fbf35ced0d59fd4a
-  dynamic.linux-arm64.max-instances: "50"
+  dynamic.linux-arm64.max-instances: "70"
```
hugares, line 69: To keep things simple, here is what I propose. We have 32 AWS platform, one is default arm, one default amd and the rest is flavors of those 2.
Let's set all the non default flavors to 5, this sum up to 150, leaving 100 for default arm and amd. Based on the numbers I provided in slack, arm is used way more so we could do 70 arm and 30 amd. If in future we add new platform, we could take capacity from amd

manish-jangra: done

PR #7312, opened 2025-07-24
add pulp-access-controller component to public prod

sadlerap:
It appears that pulp should be in all clusters, not just the internal clusters.  It's already in the staging clusters and internal production, but we didn't remove the deletion patch from the public production overlays in #6561.

Remove the deletion patch so that pulp is deployed to the public production clusters.

Fixes: fbef70802 ("pulp to production (#6561)")


hugares (approved):
/lgtm

PR #7318, opened 2025-07-25
fix: adding rbac needed to run tests

tisutisu:
After moving test execution to rh03 cluster, appstudio-pipeline sa missing permission to get image-repositories causing the tests to fail, adding the needed ClusterRole and RoleBiding.


chmeliik (approved):
/lgtm

PR #7342, opened 2025-07-28
KubeArchive: install on stone-stage-p01

rh-hemartin:


enkeefe00 (commented):
What about un-commenting stone-stage-p01 in the base kubearchive appset?

rh-hemartin:
I need an /approve too, not just /lgtm

PR #7349, opened 2025-07-28
Revert "restrict update event for integration's ClusterPolicy"

filariow:
Reverts redhat-appstudio/infra-deployments#7298

`celPreconditions` is an immutable field. That's not the right way of updating the policy. We need to follow a "recreate" approach.


sadlerap (approved):
/lgtm

PR #7355, opened 2025-07-28
KONFLUX-134: Cancel in-progress PipelineRuns for updated pull requests (prod)

ralphbean:
This should reduce some load on the cluster when pull requests are updated over and over again. Earlier unnecessary PipelinesRuns will be cancelled and impose less load.

This was deployed to staging a few weeks ago in https://github.com/redhat-appstudio/infra-deployments/pull/7079. This catches things up in prod.

Additionally, I re-ordered a few settings on different clusters to make them consistent, and updated the hack/new-cluster/templates/ to reflect this change.


aThorp96 (approved):
/lgtm

PR #7371, opened 2025-07-29
add ability to view pac repositories

scoheb:
- The legacy components/authentication/base/everyone-can-view.yaml seemed to include this RBAC.
- it helps users determine if a Repository might point to be deleted secret for example.


scoheb:
> I don't see why this would be included with this cluster role. This cluster role is for the operator to bind to. If we are reusing it in e2e, a new role should be created for the e2e part IMO

Good point. But tram members running this locally would potentially also need it.

But I can refactor to have it separate from existing cluster role.

PR #7380, opened 2025-07-30
KubeArchive: install kflux-rhel-p01

rh-hemartin:


rh-hemartin:
@maruiz93 @enkeefe00 need an approval as well

mshaposhnik:
FYI there is a parallel PR to enable cert-manager on  ehrl-p01: https://github.com/redhat-appstudio/infra-deployments/pull/7495
I dont know if that may simplify things for you, but just FYI

rh-hemartin:
> FYI there is a parallel PR to enable cert-manager on rhel-p01: #7495 I dont know if that may simplify things for you, but just FYI

I want to keep this consistent, when cert-manager is enabled everywhere else and we are also everywhere, then I may do the switch. Thanks :D

mshaposhnik (approved):

PR #7385, opened 2025-07-30
KubeArchive: add vacuum to prevent missing messages

rh-hemartin:


rh-hemartin:
@ggallen give it a look

components/kubearchive/development/vacuum.yaml
```diff
@@ -0,0 +1,34 @@
...
+  name: vacuum-config-all
+spec:
+  namespaces: {}
+---
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: vacuum-all
+spec:
+  schedule: "* */1 * * *"
```
ggallen, line 14: Should we start with a less aggressive schedule rather than hourly?  We're not sure of the impact on the system when a vacuum runs.  Maybe let's start with once or twice a day and see how it goes?

rh-hemartin: I changed it to run once a day, close to midnight but not exactly to avoid overlapping other stuff.

ggallen: Shouldn't this be `* */24 * * *"?  Won't what you have still run once an hour?

rh-hemartin: "20 1 * * *" makes it run at 1:20 AM, so once a day.

ggallen: Sorry, I didn't see the changes.

ggallen (approved):

maruiz93 (approved):

PR #7390, opened 2025-07-30
KubeArchive: enable Kyverno policies

rh-hemartin:
I forgot this existed, adding policies to clusters we were installed recently.


sadlerap (approved):
/lgtm

PR #7409, opened 2025-07-31
KFLUXINFRA-2015: Enable newer arm64 instance types (m7g+)

mikebonnet:
Add large disk instances of the latest `arm64` and `amd64` instance types available in AWS. These instance types support newer cpu features required to build and test `vllm`. And the higher network and disk performance will accelerate the very resource-intensive `vllm` builds.


manish-jangra (approved):
/approve
/lgtm

PR #7414, opened 2025-07-31
chore(KFLUXSPRT-3661): Add namespace expression konflux-user-support

Kousalya1998:


Kousalya1998:
> Do you also need `kube_deployment_status_replicas_ready` and `kube_deployment_spec_replicas`?
> 
> Double-checking as we want to minimize unnecessary metrics forwarding when possible.

 wasn't completely sure if this was needed. But after checking what other teams had done in comparable scenarios, I decided it made sense to include it in our setup as well. wdyt?

pacho-rh (approved):
/lgtm
/approve

PR #7420, opened 2025-07-31
member-operator: split config for each cluster

sadlerap:
We're decommissioning kubesaw from the clusters, and to reduce risk, we'd like to take each cluster one at a time.  This means each cluster needs its own kustomization manifest, so that we can change things for one cluster but not the others (i.e. make changes to stone-prod-p02's configuration but leave stone-prd-rh01 alone).

Split the configuration by populating the clusterDir value in the ApplicationSet and populating the respective directories with kustomization manifests.  Move the existing configurations for both staging and production into a base/ directory, and update their references to the shared base directories as well.

These clusters now have their config split:
- stone-stg-rh01
- kflux-stg-es01
- stone-stage-p01
- stone-prd-rh01
- kflux-prd-es01
- kflux-prd-rh02
- stone-prod-p01
- stone-prod-p02
- kflux-ocp-p01
- kflux-prd-rh03
- kflux-rhel-p01
- kflux-osp-p01


meyrevived (approved):

meyrevived (approved):
/lgtm

PR #7425, opened 2025-08-01
KubeArchive: keep event_type

rh-hemartin:
We need to keep `event_type` for `kubearchive_cloudevents_total`.


TominoFTW (approved):
/lgtm
/approve

PR #7432, opened 2025-08-01
Stage p01 remove member operator subscription

sadlerap:
In order to remove member-operator from stone-stage-p01, we need to first remove the subscription.  This prevents OLM from installing any further updates to member-operator and allows us to safely scale down the deployment replica count to 0.

Requires #7420.


meyrevived (approved):
/lgtm

meyrevived (approved):
/lgtm

PR #7438, opened 2025-08-04
KFLUXINFRA-2013: deploy cert-manager on rh01 and p02

gbenhaim:


components/cert-manager/production/empty-base/.gitempty
mshaposhnik, line 1: is tha needed?

mshaposhnik (approved):

components/cert-manager/production/empty-base/.gitempty
gbenhaim: In this PR I removed it.
It was needed in order to deploy cert-manager to only a subset of the production clusters. I didn't want to deploy to all of them at once.

PR #7442, opened 2025-08-04
KONFLUX-8860: Swapping PPC Nodes for RHEL Environment

manish-jangra:
This commit will replace the existing ppc nodes with new nodes in Washington DC


sadlerap (approved):
/lgtm

PR #7444, opened 2025-08-04
add the OWNER for konflux-info

testcara:
As the subject.


testcara:
@ scoheb could you help to have a look?

PR #7448, opened 2025-08-04
konflux-ci: add gangway-token & link to SA

Prucek:
After this PR is merged, there is no need to manually add the gangway token in the users namespace and the pipelines that will run the [run-prowjob](https://github.com/openshift/konflux-tasks/tree/main/tasks/run-prowjob) task will use this token.


gbenhaim (changes requested):
I asked several question in slack about this change.
In addition, if we will choose to store the secret on vault, you would need to create an additional. ClusterSecretStore that has an allow list for the namespaces which need to consume this secret.

PR #7470, opened 2025-08-05
SRVKP-8255: Patch PAC images on stage

gbenhaim:
Continue to debug the reporting to SCM issue when using Kueue and PAC.


enarha (commented):
/lgtm

PR #7499, opened 2025-08-06
feat(RELEASE-1437): archive and delete Releases

davidmogar:
This commit configures KubeArchive for Releases. With this, all Releases will be archived. In addition, completed releases older than 1 week will get pruned.


davidmogar:
Please @johnbieren take a look. @rh-hemartin already reviewed the change so the only left thing is to review if the specified selectors and periods make sense for us.

davidmogar:
> > Please @johnbieren take a look. @rh-hemartin already reviewed the change so the only left thing is to review if the specified selectors and periods make sense for us.
> 
> Yes, the deleteWhen lgtm. We could probably go even less than 5 days if kubearchive things show up in the UI, but it is a fine starting point. I am not familiar with the vacuum stuff (and didn't notice it in the kubearchive presentation yesterday). But I guess you just wanted review on the deleteWhen? If so, I'll approve

The vacuum is just a requirement for the deletion to work. Hector already reviewed that part. Regarding the 5 days, I fully agree. Let's see it working properly first though. Users not aware of the change will start complaining about "missing releases" whenever they try to fetch them with kubectl/oc.

PR #7509, opened 2025-08-06
o11y: Split MonitoringStack by environment

pacho-rh:
This is in attempt to speed up the pipeline whenever changes to this MonitoringStack definition is made. This is especially crucial for if we need to revert a breaking change to this definition. Additional, this allows us to test changes to this definition in dev or stage first before applying them to production.


TominoFTW (approved):

TominoFTW:
Approved, I am just wandering for what is the `development` directory? Is that for local cluster development?

With that its just the question if pipeline of adding something should be just: update `staging` => check if its how you desired (change/revert) => update `production`?

TominoFTW:
I am thinking if we can just create a "symlink" to the file in staging perhaps? with `../../../staging/base/monitoringstack/monitoringstack.yaml` path? I think something similar worked for playbooks

TominoFTW:
Hmm, I am worried that it gets more and more confusing with that change 🤔

But I don't really see any other way out of that tho, so if that is working I am okay with using that. 👍

TominoFTW:
> Or better yet:
One more thought for this:
```
.
├── base/  # Production only
├── shared/  # Shared config for dev/staging
│   └── monitoringstack/
│       ├── monitoringstack.yaml
│       └── kustomization.yaml
├── development/ # Ref to shared
└── staging/ # Ref to shared
```
Could something like this be done?

pacho-rh:
> Could something like this be done?

I think this runs into the same problem. It could be assumed that `shared` also include resources for `production`

pacho-rh:
> then `stg-dev-(shared)-monitoringstack` 😆
> 
> I just want to completely omit the idea of having nested folders when symlinks don't work 🤔

I agree with avoiding nested folders. Made a new folder for holding staging and development common resources and moved the MonitoringStack definition under there.

TominoFTW (approved):
/lgtm
/approve

PR #7512, opened 2025-08-06
export kubearchive traces to cluster otlp collector

skoved:
export kubearchive traces to cluster otlp collector so traces can be viewed in signalfx

### DIFF
--- a/components/knative-eventing/staging/stone-stg-rh01/kustomization.yaml
+++ b/components/knative-eventing/staging/stone-stg-rh01/kustomization.yaml
@@ -3,3 +3,15 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
   - ../../base
+
+patches:
+  - patch: |-
+      apiVersion: v1
+      kind: ConfigMap
+      metadata:
+        name: config-tracing
+        namespace: knative-eventing
+      data:
+        backend: "zipkin"
+        zipkin-endpoint: "http://otel-collector.product-kubearchive.svc.cluster.local:9411/api/v2/spans"
+        sample-rate: "0.1"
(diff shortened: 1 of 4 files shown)


rh-hemartin:
You need to change the `KUBEARCHIVE_OTEL_MODE` of each of KubeArchive's components to `delegated`. These are set in the `development` folder and staging inherits from that folder.

rh-hemartin:
And then you need to change Knative Eventing. I will suggest that you change it here as well. You need to create a ConfigMap in the `knative-eventing` staging folder similar to the example one they provide:

https://knative.dev/docs/eventing/accessing-traces/#configuring-tracing

```
apiVersion: v1
kind: ConfigMap
metadata:
  name: config-tracing
  namespace: knative-eventing
data:
  backend: "zipkin"
  zipkin-endpoint: "http://zipkin.istio-system.svc.cluster.local:9411/api/v2/spans"
  sample-rate: "0.1"
```

The `konflux-otel` collector does not have a zipkin endpoint see [this](https://console-openshift-console.apps.stone-stg-rh01.l2vh.p1.openshiftapps.com/k8s/ns/konflux-otel/configmaps/open-telemetry-opentelemetry-collector) so you need to point Knative Eventing to KubeArchive's collector, and then the KubeArchive's collector will send the traces from Knative Eventing to the `konflux-otel` collector.

Related to this, in the zipkin receiver configuration you added, you need to specify an endpoint. By default is `localhost`, so Knative Eventing won't be able to send traces. You need it to be `0.0.0.0:9441). See https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/zipkinreceiver

skoved:
@rh-hemartin I've made changes based on your feedback. I ran the kustomize command for both kubearchive and knative-eventing on stone-stg-rh01 and the results look the way that I expect

PR #7531, opened 2025-08-07
stone-stg-host: remove host-operator components

sadlerap:
Populate the toolchain-host-operator component with nothing but a namespace.


gbenhaim:
@sadlerap I don't think that restore from backup fully workd. I tried to trigger a pipeline from one of my repo which is connected to stage and it wasn't triggered. You can also check it by triggering a pipeline from the multi arch samples repository.

PR #7551, opened 2025-08-08
KubeArchive logging: remove application for clusters other than staging

rh-hemartin:


sadlerap (approved):
/lgtm

lcarva (approved):
Talking to Hector in slack, a previous change caused kubearchive logging to be deployed to unintended clusters. Those clusters now have multiple "Create Container" errors. This change addresses that. This seems like a low risk change, and multiple folks that have better expertise than I do are on PTO. I'll go ahead and approve. Hector will monitor for any issues that may occur.

PR #7576, opened 2025-08-11
mintmaker: set GOMEMLIMIT to avoid controller OOMKilled crashes

qixiang:
Set GOMEMLIMIT=7GiB so the Go runtime respects Kubernetes memory limits. Although the controller is allocated 8GiB memory, in large clusters (e.g., p02 with 5000+ components) it can still be OOMKilled, possibly due to the Go runtime ignoring container limits.

### DIFF
--- a/components/mintmaker/production/base/manager_patch.yaml
+++ b/components/mintmaker/production/base/manager_patch.yaml
@@ -15,3 +15,6 @@ spec:
           requests:
             cpu: 100m
             memory: 8Gi
+        env:
+        - name: GOMEMLIMIT
+          value: "7GiB"


qixiang:
Since earlier today, the mintmaker controller in the p02 cluster has been repeatedly crashing due to OOMKilled errors. I’m not certain of the exact root cause, but it appears to be related to the Go runtime not respecting container memory limits. I’m setting `GOMEMLIMIT` to check whether this helps.

qixiang:
> do you think we could log the memory in controller via https://pkg.go.dev/runtime/metrics ?

I’m currently investigating and analyzing the controller’s memory usage, and I will include this in the research plan

PR #7603, opened 2025-08-12
gpu instances: make /var/run/cdi world-writable

mikebonnet:
Both `/etc/cdi` and `/var/run/cdi` are valid locations for CDI config, with the latter overriding the former. Make both world-writable so config can be modified by users as necessary.

No need to `su - ec2-user`, Konflux workloads run as a newly-created random user. Making the directory world-writable is sufficient.


mikebonnet:
What needs to be done to get this merged?

manish-jangra (approved):
/approve
/lgtm

PR #7608, opened 2025-08-12
give konflux admins rights on crds and webhooks

sadlerap:
As a part of the kubesaw removal, we're going to need to be able to make changes to CRDs and ValidatingWebhookConfigurations on the clusters (mostly deleting a handful, but read permissions would make things easier).  We could take a few approaches to this, including defining a Job to automate this removal process, but if anything goes wrong, admins need to be able to step in and restore things on the clusters. Therefore, give members of the konflux-admins group more permissions to administrate these resources.

### DIFF
--- a/components/authentication/base/konflux-admins.yaml
+++ b/components/authentication/base/konflux-admins.yaml
@@ -325,8 +325,13 @@ rules:
     resources:
       - customresourcedefinitions
     verbs:
-      - list
+      - create
       - get
+      - list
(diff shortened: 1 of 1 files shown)


hugares:
@sadlerap Can you rebase the PR

sadlerap:
@hugares done

I also added `create` permissions just in case.

hugares (approved):
/lgtm

PR #7615, opened 2025-08-13
Update PAC image on rh02 and rh03 & Enable queuing for all pipelineruns on rh02 and rh03

gbenhaim:
- Enable queuing for all pipelineruns on rh02 and rh03
- update PAC image on rh02 and rh03. The images contain a fix required for enabling Kueue on all pipelineruns.


enarha (commented):
/lgtm

PR #7644, opened 2025-08-14
fix(KAR-577): too many requests error in vector pod

maruiz93:
Trying to make gemini error in https://github.com/redhat-appstudio/infra-deployments/pull/7614 go away.
It has the same changes.


maruiz93:
The original PR could be finally merged

PR #7725, opened 2025-08-19
Set memory request and limits to same values for ESO

hugares:
The main pod is getting OOMkilled due to it's usage above requests even if below limit. Increase the request to 512Mi to make this guaranteed allocation. Also reduce limit of other 2 pods to match their request.

Those number were tuned based on production rh01 and p02 where we have the most load.

[KONFLUX-9158](https://issues.redhat.com//browse/KONFLUX-9158)


sadlerap (approved):
/lgtm

PR #7729, opened 2025-08-19
rhel: remove all member-operator components

sadlerap:
The next step to removing kubesaw from a cluster after the member-operator deployment has deleted is to remove all other resources managed by the sandbox component.  We can make this change for the rhel cluster.


hugares:
this one is not used in dev overlay so by-passing the e2e validation

PR #7762, opened 2025-08-21
Organising Stage Metrics

rubenmgx:
Since we are uploading a lot of metrics lately, this might be tricky to idenfity sometimes which are there and not. I have reorganisez all metrics based on pod, containers, etc...for a better management and visualization.
Doing stage first to ensure nothing changes (shouldnt be).


kubasikus (approved):
/lgtm
/approve

PR #7767, opened 2025-08-21
Increase quota for linux extra fast on rh01

hugares:
This one is hitting 100% usage quite often so double it up until we have the real fix which is to not limit individual profile but limit at the subnet level instead.


sadlerap (approved):
/lgtm

PR #7773, opened 2025-08-21
KFLUXINFRA-2161: Control AWS IP requests using Kueue

gbenhaim:
For each requested platform from AWS, request an IP by specifying the "aws-ip" resource in the pipeline's annotation using tekton-kueue CEL expressions.

This change will affect only the development overlay.

Requires: https://github.com/konflux-ci/tekton-kueue/pull/91
Assisted-By: Cursor

### DIFF
--- a/components/kueue/development/tekton-kueue/config.yaml
+++ b/components/kueue/development/tekton-kueue/config.yaml
@@ -6,11 +6,35 @@ cel:
         has(pipelineRun.spec.params) &&
         pipelineRun.spec.params.exists(p, p.name == 'build-platforms') ?
         pipelineRun.spec.params.filter(
-          p, 
+          p,
           p.name == 'build-platforms')[0]
         .value.map(
           p,
-          annotation("kueue.konflux-ci.dev/requests-" + replace(replace(p, "/", "-"), "_", "-"), "1")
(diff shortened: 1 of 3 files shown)


components/kueue/development/tekton-kueue/config.yaml
```diff
@@ -6,11 +6,35 @@ cel:
...
+        .value.filter(
+          p,
+          !(
+            p in [
+              'linux/ppc64le',
+              'linux/s390x',
+              'linux/x86-64',
+              'local',
+              'localhost',
+              'linux/amd64'
```
filariow, line 32: this is on AWS, we should not filter it out

```suggestion
```

components/kueue/development/tekton-kueue/config.yaml
```diff
@@ -24,7 +47,34 @@ cel:
...
+      .filter(
+        p,
+        !(
+          p[0].value in [
+            'linux/ppc64le',
+            'linux/s390x',
+            'linux/x86_64',
+            'local',
+            'localhost',
+            'linux/amd64'
```
filariow, line 71: ditto

```suggestion
```

gbenhaim:
@filariow this one is ready IMO. Can you please take a a look? After merging, would you be able to promote the change to staging ?

PR #7783, opened 2025-08-25
Removing duplicates from stage stack file

rubenmgx:
There are metrics which are duplicated and just pointing to individual namespaces. We have included all namespaces under same metrics to reduce number of lines and clarity.
Doing this first in stage.


martysp21 (approved):
/lgtm

PR #7786, opened 2025-08-25
feat(SPRE-1626) Adding namespace release-service

rubenmgx:
Adding missing namespace to the current metrics we are using to monitor namespace activity. First, stage


martysp21 (approved):
/lgtm

PR #7789, opened 2025-08-25
kflux-osp-p01: remove member operator installation

sadlerap:
Removes all components and resources from the member sandbox component for kflux-osp-p01 except for the namespace, which will be removed at a later date.


hugares (approved):
/lgtm

PR #7806, opened 2025-08-25
remove appstudio-pipelines-runner from sandbox component

sadlerap:
It's a duplicate and managed by konflux-rbac, so delete it from the sandbox component.  Also remove the patches that filter out this clusterrole in konflux-rbac on clusters with kubesaw already removed.


hugares (approved):
/lgtm

PR #7817, opened 2025-08-26
KFLUXINFRA-2048 work part II - release to public production clusters

meyrevived:
Second commit to KFLUXINFRA-2048 - releasing the small update with only Gal-L's update to metrics to the public production clusters. No harm was observed to the metrics on the clusters where the release was done and this update is transparent for users.


filariow (approved):
/lgtm

PR #7821, opened 2025-08-26
Stop alerts on pentest cluster by disabling metrics export v2

kubasikus:
Previous attempt to stop metric export did not work, as removing the cluster name from the list only caused the clusterDir value to not be set. The Application is still generated via the 'clusters' generator. Attempt to create an Application that manages no resources by modifying the cluster-specific component kustomization.


TominoFTW (approved):
/lgtm
/approve

PR #7822, opened 2025-08-26
Deploy kubearchive in pentest-p01

maruiz93:


maruiz93:
@manish-jangra can you please confirm that the external secret path to access the database is ok?

PR #7824, opened 2025-08-26
use Group/Version/Kind in deny-virtual-domain

filariow:
We are facing the validation error `admission webhook "validate-policy.kyverno.svc" denied the request: no unique match for kind Namespace`.

This change adds Group and Version in the ClusterPolicy's match.

Follows up to #7600


sadlerap (approved):
/lgtm

PR #7830, opened 2025-08-26
Revert "Stop alerts on pentest cluster by disabling metrics export (#…

kubasikus:
…7791)"

This reverts commit a8a716fcf05be1f69716ee2f0dae910e37c66b53.

The clusterDir variable needs to be set so that ArgoCD Application can use the updated "empty" kustomize directory for pentest-p01 cluster.


manish-jangra (approved):
/approve
/lgtm

PR #7796, opened 2025-08-25
Increase registry pull QPS to avoid 'pull QPS exceeded' errors

mshaposhnik:
Debugging some failed  TaskRun pods shows tghe folowing error:

```
2025-08-23 00:08:47E0823 00:08:36.232130    2669 pod_workers.go:1298] "Error syncing pod, skipping" err="failed to \"StartContainer\" for \"prepare\" with ErrImagePull: \"pull QPS exceeded\"" pod="rhodf-tenant/odr-operator-4-20-on-push-c2mlv-clone-repository-pod" podUID="03c1ff62-b2a5-4ed3-aeac-035af202b5b0"
```

As recommended in https://access.redhat.com/solutions/4767061, we're increasing the QPS limit in kubelet config.

Issue: https://issues.redhat.com/browse/KONFLUX-9836

PR #7855, opened 2025-08-27
Add preview.sh expected final blank line

maruiz93:
Yesterday along with some changes in https://github.com/redhat-appstudio/infra-deployments/pull/7796 the final blank line at the end of `argo-cd-apps/overlays/development/delete-applications.yaml` was removed.

This is causing issues when running `hack/preview.sh` script with `DEPLOY_ONLY` parameter.

The scripts modifies this file appending more applications to be deleted and the appending is done wrong if there is no blank line.

There is probably a fix that should be done for the `preview.sh` script to prevent this but for now I'm recovering the blank line.


hugares (approved):
/lgtm

PR #7858, opened 2025-08-27
staging: use exclusion lists for resources

sadlerap:
Rather than only including a select few resources in our tenant backups, let's include everything and filter out the resources we don't want.  For now, that means excluding pipelineruns, taskruns, and pods.

The first commit (c569e94) is a formatting change so that the kustomization manifests play better with `kustomize edit`.

The second commit (2f321be) adds patches to implement this on the staging clusters.  Until we're confident in this change and have validated that it works, hold off on rolling this out to the production clusters.


sadlerap:
> We need to update SOP to change restore procedure to only restore resources that we cherry picked in the backup before this change

ack, I'll file an MR to do that once this lands in staging

PR #7861, opened 2025-08-27
Update the UI banner with some whimsy

ralphbean:
This is so wonderful, I just want to share it with the world.

We don't have to keep it up long. Perhaps a week and a half?

(We can use this to advertise other things - new features or important fixes, too.)


testcara (commented):
Nice work. LGTM. Thank you.

PR #7875, opened 2025-08-28
Add o11y team members

kubasikus:
Allow o11y to approve changes related to their service ApplicationSets, which are located outside of the component folders.


manish-jangra (approved):
/approve
/lgtm

PR #7880, opened 2025-08-28
Allow up to 250 VMs of each AWS type

hugares:
The reason we reduced number of VMs was for the sum of them all to not exceed 250, which is the number of IPs we have on AWS side for MPC VMs.

The problem is this is artificially creating queues and user build are stuck in pending stated for up to hours sometime. We are working on a change to have a resource queue to make sure we do not create more than 250 VMs at the time otherwise we will get into provisioning issue because of lack of IPs.

We were supposed to land resource queue counting IPs before allowing MPC to create more VMs but the problem of long waiting time in the queue is worse than original problem being solved. For that reason, allow right away MPC to create up to 250 VMs of each type to unblock users and ip queue change will follow soon after.

Do the change on p02 first, other clusters will be change in another change.

### DIFF
--- a/components/kueue/production/stone-prod-p02/queue-config/cluster-queue.yaml
+++ b/components/kueue/production/stone-prod-p02/queue-config/cluster-queue.yaml
@@ -48,37 +48,37 @@ spec:
     - name: platform-group-1
       resources:
       - name: linux-amd64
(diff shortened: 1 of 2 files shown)


hugares:
/hold
I missed some arch

PR #7886, opened 2025-08-29
Refactor vector kubearchive component

maruiz93:
Improve readability and maintainance.

This PR intents to improve the maintainability but I found a problem in the process: kustomize complaints when `HelmChartInflationGenerator` points to a file outside of the directory.

To check if everything looks like it should I made the following before and after the changes and checked that the outcome is equivalent.
```
kubectl kustomize --enable-helm components/vector-kubearchive-log-collector/development -o <tmp-dir>
kubectl kustomize --enable-helm components/vector-kubearchive-log-collector/staging/stone-stg-rh01 -o <tmp-dir>
```

This made me replicate the following files into the different cluster directories instead of having just one under `base/`:
- `loki-helm-values.yaml`
- `vector-helm-values.yaml`
- `grafana-helm-values.yaml` 

The same happened with `loki-helm-stg-values.yaml` that I had to place under the cluster directory instead of under `staging/base` to be reused.

I will look if there is a way of bypassing this problem because otherwise we will have the burden of keeping these duplicated files in sync.

EDIT: For now I think we can go with repeating the common values among environments and clusters.
It's the same route that `kyverno` component went.


components/vector-kubearchive-log-collector/development/loki-helm-dev-values.yaml
```diff
@@ -0,0 +1,189 @@
...
+
+memcachedIndexQueries:
+  enabled: false
+
+memcachedIndexWrites:
+  enabled: false
+
+memcachedExporter:
+  enabled: false
+
```
mafh314, line 187: not sure but maybe the yaml lint will say something about this few white lines hehehe

mafh314:
small comment.
/lgtm
/approve

PR #7895, opened 2025-08-31
KAR-589: setup stone-prod-p02 config for kubearchive-logging

olegbet:

rh-pre-commit.version: 2.3.2


components/vector-kubearchive-log-collector/production/stone-prod-p02/loki-helm-values.yaml
```diff
@@ -0,0 +1,266 @@
+---
```
maruiz93, line 1: This file has unneeded config that it's going to be overwritten by `loki-helm-prod-values.yaml`. Can you please make sure that this file has exactly the same contents as`staging/stone-stg-rh01/loki-helm-values.yaml` and `development/loki-helm-values`? The three files should be identical.

Remember that after the refactor I made the idea is that for each generator we should have:
- `<component>-helm-values.yaml`  --> with exactly the same content for every cluster
- `<component>-helm-<env>-valyes.yaml --> should be identical for all the clusters under staging, and identical for all the clusters under production but potentially different between staging and production.

components/vector-kubearchive-log-collector/production/stone-prod-p02/vector-helm-values.yaml
```diff
@@ -0,0 +1,165 @@
...
+    address: 127.0.0.1:8686
+    playground: false
+  sources:
+    k8s_logs:
+      type: kubernetes_logs
+      rotate_wait_secs: 5
+      glob_minimum_cooldown_ms: 500
+      max_line_bytes: 3145728
+      auto_partial_merge: true
+      extra_label_selector: "app.kubernetes.io/managed-by in (tekton-pipelines,pipelinesascode.tekton.dev)"
```
maruiz93, line 23: Please make sure that the content of this file is exactly the same of `staging/stone-stg-rh01/vector-helm-values.yaml` and `development/vector-helm-values.yaml`. The three of them should be identical.

components/vector-kubearchive-log-collector/staging/stone-stg-rh01/loki-helm-values.yaml
```diff
@@ -27,7 +27,6 @@ loki:
         index:
           prefix: loki_index_
           period: 24h
-  # Configure compactor to use writable volumes
```
maruiz93, line 30: I don't think this change is needed

components/vector-kubearchive-log-collector/staging/base/kustomization.yaml
```diff
@@ -14,7 +14,6 @@ commonAnnotations:
 resources:
 - ../../base
 - loki-secret.yaml
-- sa.yaml
```
maruiz93, line 17: any reason for removing this line?

PR #7915, opened 2025-09-02
fix: use pac latest images in development overlays

tisutisu:
Using the same config used in stage/prod cluster for pipeline-service component so that rerun of push pipelineruns work on development clusters.


mmorhun (approved):
/lgtm

PR #7924, opened 2025-09-02
disable namespace-lister's group support in stg

filariow:
We don't want to return the list of all namespaces shared with the system:authenticated group (the only group supported nowadays).

We just want to show the list of namespaces directly shared with the user.


hugares (commented):
/lgtm

PR #7927, opened 2025-09-02
sandbox: remove host subscription from prod-p01

sadlerap:
Needed for safely removing kubesaw from the cluster.


hugares (approved):
/lgtm

PR #7934, opened 2025-09-02
Fix gosmee payload size limit to handle MintMaker MRs > 1MB

kelchen123:
Use --sse-buffer-size and --max-body-size flags from gosmee v0.28.0 to resolve https://issues.redhat.com/browse/KFLUXVNGD-447

### DIFF
--- a/components/smee-client/base/deployment.yaml
+++ b/components/smee-client/base/deployment.yaml
@@ -17,13 +17,15 @@ spec:
         - name: shared-health
           emptyDir: {}
       containers:
-        - image: "ghcr.io/chmouel/gosmee:v0.26.1"
+        - image: "ghcr.io/chmouel/gosmee:v0.28.0"
           imagePullPolicy: Always
           name: gosmee
           args:
             - "client"
             - TBA
             - "http://localhost:8080"
+            - "--sse-buffer-size"
+            - "2097152"
           volumeMounts:
             - name: shared-health
               mountPath: /shared
--- a/components/smee/base/deployment.yaml
+++ b/components/smee/base/deployment.yaml
@@ -19,10 +19,10 @@ spec:
         - name: shared-health
           emptyDir: {}
       containers:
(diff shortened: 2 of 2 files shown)


amisstea (changes requested):
I think we need to roll these changes out in dev and staging first and do some testing before we touch production. There are some [important caveats](https://github.com/chmouel/gosmee/blob/main/README.md#%EF%B8%8F-important-caveats) about resource requests and limits to consider.

amisstea (commented):
Let's also make sure these settings get deployed from the dev overlay as well.

kelchen123:
> Let's also make sure these settings get deployed from the dev overlay as well.

i think smee is deleted from dev environments? Not sure why

ie in line 41 from argo-cd-apps/overlays/development/delete-applications.yaml:

kind: ApplicationSet
metadata:
  name: smee
$patch: delete

amisstea:
> > Let's also make sure these settings get deployed from the dev overlay as well.
> 
> i think smee is deleted from dev environments? Not sure why

I see now there's no `development` overlay for the `smee` and `smee-client` components. I had assumed they existed :upside_down_face:. We can focus on just staging then.

components/smee/staging/stone-stg-host/gosmee-v0.28.0-patch.yaml
```diff
@@ -0,0 +1,11 @@
...
+- op: replace
+  path: /spec/template/spec/containers/0/image
+  value: "ghcr.io/chmouel/gosmee:v0.28.0"
+- op: replace
+  path: /spec/template/spec/containers/0/args
+  value: ["server", "--address", "0.0.0.0", "--max-body-size", "2097152"]
+- op: replace
+  path: /spec/template/spec/containers/1/image
+  value: "ghcr.io/chmouel/gosmee:v0.28.0" 
+
```
amisstea, line 11: There's some extra whitespace here apparently.

The rest looks good :+1: .

PR #7937, opened 2025-09-02
sandbox: remove host operator components from clusters

sadlerap:
Deletes all components from the `toolchain-host-operator` namespaces in all clusters with a host-operator deployment.


hugares (approved):
/lgtm

PR #7949, opened 2025-09-03
Fix kubelet app to avoid unwanted fields modification

mshaposhnik:


argo-cd-apps/base/all-clusters/infra-deployments/kubelet-config/kubelet.yaml
```diff
@@ -26,13 +26,20 @@ spec:
         targetRevision: main
       destination:
         server: '{{server}}'
+      ignoreDifferences:
+        - group: machineconfiguration.openshift.io
+          kind: KubeletConfig
+          jsonPointers:
+            - /spec/autoSizingReserved
+            - /spec/machineConfigPoolSelector
```
hugares, line 34: humm, I am a bit concerned with that approach. If eventually the thing owning this resource add another field, it will be remove by this config

hugares (approved):
/lgtm

hugares:
This application is not deployed in dev overlay, merging now

PR #7952, opened 2025-09-03
sandbox: remove remaining production manifests

sadlerap:
Removes all resources from the toolchain-member-operator namespaces in the remaining production clusters that have a member-operator deployment.  While we're at it, let's unify everything underneath a single `empty/` manifest tree.


hugares (approved):
/lgtm

hugares (approved):
/lgtm

PR #7953, opened 2025-09-03
Remove Kubesaw from CI jobs

gbenhaim:
Don't deploy Kubesaw as part of the CI job which runs the e2e tests since we are not going to use it anymore.


hack/bootstrap-cluster.sh
```diff
@@ -73,15 +65,15 @@ main() {
 }
 
 print_help() {
-    echo "Usae: $0 MODE [-t|--toolchain] [-kc|--keycloak] [-o|--obo] [-e|--eaas] [-h|--help]"
+    echo "Usae: $0 MODE [-o|--obo] [-e|--eaas] [-h|--help]"
     echo "  MODE             upstream/preview (default: upstream)"
-    echo "  -t, --toolchain  (only in preview mode) Install toolchain operators"
-    echo "  -kc, --keycloak  (only in preview mode) Configure the toolchain operator to use keycloak deployed on the cluster"
+    echo "  -o, --obo        (only in preview mode) Install Observability operator and Prometheus instance for federation"
+    echo "  -e  --eaas       (only in preview mode) Install environment as a service components"
```
hugares, line 71: why adding those 2 lines, 70-71 duplicate 72-73

sadlerap (commented):
~~We call `hack/bootstrap-member-cluster.sh` and `hack/bootstrap-common-cluster.sh`, should those calls be removed as well?~~

Edit: filed #8161 to remove these.

hack/bootstrap-cluster.sh
gbenhaim: fixed

PR #7957, opened 2025-09-04
feat: add the infra team as owners

testcara:
To ensure the review does not reply on UI team, let us expand the owners.


testcara:
Needs one /lgtm :-), @caugello could you help?

PR #7958, opened 2025-09-04
ocp-p01: Support user defined workload priority class

gbenhaim:
Allow the ART team to decide their workloads priorities. If a priority isn't defined on the PLR using a label, choose the priority using the same way like on the other clusters.

Adjust the test to test tekton-kueue configs from all of the overlays.

Assisted-By: Cursor


filariow (approved):
/lgtm

PR #7969, opened 2025-09-04
Remove clamav outage banner

tnevrlka:
The clamav outage is over, so the banner informing about it is no longer needed.


hugares (approved):
/lgtm

PR #7983, opened 2025-09-04
Revert "Fix gosmee payload size limit to handle MintMaker MRs > 1MB"

kelchen123:
Reverts redhat-appstudio/infra-deployments#7934


amisstea (approved):
/lgtm

PR #7990, opened 2025-09-05
Monitor Github/Gitlab availability

HozifaWasfy:
jira: [CWFHEALTH-4346](https://issues.redhat.com//browse/CWFHEALTH-4346)


components/mintmaker/staging/blackbox/blackbox-deployment.yaml
```diff
@@ -0,0 +1,70 @@
...
+
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
+    app.kubernetes.io/name: mintmaker
+    app.kubernetes.io/managed-by: kustomize
+  name: git-platforms-exporter
+  namespace: mintmaker
```
gbenhaim, line 33: Why do you deploy it in the mintmaker namespace? I think the o11y team is going to deploy a Blackbox exporter as well cc @mftb

components/mintmaker/staging/blackbox/git-platforms.yaml
```diff
@@ -0,0 +1,74 @@
...
+      namespace: mintmaker
+      selector:
+        matchLabels:
+          app: git-platforms-exporter
+      port: http 
+  module: http_2xx
+  targets:
+    staticConfig:
+      static:
+        - https://github.com
```
gbenhaim, line 24: Most of the times when the is an outage in github this endpoint is available. Is there  better way to check for availability? Same question about gitlab.

components/mintmaker/staging/blackbox/blackbox-deployment.yaml
HozifaWasfy: I want to monitor the availability of git platforms from mintmaker cluster

components/mintmaker/staging/blackbox/git-platforms.yaml
HozifaWasfy: > Most of the times when there is an outage in github this endpoint is available. Is there better way to check for availability? Same question about gitlab.

this is a good point, let me see what I can improve.

mftb:
@gcpsoares @kubasikus Please take a look at this

PR #7996, opened 2025-09-05
Fix overlays deployed for the crossplane-config component

amisstea:
All overlays were deploying the staging configuration which just happens to be the same across all 3 envs.


hugares (approved):
/lgtm

PR #8000, opened 2025-09-05
feat(kite): add bridge operator manifests

CryptoRodeo:
Add the KITE bridge operator manifests so it can be deployed.

### DIFF
--- a/components/knative-eventing/production/base/kustomization.yaml
+++ b/components/knative-eventing/production/base/kustomization.yaml
@@ -24,7 +24,7 @@ patches:
                   memory: 100Mi
                 limits:
                   cpu: 150m
-                  memory: 200Mi
+                  memory: 512Mi
 - patch: |-
     apiVersion: apps/v1
     kind: Deployment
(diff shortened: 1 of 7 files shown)


gbenhaim:
Why do we need an additional service for storing piplinerun status? We have kubeaechive or tetkon results. Why can't we use them?

gbenhaim:
> I'm also working with MintMaker and the Konflux Build team so they can create controllers to watch their resources as well

Which resources?

> KubeArchive and Tekton Results are great for storing historical data, but they don't support updating existing records AFAIK.

What is the flow? When are you going to update records. Let's say that a pipeline run failed and your operator created an issue for it. What's next?

> The bigger picture is that this will power the issues dashboard in Konflux, with teams contributing controllers that watch various resources and automatically managing their issue records
 
 Is there an ADR fot this plan? If not, I suggest to create one before proceeding with the deployment.
.

CryptoRodeo:
@gbenhaim I have an ADR you can view here: https://github.com/konflux-ci/architecture/pull/255

CryptoRodeo:
Converting to draft until final ADR decision

PR #8027, opened 2025-09-08
PVO11Y-4774 - Adding Prometheus metrics to determine Prometheus healt…

mike-kingsbury:
…h within Konflux cluster

This enable an existing Prometheus dashboard in AppSRE Grafana as well as add'l metrics to satisfy the goals of PVO11Y-4770.


pacho-rh (approved):
LGTM

PR #8055, opened 2025-09-09
Fix gosmee payload size limit to handle MintMaker MRs > 1MB in prod

kelchen123:
Use gosmee v0.28.0 flags --sse-buffer-size and --max-body-size to resolve [KFLUXVNGD-447](https://issues.redhat.com//browse/KFLUXVNGD-447)


yftacherzog:
@kelchen123 did you get to test it in konflux-ci/konflux-ci?

PR #8070, opened 2025-09-10
Remove custom code review and use Gemini API

enkeefe00:
Remove the custom Gemini code review action
(https://github.com/sshnaidm/gemini-code-review-action) and use Google's Gemini API workflow
(https://github.com/google-github-actions/run-gemini-cli) directly. This workflow uses an API key so it should still implement Red Hat's security requirements.


meyrevived:
Forgot to /lgtm

PR #8115, opened 2025-09-14
KFLUXINFRA-2182: Deprecated Restic Backups in OADP 1.5.0

eisraeli:
https://docs.redhat.com/en/documentation/openshift_container_platform/4.19/html/backup_and_restore/oadp-application-backup-and-restore#deprecated-features-1-5-0_oadp-1-5-release-notes


manish-jangra:
@eisraeli this will make changes to all the clusters in one go. Do you think, we should introduce patches for each overlay?

PR #8131, opened 2025-09-15
set Prune=false on core ClusterPolicies in stg

filariow:
Deleting these policies can cause security problems.

After this change is merged to remove the policies, either a manual intervention or an extra step to remove the annotation is required.


sadlerap (approved):
/lgtm

PR #8132, opened 2025-09-15
set Prune=false on core ClusterPolicies in prod

filariow:
Requires:
- [x] #8131

Deleting these policies can cause security problems.

After this change is merged, to remove these policies, either a manual intervention or an extra step to remove the annotation is required.

Signed-off-by: Francesco Ilario filario@redhat.com


filariow:
/hold

for #8131

PR #8145, opened 2025-09-16
Give users permission to view Kueue Workloads

gbenhaim:
Until we get the status of the Workload copied into the status of the Pipelinerun, let users see Workloads so they can understand the status of the Pipelinerun in the context of queuing.


aThorp96 (approved):
/lgtm

PR #8155, opened 2025-09-16
increase loki stream rate

olegbet:

rh-pre-commit.version: 2.3.2


maruiz93 (commented):
What do you think of adding also `ingestion_rate_strategy` to `local`?

related docs: https://grafana.com/docs/loki/latest/configure/#limits_config

PR #8160, opened 2025-09-16
Remove production smee server

enkeefe00:
KFLUXINFRA-1798


argo-cd-apps/base/host/smee/kustomization.yaml
```diff
@@ -1,6 +1,3 @@
 apiVersion: kustomize.config.k8s.io/v1beta1
```
hugares, line 1: why not deleting this kustomization file and parent folder argo-cd-apps/base/host/smee

components/smee/production/stone-prd-host1/kustomization.yaml
```diff
@@ -1,16 +1,4 @@
 ---
 apiVersion: kustomize.config.k8s.io/v1beta1
```
hugares, line 2: remove this file also?

PR #8035, opened 2025-09-08
Remove kubesaw manifests

sadlerap:
Remove all of kubesaw's manifests and references to said manifests from infra-deployments.

PR #8161, opened 2025-09-16
remove more sandbox references

sadlerap:
There are a few scripts that reference sandbox components underneath the hack/ directory.  Since sandbox has been removed as of #8035, we can remove these.


hugares (approved):
/lgtm

enkeefe00:
@gemini-cli /review

hack/bootstrap-cluster.sh
```diff
@@ -38,8 +38,6 @@ main() {
     done
 
     "${ROOT}/hack/deploy-argocd.sh"
-    "${ROOT}/hack/bootstrap-host-cluster.sh"
-    "${ROOT}/hack/bootstrap-member-cluster.sh"
```
gbenhaim, line 42: we need to keep this one since it deploys required secrets.

hack/bootstrap-host-cluster.sh
```diff
@@ -1,22 +0,0 @@
-#!/bin/bash -e
-
-declare -r ROOT="${BASH_SOURCE[0]%/*}"
-
-main() {
```
gbenhaim, line 5: @psturc @flacatus do we still need to run the scripts below?

gbenhaim (changes requested):

PR #8162, opened 2025-09-16
Update stage monitoring access

gcpsoares:
- Include new group to access stage monitoring namespaces.


raks-tt (approved):
This looks good to me. This should give us access to debug in the monitoring component in stage.

kubasikus (changes requested):
This should be a patch. Also, will the existing Role permissions be enough for your testing? What kind of resources do you intend to work with?

PR #8184, opened 2025-09-17
Remove trigger for pull_request open event

enkeefe00:
Since most pull requests (PRs) are opened with a forked branch, the permissions being used are read-only by default because of GitHub. This causes errors to occur when the workflow attempts to write a comment to the PR. These permissions cannot be changed even with explicit configuration in the workflow, so the user will be expected to invoke the review workflow manually via a comment starting with `@gemini-cli /review`.

An additional workflow has been added to add a helper comment to the PR when it is opened. This comment will help the user get started with the Gemini AI assistant.

### DIFF
--- a/.github/workflows/gemini-dispatch.yml
+++ b/.github/workflows/gemini-dispatch.yml
@@ -7,7 +7,7 @@ on:
   pull_request_review:
     types:
       - 'submitted'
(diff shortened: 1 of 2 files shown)


sadlerap (approved):
/lgtm

PR #8185, opened 2025-09-17
Remove adm64 from local platforms on staging

hugares:
The intent was to migrate amd64 to local platforms but the migration was never completed, it was only done on staging. Revert that back to staging and prod are configured the same way.

### DIFF
--- a/components/kueue/staging/stone-stage-p01/queue-config/cluster-queue.yaml
+++ b/components/kueue/staging/stone-stage-p01/queue-config/cluster-queue.yaml
@@ -60,7 +60,7 @@ spec:
     - name: platform-group-1
       resources:
       - name: linux-amd64
-        nominalQuota: '1000'
+        nominalQuota: '250'
       - name: linux-arm64
         nominalQuota: '250'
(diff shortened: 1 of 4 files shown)


scoheb (approved):
/approve
/lgtm

sadlerap (commented):
Do we also need to update the kueue configurations to account for these platforms now being hosted on AWS?  I think they would count toward the ip limitations we have.

hugares:
> Do we also need to update the kueue configurations to account for these platforms now being hosted on AWS? I think they would count toward the ip limitations we have.

I did, there is a change to set max from 1000 to 250

hugares:
> > Do we also need to update the kueue configurations to account for these platforms now being hosted on AWS? I think they would count toward the ip limitations we have.
> 
> I did, there is a change to set max from 1000 to 250

see https://github.com/redhat-appstudio/infra-deployments/pull/8185/files#diff-5b42d118fb961761c67667633e29f51778dd842779fbee7886dd32f3d96d75ccR63

PR #8191, opened 2025-09-18
feat(SRVKP-8847) Exposing Stage Metrics

rubenmgx:
To continue with the SLO work, we need to create a dashboard for openshift-pipelines team. There is already a created dashboard in each cluster, but we need to create one in app-sre grafana instances, and for that we need to expose few metrics.


FaisalAl-Rayes (approved):
/lgtm
/approve

PR #8210, opened 2025-09-18
feat: collect metrics from nginx in proxy pod

JoaoPedroPP:
Add `/nginx_status` to collect HTTP metrics from Konflux UI pod


JoaoPedroPP:
Close because it was a mirror of #8236

PR #8216, opened 2025-09-18
kyverno: bump to v1.15.2 on staging

sadlerap:
Upgrade kyverno from v1.13.4 to v1.15.2 on the staging clusters.

The PodDisruptionBudgets were having a hard time passing `kube-linter` checks (kube-linter wasn't happy with the `minAvailable` field), so modifications were needed to ensure they pass without issue.

PR #8233, opened 2025-09-19
ci: bump kubelinter to v0.7.6

sadlerap:
Needed for #8216, since it appears the current version (v0.7.2) gives false warnings about pod disruption budgets.


hugares (approved):
/lgtm

PR #8237, opened 2025-09-19
Fix git-probe labels and url

HozifaWasfy:


FernandesMF (approved):
Makes sense and looks good to me!

PR #8251, opened 2025-09-22
Change RHEL AMIs to Cloud Access

p8r-the-gr8:
Use Red Hat Cloud Access (BYOS) AMIs instead of regular PAYG AMIs that incur licensing fees.
The following changes were made:

ami-026ebd4cfe2c043b2 (RHEL-9.2.0_HVM-20230503-x86_64-41-Hourly2-GP2) -> ami-01aaf1c29c7e0f0af (RHEL-9.6.0_HVM-20250910-x86_64-0-Access2-GP3)

ami-03d6a5256a46c9feb (RHEL-9.2.0_HVM-20230503-arm64-41-Hourly2-GP2) -> ami-06f37afe6d4f43c47 (RHEL-9.6.0_HVM-20250910-arm64-0-Access2-GP3)

### DIFF
--- a/components/multi-platform-controller/staging-downstream/host-config.yaml
+++ b/components/multi-platform-controller/staging-downstream/host-config.yaml
@@ -52,7 +52,7 @@ data:
   # cpu:memory (1:4)
   dynamic.linux-arm64.type: aws
   dynamic.linux-arm64.region: us-east-1
-  dynamic.linux-arm64.ami: ami-03d6a5256a46c9feb
+  dynamic.linux-arm64.ami: ami-06f37afe6d4f43c47
   dynamic.linux-arm64.instance-type: m6g.large
   dynamic.linux-arm64.instance-tag: stage-arm64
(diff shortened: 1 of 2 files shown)


hugares (changes requested):
@p8r-the-gr8 Please split this change. First let's start with staging and validate that MPC builds are still working before touching the production clusters.

p8r-the-gr8:
Hi @hugares, I reworked the PR to only touch the staging files.

components/multi-platform-controller/staging-downstream/host-config.yaml
```diff
@@ -406,7 +406,7 @@ data:
 
   dynamic.linux-root-amd64.type: aws
   dynamic.linux-root-amd64.region: us-east-1
-  dynamic.linux-root-amd64.ami: ami-026ebd4cfe2c043b2
```
mshaposhnik, line 409: I think there is an particular reason why the `root`-able machines have an different AMI... (I don't remember it, through...)

PR #8252, opened 2025-09-22
KubeArchive: downscale to 0 on stage-stg-rh01 for tests

rh-hemartin:
We want to do some load tests on Konflux with KubeArchive disabled.


skoved (approved):
/lgtm

PR #8264, opened 2025-09-22
staging: remove sprayproxy from host clusters

sadlerap:
With the completion of [KFLUXINFRA-1790](https://issues.redhat.com//browse/KFLUXINFRA-1790)[^1], we no longer need sprayproxy on the host clusters.

Remove it from the staging overlay.

[^1]: https://issues.redhat.com/browse/KFLUXINFRA-1790

Part of [KFLUXINFRA-2240](https://issues.redhat.com/browse/KFLUXINFRA-2240).

### DIFF
--- a/argo-cd-apps/overlays/production-downstream/delete-applications.yaml
+++ b/argo-cd-apps/overlays/production-downstream/delete-applications.yaml
@@ -1,11 +1,4 @@
 ---
-# Downstream deployment has the host and member operators deployed on the same cluster
-apiVersion: argoproj.io/v1alpha1
-kind: ApplicationSet
-metadata:
-  name: sprayproxy
-$patch: delete
----
 apiVersion: argoproj.io/v1alpha1
 kind: ApplicationSet
 metadata:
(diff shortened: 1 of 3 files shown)


argo-cd-apps/overlays/production-downstream/delete-applications.yaml
```diff
@@ -1,11 +1,4 @@
 ---
-# Downstream deployment has the host and member operators deployed on the same cluster
-apiVersion: argoproj.io/v1alpha1
-kind: ApplicationSet
-metadata:
-  name: sprayproxy
-$patch: delete
----
```
filariow, line 8: this is impacting production. Should we revert it?

filariow (changes requested):

argo-cd-apps/overlays/production-downstream/delete-applications.yaml
sadlerap: No, `sprayproxy` isn't in the `staging-downstream` or `production-downstream` overlays, so that change shouldn't be reverted.  This is just moving the deletion patch from their overlays to the `konflux-public-staging`, which they inherit from.

As for `konflux-public-production`, it isn't impacted by this change.

filariow: thanks for the explanation

PR #8311, opened 2025-09-25
fix(KFLUXVNGD-481): Add argocd permissions to trust-manager

avi-biton:
Add argocd permissions to trust-manager


amisstea (approved):
/lgtm

PR #8284, opened 2025-09-24
Change RHEL AMIs to Cloud Access for prd-rh01

p8r-the-gr8:
Use Red Hat Cloud Access (BYOS) AMIs instead of regular PAYG AMIs that incur licensing fees.
The following changes were made:

ami-026ebd4cfe2c043b2 (RHEL-9.2.0_HVM-20230503-x86_64-41-Hourly2-GP2) -> ami-01aaf1c29c7e0f0af (RHEL-9.6.0_HVM-20250910-x86_64-0-Access2-GP3)

ami-03d6a5256a46c9feb (RHEL-9.2.0_HVM-20230503-arm64-41-Hourly2-GP2) -> ami-06f37afe6d4f43c47 (RHEL-9.6.0_HVM-20250910-arm64-0-Access2-GP3)

This PR continues the work started in https://github.com/redhat-appstudio/infra-deployments/pull/8251 but limits the changes to the prd-rh01 cluster.

PR #8328, opened 2025-09-26
Revert "Change RHEL AMIs to Cloud Access for prd-rh01 (#8284)"

littlejawa:
This commit reverts PR #8284
I understand this has an impact as it is bringing back PAYG AMIs that incur licensing fees.
But it seems that this change has broken the build for our images, and we are currently blocked from doing a release, that needs to happen before Oct 1st.

We have started discussing this on a slack channel, and tried to ping the author and approver (@p8r-the-gr8 and @hugares) on it, but didn't have feedback.
See https://redhat-internal.slack.com/archives/C04PZ7H0VA8/p1758792674077519?thread_ts=1758621485.172029&cid=C04PZ7H0VA8

If no quick alternative is found to fix our problem, I strongly suggest to merge this until a proper fix is found.

Any suggestion is welcome.

## Full rationale
- we have a custom build task that is using one of those VMs (see: https://github.com/confidential-devhub/coco-podvm-scripts/blob/main/task/build-dm-verity-image/0.1/build-dm-verity-image.yaml)
- starting Wednesday Sept. 24th, this task has started to fail systematically
- the change that I am reverting here happened at 6:30pm on that day.
- the first PR that caused a build issue was created at 6:45pm on that day.
  See: https://github.com/confidential-devhub/coco-podvm-scripts/pull/87
- all subsequent build attempts, even on PRs that were passing before, have failed with the same error since then
- there was no code change in our task that can explain the error

The error happens during a "podman build" execution, and reports this message:
```
STEP 2/5: RUN subscription-manager register --org "$(cat /activation-key/org)" --activationkey "$(cat /activation-key/activationkey)"
/bin/sh: error while loading shared libraries: /lib64/libc.so.6: cannot apply additional memory protection after relocation: Permission denied
Error: building at STEP "RUN subscription-manager register --org "$(cat /activation-key/org)" --activationkey "$(cat /activation-key/activationkey)"": while running runtime: exit status 127
```

Based on the error message, we've tried to apply the workaround suggested by https://access.redhat.com/solutions/7021610
See https://github.com/confidential-devhub/coco-podvm-scripts/pull/88
But it didn't work.

In any case, I suspect that the fix should belong to the VM, and not to every task that are running on it.

### DIFF
--- a/components/multi-platform-controller/production/stone-prd-rh01/host-config.yaml
+++ b/components/multi-platform-controller/production/stone-prd-rh01/host-config.yaml
@@ -59,7 +59,7 @@ data:
   # cpu:memory (1:4)
   dynamic.linux-arm64.type: aws
   dynamic.linux-arm64.region: us-east-1
-  dynamic.linux-arm64.ami: ami-06f37afe6d4f43c47
+  dynamic.linux-arm64.ami: ami-03d6a5256a46c9feb
   dynamic.linux-arm64.instance-type: m6g.large
   dynamic.linux-arm64.instance-tag: prod-arm64
   dynamic.linux-arm64.key-name: konflux-prod-ext-mab01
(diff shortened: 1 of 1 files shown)


hugares (approved):
/lgtm
/ok-to-test

PR #8329, opened 2025-09-26
Add pod debugging permissions to mintmaker team

qixiang:
Adds pods/attach, pods/exec, and pods/log subresources to enable debugging capabilities for the konflux-mintmaker-team.


qixiang:
@staticf0x this should allow us rsh into containers for troubleshooting.

PR #8342, opened 2025-09-26
ci: fix kube-linter artifact upload

sadlerap:
For unfathomable reasons, github actions requires a call to failure() for the if condition on artifact upload to be processed correctly. Without it, the condition never evaluates to true and artifact upload is skipped.

Add a call to failure() in the conditional for kube-linter's artifact upload to fix this issue.


hugares (approved):
/lgtm

PR #8351, opened 2025-09-28
Remove Keycloak component

gbenhaim:
The Keycloak component was used when running the old UI and Kubesaw. The old UI and Kubesaw are not being used anymore, thus the Keycloak component can be removed.


hugares (approved):
/lgtm

PR #8352, opened 2025-09-29
Check that LocalQueue CRD exist before policy sync

gbenhaim:
I was observed the syncing the Kyverno policy for creating LocalQueue can have great delay if it starts before the LocalQueue CRD is installed on the cluster.

Add a PreSync hook the will delay the sync of the policies until the LocalQueue CRD is installed on the cluster.

Assisted-By: Cursor

### DIFF
--- a/components/policies/development/kueue/pre-sync-hook.yaml
+++ b/components/policies/development/kueue/pre-sync-hook.yaml
@@ -1,0 +1,88 @@
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: kueue-crd-checker
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: kueue-crd-checker
+rules:
+- apiGroups: ["apiextensions.k8s.io"]
+  resources: ["customresourcedefinitions"]
+  verbs: ["get", "list"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: kueue-crd-checker
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: kueue-crd-checker
+subjects:
+- kind: ServiceAccount
+  name: kueue-crd-checker
+  namespace: will-be-replaced-by-kustomization
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: kueue-crd-check
+  annotations:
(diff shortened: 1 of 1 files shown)


gbenhaim:
@gemini-cli /review

components/policies/development/kueue/pre-sync-hook.yaml
```diff
@@ -0,0 +1,104 @@
...
+              exit 1
+            fi
+
+            echo "Required CRDs not found yet. Waiting ${INTERVAL}s... (${ELAPSED}s/${TIMEOUT}s elapsed)"
+            sleep $INTERVAL
+          done
+
+          echo "SUCCESS: All required CRDs found!"
+          echo "Pre-sync hook completed successfully."
+          exit 0
```
filariow, line 96: To simplify this logic, I'd suggest to:
* use `kubectl wait --for=create` instead of polling on `get`
* scope to just a single CustomResourceDefinition
* take `TIMEOUT` and `CRD` from configuration

```suggestion
        env:
        - name: CRD
          value: localqueues.kueue.x-k8s.io
        command:
        - /bin/bash
        - -c
        - |
          TIMEOUT=${TIMEOUT:-300}  # default to 5 minutes timeout
          
          echo "Checking CRD ${CRD} exists..."
          kubectl wait --for=create "${CRD}" --timeout ${TIMEOUT}
```

components/policies/development/kueue/pre-sync-hook.yaml
```diff
@@ -0,0 +1,104 @@
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: kueue-crd-checker
+  annotations:
+    argocd.argoproj.io/hook: PreSync
+    argocd.argoproj.io/hook-delete-policy: BeforeHookCreation
```
filariow, line 8: Should we use `HookSucceeded` instead? It seems to me that with `BeforeHookCreation` this resources will always be around, just updated before the next wave. I'd rather prefer them to be gone if everything applied correctly

https://argo-cd.readthedocs.io/en/stable/user-guide/sync-waves/#hook-lifecycle-and-cleanup

filariow (changes requested):

components/policies/development/kueue/pre-sync-hook.yaml
filariow: thanks for the explanation

PR #8355, opened 2025-09-29
KubeArchive: eventing-free version on stone-stage-p01

rh-hemartin:
Changing KubeArchive in stone-stage-p01 to test a new version of KubeArchive without disturbing UI development on stone-stg-rh01.


rh-hemartin:
> I don't see the kubearchive-operator-config ConfigMap with the use-knative field. Is this because the version being deployed is hardcoded directly without having knative?

This a new branch Greg made where he directly removed knative-eventing, so no ConfigMap or flag to toggle Knative Eventing on-off.

PR #8362, opened 2025-09-29
kflux-ocp-p01: fix object too large errors

olegbet:

rh-pre-commit.version: 2.3.2


maruiz93 (changes requested):
this won't fix the problem, the issue is with the cache, the max_line_size in loki isn't the issue as vector is configured to send lines of a maximum of 1 MB

PR #8365, opened 2025-09-29
feat(PVO11Y-4928): Implement Otel sidecar for Konflux UI

martysp21:
[KONFLUX-5285](https://issues.redhat.com//browse/KONFLUX-5285)
This is an implementation of OTEL sidecar for Konflux UI Nginx proxy in order to collect all necessary HTTP traffic data and convert them into metrics. These metrics will be then used to create alert for inaccessibility. 

**The otel image will change before this PR is merged to the [otel-collector](https://github.com/release-engineering/otel-collector-sp)-sp after it is updated**

Changes:
nginx.conf
- added a new logformat that will be saved into access.log
- The new logformat is the same as the default ngingx format "combined", I redefined it for better readability and easier future maintenance
- access.log is located at /var/log/nginx/access.log and stores all the logs of in the combined_cusotm format

otel-collector-conf.yaml
- configuration for the otel collector
- takes logs from the access.log and parses them using regex
- the data are then transformed into metrics
- this is the main file to change how the output lookslike and will probably need most of the refinement

proxy.yaml
- added sidecar to proxy deployment
- sidecar uses the otel collector image 
- mounts the nginx logs volume
- Added a new port in the Service configuration to allow access to the metrics port of the otel collector by the prometheus
- Added Service monitor to define scraping definition for Prometheus using UWM -- More at: https://github.com/redhat-appstudio/infra-deployments/tree/main/components/monitoring/prometheus

The changes can be also found at: https://docs.google.com/document/d/1RTvHJl90rvbVMxM6a8ag9dWTGTcTIow_zapaxCYEmzE/edit?tab=t.0


martysp21:
The tests run in the Draft of this PR do not seem to be related with the changes made by this PR.
Asking for advice from Konflux Ui when the PR will be opened.

components/konflux-ui/staging/base/proxy/proxy.yaml
```diff
@@ -150,6 +150,44 @@ spec:
           readOnlyRootFilesystem: true
           runAsNonRoot: true
           runAsUser: 1001
+      - image: otel/opentelemetry-collector-contrib:0.85.0
```
raks-tt, line 153: we should use the same image that we are using in the konflux -> https://github.com/redhat-appstudio/infra-deployments/blob/main/components/tracing/otel-collector/production/otel-collector-helm-values.yaml#L94

components/konflux-ui/staging/base/proxy/proxy.yaml
```diff
@@ -150,6 +150,44 @@ spec:
...
+        name: otel-collector
+        command:
+        - "/otelcol-contrib"
+        - "--config=/conf/otel-collector-conf.yaml"
```
raks-tt, line 157: This means, the command will be -> /usr/local/bin/otel-collector-sp --config /etc/otelcol-contrib/config.yaml.

Can we please test this @martysp21 locally.

martysp21: Changed and tested. Please double check that it is mounted correctly, because that part is different from local testing.

raks-tt (approved):
LGTM

mftb (approved):
/lgtm
/approve

PR #8401, opened 2025-09-30
KAR-622: setup kflux-osp-p01 kubearchive logging config

olegbet:

rh-pre-commit.version: 2.3.2


maruiz93 (changes requested):
Please make sure you also add the external secret and the configmap for kubearchive-logging in the `kubearchive` konflux component

components/kubearchive/production/kflux-osp-p01/kustomization.yaml
maruiz93, line 12: you need to add also the configmap kubearchive-logging, as well as delete the already provided `kubearchive-logging` ConfigMap and Secret. Please taka a look at other clusters like the following: https://github.com/redhat-appstudio/infra-deployments/blob/main/components/kubearchive/production/kflux-ocp-p01/kustomization.yaml#L15-L47

PR #8414, opened 2025-09-30
Allow bigger chunks of logs in loki

maruiz93:
With the recent enabled logs we are now seeing the issue that was causing some logs to be missing:

```	
level=debug ts=2025-09-30T15:25:51.87843779Z caller=http.go:108 org_id=kubearchive msg="push request failed" code=500 err="rpc error: code = ResourceExhausted desc = grpc: received message larger than max (10361786 vs. 4194304)"
```

I'm updating the GRPC max size limits for server and client following the config shown in https://grafana.com/docs/loki/latest/configure/#grpc_client

Note that in vector we have the maximum chunk size to be sent to loki set in 10MB, so that is the main issue. If we find problems with this setup we will need to reduce the chunks sent from vector to loki.


olegbet (approved):
/lgtm

PR #8417, opened 2025-09-30
feat(SPRE-1268): updated monitoringstack endpoints for kube_pod_container_status_terminated_reason

peet-rh:
this is so we can check for OOM events from multiple prod namespaces and not just one


TominoFTW (approved):
/lgtm
/approve

PR #6449, opened 2025-05-27
Add testplatform ProviderConfig in Staging

danilo-gemoli:
Same as https://github.com/redhat-appstudio/infra-deployments/pull/6446 but for the staging environment.

/cc @amisstea

PR #8425, opened 2025-10-01
Add testplatform ProviderConfig in Production

danilo-gemoli:
Production counterpart for https://github.com/redhat-appstudio/infra-deployments/pull/6449

/cc @amisstea


components/crossplane-control-plane/production/testplatform-provider-config.yaml
amisstea, line 1: We're missing a reference to this file in `kustomization.yaml`.

PR #8430, opened 2025-10-01
Set common-secret=true on e2e-test pull secret

chmeliik:
The 'registry-redhat-io-pull-secret' is needed for e2e test pipelines that pull base images from registry.redhat.io. Previously, this secret was linked to the 'appstudio-pipeline' Service Account.

After recent changes that remove appstudio-pipeline compatiblity, it has become necessary to turn this secret into a "common" secret so that it gets linked to all Service Accounts created for e2e tests.


chmeliik:
@gemini-cli /review Check if this is valid ExternalSecret syntax. Provide sources.

PR #8434, opened 2025-10-01
Make host-config dynamically provisioned (part 2 - internal prod)

mshaposhnik:
This PR moves the approach from handwritten host-config.yaml to a generated ones using Helm templating.

We will have a sinlge template which contains all existing dynamic platforms configuration with theis default configuration values, as far as the rest of the fields (instancetags, etc) with default values too.
Typically, this template does not require manual editing, except the adding/removing flavours or amending default values for the existing VM flavours.

Each cluster overlay will contain the host-values file, wich configured (once for all flavours) the defult values for dynamic parameters (such as AMI, ssh key etc) and have a list of dynamic flavors and static machines. It is also allows to override any field from general template (like instance tags etc) and any property of any dynamic flavour machine, if needed.

In this PR, we do not equalize the list of supported VM flavours for all clusters.
They are still genegated dynamically per-cluster by collecting the all keys from dynamicConfigs section of the values file.

If you wanna see diff, you can generate config locally (paths are considered being in cluster folder):
```
helm template ../../base/host-config-chart/   --namespace multi-platform-controller   -f host-values.yaml > out.yaml
```


mshaposhnik:
@oswcab I acknowledged your comments, thise user-scripts worth reviewing, I will do that separately and centralizedly for all platforms.

PR #8441, opened 2025-10-02
Allow larger items in mem cache

maruiz93:
This PR tries to solve the issue seen in several clusters with these messages in the ingester logs:

level=warn ts=2025-10-01T17:47:34.128244414Z caller=background.go:200 msg="backgroundCache writeBackLoop Cache.Store fail" err="server=192.168.188.6:11211: memcache: unexpected response line from \"set\": \"SERVER_ERROR object too large for cache\\r\\n\""

This problem doesn't prevent the archival of logs but it does decrease the performance


olegbet (approved):
/lgtm

PR #8462, opened 2025-10-03
Aplly Validating Admission Policy on repository validation on stg

mshaposhnik:
First attempt to replace repositoy Validator webhook with `ValidatingAdmissionPolicy` objects (staging only)


components/repository-validator/base/kustomization.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - validating-admission-policy.yaml
+  - validating-admission-policy-binding.yaml
+
+namespace: repository-validator
```
hugares, line 7: this is causing to apply a NS on the cluster scoped resources, i.e. ValidatingAdmissionPolicy and ValidatingAdmissionPolicyBinding. Remove that from here and from the kustomization files in staging folder.

hugares (changes requested):

components/repository-validator/base/kustomization.yaml
mshaposhnik: Good catch, fixed

PR #8470, opened 2025-10-03
Adjust the cache proxy settings in dev and staging

amisstea:


amisstea:
This requires a new release of the caching proxy helm chart which won't happen until Monday. Keeping this in draft state until then.

hmariset (approved):
LGTM

PR #8475, opened 2025-10-06
KubeArchive: using watchers instead of informers

rh-hemartin:
We changed KubeArchive to use informers instead of Knative Eventing, but informers consumed too much memory. We are switching to watchers that do not have cache to see if we can reduce memory footprint.

### DIFF
--- a/components/kubearchive/development/kubearchive.yaml
+++ b/components/kubearchive/development/kubearchive.yaml
@@ -5,7 +5,7 @@ metadata:
     app.kubernetes.io/component: namespace
     app.kubernetes.io/name: kubearchive
     app.kubernetes.io/part-of: kubearchive
-    app.kubernetes.io/version: no-eventing-1a13a90
+    app.kubernetes.io/version: watchers-539215e
   name: kubearchive
 ---
 apiVersion: apiextensions.k8s.io/v1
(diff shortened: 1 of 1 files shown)


components/kubearchive/development/kubearchive.yaml
```diff
@@ -1552,18 +1558,13 @@ spec:
     spec:
       containers:
         - args:
-            - set -o errexit;
-              git clone https://github.com/kubearchive/kubearchive --depth=1 --branch=${KUBEARCHIVE_VERSION} /tmp/kubearchive;
-              cd /tmp/kubearchive;
-              export QUOTED_PASSWORD=$(python3 -c "import urllib.parse; print(urllib.parse.quote('${DATABASE_PASSWORD}', ''))");
-              curl --silent -L https://github.com/golang-migrate/migrate/releases/download/${MIGRATE_VERSION}/migrate.linux-amd64.tar.gz | tar xvz migrate;
-              ./migrate -verbose -path integrations/database/postgresql/migrations/ -database postgresql://${DATABASE_USER}:${QUOTED_PASSWORD}@${DATABASE_URL}:${DATABASE_PORT}/${DATABASE_DB} up
+            - set -o errexit; git clone https://github.com/kubearchive/kubearchive --depth=1 --branch= /tmp/kubearchive; cd /tmp/kubearchive; export QUOTED_PASSWORD=$(python3 -c "import urllib.parse; print(urllib.parse.quote('', ''))"); curl --silent -L https://github.com/golang-migrate/migrate/releases/download//migrate.linux-amd64.tar.gz | tar xvz migrate; ./migrate -verbose -path integrations/database/postgresql/migrations/ -database postgresql://:@:/ up
```
maruiz93, line 1561: why this change?

PR #8498, opened 2025-10-06
kfluxinfra-2359 - Add instance with bigger disk to 'stone-prod-p02'

oswcab:
As part of the investigation of a user ticket where the VM seems to have run out of space while building [1],one of the agreed debug actions is to provide a VM with bigger disk to see if this fixes their problem. If it does, we would probably create a new instance with this amount of disk space but in a type m instance, as they're using now.

[1] https://redhat-internal.slack.com/archives/C04PZ7H0VA8/p1759754081471249?thread_ts=1759330105.097669&cid=C04PZ7H0VA8


sadlerap (approved):
/lgtm

PR #8515, opened 2025-10-07
Revert "KubeArchive: using watchers instead of informers"

rh-hemartin:
Reverts redhat-appstudio/infra-deployments#8475 as the implementation went crazy logging some issues. I'm not sure what is happening but it is logging 72k lines per minute, so I will revert it.


rh-hemartin:
I will move forward instead and add a debug statement

PR #8524, opened 2025-10-07
Fix CEL expression in validation admission policy

mshaposhnik:
Fix CEL expression in validation admission policy


hugares (approved):
/lgtm

PR #8536, opened 2025-10-07
Use just list of items instead of json for validator

mshaposhnik:
It tuns out that CEL expression cannot decode the json from  data field in config map...


hugares (approved):
/lgtm

PR #8542, opened 2025-10-08
feat: adding konflux UI whatson env variables

StanislavJochman:


enkeefe00 (approved):

components/konflux-ui/staging/stone-stage-p01/kustomization.yaml
```diff
@@ -13,6 +13,9 @@ configMapGenerator:
     files:
       - kite.conf
     behavior: merge
+  - name: ui-runtime-env-variables
```
gbenhaim: Should it be part of the konflux-info config map?

components/konflux-ui/staging/base/proxy/proxy.yaml
```diff
@@ -52,8 +52,59 @@ spec:
...
+            # Generate runtime-config.js with PUBLIC_ variables
+            out=/mnt/static-content/runtime-config.js
+            printf "window.KONFLUX_RUNTIME = window.KONFLUX_RUNTIME || {};\n" > "$out"
+            env | grep -E '^PUBLIC_' | while IFS='=' read -r name value; do
+              esc=$(printf '%s' "$value" | sed -e 's/\\\\/\\\\\\\\/g' -e 's/\"/\\\"/g')
+              printf 'window.KONFLUX_RUNTIME.%s = "%s";\n' "$name" "$esc" >> "$out"
+            done
+
+            # Apply cluster-specific variables from ConfigMap files to runtime-config.js, if present
+            if [ -d /mnt/ui-runtime-env-variables ]; then
```
gbenhaim: I would prefer to keep the secret with only secret data and to store the rest of the data in the konflux-info config map.

gbenhaim:
How do we ensure users doesn't abuse the shared token? I guess there is q limit to the amount of requests we can do to the backend. I believe that with the current approach a user can consume all the credits.

components/konflux-ui/staging/base/proxy/proxy.yaml
```diff
@@ -52,8 +52,55 @@ spec:
           requests:
             cpu: 10m
             memory: 64Mi
+      - name: render-runtime-config
+        image: registry.access.redhat.com/ubi9/ubi@sha256:66233eebd72bb5baa25190d4f55e1dc3fff3a9b77186c1f91a0abdb274452072
+        envFrom:
```
gbenhaim: this should be removed. reading the secret shouldn't be needed here.

components/konflux-ui/staging/base/external-secret.yaml
```diff
@@ -0,0 +1,28 @@
+---
+apiVersion: external-secrets.io/v1beta1
+kind: ExternalSecret
+metadata:
+  name: watson-config
+  namespace: konflux-ui
```
gbenhaim: you would need to add the `konflux-ui` namespace to the allow list in the ClusterSecretStore https://github.com/redhat-appstudio/infra-deployments/blob/main/components/cluster-secret-store/base/appsre-stonesoup-vault-secret-store.yaml#L29

components/konflux-ui/staging/base/proxy/proxy.yaml
```diff
@@ -66,13 +113,21 @@ spec:
 
             chmod 640 /mnt/nginx-generated-config/auth.conf
 
+            # Generate watson-auth.conf with Basic auth from Watson API key
+            watson_basic=$(cat /mnt/watson-api-key/API_KEY)
```
sahil143: Add the base64 encoding and append it to `apikey: <base64-token>` that will be your watson_basic.

PR #8556, opened 2025-10-08
fix(PVO11Y-4928): Remove run as user in kustomize

martysp21:
After merged PR https://github.com/redhat-appstudio/infra-deployments/pull/8365
There was a problem with argoCD not allowing runAsUser directive, but this directive is enforced by GitLab linter. 
I need to update files to remove that directive when kustomize applies


hugares (approved):
/lgtm

PR #8740, opened 2025-10-20
full read permissions on RoleBindings

filariow:
This change provides full read access to RoleBindings to all personas which were already allowed to `list` RoleBindings.


gbenhaim:
Why those extra permissions are needed?

PR #10850, opened 2026-03-12
add tokens in tekton-kueue

filariow:
This proposal adds the new ephemeral resource `konflux-ci.dev/token` in tekton-kueue. This is to weight PipelineRuns by known patterns in tekton-kueue and better balance them.

As an example, we noticed that Build PipelineRuns are usually heavier than Non-Build PipelineRuns, so we can identify Build PipelineRuns and mark them to require 2 tokens. The other PipelineRuns will only take 1 instead.

Assuming both `tekton.dev/pipelineruns` and `konflux-ci.dev/token` are 600, we could have in execution:
* 300 Build PipelineRuns
* 200 Build PipelineRuns + 200 Non-Build PipelineRuns
* 100 Build PipelineRuns + 400 Non-Build PipelineRuns
* 600 Non-Build PipelineRuns


gbenhaim (approved):

components/kueue/development/tekton-kueue/config.yaml
```diff
@@ -92,6 +92,31 @@ cel:
...
+    # * 100 Build PipelineRuns + 400 Non-Build PipelineRuns
+    # * 600 Non-Build PipelineRuns
+    - |
+        pacEventType == 'push' ||
+          pacEventType == 'pull_request' ||
+          pacEventType == "Merge_Request" ||
+          pacEventType == 'test-comment' ||
+          pacEventType == 'retest-comment' ||
+          pacEventType == 'retest-all-comment' ||
+          pacEventType == 'ok-to-test-comment'  ? resource('konflux-ci-dev-token', 2) :
```
gbenhaim, line 117: @filariow I now recall the issue the HB team has is that they have a lot of small test pipelines. With this approach we give the same wight for both build and test pipelines. Does it make sense to give less wight for test pipelines?

gbenhaim: @filariow can you please check this comment?

filariow: this configuration is just for the development and staging overlays. We can leave it as is here and have a different one in each prod cluster tailored on their specific need. As an example, in the case of HB we can set:

* tokens: 1000
* test pipelines: 1
* build pipelines: 5

sadlerap (approved):

filariow:
/hold

looking at Qodo's feedback

sadlerap (approved):

PR #10940, opened 2026-03-18
kueue: update checks for tenant-release priority in prod

sadlerap:
The checks for a release pipeline in a tenant namespace are too strict, since users can see the `pipelines.appstudio.openshift.io/type` label to take values such as `tenant-collectors` or `final`.  These pipelines should be assigned the konflux-tenant-release priority instead of falling through to the default priority.

Furthermore, there's a better check for a release pipeline in a tenant: the value of the `release.appstudio.openshift.io/namespace` label should match the namespace of the pipelinerun.  Release pipelines running in the dedicated release namespace will fail this check.

Update the conditions for the `konflux-tenant-release` priority to only check for the namespace condition as well as the release service label.

* * *

Production version of #10902.


sadlerap:
/hold

Waiting for some additional testing before merging.

gbenhaim:
@sadlerap what is the next step for this pr?

PR #10994, opened 2026-03-23
Add Tekton infra noise filters to staging

raks-tt:
Filter transient reconcilation errors that resolve automatically via retry mechanism.

This is to filter the transient reconcilation errors that resolve automatically.

### DIFF
--- a/components/monitoring/logging/staging/base/patches/configure-logforwarder-filters-patch.yaml
+++ b/components/monitoring/logging/staging/base/patches/configure-logforwarder-filters-patch.yaml
@@ -36,9 +36,13 @@
     type: drop
     drop:
       - test:
+          - field: .kubernetes.namespace_name
+            matches: 'openshift-pipelines|tekton-.*'
           - field: .message
             matches: '(?i)Operation cannot be fulfilled on (pipelineruns?|taskruns?)\.tekton\.dev'
       - test:
+          - field: .kubernetes.namespace_name
+            matches: 'openshift-pipelines|tekton-.*'
           - field: .message
             matches: '(?i)Reconcile error'
           - field: .message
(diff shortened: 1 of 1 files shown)


pacho-rh:
These tests should be narrowed down specifically to tekton logs. Thinking including a match on the originating namespace for test should do.
Something like:
```yaml
...
    drop:
      - test:
          - field: .kubernetes.namespace_name
            matches: "openshift-pipelines|tekton-.*"
          - field: .message
            matches: '(?i)Operation cannot be fulfilled on (pipelineruns?|taskruns?)\.tekton\.dev'
...
```

components/monitoring/logging/staging/base/patches/configure-logforwarder-filters-patch.yaml
```diff
@@ -27,3 +27,87 @@
...
+- op: add
+  path: /spec/pipelines/0/filterRefs/-
+  value: drop-finalizer-not-found
+- op: add
+  path: /spec/pipelines/0/filterRefs/-
+  value: drop-resource-request-in-progress
+- op: add
+  path: /spec/pipelines/0/filterRefs/-
+  value: drop-tekton-label-update-failures
+# audit index filters (openshift-pipelines, tekton-results infrastructure namespaces)
```
pacho-rh: The infrastructure logs (included in /spec/pipelines/1) does not include tekton-.* namespaces. Here is the doc that outlines what namespaces are captured by it: https://docs.redhat.com/en/documentation/red_hat_openshift_logging/6.1/html-single/configuring_logging/index#clf-inputs_configuring-log-forwarding

Let's move the mention of that namespace under the comment for the application index filters (line 88) to avoid confusion.

raks-tt: Right it is in the application index.

martysp21 (approved):
/lgtm

PR #11036, opened 2026-03-26
Fix incorrect probe URLs for kflux-osp-p01 cluster

merlyel:
Fix incorrect probe URLs for kflux-osp-p01 cluster causing false alarms

Changes between two commits (fda35526ed8548d1c9b32bc128d903cb2a9b6c6a and 4a2a032cf45fb1a64fefd7e39f222f74af7f114c) include fixes in README, changes in stage clusters probes, and adding kflux-fedora-01 cluster probes. So the change should affect only probes of kflux-osp-p01


merlyel:
@eedri "external repo" is the expected and long-time used one (`internal-infra-deployments`).
As for those referenced changes, they include my fixes for probe URLs for `kflux-osp-p01` cluster, fix for probes of public clusters, and changes to `kflux-fedora-01` cluster.  But this PR includes the changes only for `kflux-osp-p01` cluster (`components/monitoring/blackbox/production/kflux-osp-p01/kustomization.yaml` file), so other cluster will not be affected

eisraeli (approved):

PR #11039, opened 2026-03-26
policies: don't remove generated resources on deletion

sadlerap:
If a generation policy needs to be deleted (which happens anytime we need to update it, since they're immutable), kyverno by default will delete all generated resources.  In pretty much all circumstances, these deletions will cause disruption for users and break some assumptions some components have.

However, kyverno does support leaving these generated resources in place with the `generate.orphanDownstreamOnPolicyDelete` field set to `true`.  Set that field for all generation policies in the staging/development overlays.

Fixes: [KFLUXINFRA-3318](https://redhat.atlassian.net/browse/KFLUXINFRA-3318)

[KFLUXINFRA-3318]: https://redhat.atlassian.net/browse/KFLUXINFRA-3318?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/components/policies/development/integration/bootstrap-namespace/.chainsaw-test/chainsaw-test.yaml
+++ b/components/policies/development/integration/bootstrap-namespace/.chainsaw-test/chainsaw-test.yaml
@@ -327,3 +327,74 @@ spec:
         expect:
         - check:
             ($error != null): true
+---
+# yaml-language-server: $schema=https://raw.githubusercontent.com/kyverno/chainsaw/main/.schemas/json/test-chainsaw-v1alpha1.json
+apiVersion: chainsaw.kyverno.io/v1alpha1
+kind: Test
+metadata:
+  name: orphan-resources-on-policy-deletion
+spec:
+  description: |
+    tests that resources created by this policy remain when the policy is deleted
+  concurrent: false
+  namespace: 'generate-new-namespace'
+  bindings:
+  - name: suffix
+    value: orphan
+  steps:
+  - name: given-konflux-integration-runner-clusterrole-exists
+    try:
+    - apply:
+        file: resources/actual-konflux-integration-runner-clusterrole.yaml
+  - name: given-kyverno-has-permission-on-resources
+    try:
+    - apply:
+        file: ../kyverno-rbac.yaml
+  - name: given-cluster-policy-is-ready
+    try:
+    - apply:
+        file: ../bootstrap-namespace.yaml
+    - assert:
+        file: chainsaw-assert-clusterpolicy.yaml
+  - name: when-konfluxcidev-labeled-namespace-is-created
+    try:
+    - apply:
+        file: resources/actual-namespace-konfluxcidev.yaml
+        template: true
+  - name: then-serviceaccount-is-created
+    try:
+    - assert:
+        file: resources/expected-integration-serviceaccount.yaml
+        template: true
+  - name: then-rolebinding-is-created
+    try:
+    - assert:
+        file: resources/expected-integration-rolebinding.yaml
+        template: true
+  - name: when-the-policy-is-deleted
+    try:
+    - delete:
+        file: ../bootstrap-namespace.yaml
+    - wait:
+        apiVersion: kyverno.io/v1
+        kind: ClusterPolicy
+        name: init-ns-integration
+        for:
+          deletion: {}
+    - script:
+        content: |
+          while true; do
+            if [ "$(kubectl get updaterequests -n konflux-kyverno -o yaml | yq '.items | length')" -eq 0 ]; then
+              exit 0
+            fi
+            sleep 1
+          done
+        timeout: 30s
+  - name: then-the-resouces-remain
+    try:
+    - assert:
+        file: resources/expected-integration-serviceaccount.yaml
+        template: true
+    - assert:
+        file: resources/expected-integration-rolebinding.yaml
+        template: true
--- a/components/policies/development/konflux-rbac/bootstrap-tenant-namespace/.chainsaw-test/chainsaw-test.yaml
+++ b/components/policies/development/konflux-rbac/bootstrap-tenant-namespace/.chainsaw-test/chainsaw-test.yaml
@@ -314,3 +314,85 @@ spec:
     - assert:
         file: resources/actual-existing-resources.yaml
         template: true
+---
+# yaml-language-server: $schema=https://raw.githubusercontent.com/kyverno/chainsaw/main/.schemas/json/test-chainsaw-v1alpha1.json
+apiVersion: chainsaw.kyverno.io/v1alpha1
+kind: Test
+metadata:
+  name: orphan-resources-on-policy-deletion
+spec:
+  description: |
+    tests that the resources remain in a tenant namespace when the cluster policy is deleted
+  concurrent: false
+  namespace: 'orphan-resources'
+  bindings:
+  - name: suffix
+    value: orphan
+  steps:
+  - name: given-appstudio-pipeline-clusterrole-exists
+    try:
+    - apply:
+        file: ./resources/actual-appstudio-pipeline-clusterrole.yaml
+  - name: given-kyverno-has-permission-on-resources
+    try:
+    - apply:
+        file: ../kyverno_rbac.yaml
+  - name: given-cluster-policies-are-ready
+    try:
+    - apply:
+        file: ../bootstrap-tenant-namespace-np-ocpconsole-clusterpolicy.yaml
+    - apply:
+        file: ../bootstrap-tenant-namespace-np-ocpingress-clusterpolicy.yaml
+    - apply:
+        file: ../bootstrap-tenant-namespace-np-ocpmonitoring-clusterpolicy.yaml
+    - apply:
+        file: ../bootstrap-tenant-namespace-np-olm-clusterpolicy.yaml
+    - apply:
+        file: ../bootstrap-tenant-namespace-np-samenamespace-clusterpolicy.yaml
+    - apply:
+        file: ../bootstrap-tenant-namespace-rbcm-clusterpolicy.yaml
+    - assert:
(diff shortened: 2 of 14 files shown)


components/policies/development/konflux-rbac/bootstrap-tenant-namespace/.chainsaw-test/chainsaw-test.yaml
```diff
@@ -314,3 +314,76 @@ spec:
...
+    - wait:
+        apiVersion: kyverno.io/v1
+        kind: ClusterPolicy
+        for:
+          deletion: {}
+  - name: then-the-resources-remain
+    try:
+    - assert:
+        file: resources/expected-resources.yaml
+        template: true
```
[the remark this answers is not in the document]
sadlerap: @filariow fixed in 88b0ed1.

hugares (approved):
/lgtm

components/policies/development/kueue/queue-config/.chainsaw-test/chainsaw-test.yaml
```diff
@@ -397,3 +397,81 @@ spec:
...
+    - script:
+        content: |
+          while true; do
+            if [ "$(kubectl get updaterequests -n konflux-kyverno -o yaml | yq '.items | length')" -eq 0 ]; then
+              exit 0
+            else
+              sleep 1
+            fi
+          done
+        timeout: 30s
```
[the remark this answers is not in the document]
enkeefe00: @sadlerap is this a valid concern?

components/policies/development/integration/bootstrap-namespace/.chainsaw-test/chainsaw-test.yaml
```diff
@@ -327,3 +327,75 @@ spec:
...
+    - wait:
+        apiVersion: kyverno.io/v1
+        kind: ClusterPolicy
+        name: init-ns-integration
+        for:
+          deletion: {}
+    - script:
+        content: |
+          while true; do
+            if [ "$(kubectl get updaterequests -n konflux-kyverno -o yaml | yq '.items | length')" -eq 0 ]; then
```
filariow, line 387: this introduces a dependency on an external tool (`yq`). Do we have any kyverno chainsaw native way to do that?

components/policies/development/integration/bootstrap-namespace/.chainsaw-test/chainsaw-test.yaml
```diff
@@ -327,3 +327,75 @@ spec:
...
+              exit 0
+            else
+              sleep 1
+            fi
```
filariow: I think we can simplify this

```suggestion
            fi
            sleep 1
```

hugares (approved):
/lgtm

components/policies/development/integration/bootstrap-namespace/.chainsaw-test/chainsaw-test.yaml
```diff
@@ -327,3 +327,75 @@ spec:
...
+metadata:
+  name: orphan-resources-on-policy-deletion
+spec:
+  description: |
+    tests that resources created by this policy remain when the policy is deleted
+  concurrent: false
+  namespace: 'generate-new-namespace'
+  bindings:
+  - name: suffix
+    value: konfluxcidev
```
filariow: It looks like we forgot to update the namespace `suffix`


```suggestion
  namespace: 'generate-new-namespace'
  bindings:
  - name: suffix
    value: orphan
```

components/policies/development/kueue/queue-config/.chainsaw-test/chainsaw-test.yaml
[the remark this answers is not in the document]
sadlerap: Qodo is wrong here, yq is included in the base image for github actions: https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md

components/policies/development/integration/bootstrap-namespace/.chainsaw-test/chainsaw-test.yaml
```diff
@@ -327,3 +327,75 @@ spec:
...
+    - script:
+        content: |
+          while true; do
+            if [ "$(kubectl get updaterequests -n konflux-kyverno -o yaml | yq '.items | length')" -eq 0 ]; then
+              exit 0
+            else
+              sleep 1
+            fi
+          done
+        timeout: 30s
```
filariow, line 392: if any other test is executed concurrently this could become flaky -that's not the case as of today. Btw, I think we can use UpdateRequests' labels to restrict to the UpdateRequests for the ClusterPolicy we deleted.

components/policies/development/konflux-rbac/bootstrap-tenant-namespace/.chainsaw-test/chainsaw-test.yaml
```diff
@@ -314,3 +314,86 @@ spec:
...
+    - delete:
+        file: ../bootstrap-tenant-namespace-np-samenamespace-clusterpolicy.yaml
+    - delete:
+        file: ../bootstrap-tenant-namespace-rbcm-clusterpolicy.yaml
+    - wait:
+        apiVersion: kyverno.io/v1
+        kind: ClusterPolicy
+        for:
+          deletion: {}
+    - script:
```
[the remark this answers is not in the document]
sadlerap: By construction, these are the only cluster policies on the cluster, so waiting for all active deletions to succeed is enough here.

components/policies/development/integration/bootstrap-namespace/.chainsaw-test/chainsaw-test.yaml
filariow, line 387: AIU how kyverno works, we have no great way to check if the ClusterPolicy was cleaned-up or not. Checking now if the UpdateRequests are 0 could not work well all the time because it could take some time to Kyverno to create them after the ClusterPolicy is removed from etcd. That's because it seems to me Kyverno allows you to remove the ClusterPolicy from k8s immediately, and use a cached version of the policy to run the cleanup operation. For the scope of this test I think it could work, but maybe we can think of a better approach.

components/policies/development/kueue/queue-config/.chainsaw-test/chainsaw-test.yaml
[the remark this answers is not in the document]
filariow: yq is preinstalled in github runners. We can assume it's installed already until we rely on GitHub Runners 😄

components/policies/development/integration/bootstrap-namespace/.chainsaw-test/chainsaw-test.yaml
filariow: Or at least execute this check consistently over some time. Like expecting it to pass 5 consecutive times with 1s interval.

filariow (changes requested):

components/policies/development/integration/bootstrap-namespace/.chainsaw-test/chainsaw-test.yaml
sadlerap: The purpose of checking for updaterequests is to check that kyverno isn't still processing anything, not necessarily to check for the cleanup of the cluster policy.  This is needed to ensure that we aren't racing with kyverno when we check for the existence of the resource later in the test.

components/policies/development/integration/bootstrap-namespace/.chainsaw-test/chainsaw-test.yaml
sadlerap: All of these tests are setup to not support concurrency anyways, since we're deleting the cluster policy (which is sure to cause the other tests to fail).

components/policies/development/integration/bootstrap-namespace/.chainsaw-test/chainsaw-test.yaml
sadlerap: Not to my knowledge, no.  I'm trying to enforce that kyverno is no longer processing anything, and this is the best I could come up with.

components/policies/development/integration/bootstrap-namespace/.chainsaw-test/chainsaw-test.yaml
sadlerap: Fixed in cb3bdaf66.

components/policies/development/integration/bootstrap-namespace/.chainsaw-test/chainsaw-test.yaml
sadlerap: Fixed in cb3bdaf66.

filariow (approved):

components/policies/development/konflux-rbac/konflux-support-access/.chainsaw-test/chainsaw-test.yaml
```diff
@@ -384,3 +384,169 @@ spec:
...
+    - name: Delete the cluster policy
+      try:
+        - delete:
+            file: ../generate-support-clusterrolebinding-clusterpolicy.yaml
+        - wait:
+            apiVersion: kyverno.io/v1
+            kind: ClusterPolicy
+            name: generate-konflux-support-crb
+            for:
+              deletion: {}
```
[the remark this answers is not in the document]
filariow: @sadlerap fya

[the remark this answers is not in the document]
sadlerap: Fixed in 2cff8aee9

PR #11076, opened 2026-03-31
fix: OIDC auth, add patch coverage config

psturc:
* Switch from pull_request_target with token to pull_request with OIDC, eliminating the security risk of running untrusted PR code with secrets
* Add push trigger on main for baseline coverage
* Enable explicit patch coverage status and PR comment layout in codecov.yml.

### DIFF
--- a/.github/workflows/codecov.yaml
+++ b/.github/workflows/codecov.yaml
@@ -39,7 +39,7 @@ jobs:
     permissions:
       id-token: write
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v6
 
       - name: Setup Python
         uses: actions/setup-python@v4
--- a/.github/workflows/codecov.yaml
+++ b/.github/workflows/codecov.yaml
@@ -10,7 +10,7 @@ jobs:
     permissions:
       id-token: write
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v6
 
       - name: Setup Go
         uses: actions/setup-go@v5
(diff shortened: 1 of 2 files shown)


eedri:
@eisraeli @filariow can you help review?

.github/workflows/codecov.yaml
```diff
@@ -1,12 +1,16 @@
...
+  pull_request:
     types: [opened, synchronize, reopened]
 jobs:
   go-tests:
     runs-on: ubuntu-latest
+    permissions:
+      id-token: write
     steps:
-      - uses: actions/checkout@main
+      - uses: actions/checkout@v4
```
filariow: V4 is using a deprecated version of nodejs. Can we use V6 instead?

```suggestion
      - uses: actions/checkout@v6
```

filariow (changes requested):

psturc:
Hi @eedri @filariow can you please re-review and approve when you get a chance? thanks!

.github/workflows/codecov.yaml
```diff
@@ -26,14 +30,16 @@ jobs:
...
           flags: go
           files: ./go-coverage.txt,./coverage.txt
 
   python-tests:
     runs-on: ubuntu-latest
+    permissions:
+      id-token: write
     steps:
-      - uses: actions/checkout@main
+      - uses: actions/checkout@v4
```
filariow:
```suggestion
      - uses: actions/checkout@v6
```

filariow (changes requested):

.github/workflows/codecov.yaml
psturc: 🤦

filariow (approved):

PR #11124, opened 2026-04-02
Replace smee.io with hook.pipelinesascode.com in development config

MartinBasti:
smee.io does not properly preserve webhook signatures, which causes Forgejo webhook signature validation to fail. hook.pipelinesascode.com uses gosmee on the server side and correctly forwards the original webhook headers and payload.


psturc (approved):
/lgtm

amisstea (approved):

PR #11131, opened 2026-04-02
limit PipelineRuns in mintmaker namespace to 5000 in staging

filariow:
If the cluster is not able to process the MintMaker's PipelineRuns they will pile up harming the cluster stability. Enforcing a maximum number of concurrent PipelineRuns in the namespace could help mitigate this problem.

This change enforces a reasonably high value to avoid interfering with MintMaker's operativy and act as a safety net if something is going sideways.

rh-pre-commit.version: 2.3.2

### DIFF
--- a/components/mintmaker/development/kustomization.yaml
+++ b/components/mintmaker/development/kustomization.yaml
@@ -2,6 +2,7 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
   - ../base
+  - plrs-resource-quota.yaml
   - https://github.com/konflux-ci/mintmaker/config/default?ref=85962df3895522eb8df1d39438b0b85bcde3438a
   - https://github.com/konflux-ci/mintmaker/config/renovate?ref=85962df3895522eb8df1d39438b0b85bcde3438a
 
--- a/components/mintmaker/development/plrs-resource-quota.yaml
+++ b/components/mintmaker/development/plrs-resource-quota.yaml
@@ -1,0 +1,9 @@
+---
+apiVersion: v1
+kind: ResourceQuota
+metadata:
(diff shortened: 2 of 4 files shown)


filariow:
> The ResourceQuota hard-codes the `namespace: mintmaker`. Verify this is intended for the development overlay as well, since it will force the object into `mintmaker` regardless of the kustomization/target namespace configuration and could unintentionally affect other environments.

this is not correct, kustomize will replace it.

> The staging base kustomization already sets namespace: mintmaker, while the ResourceQuota manifest also specifies metadata.namespace. Consider relying on kustomize namespace injection (omitting metadata.namespace) to avoid future drift if the namespace changes or the manifest is reused elsewhere.

that's NOT a great point in our case. The standard for this component is to have `namespace: mintmaker`, it's more important for this change to adhere at that standard.

staticf0x:
AFAIK this was superseded by https://github.com/konflux-ci/mintmaker/pull/489, is this limit still necessary after the change on MintMaker's side?

filariow:
> AFAIK this was superseded by [konflux-ci/mintmaker#489](https://github.com/konflux-ci/mintmaker/pull/489), is this limit still necessary after the change on MintMaker's side?

Yes, this one is still needed. https://github.com/konflux-ci/mintmaker/pull/489 helps reducing the load MM generates by avoiding duplicates, while this change is a safety net to prevent too many PLRs to accumulate in MM's namespace. 
The 5K threshold is very high and is meant to catch scenarios in which a major problem occurred and creating too many PLRs could harm the cluster.
Examples are the cases in which the pruner is misbehaving and it's not able to prune completed PLRs, or if a bug is introduced in MM.

staticf0x (approved):

PR #11165, opened 2026-04-06
convert pipelines staging PaC units from Gi to Mi

aThorp96:
For some reason 2.5Gi was not syncing when the value was 2560Mi, however it was still being considered "out of sync" and causing an alert to fire

image


mshaposhnik (approved):

ab-ghosh (approved):

hugares (approved):
/lgtm

PR #11169, opened 2026-04-06
fix(PVO11Y-5132): Set GOMEMLIMIT for stage in-cluster Grafana

pacho-rh:
This is an attempt to prevent the in-cluster Grafana instances from being OOMKilled if a user run expensive queries. Setting GOMEMLIMIT to 80% of the container's resource limit. Applying and validating this change in stage clusters first.


martysp21 (approved):
/lgtm

PR #11204, opened 2026-04-08
feat(RELEASE-2387): move release dashboard to per env overlays

seanconroy2021:
The shared base directory components/monitoring/grafana /base/dashboards/ is used by all three environments via ../base. This commit moves the release dashboard out of base into the per env overlay. The rest of base stays as is so other teams aren’t affected.

Changes:
* Removed release/ from base/dashboards/kustomization.yaml
* Created dashboards/release/ in each overlay
* Added dashboards/ resource into each overlay
* Per env dashboards/ follows the same layout as base so other teams can do the same

kustomize build output is the same before and after for all three environments.

Assisted-by: Qodo


eedri:
looks good, @enkeefe00 can you check?

seanconroy2021:
> looks good, @enkeefe00 can you check?

Any updates on this :)

components/monitoring/grafana/base/dashboards/kustomization.yaml
filariow, line 1: I'd suggest to either update the [README](https://github.com/redhat-appstudio/infra-deployments/blob/main/components/monitoring/grafana/README.md) or to dedicate the base folder to the development environment only and migrate it entirely under the `development` folder later

filariow (dismissed):
mostly lgmt

eedri:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @eedri |
| **Timestamp** | 2026-04-13T09:46:10.026Z |

components/monitoring/grafana/base/dashboards/kustomization.yaml
seanconroy2021: You mean updating README you want me to add a deprecation notice to base explaining that dashboards should move to per env overlays going forward. I wouldn't want to migrate base fully into development though since it's shared across teams and that would risk breaking their promotion scripts. It would also break some of our own scripting logic which expects the dev/staging/production overlay not base. That's why I only moved the release dashboard out and left everything else in base untouched. So follow up need to be each team moving out of base then only keeping RBAC and Config

filariow: That makes sense to me. However, after this PR is merged, by following the guidelines in the README you won't have the release dashboard. Adding a note there explaining to apply `development/base/release` also could be enough.
Alternatively, we can skip creating the `development/base` at all and use `base` from the `development` overlay only.

seanconroy2021: Pefect updated the README for the new change let me know need any other changes :)

filariow (commented):
lgtm, waiting for a review from the component owners

components/monitoring/grafana/README.md
```diff
@@ -11,8 +11,16 @@ Note: The steps below should be handled by Argo CD
...
+    $ kustomize build components/monitoring/grafana/development | oc apply -f -
+    ```
+
+    Replace `development` with `staging` or `production`. 
+
+## Dashboard migration from base to per environment overlays
+
+The `base/dashboards/` directory is being deprecated. Team dashboards should be moved from `base/dashboards/` into per environment overlays `development/dashboards/`, `staging/dashboards/` and `production/dashboards/`. This allows each environment to reference its own dashboard source independently.
+
+The `release/` dashboard has already been migrated and can be used as a reference for the new structure. The remaining dashboards in `base/dashboards/` will continue to work via `../base` but each team should plan to migrate.
```
kubasikus, line 26: This is news to me. Where was this agreed? Do the teams have tickets in their backlogs for this?

seanconroy2021: https://github.com/redhat-appstudio/infra-deployments/pull/11083#issuecomment-4168692760
I dont know about other teams I just know that able to merge it has to be split into seperate env.

eedri: I think this was detected recently as we see start reviewing PRs going to production and noticed many PRs are being sent in parallel to development, staging and prod. that goes against the request following the moratorium, to test and merge for develpoment, then staging and lastly prod.

components/monitoring/grafana/production/dashboards/release/OWNERS
```diff
@@ -0,0 +1,27 @@
+# See the OWNERS docs: https://go.k8s.io/owners
```
kubasikus: Is it possible to deduplicate the OWNERS file by putting it higher and using a filter to select which overlays/directories it applies to (Is DRY and sets a good precedent for future migrations)?

components/monitoring/grafana/README.md
kubasikus: I see. I think we can keep it here then.

components/monitoring/grafana/production/dashboards/release/OWNERS
seanconroy2021: I would agree. Does Prow support filtering in the owner file? I have only ever seen the use of owner aliases, which is not used in this repo. 

Would this work then would be at `components/monitoring/grafana/OWNERS`
```
# See the OWNERS docs: https://go.k8s.io/owners

filters:
  ".*/dashboards/release/.*":
    approvers:
    - seanconroy2021
  
    reviewers:
    - seanconroy2021
```

FaisalAl-Rayes: according to the [docs](https://www.kubernetes.dev/docs/guide/owners/#filters) it seems supported when it says


"The filters key is a map whose keys are [Go regular expressions](https://golang.org/pkg/regexp/#pkg-overview) and whose values are [**simple OWNERS configurations**](https://www.kubernetes.dev/docs/guide/owners/#owners)"


just to note it also says that " If you set filters you must not set a [simple OWNERS configuration](https://www.kubernetes.dev/docs/guide/owners/#owners) outside of filters"

kubasikus: Based on my reading of https://github.com/kubernetes/community/blob/main/contributors/guide/owners.md#filters , I think this is correct.

seanconroy2021: Thanks guy have added the new OWNER file now :)

johnbieren:
LGTM

components/monitoring/grafana/production/dashboards/release/OWNERS
kubasikus: Thanks for adding this :)

kubasikus (approved):

PR #11239, opened 2026-04-10
Externalize MPC embedded scripts - set 2 - KFLUXINFRA-3057

celeztyne:
The purpose of this PR is to externalize a part of the remaining inline cloud-init user-data shell scripts from MPC host-values.yaml files into standalone .sh files under base/host-config-chart/files/ as done in [KFLUXINFRA-2995](https://redhat.atlassian.net/browse/KFLUXINFRA-2995)
- Some of the scripts in base/host-config-chart/files were already created, so I've just added the one for linux-test-amd64
- Removed the user-data inline scripts from the requested configuration as per  [KFLUXINFRA-3057](https://redhat.atlassian.net/browse/KFLUXINFRA-3057)

[KFLUXINFRA-2995]: https://redhat.atlassian.net/browse/KFLUXINFRA-2995?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[KFLUXINFRA-3057]: https://redhat.atlassian.net/browse/KFLUXINFRA-3057?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


components/multi-platform-controller/base/host-config-chart/files/linux-test-amd64-init.sh
```diff
@@ -0,0 +1,42 @@
...
+
+#cloud-config
+cloud_final_modules:
+  - [scripts-user, always]
+
+--//
+Content-Type: text/x-shellscript; charset="us-ascii"
+MIME-Version: 1.0
+Content-Transfer-Encoding: 7bit
+Content-Disposition: attachment; filename="userdata.txt"
```
oswcab: Including the cloud-init multipart MIME headers mixed with the actual bash script defeats the main goals of     externalizing the scripts that's allowing to run static analysis on them (shellcheck fails on MIME headers, for example). Also, if ever we want to perform a manual execution for testing, it won't work.
We need to separate the MIME headers from the script and still guarantee they're present.

components/multi-platform-controller/production/stone-prd-rh01/host-values.yaml
oswcab, line 1: Not sure if we should include production in the same change but my impression is we should not. We need to make sure first that our change works as expected on staging and, after validation, apply it to production. @filariow WDYT?

oswcab (commented):
Please see my proposal how to separate the MIME changes from the bash script on this branch: https://github.com/redhat-appstudio/infra-deployments/pull/11242/changes

components/multi-platform-controller/production/stone-prd-rh01/host-values.yaml
celeztyne: OK, I can include here only Stage related and I'm opening a second PR for Production

components/multi-platform-controller/base/host-config-chart/files/linux-test-amd64-init.sh
celeztyne: @oswcab OK we can go ahead with your Proposal, then I'll open another PR for stone-prd-rh01 - prod

oswcab: My goal was to show it can be a better way of doing this, not to take your task over :) If you like the proposal I did, feel free to take the code and update your PR with it.

celeztyne: OK got it, I thought that because it would be not only a new way but even correcting what was done in [KFLUXINFRA-2995](https://redhat.atlassian.net/browse/KFLUXINFRA-2995) / https://github.com/redhat-appstudio/infra-deployments/pull/10720
I'll take the suggestions/corrections

components/multi-platform-controller/production/stone-prd-rh01/host-values.yaml
celeztyne: mm, Actually if we apply these changes would be more difficult to divide this PR in Stage/Prod, because changes are affecting files in ../base/host-config-chart/{files,templates}/* that are used for both environments so we would have to undo Prod changes done in https://github.com/redhat-appstudio/infra-deployments/pull/10720 (most of the changes) or make files there somehow for each env.

Please @oswcab @filariow let me know WDYT

oswcab: I think we can do it in two steps: first, we keep the production file as it is now and we only apply the extraction to the staging files. Once we've verified it works in staging, we can do the same change in production in a follow up PR.

celeztyne: @oswcab seems that making changes for only staging might take more changes than expected: production, staging and other environments are accessing the scripts in the same path, for example, making changes in base/host-config-chart/files/linux-c6gd2xlarge-arm64-init.sh is going to affect not only staging but  production and production downstream, etc. 

I have explored several options externalizing the scripts (to not be inline) as requested and the less difficult IMO is to create and use a new "base" directory only for staging e.g. base-staging""  kustomization file  ../multi-platform-controller/staging/kustomization.yaml. The change would look like:
```
resources:
- ../base-staging
```
.. and then copying there the existent files from ../base and update them for staging. I asked Claude about this and told me I can even complement this with symlinks to reduce the necessary sync. 

Anyway, although this seems not very elegant that would be temporary as when we will be sure the externalization of the scripts work we would back these changes, right?. I can make an example PR if its not clear.

I hope there would be a less complicated way to do this, WDYT?

oswcab: Well, I know one of the goals on this repository is to completely separate the production environment from development/staging. If that implies having a new `base` folder, it's OK IMO. In fact, we have already done so in some components already.

celeztyne: OK, could you point for me please some of those examples so I can do it in the same way? I guess I'll bring this to the infra chat in slack to have a broader audience and opinions.

celeztyne: Right @oswcab , I have created an overlay for the kustomization in a different way. I've added a new directory in ../multi-platform-controller/base/host-config-chart-stage and then change the reference in ../multi-platform-controller/staging/kustomization.yaml from host-config-chart to host-config-chart-stage:
```
helmCharts:
- name: host-config-chart-stage
  releaseName: host-config
```
Is in the last commit, do you think this would work?

components/multi-platform-controller/staging/base/host-config-chart/Chart.yaml
oswcab, line 1: I think `base/` should remain environment-neutral. Placing `host-config-chart-stage/` alongside `host-config-chart/` in `base/` mixes environment concerns into what's supposed to be shared code. This breaks the repository  convention as other components (e.g., build-service, authentication) never put environment-specific resources in
`base/`.
Instead, I think we should duplicate the chart into `staging/base/host-config-chart/` and `production/base/host-config-chart/` so each environment owns its chart independently. This is the direction we want for this repo: full▎`staging/production` separation to prevent staging changes from leaking into production.

components/multi-platform-controller/staging/kustomization.yaml
oswcab, line 1: I see `staging-downstream` is still using the production chart. It references `host-config-chart` (production) instead of `host-config-chart-stage`. If staging and production charts are meant to differ (and they do — cloud-init format, test platform support), the `staging-downstream` overlay is getting the wrong one.

components/multi-platform-controller/staging/base/host-config-chart/Chart.yaml
celeztyne: OK I'll check that and try to separate the environments like that

components/multi-platform-controller/staging/kustomization.yaml
celeztyne: Yes I didn't change that (yet) but I was planning to included if my "proposal" got accepted . I'll make staging-downstream to reference staging changes in the next commits.

celeztyne:
@oswcab I have taken your suggestion and created a staging/base for charts, that is also referenced/used by staging-downstream.
I skipped production as we spoke about test first in non-prod, but if anything please let me know.

components/multi-platform-controller/staging/host-values.yaml
```diff
@@ -49,51 +49,8 @@ dynamicConfigs:
...
-      sed -n 's,.*\(ssh-.*\s\),\1,p' /root/.ssh/authorized_keys > /home/ec2-user/.ssh/authorized_keys
-      chown ec2-user /home/ec2-user/.ssh/authorized_keys
-      chmod 600 /home/ec2-user/.ssh/authorized_keys
-      chmod 700 /home/ec2-user/.ssh
-      restorecon -r /home/ec2-user
-
-      --//--
-
+  linux-c6gd2xlarge-arm64: {}
+      # user-data loaded from base/host-config-chart/files/linux-c6gd2xlarge-arm64-init.sh
```
oswcab, line 53: Nit: this comment references `base/host-config-chart/files/linux-c6gd2xlarge-arm64-init.sh` but since the staging chart was moved to `staging/base/`, the correct path would be `staging/base/host-config-chart/files/linux-c6gd2xlarge-arm64-init.sh`.

Also, the previous qodo-code-review flagged that the original comment was missing the `linux-` prefix — that's been fixed in the filename but the directory path is now stale.

components/multi-platform-controller/staging/host-values.yaml
```diff
@@ -116,57 +73,7 @@ dynamicConfigs:
...
-
-      # Configure ec2-user SSH access
-      chown -R ec2-user /home/ec2-user
-      sed -n 's,.*\(ssh-.*\s\),\1,p' /root/.ssh/authorized_keys > /home/ec2-user/.ssh/authorized_keys
-      chown ec2-user /home/ec2-user/.ssh/authorized_keys
-      chmod 600 /home/ec2-user/.ssh/authorized_keys
-      chmod 700 /home/ec2-user/.ssh
-      restorecon -r /home/ec2-user
-      --//--
+    # user-data loaded from base/host-config-chart/files/linux-g64xlarge-amd64-init.sh
```
oswcab, line 76: Same as above — the comment path `base/host-config-chart/files/linux-g64xlarge-amd64-init.sh` should be updated to `staging/base/host-config-chart/files/linux-g64xlarge-amd64-init.sh` to reflect the new chart location.

components/multi-platform-controller/staging/base/host-config-chart/templates/host-config.yaml
```diff
@@ -0,0 +1,1181 @@
...
+  dynamic.linux-root-arm64.disk: {{ index $config "disk" | default "200" | quote }}
+  dynamic.linux-root-arm64.check-interval: {{ (index $config "check-interval") | default "60" | quote }}
+  dynamic.linux-root-arm64.allocation-timeout: "1200"
+  {{- if (index $config "iops") }}
+  dynamic.linux-root-arm64.iops: {{ index $config "iops" | quote }}
+  {{ end }}
+  {{- if (index $config "throughput") }}
+  dynamic.linux-root-arm64.throughput: {{ index $config "throughput" | quote }}
+  {{ end }}
+  {{- if (index $config "user-data") }}
```
oswcab, line 966: Question: `linux-root-arm64` only supports inline `user-data` (no `.Files.Get` fallback), while the equivalent `linux-root-amd64` at line 992 has the external file fallback pattern with `linux-root-amd64-init.sh`. Is this intentional?

If `linux-root-arm64` is supposed to have the same init script behavior, it should get the same external file treatment. If it genuinely doesn't need custom user-data, then this is fine — just wanted to confirm.

oswcab (commented):
Thanks for addressing the previous feedback — the MIME/script separation via `_cloud-init-userdata.tpl` and the staging chart isolation look good. A couple of small items to fix and one question.

components/multi-platform-controller/staging/base/host-config-chart/templates/host-config.yaml
celeztyne: yes, IMO should have the same behavior, and I can updated it, just that this original request was only for linux-c6gd2xlarge-arm64, linux-g64xlarge-amd64 and linux-test-amd64 (as per the task summary). I agree, logically this change should be done for all flavors, even for linux-root-amd64 that was done by Max in [KFLUXINFRA-2995](https://redhat.atlassian.net/browse/KFLUXINFRA-2995) before we changed how is organized (including Staging and Prod separation). 
There is another task [KFLUXINFRA-3056](https://redhat.atlassian.net/browse/KFLUXINFRA-3056) that might also need to update requirements after these changes.

celeztyne: @mshaposhnik  as you have work with this recently my guess is that you might be familiar with the issue, do you have any thoughts about Hector's comments on linux-root-arm64 and custom user-data?

components/multi-platform-controller/staging/host-values.yaml
```diff
@@ -49,51 +49,8 @@ dynamicConfigs:
...
-      sed -n 's,.*\(ssh-.*\s\),\1,p' /root/.ssh/authorized_keys > /home/ec2-user/.ssh/authorized_keys
-      chown ec2-user /home/ec2-user/.ssh/authorized_keys
-      chmod 600 /home/ec2-user/.ssh/authorized_keys
-      chmod 700 /home/ec2-user/.ssh
-      restorecon -r /home/ec2-user
-
-      --//--
-
+  linux-c6gd2xlarge-arm64: {}
+      # user-data loaded from staging/base/host-config-chart/files/linux-c6gd2xlarge-arm64-init.sh
```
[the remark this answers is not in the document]
celeztyne: @oswcab this Qodo required action seems to not match your suggestion above (which I've applied). I understand Qodo see these paths from a "local" perspective (as standing in "components/multi-platform-controller/staging/" path). Any thoughts?

components/multi-platform-controller/staging/base/host-config-chart/templates/host-config.yaml
mshaposhnik: There is no particular reason not supporting the user-data files in `linux-root-arm64` other than just absense of this file and/or content for it :) So, in case you have a desire to add the support for possible future, I have no objections with that.

components/multi-platform-controller/staging/base/host-config-chart/templates/host-config.yaml
```diff
@@ -0,0 +1,1181 @@
+apiVersion: v1
```
mshaposhnik, line 1: Sorry, can you state for me which problem we're solving by duplicating this file?
Because imo its duplicating cancels the main idea of whole templating -  simplicity of changes and single source of truth.
 cc @oswcab

components/multi-platform-controller/staging/host-values.yaml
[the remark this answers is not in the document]
oswcab: We can leave it as qodo suggests as, indeed, it assumes the reader acknowledges is under `staging` already.

components/multi-platform-controller/staging/base/host-config-chart/templates/host-config.yaml
oswcab: One of the stated goals in this repository, AFAIK, it's to completely separate `staging` manifests from `production` ones, so there's no even a possibility of a change in `staging` ever affecting a `production` environment. We have already done that duplication in other components and, even if I agree it could contradict the initial goal of having templates, seems the easier/faster way to clearly do this separation.
IIUC, the feature about ring deployments will provide directives for the layout of the different deployment stages for a component (development, staging and production) that will probably change what we're currently doing. @filariow, please keep me honest here.

components/multi-platform-controller/staging/host-values.yaml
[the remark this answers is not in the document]
celeztyne: OK, Thanks @oswcab

components/multi-platform-controller/staging/base/host-config-chart/files/linux-c6gd2xlarge-arm64-init.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash -ex
```
filariow, line 1: @mshaposhnik for security reason, should we drop the `x`? I'm thinking about public tenant namespaces.

should we also standardize these scripts around `set -euo pipefail`?

components/multi-platform-controller/staging/base/host-config-chart/templates/host-config.yaml
filariow: we don't want changes to templates to impact all environments at the same time. We should definitely have one version of it for staging and another one for production. Development should only target staging. Once the change is tested in staging, we'll `cp -r` the staging folder into production's one and promote the change to production.

celeztyne:
- qodo code finds in the reviews  fixed
- host config Staging charts (including staging-downstream) separated from Production
- Suggestions from konflux-infra team reviewers done

@oswcab Any other change needed for this PR?

components/multi-platform-controller/staging/base/host-config-chart/files/linux-c6gd2xlarge-arm64-init.sh
celeztyne: In case this is wanted, should we include it in this PR? @filariow @mshaposhnik

oswcab:
> * qodo code finds in the reviews  fixed
> * host config Staging charts (including staging-downstream) separated from Production
> * Suggestions from konflux-infra team reviewers done
> 
> @oswcab Any other change needed for this PR?

I still saw some bits missing but instead of posting yet another comment, decided to push the fixes myself. 

BTW, when working in a long term branch like this one, it's better to rebase it on top of the main branch instead of merging the main branch into yours, as the resulting history of the latter is more difficult to read and follow. If ever need/want to know how to do it, let me know and schedule a half an hour conversation and I can show you.

Last thing, please check the commands used to verify the resulting manifests as it's very important to check the result before thinking about merging a commit in this repo.

celeztyne:
> Last thing, please check the commands used to verify the resulting manifests as it's very important to check the result before thinking about merging a commit in this repo.
@oswcab I have tested building rendering stage and stage-downstream (command kustomize build --enable-helm components/multi-platform-controller/[staging,staging-downstream] ) in the main origin and then using the files in this PR, then compared the before/after and aside of the improvements suggested by Qodo-review in the new version the generated output is the same. 
I did not test production because those files were not touched.
I don't know if that is enough or should I try dry run or in some dev environment

celeztyne:
> I still saw some bits missing but instead of posting yet another comment, decided to push the fixes myself.
> 
Yes sorry I missed deleting the inline scripts in stage-downstream

> BTW, when working in a long term branch like this one, it's better to rebase it on top of the main branch instead of merging the main branch into yours, as the resulting history of the latter is more difficult to read and follow. If ever need/want to know how to do it, let me know and schedule a half an hour conversation and I can show you.
> 
Yes I think I messed with the upstream locally and then I was in a hurry and rebased using the "update branch" button in here, I should rather fetch upstream and rebase. Thanks for the offer, I'll ping you later just to clarify.

oswcab:
> > Last thing, please check the commands used to verify the resulting manifests as it's very important to check the result before thinking about merging a commit in this repo.
> > @oswcab I have tested building rendering stage and stage-downstream (command kustomize build --enable-helm components/multi-platform-controller/[staging,staging-downstream] ) in the main origin and then using the files in this PR, then compared the before/after and aside of the improvements suggested by Qodo-review in the new version the generated output is the same.
> > I did not test production because those files were not touched.
> > I don't know if that is enough or should I try dry run or in some dev environment

That should be enough, is the same verification I did.

oswcab:
> > I still saw some bits missing but instead of posting yet another comment, decided to push the fixes myself.
> 
> Yes sorry I missed deleting the inline scripts in stage-downstream
> 
> > BTW, when working in a long term branch like this one, it's better to rebase it on top of the main branch instead of merging the main branch into yours, as the resulting history of the latter is more difficult to read and follow. If ever need/want to know how to do it, let me know and schedule a half an hour conversation and I can show you.
> 
> Yes I think I messed with the upstream locally and then I was in a hurry and rebased using the "update branch" button in here, I should rather fetch upstream and rebase. Thanks for the offer, I'll ping you later just to clarify.

Yeah, no worries, is just a suggestion for future changes.

components/multi-platform-controller/staging/base/host-config-chart/files/linux-c6gd2xlarge-arm64-init.sh
filariow: no we can have a follow up PR for this

celeztyne:
Because this PR became too messy and I'm not able to rebase properly I've opened https://github.com/redhat-appstudio/infra-deployments/pull/11858 and closing this one

PR #11253, opened 2026-04-13
PVO11Y-5159 Drop high cardinality metric and label

kubasikus:
This metric is known to have high cardinality due to the pod label. It is also not useful, so it will be dropped entirely.

Assisted-by: Cursor


martysp21 (approved):
/lgtm

martysp21 (approved):
/lgtm

PR #11303, opened 2026-04-15
Add tekton-ms Prometheus as Grafana datasource in staging

martysp21:
Add a new GrafanaDatasource CR (prometheus-tekton-ms-ds) pointing to the appstudio-tekton-ms Prometheus instance in the appstudio-tekton namespace. This enables Grafana to query Tekton Pipelines controller metrics scraped by the dedicated MonitoringStack. Uses the existing metrics-reader ServiceAccount token for authentication over HTTPS. 

Scoped to staging clusters rh01 and p01 as they are the only clusters which have tekton data.
A env split needed to happen so specific env and clusters could be affected. This is the cause for production change along with staging change. This operation cannot be split.


components/monitoring/grafana/staging/kustomization.yaml
```diff
@@ -2,8 +2,8 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
   - ../base
-  - dashboards
   - https://github.com/redhat-appstudio/o11y/config/exporters/monitoring/grafana/base?ref=0d3c45516968bffb24196952c3273cebd6df5d4f
+  - tekton-ms-datasource.yaml
```
kubasikus: We want to avoid deploying to EaaS clusters, so this will need to be split by different clusters.

martysp21: Done, I had to change the structure of the grafana to allow specific cluster deploy but it should work like this and deploy only for stage p01 and rh01.

components/monitoring/grafana/staging/tekton-ms-datasource/tekton-ms-datasource.yaml
```diff
@@ -0,0 +1,30 @@
+apiVersion: grafana.integreatly.org/v1beta1
```
kubasikus, line 1: To avoid duplicating this file, can we put it into monitoring/grafana/staging/tekton-ms folder?

martysp21: Yeah I thought the same after reviewing the code. 
It is now fixed as a new resource.

kubasikus (approved):

kubasikus:
Looks great, thanks! We'll need to get this through production approval through. Risk should be low, since this is change only modifies the way we deploy grafana into Konflux clusters. Disaster recovery is rolling back this PR.

martysp21:
@gbenhaim Could you take a look?
I hope I did understand the new procedure correctly and that I should be pinging you.

Anyway I know that email stated that there should not be production and staging label at once, but it is necessity to do this in one PR as separating it would cause problems. The production change is only a structural change not a functional one.

martysp21:
tlsSkipVerify problem:
Not really a problem, everything stays inside of cluster. This solution is used from already deployed components.
AuthWiring problem:
metrics-reader SA is dedicated SA for this purpose
Cluster targeting:
Yes it was checked that all will be built as before with stage clusters p01 and rh01 having the datasource on top of others.

nhaklik (approved):

nhaklik:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @nhaklik |
| **Timestamp** | 2026-04-21T12:15:09.494Z |

kubasikus (approved):

martysp21:
@filariow Please can you take a look?

kubasikus (approved):

enkeefe00 (approved):

kubasikus (approved):

nhaklik:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | REVERTED TO PENDING |
| **Reviewer** | @nhaklik |
| **Timestamp** | 2026-04-27T13:00:39.101Z |

nhaklik:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @nhaklik |
| **Timestamp** | 2026-04-27T13:00:52.963Z |

PR #11312, opened 2026-04-15
Log optimization staging fix

raks-tt:


components/monitoring/logging/staging/base/patches/configure-logforwarder-filters-patch.yaml
```diff
@@ -38,15 +38,20 @@
...
       - test:
           - field: .kubernetes.namespace_name
             matches: 'openshift-pipelines|tekton-.*'
-          - field: .structured.msg
+          - field: .message
             matches: '(?i)Reconcile error'
-          - field: .structured.error
+          - field: .message
             matches: '(?i)(pipelinerun|taskrun|tekton)'
+      - test:
```
pacho-rh, line 50: I think the first test in this filter should already be covering this new test.

- `tekton-kueue` should be covered by the `tekton-.*` match
- The message match are largely the same. AFAIK, the message will match as long as it contains the specific substring: https://vector.dev/docs/reference/vrl/functions/#match.

Is this not the case when you examined the logs?

raks-tt: yes that is because `tekton-kueue` log structure is quite different. So it doesn't match the condition `- field: .structured.error ` for `Operation cannot be fulfilled on`. Hence needed to apply it on the `message` field

pacho-rh: Ah I see. Apologies, I missed the differing fields between the two matches.

pacho-rh (approved):
/lgtm

PR #11298, opened 2026-04-15
KFLUXINFRA-3534 Revert staging etcd-defrag cronjob to previous config

glevi-rh:
## What
Reverts the staging etcd-defrag cronjob to the configuration before fbbc893ae, restoring the `FRAGMENTATION_THRESHOLD`, `DISK_USAGE_THRESHOLD`, and `RECLAIM_SPACE_THRESHOLD` env vars. Updates the image to `sha256:7a19f7b9b7bbdeb4f8d51e06d3319ac5c0555094fcbf2093d6e4d71f136afb55` and sets `DISK_USAGE_THRESHOLD` to 75 (just below etcd-shield's 80% threshold).

## Why
The built-in etcd-defrag of the etcd-operator doesn't handle well the case of an overloaded cluster with low fragmentation percentage.

## Validation
We had an outage in `stone-prd-rh01` and reverting to this configuration fixed the issues. Deploying to staging first to validate before rolling out to production.

## Risk Assessment
**Risk Level:** Low

Reverting to a previously known-good configuration in staging only. etcd-shield is confirmed deployed to staging clusters and will provide protection at the 80% threshold.

PR #11317, opened 2026-04-15
KFLUXINFRA-3534 Revert etcd-defrag production cronjob for stone-prd-rh01, stone-prod-p02, kflux-ocp-p01 (ring 1)

glevi-rh:
## What
Reverts the etcd-defrag production cronjob for ring 1 clusters (`stone-prd-rh01`, `stone-prod-p02`, `kflux-ocp-p01`) to the previous configuration using `FRAGMENTATION_THRESHOLD`, `DISK_USAGE_THRESHOLD`, and `RECLAIM_SPACE_THRESHOLD` env vars. Updates the image to `sha256:7a19f7b9b7bbdeb4f8d51e06d3319ac5c0555094fcbf2093d6e4d71f136afb55` and sets `DISK_USAGE_THRESHOLD` to 75 (just below etcd-shield's 80% threshold).

For `kflux-ocp-p01`, a cronjob.yaml patch was added since it inherits from the production base which still uses the old config. This will be cleaned up in ring 2 when the base itself is updated.

## Why
The built-in etcd-defrag of the etcd-operator doesn't handle well the case of an overloaded cluster with low fragmentation percentage.

## Validation
- Staging PR: https://github.com/redhat-appstudio/infra-deployments/pull/11298 — though not very helpful for validation since both staging clusters had only ~1 defrag in the last week, which makes it less effective to test
- `kustomize build` ran successfully for `configs/etcd-defrag/production/kflux-ocp-p01/`
- We had an outage in `stone-prd-rh01` and reverting to this configuration fixed the issues

## Risk Assessment
**Risk Level:** Medium

Production change affecting 3 clusters. However, this reverts to a previously known-good configuration that already resolved the outage. Ring 2 will follow separately for the remaining production clusters.


hugares (approved):
/lgtm

mcada:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @mcada |
| **Timestamp** | 2026-04-16T10:16:36.378Z |
| **Risk Level** | medium |

This is a revert to previously stable defrag, important to avoid further etcd incidents.

While the qodo suggestion sounds reasonable, I will ask the team to address it separately - if we want to refactor how the cronjob is configured. We know that this revert worked perfectly in the past.

configs/etcd-defrag/production/kflux-ocp-p01/kustomization.yaml
```diff
@@ -2,3 +2,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
 - ../base
+patches:
```
filariow: can you keep this consistent with what we have for `rh01` and `p02` by having a full base folder here too? You can just copy the `base` folder from one `rh01` and `p02`. They are meant to be identical.

Then ring2 deployment will promote one of these identical `base` folders as `production/base`

filariow (changes requested):

filariow (approved):

PR #11345, opened 2026-04-17
Disable leader election in production clusters

raks-tt:
Increase CPU resources to reduce the throttling and disable leader election from CSV patch and a job that applies directly to the cli arg via the Deployment


nhaklik:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @nhaklik |
| **Timestamp** | 2026-04-23T16:27:07.220Z |

TominoFTW (approved):

PR #11365, opened 2026-04-20
Deploy insights ClusterSecretStore to p02

hugares:
So far, insights ClusterSecretStore was only deployed to stone-prd-rh01 but the insights team need it on stone-prod-p02 too.

Move the definition of the store into production/insights folder so it can be included in both cluster without duplicating it and patch it in each cluster to include the specific namespaces that can use it.

[KFLUXINFRA-3624](https://redhat.atlassian.net/browse/KFLUXINFRA-3624)

[KFLUXINFRA-3624]: https://redhat.atlassian.net/browse/KFLUXINFRA-3624?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Risk Assessment
**Risk Level:** Low
**Description:** Deploy insights ClusterSecretStore to p02 following user request.
**Rollback:** Revert PR

## Validation
This cannot be tested on staging as we do not deploy the insights store elsewhere than rh01 and p02 with this change.

I validated that the change will not affect other cluster and make sure that p02 is getting the indented change by running kustomize build before and after my change.


sadlerap (approved):

components/cluster-secret-store-rh/production/stone-prd-rh01/insights-conditions-patch.yaml
```diff
@@ -0,0 +1,14 @@
+---
+apiVersion: external-secrets.io/v1
+kind: ClusterSecretStore
+metadata:
+  name: insights-appsre-vault
+spec:
+  conditions:
+    - namespaces:
```
filariow, line 8: do we need to limit namespaces also in p02?

filariow (changes requested):

components/cluster-secret-store-rh/production/stone-prd-rh01/insights-conditions-patch.yaml
hugares: yes, we do limit them in p02, currently only one NS is allowed https://github.com/redhat-appstudio/infra-deployments/pull/11365/changes#diff-1a23e2d4ea0b473a75dc13e3d81ce274c3fe181c4bed21a7f044569a0ab5cfaaR9

filariow: nice, I missed that. Thanks!

filariow (approved):

eedri:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @eedri |
| **Timestamp** | 2026-04-21T13:58:51.455Z |
| **Risk Level** | low |

hugares:
there is a change ongoing to disable make the failing job optional until it is stabilized, merging the change now

PR #11367, opened 2026-04-20
fix(PVO11Y-5132): Set GOMEMLIMIT for in-cluster Grafana

pacho-rh:
This is an attempt to prevent the in-cluster Grafana instances from being OOMKilled if a user run expensive queries. Setting GOMEMLIMIT to 80% of the container's memory resource limit.
This was applied and tested in stage per redhat-appstudio/infra-deployments#11169


TominoFTW (approved):

pacho-rh:
/hold

Putting this hold as we want to make adjustments in stage. Using request as a soft-limit which GOMEMLIMIT will reference for the value.

nhaklik:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @nhaklik |
| **Timestamp** | 2026-04-23T20:12:26.892Z |

TominoFTW (approved):

PR #11400, opened 2026-04-22
Add `prod/*` labels to list of managed labels

enkeefe00:
This commit addresses the issue where pre-existing `prod/` labels were not being pruned after a PR rebase. "prod/" labels are now managed by the `env-detector` tool and unit tests have been added to simulate the rebase scenario.

Assisted-by: Cursor

### DIFF
--- a/infra-tools/internal/github/labels.go
+++ b/infra-tools/internal/github/labels.go
@@ -16,6 +16,7 @@ var LabelPrefixes = []string{
 	"environment/",
 	"cluster/",
 	"infra/",
+	"prod/",
 }
 
 // HoldProductionLabel is applied when the PR affects the production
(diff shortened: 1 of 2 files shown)


infra-tools/internal/github/labels.go
```diff
@@ -27,6 +28,10 @@ const HoldProductionLabel = "infra/hold-production"
 // environment, signalling that explicit approval is required before merging.
 const NeedsApprovalProductionLabel = "prod/needs-approval"
 
+// ApprovedProductionLabel is applied when the PR affects the production
+// environment, signalling that the PR has been approved.
+const ApprovedProductionLabel = "prod/approved"
```
manish-jangra, line 33: looks like the const `ApprovedProductionLabel` is declared but not used anywhere.

enkeefe00: It's used in tests but put it here so it would be grouped with the other constants.

mshaposhnik (approved):

PR #11408, opened 2026-04-23
[KFLUXDP-883] Add GH Action to warn if prod is affected by stage changes

p8r-the-gr8:
Our end goal is to implement ring deployment where ring 0 is stage.
Changes must be validated in staging before they can be promoted to production.

As an early step towards this goal, this change adds a new GH Action that checks if a PR is modifying both staging and production overlays at the same time.

The PR also has the policy document: docs/deployment/ring-deployment.md

| PR scope  | Action output |
| ------------- | ------------- |
| staging only  | Passes  |
| prod only  | Passes  |
| base files | Passes with warning |
| stage & prod | Fails with message |

Examples can be seen here:
https://github.com/p8r-the-gr8/infra-deployments/pulls

### DIFF
--- a/.github/workflows/enforce-ring-deployment.yaml
+++ b/.github/workflows/enforce-ring-deployment.yaml
@@ -1,0 +1,83 @@
+name: Enforce Ring Deployment
+
+on:
+  pull_request_target:
+    types: [opened, synchronize, reopened]
+    branches: [main]
+    paths:
+      - 'components/**'
+      - 'argo-cd-apps/**'
+      - 'configs/**'
+      - '.github/workflows/enforce-ring-deployment.yaml'
+
+jobs:
+  ring-deployment-check:
+    name: Check ring deployment policy
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: write
+      issues: write
+    steps:
+      # Check out the BASE branch (trusted code) so we never execute PR code.
+      - uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+
+      - uses: actions/setup-go@v5
+        with:
+          go-version-file: infra-tools/go.mod
+          cache-dependency-path: infra-tools/go.sum
+
+      - name: Setup Kustomize
+        uses: multani/action-setup-kustomize@v1
+        with:
+          version: 5.6.0
+
+      # Build from the trusted base branch before switching to PR code.
+      - name: Build env-detector (from base branch)
+        working-directory: infra-tools
+        run: go build -o bin/env-detector ./cmd/env-detector
+
+      # Fetch the GitHub-synthesized merge commit so the tool analyses the
+      # post-merge state, while the binary remains the trusted base build.
+      - name: Checkout PR merge ref
+        run: |
+          git fetch origin pull/${{ github.event.pull_request.number }}/merge:pr-merge
+          git checkout pr-merge
+
+      - name: Check ring deployment policy
+        id: ring-check
+        continue-on-error: true
+        working-directory: infra-tools
+        run: |
+          ./bin/env-detector \
+            --repo-root=.. \
+            --base-ref=origin/${{ github.event.pull_request.base.ref }} \
+            --enforce-ring-deployment \
+            --ring-report-file=/tmp/ring-report.md \
+            --dry-run
+
+      - name: Post or update PR comment
+        if: always() && steps.ring-check.outcome != 'skipped'
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          COMMENT_MARKER="<!-- ring-deployment-check -->"
+
+          # Delete any previous comment from this workflow
+          gh api "repos/${{ github.repository }}/issues/${{ github.event.pull_request.number }}/comments" \
+            --paginate --jq ".[] | select(.body | contains(\"${COMMENT_MARKER}\")) | .id" \
+            | xargs -r -I {} gh api "repos/${{ github.repository }}/issues/comments/{}" -X DELETE
+
+          # Post a new comment only if the report file was generated
+          if [ -s /tmp/ring-report.md ]; then
+            echo "${COMMENT_MARKER}" >> /tmp/ring-report.md
+            gh pr comment ${{ github.event.pull_request.number }} \
+              --repo ${{ github.repository }} \
+              --body-file /tmp/ring-report.md
+          fi
+
+      - name: Fail on ring deployment violation
+        if: always() && steps.ring-check.outcome == 'failure'
+        run: exit 1
--- a/infra-tools/cmd/env-detector/main.go
+++ b/infra-tools/cmd/env-detector/main.go
@@ -13,6 +13,7 @@ import (
 	"path/filepath"
 	"slices"
 	"sort"
+	"strings"
 	"syscall"
 
 	charmlog "github.com/charmbracelet/log"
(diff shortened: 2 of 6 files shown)


flacatus:
cc @eedri @filariow @oswcab

eedri:
before we merge this, i want to make sure we communicate this is coming for all developers on email and slack so teams can get ready, i don't want anyone to be surprised ( even though they shouldn't )

.github/workflows/enforce-ring-deployment.yaml
```diff
@@ -0,0 +1,84 @@
...
+
+on:
+  pull_request_target:
+    types: [opened, synchronize, reopened]
+    branches: [main]
+    paths:
+      - 'components/**'
+      - 'argo-cd-apps/**'
+      - 'configs/**'
+      - 'infra-tools/**'
```
oswcab: Do we really need to add this folder here? I mean, it contains just tools like its name reflects, and t they are not deployed to any cluster, just run in CI, so not sure the ring deployment workflow applies. Besides, including this folder contradicts what the documentation says.

.github/workflows/enforce-ring-deployment.yaml
```diff
@@ -0,0 +1,84 @@
...
+  ring-deployment-check:
+    name: Check ring deployment policy
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: write
+      issues: write
+    steps:
+      # Check out the BASE branch (trusted code) so we never execute PR code.
+      - uses: actions/checkout@v4
```
oswcab: `actions/checkout@v6` is the latest stable major version, released to support Node.js 24. 
Node.js 20, the one used by this version, is deprecated on GitHub Actions runners as of June 2026 so you will see a warning every time you run this workflow.

infra-tools/cmd/env-detector/main.go
```diff
@@ -24,15 +25,17 @@ import (
...
+		repoRoot             = flag.String("repo-root", ".", "Path to the repository root")
+		baseRef              = flag.String("base-ref", "main", "Base git ref to compare against")
+		overlaysDir          = flag.String("overlays-dir", "argo-cd-apps/overlays", "Path to overlays directory relative to repo root")
+		dryRun               = flag.Bool("dry-run", false, "Print results without calling GitHub API")
+		prNumber             = flag.Int("pr-number", 0, "PR number to label (required if not --dry-run)")
+		githubToken          = flag.String("github-token", "", "GitHub token (required if not --dry-run)")
+		repo                 = flag.String("repo", "", "GitHub repository in owner/repo format (required if not --dry-run)")
+		clusterLabels        = flag.Bool("cluster-labels", false, "Include cluster/<name> labels in addition to environment labels")
+		logFile              = flag.String("log-file", "", "Write debug-level logs to this file (in addition to INFO-level logs on stdout)")
+		enforceRingDeploy    = flag.Bool("enforce-ring-deployment", false, "Fail when both staging and production overlays are directly modified in the same PR")
```
oswcab, line 37: If we want to enforce this ring deployment, shouldn't we set the default as `true` here?

infra-tools/cmd/env-detector/main.go
```diff
@@ -136,15 +139,32 @@ func main() {
 
 	printSummary(result, labels, headSHA, baseSHA)
 
-	if *dryRun {
-		return
+	if !*dryRun {
```
oswcab, line 142: What happens if `dryRun` is `true`? As the code it is now, I see it will continue executing the rest of the code, instead of returning as originally intended.

.github/workflows/enforce-ring-deployment.yaml
```diff
@@ -0,0 +1,84 @@
+name: Enforce Ring Deployment
+
+on:
+  pull_request_target:
+    types: [opened, synchronize, reopened]
+    branches: [main]
+    paths:
+      - 'components/**'
+      - 'argo-cd-apps/**'
```
gbenhaim, line 9: Changes here can affect all environments at once since we use one application set for each component. If we want to enforce ring deployments on the Argo cd apps, we should plan it and refactor the argocd manifests.

gbenhaim (changes requested):

gbenhaim:
Do we really want ring 0 to be stage? It means that changes can skip the development overlay which is what we test in the ci.

flacatus:
@gbenhaim should we have this check until we start to refactor the manifests? I believe at least for now we can start with something small to ensure stage/prod overlays are not going at the same time.

flacatus:
> Do we really want ring 0 to be stage? It means that changes can skip the development overlay which is what we test in the ci.

An Alternative since we want to add the operator to deploy components i think development is ring 0 where we already start to run a set of tests(conformance tests)

filariow:
> > Do we really want ring 0 to be stage? It means that changes can skip the development overlay which is what we test in the ci.
> 
> An Alternative since we want to add the operator to deploy components i think development is ring 0 where we already start to run a set of tests(conformance tests)

In some of our components the staging overlay is referencing the development overlay and applying at most few necessary patches. That's because it usually doesn't make much sense to deploy to development (effectively just the e2e test environment) without pushing to staging too. I would support considering stage 0 development+staging.

.github/workflows/enforce-ring-deployment.yaml
p8r-the-gr8: I wasn't sure which directories should be in scope (besides `components`) so that's how that ended up there.
I'll remove it.

.github/workflows/enforce-ring-deployment.yaml
p8r-the-gr8: Good catch, most of the other actions use v4 as well so that's how it ended up there.
I'll update it to v6.

infra-tools/cmd/env-detector/main.go
p8r-the-gr8: This way the tool's current behaviour (as used by [pr-env-label](https://github.com/redhat-appstudio/infra-deployments/blob/main/.github/workflows/pr-env-label.yaml#L63-L69)) is preserved and the new feature is enabled by the flag.

infra-tools/cmd/env-detector/main.go
p8r-the-gr8: The original intention was that if `dryRun == true` to *not* perform the [PR labeling operation](https://github.com/redhat-appstudio/infra-deployments/blob/main/infra-tools/cmd/env-detector/main.go#L143-L148).

This is still true with the new changes, as the labeling code block depends on a [`dryRun == false`](https://github.com/p8r-the-gr8/infra-deployments/blob/461611bbd99316a46547341364ebb6838bb37695/infra-tools/cmd/env-detector/main.go#L142-L148) check.

The new functionality is also behind a [`enforceRingDeploy == true`](https://github.com/p8r-the-gr8/infra-deployments/blob/461611bbd99316a46547341364ebb6838bb37695/infra-tools/cmd/env-detector/main.go#L150-L166) check so it's not executed unless explicitly requested.

With these changes, if  `dryRun == true` and `enforceRingDeploy == false` the only change in behaviour is that now ["Done!" will be printed](https://github.com/p8r-the-gr8/infra-deployments/blob/461611bbd99316a46547341364ebb6838bb37695/infra-tools/cmd/env-detector/main.go#L168C2-L168C20) at the end, which I think makes sense even on a dry run.

.github/workflows/enforce-ring-deployment.yaml
p8r-the-gr8: I wasn't sure if the dir should be in scope or not, so I added it there.
Let me know if you'd rather exclude it.

flacatus:
I think the scope of this PR is not to define an architecure of ring deployments. Lets put a basic check that in the future can be expanded once we have all ring deployments manifests in place

flacatus:
@gbenhaim @filariow Can we take another look?

PR #11460, opened 2026-04-27
feat(SPRE-5143): iib prod probe deployment

gcpsoares:
- Deploy new iib probes.
- Tested and validated with probes in stage (available in stage [grafana dashboard](https://grafana.app-sre.devshift.net/goto/3K3JwdTvR?orgId=1).
- This probe will only be available for private clusters (by design).


TominoFTW (approved):

jkriz-rh:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @jkriz-rh |
| **Timestamp** | 2026-04-28T12:54:52.431Z |

Qodo comment is a false positive here, similar probe is already on other clusters and there is a low probability of any issue, as discussed with Gabriel

PR #11497, opened 2026-04-28
Increase artifact registry proxy resources in production

amisstea:
**What/Why**

Increase artifact registry proxy resources in production.

This also sets memory requests==limits which is a best practice for scheduling purposes.

Assisted-by: Claude claude-opus-4-6

**Validation**

These limits allowed for _at least_ 100 parallel req/s in staging without error. Further performance testing will be performed.

**Risk Assessment**

Risk Level: Low
Description: An increase in resources should be harmless
Rollback: Revert the commit

### DIFF
--- a/components/squid/production/squid-helm-generator.yaml
+++ b/components/squid/production/squid-helm-generator.yaml
@@ -32,10 +32,10 @@ valuesInline:
       size: 51200
     resources:
       requests:
-        cpu: 100m
-        memory: 128Mi
+        cpu: "2"
+        memory: 2Gi
       limits:
-        cpu: 200m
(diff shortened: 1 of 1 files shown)


hmariset (approved):
/lgtm

svatares:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @svatares |
| **Timestamp** | 2026-04-28T19:02:05.182Z |

Approving the change. The "bug" Qodo flagged regarding resources isn't a bug, but highlights high resource requirements that are required in this case.

PR #11582, opened 2026-05-03
Revert kubearchive v1.21.3 upgrade in stone-prod-p02

maruiz93:
## Summary
- Revert #11458 (Upgrade kubearchive to v1.21.3 in stone-prod-p02)
- Revert #11581 (ITN-2026-00132 fix for the stale kubearchive DB migration job)
- Restores kubearchive to v1.20.0 with MIGRATION_VERSION=5

## Reason
The database migration (07_08_populate_label_tables) is processing ~225M resource IDs and requires 2+ days to complete. During the migration, the api-server and sink pods are down with `CreateContainerConfigError` / schema version mismatch because the DB has advanced past what v1.20.0 supports, and ArgoCD sync waves prevent updating the Deployments to v1.21.3 until the migration Job completes.
We shouldn't also for the deployments going to v1.21.3 until the migration isn't finished.

## Test plan
- [ ] ArgoCD syncs successfully for stone-prod-p02
- [ ] Migration job runs down migrations to roll back DB schema
- [ ] KubeArchive API, operator, and sink pods are running on v1.20.0
- [ ] Verify kubearchive API is functional

## Risk assessment
- The down migration drops the new label tables and restores the GIN index. No data loss — the label tables only contained derived data from existing resources.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #11583, opened 2026-05-03
Use v1.21.3 migration image for DB schema rollback

maruiz93:
## Summary
- Override migration job image to v1.21.3 so it can run the down migrations from schema version 7 back to 5
- The v1.20.0 image (restored by #11582) doesn't contain the down migration scripts for version 7

## Reason
After reverting the v1.21.3 upgrade (#11582), the migration job panics because the v1.20.0 image doesn't have the down migration files needed to roll back from schema version 7 to 5. The v1.21.3 migration image has all the required scripts.

## Test plan
- [ ] Migration job starts and runs down migrations successfully
- [ ] DB schema version returns to 5
- [ ] KubeArchive API and sink pods start on v1.20.0

## Risk assessment
- Low. Only the migration job image is changed. Once the rollback completes, this image override can be removed in a follow-up PR.

### DIFF
--- a/components/kubearchive/production/stone-prod-p02/kustomization.yaml
+++ b/components/kubearchive/production/stone-prod-p02/kustomization.yaml
@@ -131,6 +131,7 @@ patches:
           spec:
             containers:
               - name: migration
+                image: quay.io/kubearchive/postgresql:v1.21.3@sha256:2ce1545b19d29bd36c1022de13f451f14847aed04d62ab896a3254d74f7c0d39
                 env:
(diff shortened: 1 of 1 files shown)


olegbet (approved):
/lgtm

eedri:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @eedri |
| **Timestamp** | 2026-05-03T16:51:07.871Z |
| **Risk Level** | low |

approving due to the urgency after the recent outage.

PR #11604, opened 2026-05-06
UI proxy replace long lived token with rotated ones

mshaposhnik:
Downstream version of https://github.com/konflux-ci/konflux-ci/pull/6780

```
We want to migrate from Long-Lived Tokens to Short-Lived Tokens.

Unfortunately, NGINX is not automatically reloading its configuration, so we need to introduce some logic to rebuild the configuration and hot reload it.

This PR introduces a mechanism to regenerate and reload the configuration with a short-lived token provided by Kubernetes. A time based reload approach was preferred to an inotify-based one for its simplicity.
```

Assisted-by: Cursor


components/konflux-ui/staging/stone-stage-p01/remove-run-as-user-proxy-patch.yaml
```diff
@@ -8,11 +8,11 @@
 - op: remove
```
mshaposhnik, line 8: Why don't we remove it from main `Deployment` descriptor instead of removing by patches in every overlay?

filariow: good call. @gbenhaim do you recall about any specific reason for this?

components/konflux-ui/staging/stone-stage-p01/remove-run-as-user-proxy-patch.yaml
```diff
@@ -8,11 +8,11 @@
...
   path: /spec/template/spec/initContainers/2/securityContext/runAsUser
 
-- op: remove
-  path: /spec/template/spec/containers/0/securityContext/runAsUser
-
```
filariow, line 13: I would suggest to keep this consistent with the others. Hence keep this pattern also for container 0.

components/konflux-ui/staging/stone-stg-rh01/remove-run-as-user-proxy-patch.yaml
```diff
@@ -8,11 +8,11 @@
 - op: remove
   path: /spec/template/spec/initContainers/2/securityContext/runAsUser
 
-- op: remove
-  path: /spec/template/spec/containers/0/securityContext/runAsUser
-
```
filariow, line 13: ditto

components/konflux-ui/staging/base/proxy/proxy.yaml
```diff
@@ -132,14 +133,47 @@ spec:
             cpu: 10m
             memory: 64Mi
       containers:
+      - name: generate-nginx-configs-loop
+        image: quay.io/konflux-ci/konflux-ui:68500b32e57278bf33ac18d4ef631ef243abb579
```
filariow: can we use sha256 instead?

filariow (changes requested):
nice work! a few  minor feedback 😄

components/konflux-ui/staging/base/proxy/proxy.yaml
mshaposhnik: Changed & updated (old image was 1 year old)

components/konflux-ui/staging/stone-stage-p01/remove-run-as-user-proxy-patch.yaml
mshaposhnik: \+

components/konflux-ui/staging/stone-stg-rh01/remove-run-as-user-proxy-patch.yaml
mshaposhnik: ack

filariow:
mostly lgtm. It looks like we need to update https://github.com/redhat-appstudio/infra-deployments/blob/main/hack/new-cluster/templates/konflux-ui/remove-run-as-user-proxy-patch.yaml too

components/konflux-ui/staging/stone-stage-p01/remove-run-as-user-proxy-patch.yaml
gbenhaim: just a copy paste error probably. there is no good reason for that.

gbenhaim (approved):
/approve
/lgtm

PR #11672, opened 2026-05-10
feat(KFLUXVNGD-906): development-operator overlay for Konflux operator

yftacherzog:
Introduce a development-operator Argo overlay to deploy the Konflux operator as the path that will eventually replace the individual Konflux microservice components.

The operator component splits an invariant baseline from per-environment material so we can evolve ring-style promotion without duplicating the same pins everywhere.

Today this overlay installs the operator only; follow-up work will add the Konflux CR here so the overlay drives full Konflux deployment via the operator.

Assisted-by: Cursor

Depends on https://github.com/redhat-appstudio/infra-deployments/pull/11683 to pass CI

### DIFF
--- a/argo-cd-apps/overlays/development-operator/delete-legacy-konflux-member-appsets.yaml
+++ b/argo-cd-apps/overlays/development-operator/delete-legacy-konflux-member-appsets.yaml
@@ -1,0 +1,98 @@
+# Strategic-merge deletes: remove these Argo CD ApplicationSets from the
+# effective ../development graph. Used by the development-operator overlay on
+# OpenShift so GitOps still installs shared platform apps but not the older
+# per-microservice Konflux ApplicationSets superseded by the Konflux operator.
+#
+# Only ApplicationSets that actually appear in ../development can be deleted here
+# (Kustomize requires a merge target). konflux-ui and namespace-lister are not in
+# that graph, so no delete entries for them.
+---
+apiVersion: argoproj.io/v1alpha1
+kind: ApplicationSet
+metadata:
+  name: application-api
+$patch: delete
+---
+apiVersion: argoproj.io/v1alpha1
+kind: ApplicationSet
+metadata:
+  name: has
+$patch: delete
+---
+apiVersion: argoproj.io/v1alpha1
+kind: ApplicationSet
+metadata:
+  name: release
+$patch: delete
+---
+apiVersion: argoproj.io/v1alpha1
+kind: ApplicationSet
+metadata:
+  name: integration
+$patch: delete
+---
+apiVersion: argoproj.io/v1alpha1
+kind: ApplicationSet
+metadata:
+  name: build-service
+$patch: delete
+---
+apiVersion: argoproj.io/v1alpha1
+kind: ApplicationSet
+metadata:
+  name: build-templates
+$patch: delete
+---
+apiVersion: argoproj.io/v1alpha1
+kind: ApplicationSet
+metadata:
+  name: image-controller
+$patch: delete
+---
+apiVersion: argoproj.io/v1alpha1
+kind: ApplicationSet
+metadata:
+  name: multi-platform-controller
+$patch: delete
+---
+apiVersion: argoproj.io/v1alpha1
+kind: ApplicationSet
+metadata:
+  name: project-controller
+$patch: delete
+---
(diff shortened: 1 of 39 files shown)


avi-biton:
LGTM,
One question: Can the operator coexist in the cluster with all the individual services, or will the PR get extended to remove the services and modify the CRs with the values from the kustomizations?

yftacherzog:
> LGTM, One question: Can the operator coexist in the cluster with all the individual services, or will the PR get extended to remove the services and modify the CRs with the values from the kustomizations?

@avi-biton It's going to be a process. We first deploy the operator (this PR) and the CR and some e2e tests (some tickets down the road), and once we also bridge the gaps of whatever we're missing in the operator for infra-deployments, we will start removing the individual services from one cluster at a time and replace them with the operator + CR.

Once we're done with all clusters, we will decommission the legacy development overlay.

It's going to take some time... :)

argo-cd-apps/overlays/development-operator/delete-legacy-konflux-member-appsets.yaml
```diff
@@ -0,0 +1,76 @@
...
+kind: ApplicationSet
+metadata:
+  name: internal-services
+$patch: delete
+---
+apiVersion: argoproj.io/v1alpha1
+kind: ApplicationSet
+metadata:
+  name: konflux-kite
+$patch: delete
```
gbenhaim, line 98: **Concern**: Five legacy ApplicationSets appear to be missing from this delete list. The operator has controllers for these services (verified at `operator/internal/controller/` at ref `0bace1e`), so leaving the legacy AppSets active would cause dual deployment — the operator managing them via the Konflux CR while Argo also deploys them independently.

Missing:
- `konflux-info` (operator controller: `info`, CR overlay-patch exists in this PR)
- `konflux-ui` (operator controller: `ui`, CR overlay-patch exists in this PR)
- `namespace-lister` (operator controller: `namespacelister`, CR overlay-patch exists in this PR)
- `konflux-rbac` (operator controller: `rbac`)
- `enterprise-contract` (operator controller: `enterprisecontract`)

These should be added here:

```yaml
---
kind: ApplicationSet
metadata:
  name: konflux-info
$patch: delete
---
kind: ApplicationSet
metadata:
  name: konflux-ui
$patch: delete
---
kind: ApplicationSet
metadata:
  name: namespace-lister
$patch: delete
---
kind: ApplicationSet
metadata:
  name: konflux-rbac
$patch: delete
---
kind: ApplicationSet
metadata:
  name: enterprise-contract
$patch: delete
```

Note: `cert-manager` is intentionally not listed — its legacy AppSet deploys the cert-manager operator itself, while the operator's `certmanager` controller only configures a ClusterIssuer on top.

gbenhaim (commented):
_This review was posted by **Gal's Cursor**._

argo-cd-apps/overlays/development-operator/delete-legacy-konflux-member-appsets.yaml
yftacherzog: Good catch

yftacherzog: `konflux-ui` and `namespace-lister` are already removed in the development overlay. I'm not sure I understand how they are installed there though...

yftacherzog: I added the rest you mentioned.

gbenhaim: They are not installed in development. Another benefit that the operator will bring.

avi-biton:
We can update the integration service with the new CR costumizations

yftacherzog:
> We can update the integration service with the new CR costumizations

We're pinning an older digest, so we'd need to also update that.

Also, the preview script is still not waiting for the CR readiness with this change, so the CR parts are mainly placeholders, while we will actually make it work in another PR.

gbenhaim:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @gbenhaim |
| **Timestamp** | 2026-05-12T12:07:48.215Z |

PR #11588, opened 2026-05-04
KONFLUX-13356: Upgrade RHBoK operator and tekton-kueue in dev and staging

glevi-rh:
## What
Upgrade the kueue stack in development and staging and align CRs with kueue 0.16.x conversion webhook behavior:
- Update RHBoK (Kueue operator) channel from `stable-v1.2` to `stable-v1.3`
- Update tekton-kueue to `cec0b3c` (built against kueue v0.16.6 / k8s.io v0.35.3)
- Align ClusterQueue fields with v1beta2 conversion webhook output to prevent ArgoCD drift:
  - `Borrow` (deprecated) → `MayStopSearch`
  - Remove `stopPolicy: None` (default value, not set by the webhook)
- Update Kyverno LocalQueue policy apiVersion to `v1beta2`
- Update chainsaw tests for v1beta2
- Pin CatalogSource (`redhat-operator-index:v4.18`) in development to ensure `stable-v1.3` channel availability (matching staging pattern)

CRs remain on `apiVersion: v1beta1` — full v1beta2 migration will follow in a separate PR.

[KONFLUX-13356](https://issues.redhat.com/browse/KONFLUX-13356)

## Why
To fix CVE-2026-40938 in tekton-kueue, we need to bump `tektoncd/pipeline` from v1.10.2 to v1.11.1. This bump pulls in `k8s.io` v0.35+, which breaks kueue v0.14.x due to breaking API changes in the k8s libraries.

RHBoK 1.3 is based on kueue v0.16.6, which serves v1beta2 (new storage version), v1beta1 (still supported), and v1 APIs. Additionally, RHBoK 1.2 is now out of support (per #forum-ocp-kueue).

The kueue 0.16.x conversion webhook round-trips v1beta1 resources through v1beta2, mutating deprecated field values (`Borrow` → `MayStopSearch`) and stripping defaults (`stopPolicy: None`). Without aligning the manifests to the post-conversion values, ArgoCD sees perpetual drift.

The development overlay previously relied on the default `redhat-operators` catalog, which may not include `stable-v1.3` on all OCP 4.18.x cluster versions. A pinned CatalogSource (same as staging) guarantees channel availability.

Rollout plan:
1. **Upgrade operator + tekton-kueue + align CR fields** (this PR)
2. **Migrate CRs apiVersion to v1beta2** (follow-up PR)
3. Bump tektoncd/pipeline to v1.11.1
4. Update Renovate rules

## Validation
- `kustomize build` passes for all affected paths:
[...]

PR #11692, opened 2026-05-10
KONFLUX-13356: Migrate kueue CRs to v1beta2 in dev and staging

glevi-rh:
## What
Migrate all kueue custom resources from `v1beta1` to `v1beta2` apiVersion in development and staging:
- ClusterQueue, ResourceFlavor, AdmissionCheck, WorkloadPriorityClass
- Switch staging ClusterQueues from `admissionChecks` to `admissionChecksStrategy` (prevents ArgoCD drift from conversion webhook)
- Normalize `nominalQuota: '1000'` to `1k` (Kubernetes `resource.Quantity` canonical form) to resolve ArgoCD drift on `linux-x86-64`, `local`, and `localhost` platform quotas
- Update `update-kueue-vm-quotas.py` to emit canonical `1k` notation for v1beta2 ClusterQueues while preserving string form (`'1000'`) for v1beta1 (production) files

[KONFLUX-13356](https://issues.redhat.com/browse/KONFLUX-13356)

## Why
Follow-up to PR #11588 which upgraded RHBoK to v1.3 (kueue 0.16.x). The new kueue version uses v1beta2 as the storage version. The conversion webhook round-trips v1beta1 resources through v1beta2, causing field mutations that lead to ArgoCD drift. Migrating CRs to v1beta2 eliminates the conversion layer entirely.

The `'1000'` → `1k` fix aligns the manifest with the Kubernetes `resource.Quantity` canonical form — the API server stores `1000` as `1k`, so using the canonical notation prevents drift.

## Validation
- `kustomize build` passes for all affected paths
- Chainsaw tests already updated for v1beta2 in PR #11588
- Status of staging clusters:
```
stone-stg-rh01:
- CSV: kueue-operator.v1.3.1 — Succeeded
- Operator pods: 2/2 Running 
- kueue-controller-manager: 2/2 Running
- tekton-kueue: all 5 pods Running
- ClusterQueue: 7 pending workloads being processed
- Workloads: actively being admitted and finishing
- API: kueue.x-k8s.io/v1beta2

stone-stage-p01:
- CSV: kueue-operator.v1.3.1 — Succeeded
- Operator pods: 2/2 Running
- kueue-controller-manager: 2/2 Running
- tekton-kueue: all 5 pods Running, 0 restarts
- ClusterQueue: 0 pending (all caught up)
- Workloads: actively admitted and finishing
- API: kueue.x-k8s.io/v1beta2
```

## Risk Assessment
**Risk Level:** Low
**Rollback:** Revert apiVersion back to `v1beta1` and `admissionChecksStrategy` back to `admissionChecks`
Development and staging only — no production impact. v1beta2 is the storage version in kueue 0.16.x, so this aligns manifests with how resources are actually stored.

[KONFLUX-13356]: https://redhat.atlassian.net/browse/KONFLUX-13356?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/hack/kueue-vm-quotas/update-kueue-vm-quotas.py
+++ b/hack/kueue-vm-quotas/update-kueue-vm-quotas.py
@@ -23,9 +23,12 @@ See README.md for detailed usage examples and workflow integration.
 import yaml
 import sys
 import argparse
+from decimal import Decimal
 from pathlib import Path
 from typing import Dict, List, Any
 from dataclasses import dataclass
+
+from kubernetes.utils.quantity import format_quantity
 
 
 @dataclass
@@ -154,17 +157,29 @@ def distribute_platforms(platforms: List[PlatformQuota], max_per_group: int = 16
     return groups
 
 
-def create_platform_resource_group(platforms: List[PlatformQuota]) -> ResourceGroup:
+def format_quota(quota: int) -> str:
+    """Format quota using Kubernetes resource.Quantity canonical form.
+
+    Kubernetes normalizes 1000 -> '1k' when storing resource.Quantity values.
+    Using canonical form prevents ArgoCD drift on v1beta2 ClusterQueues.
+    """
+    if quota % 1000 == 0 and quota >= 1000:
+        return format_quantity(Decimal(quota), 'k', quantize=Decimal(1))
+    return str(quota)
+
+
+def create_platform_resource_group(platforms: List[PlatformQuota], is_v1beta2: bool = False) -> ResourceGroup:
     """Create a resource group for a list of platforms."""
-    # Only include platform names in covered resources (no tekton.dev/pipelineruns)
     covered_resources = [p.name for p in sorted(platforms, key=lambda p: p.name)]
-    
-    # Create individual platform resources
+
     resources = [
-        ResourceSpec(name=p.name, nominal_quota=str(p.quota))
+        ResourceSpec(
+            name=p.name,
+            nominal_quota=format_quota(p.quota) if is_v1beta2 else str(p.quota),
+        )
         for p in sorted(platforms, key=lambda p: p.name)
     ]
-    
+
     return ResourceGroup(covered_resources, resources)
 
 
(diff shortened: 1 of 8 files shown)


hack/kueue-vm-quotas/update-kueue-vm-quotas.py
```diff
@@ -154,17 +154,29 @@ def distribute_platforms(platforms: List[PlatformQuota], max_per_group: int = 16
     return groups
 
 
-def create_platform_resource_group(platforms: List[PlatformQuota]) -> ResourceGroup:
+def format_quota(quota: int) -> str:
```
filariow, line 160: can we use something maintained by the kubernetes community instead? maybe https://github.com/kubernetes-client/python/blob/master/kubernetes/utils/quantity.py#L88C5-L88C20

filariow (changes requested):

hack/kueue-vm-quotas/update-kueue-vm-quotas.py
glevi-rh: If we do that, we'll add more dependencies for a very simple function and we need to wrap things around it, here's an example using the `quantity.py`:
```
from kubernetes.utils.quantity import format_quantity
from decimal import Decimal

def format_quota(quota: int) -> str:
    if quota % 1000 == 0 and quota >= 1000:
        return format_quantity(Decimal(quota // 1000), 'k')
    return str(quota)
```

filariow: I don't think the example is right. Also the function will take care of other cases too

filariow: Double-checking on this it seems to me you and your example are right. I understood that the function from kubernetes python library would have provided the same functionality of Go's `Canonicalize`, but it doesn't indeed.

Still it could be good to have it here. We need to update the [Requirements section of the README.md](https://github.com/redhat-appstudio/infra-deployments/blob/ce3137d2f4a021442a45e4001e657f3c37bb0b2b/hack/kueue-vm-quotas/README.md?plain=1#L127) file though

glevi-rh: Agreed — Python's `format_quantity()` doesn't auto-pick the canonical suffix like Go's `Quantity.Canonicalize()` does, but it's still good to use for forward-compatibility with upstream Kubernetes changes - as you suggested.

Updated the README requirements section and CI workflow in the latest commit (d82458f).

filariow (approved):
/lgtm

PR #11758, opened 2026-05-12
feat(KONFLUX-13961): Update perf in-cluster dashboards

jhutar:


jhutar:
## Qodo response

**Undocumented prod ref skew** - That is intentional, we are supposed to deploy to Stage first and only then to Prod.

FaisalAl-Rayes (approved):
/lgtm

PR #11761, opened 2026-05-12
dev/stage move PaC config from "spec.platforms.openshift" to "spec.platforms.kubernetes"

aThorp96:
Both are supposed to be supported, but PaC uninstalled when the below change was applied:

See: https://github.com/tektoncd/operator/pull/3337
https://redhat-internal.slack.com/archives/C06GFP9JJBV/p1778615280508879


aThorp96:
/hold

Since we're not 100% sure this is the actual problem

aThorp96:
May not be necessary after https://github.com/redhat-appstudio/infra-deployments/pull/11762

aThorp96:
Not needed after all, the spec change was a red herring

PR #11789, opened 2026-05-13
unroll build-service update, which contains change to Repository CR

rcerven:
naming

which causes issue when PaC is adding repository label to pipeline run and trimming it to 62 chars which can result at starting '.' which isn't valid value


hugares:
this is a revert, not waiting for e2e

PR #11813, opened 2026-05-14
increase timeout of kueue's policies PreSyncHook

filariow:
We are experiencing flakiness and we suspect this PreSyncHook is causing it.
What looks to be happening is that this check is started before Kueue and Tekton are available.
If they are not installed within 5min, the `policies` component will never get Synced.

This change increase its timeout to a relatively high value to prove this theory.


flacatus (approved):

hugares (approved):
/lgtm

components/policies/development/kueue/pre-sync-hook.yaml
```diff
@@ -66,7 +66,7 @@ spec:
           echo "Starting CRD check..."
 
           # Configuration
-          TIMEOUT=300  # 5 minutes timeout
+          TIMEOUT=1800  # 20 minutes timeout
```
filariow: It seems I'm not good at math

```suggestion
          TIMEOUT=1200  # 20 minutes timeout
```

hugares (approved):
/lgtm

sadlerap (approved):

gbenhaim (approved):

PR #11817, opened 2026-05-14
fix: sync fork with upstream before creating preview branch

flacatus:
preview.sh now fetches and merges upstream/main before creating the preview branch, ensuring component bumps (like build-service) are included even when the fork's branch is behind upstream.


rh-hemartin:
From KubeArchive side I approve this change!

/approve
/lgtm

PR #11864, opened 2026-05-18
fix: add pipelineruns/status RBAC for PAC watcher

mathur07:
PAC v0.46.0 (PR tektoncd/pipelines-as-code#2667) introduced informer cache transforms that strip PipelineRun Status fields. This causes the Knative-generated reconciler to call UpdateStatus() on the /status subresource, which the operator-managed watcher ClusterRole doesn't grant.

Add a supplementary ClusterRole/Binding as a workaround until the upstream PAC fix lands in config/202-watcher-role.yaml.


mathur07:
Superseded by #11865

PR #11875, opened 2026-05-19
feat(KONFLUX-14043): Enable appsre-stonesoup-vault in remaining NSs to be able to use ExternalSecrets from there

jhutar:
To be able to use ExternalSecrets from `konflux-perfscale-2-tenant` and `konflux-perfscale-3-tenant` (not only `konflux-perfscale-2-tenant`), we need to enable access to the secrets storage from there.

Once this is merged and validate, will create a followup PR to update it in Production overlay as well.

## Risk Assessment
**Risk Level:** Low
**Description:** Just adding two more namespaces to already established list
**Rollback:** Revert PR

## Validation
This is to test in Staging overlay.

### DIFF
--- a/components/cluster-secret-store/staging/rhtap-promotion-staging-patch.yaml
+++ b/components/cluster-secret-store/staging/rhtap-promotion-staging-patch.yaml
@@ -2,3 +2,9 @@
 - op: add
   path: /spec/conditions/0/namespaces/-
   value: rhtap-promotion-staging
+- op: add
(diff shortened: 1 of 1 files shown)


jhutar:
## Qodo response

* **Expanded Vault access scope** - I'm just following same path as what was there already for `konflux-perfscale-1-tenant`

manish-jangra (approved):
/approve
/lgtm

PR #11890, opened 2026-05-19
KFLUXINFRA-3827: Remove group-sync resources

enkeefe00:
## Context

Since k8s-groups has been successfully deployed on all tenant Konflux clusters (https://github.com/redhat-appstudio/infra-deployments/pull/11835 & https://github.com/redhat-appstudio/infra-deployments/pull/11816), the Group Sync Operator is no longer needed. This commit deletes all Group Sync related resources (which are no longer deployed on any tenant clusters).

## Risk
**Level**: Low
**Explanation**:  These resources aren't deployed on any clusters anyway, so deleting them from the repository shouldn't have any averse affects. Additionally, the generated resource files from the `kustomize build` command have been compared from before and after this commit; no components should be drastically affected.


enkeefe00:
Wrong branch selected

PR #11900, opened 2026-05-20
feat(KFLUXVNGD-907): add konflux operator CR

yftacherzog:
Finalize the development overlay CR for the Konflux Operator.

Also drop the appstudio-pipelines-scc from the pipeline-service for the development-operator overlay, as it is deployed by the operator.

This allows deploying Konflux using the operator in dev environments.

Assisted-by: Cursor

### DIFF
--- a/components/konflux-operator/development/invariant/release-config.yaml
+++ b/components/konflux-operator/development/invariant/release-config.yaml
@@ -7,6 +7,6 @@ spec:
   certManager:
     createClusterIssuer: true
   internalRegistry:
-    enabled: true
+    enabled: false
   defaultTenant:
-    enabled: true
+    enabled: false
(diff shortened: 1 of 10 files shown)


components/konflux-operator/development/invariant/release-config.yaml
```diff
@@ -9,4 +9,4 @@ spec:
   internalRegistry:
     enabled: true
```
yftacherzog: Should be false

gbenhaim (approved):

components/pipeline-service/development-operator/kustomization.yaml
aThorp96, line 1: I'm a little confused about this change. Is this intended to ultimately replace `components/pipeline-service/development` once it's been validated?

yftacherzog: Yes. It will happen once we migrate all clusters to deploy Konflux using the [Konflux Operator](https://github.com/konflux-ci/konflux-ci). At that point, we'll be able to decommission the then legacy development overlay that is based on Kustomize.

avi-biton:
We should add OWNERS to the `default-tenant`

PR #11901, opened 2026-05-20
SPRE-5349: add kube_deployment_status_replicas_updated metric for tekton-kueue

asafaviv-devops:
- Add kube_deployment_status_replicas_updated to the staging scrape config for tekton-kueue namespace
  - This metric is required by the TektonKueueRolloutStuck alert (o11y PR #1148) to detect stuck rollouts where new pods fail to start while old pods continue serving
  - Without this metric in the endpoints-params, the alert cannot evaluate in RHOBS since the metric is not forwarded via remote write


pumahaka (approved):

pumahaka (approved):
LGTM

PR #11928, opened 2026-05-20
KFLUXVNGD-999: increase cache TTL and add service traffic distribution to artifact-registry-proxy

amisstea:
Increase nginx cache TTL to 30d so cached S3 redirect responses are considered fresh for 30 days before being marked stale. Enable PreferClose traffic distribution to route requests to the nearest proxy replica, reducing cross-zone latency.

Assisted-by: Claude claude-opus-4-6


kelchen123 (approved):

hmariset (approved):
/lgtm

PR #11933, opened 2026-05-21
adding custom-kube-metric concurrentPLR stage - KFLUXINFRA-3577

celeztyne:
Adding a new metric to get the concurrent PLR in an specific NS based in custom-kube-state-metric. It would record the amount of unfinished PLR for a NS in an specific time.
This is to be able to get information to check if the limit to concurrent PLR of 5000 for mintmaker is reached [KFLUXINFRA-3577](https://redhat.atlassian.net/browse/KFLUXINFRA-3577)

[KFLUXINFRA-3577]: https://redhat.atlassian.net/browse/KFLUXINFRA-3577?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

If goes OK in stage will be implemented in Production too

### DIFF
--- a/components/monitoring/custom-kube-state-metrics/staging/custom-resource-state-config.yaml
+++ b/components/monitoring/custom-kube-state-metrics/staging/custom-resource-state-config.yaml
@@ -75,7 +75,7 @@ data:
                   path: [metadata, deletionTimestamp]
                   nilIsZero: true
         ###############################################################################
-        # Measures the number of concurrent PipelineRuns in an unfinished or Unknown  
+        # Measures the number of concurrent PipelineRuns in an unfinished or Unknown
         # status
         ###############################################################################
             - name: "konflux_pipelinerun_status"
--- a/components/monitoring/custom-kube-state-metrics/staging/custom-resource-state-config.yaml
+++ b/components/monitoring/custom-kube-state-metrics/staging/custom-resource-state-config.yaml
@@ -84,4 +84,4 @@ data:
                 type: Gauge
                 gauge:
                   path: [status, completionTime]
-                  nilIsZero: true                    
+                  nilIsZero: true
--- a/components/monitoring/custom-kube-state-metrics/staging/custom-resource-state-config.yaml
+++ b/components/monitoring/custom-kube-state-metrics/staging/custom-resource-state-config.yaml
@@ -75,18 +75,13 @@ data:
                   path: [metadata, deletionTimestamp]
                   nilIsZero: true
         ###############################################################################
-        # Measures the number of concurrent PipelineRuns in an unfinished or Unknown
-        # status
+        # Fetchs the concurrent PipelineRuns in an unfinished or Unknown state filtering
+        # the ones with completionTime unset (equal to zero)
         ###############################################################################
             - name: "konflux_pipelinerun_status"
-              help: "1 if PipelineRun is running (status=Unknown), 0 if finished (succeeded or failed). Sum to count concurrent runs."
+              help: "PipelineRun completion time with status label; 0 means still running. Filter status=Unknown for concurrent runs."
               each:
                 type: Gauge
                 gauge:
-                  path: [status, conditions, "[type=Succeeded]", status]
-                  nilIsZero: true
-                  valueMap:
-                    Unknown: 1
-                    True: 0
-                    False: 0
-                    
+                  path: [status, completionTime]
+                  nilIsZero: true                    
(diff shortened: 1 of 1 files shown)


components/monitoring/custom-kube-state-metrics/staging/custom-resource-state-config.yaml
```diff
@@ -74,3 +74,15 @@ data:
...
                   path: [metadata, deletionTimestamp]
                   nilIsZero: true
+        ###############################################################################
+        # Measures the number of concurrent PipelineRuns in an unfinished or Unknown
+        # status
+        ###############################################################################
+            - name: "konflux_pipelinerun_status"
+              help: "PipelineRun status: Unknown=running, True=succeeded, False=failed. Count status=Unknown for concurrent runs."
+              each:
+                type: StateSet
```
glevi-rh: As far as I remember, in our case the only type that works for `kube-state-metrics` and UWM Prometheus, is `Guage`.

celeztyne: OK I didn't know that, in that case I only need to know the PLN currently running, so I can use gauge. The idea was to have more information, but if only gauge is working is OK, I'll submit the changes

celeztyne: Updated

glevi-rh (changes requested):
**The metric won't work as configured.** I tested locally with kube-state-metrics v2.11.0 against the staging cluster:

1. **`valueMap` doesn't exist in v2.11.0** — it's silently ignored. It doesn't appear in the [docs](https://github.com/kubernetes/kube-state-metrics/blob/v2.11.0/docs/customresourcestate-metrics.md) or the source code.

2. **Gauge can't parse the status string values** — the `[type=Succeeded]` path filter works correctly and finds the condition, but the `status` field value is a string (`Unknown`, `True`, `False`). Gauge tries `strconv.ParseFloat` on it and fails:
   ```
   E0604 "kube_customresource_konflux_pipelinerun_status" err="[status,conditions,[type=Succeeded],status]: []: strconv.ParseFloat: parsing \"Unknown\": invalid syntax"
   ```
   The metric header is emitted but **zero data points** are produced.

3. There are some linting issues that are failing CI — would be good to fix those as well.

Gauge can auto-convert `"true"`/`"false"` (case-insensitive) to 1.0/0.0, but `"Unknown"` isn't handled. We need a different approach to map these values to numbers.

celeztyne:
- Changes done to the parameters to calculate the metric
- Issues with trailspaces solved

components/monitoring/custom-kube-state-metrics/staging/custom-resource-state-config.yaml
```diff
@@ -74,3 +74,14 @@ data:
                 gauge:
                   path: [metadata, deletionTimestamp]
                   nilIsZero: true
+        ###############################################################################
+        # Fetchs the concurrent PipelineRuns in an unfinished or Unknown state filtering
+        # the ones with completionTime unset (equal to zero)
+        ###############################################################################
+            - name: "konflux_pipelinerun_status"
+              help: "PipelineRun completion time with status label; 0 means still running. Filter status=Unknown for concurrent runs."
+              each:
```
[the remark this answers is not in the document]
celeztyne: Updated

components/monitoring/custom-kube-state-metrics/staging/custom-resource-state-config.yaml
glevi-rh, line 68:
```suggestion
            finalizers: [metadata, finalizers]
            status: [status, conditions, "[type=Succeeded]", reason]
```

Without it, you can't really see the status label.

glevi-rh (changes requested):

components/monitoring/custom-kube-state-metrics/staging/custom-resource-state-config.yaml
celeztyne: Discussed with Gal in a dm

celeztyne: Suggestion taken

glevi-rh (approved):
/lgtm

enkeefe00 (approved):

PR #11939, opened 2026-05-21
feat(perf-team): Populate konflux-performance group on kflux-fedora-01

jhutar:
The kflux-fedora-01 cluster uses the Fedora Account System (FAS) rather than standard Rover groups. As a result, the existing konflux-performance group used for component-maintainer access was empty, denying access to the performance team.

This commit creates a dedicated kflux-fedora-01 overlay for the component, defining an OpenShift Group named `konflux-performance` manually populated with the team members from the OWNERS file. Because this shares the same name as the group used by standard clusters, it securely re-uses the existing RoleBindings for component maintainer access and cluster-wide view permissions.

Generated-by: Gemini

### DIFF
--- a/components/perf-team-prometheus-reader/production/kflux-fedora-01/fedora-access.yaml
+++ b/components/perf-team-prometheus-reader/production/kflux-fedora-01/fedora-access.yaml
@@ -1,0 +1,8 @@
+apiVersion: user.openshift.io/v1
+kind: Group
+metadata:
+  name: konflux-performance
+users:
+  - jhutar
+  - mcharanrm
+  - smodak-rh
--- a/components/perf-team-prometheus-reader/production/kflux-fedora-01/kustomization.yaml
+++ b/components/perf-team-prometheus-reader/production/kflux-fedora-01/kustomization.yaml
@@ -1,0 +1,5 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - ../base
+  - fedora-access.yaml
--- a/argo-cd-apps/base/member/infra-deployments/perf-team-prometheus-reader/perf-team-prometheus-reader.yaml
+++ b/argo-cd-apps/base/member/infra-deployments/perf-team-prometheus-reader/perf-team-prometheus-reader.yaml
@@ -14,7 +14,9 @@ spec:
                 environment: staging
                 clusterDir: base
           - list:
-              elements: []
+              elements:
+                - nameNormalized: kflux-fedora-01
+                  values.clusterDir: kflux-fedora-01
   template:
     metadata:
       name: perf-team-prometheus-reader-{{nameNormalized}}


components/perf-team-prometheus-reader/production/kflux-fedora-01/kustomization.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - ../base
+  - fedora-access.yaml
```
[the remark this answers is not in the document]
jhutar: This is actually handled correctly by the existing Kustomize overlays. 

While the base `ApplicationSet` defaults to `environment: staging`, the production ArgoCD instances use overlays (e.g., `argo-cd-apps/overlays/production-downstream/kustomization.yaml` and `argo-cd-apps/overlays/konflux-public-production/kustomization.yaml`) which apply `production-overlay-patch.yaml`.

This patch explicitly overwrites the environment value:

```
  - op: add
    path: /spec/generators/0/merge/generators/0/clusters/values/environment
    value: production
```

Because `kflux-fedora-01` is a production cluster, ArgoCD will successfully evaluate the path as `components/perf-team-prometheus-reader/production/kflux-fedora-01` and include the new overlay.

kdelee:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @kdelee |
| **Timestamp** | 2026-05-26T11:16:48.889Z |

components/perf-team-prometheus-reader/production/kflux-fedora-01/fedora-access.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: user.openshift.io/v1
+kind: Group
+metadata:
+  name: konflux-performance
+users:
+  - jhutar
```
hugares, line 6: We have a mechanism in place to populate the k8s groups from with FAS accounts. You need to add people there: https://github.com/redhat-appstudio/internal-infra-deployments/blob/main/components/k8s-groups/production/fas/groups/konflux-performance.yaml 

@enkeefe00 Can you work Jan on this, we might need to give permission to see that repo

hugares (changes requested):

components/perf-team-prometheus-reader/production/kflux-fedora-01/fedora-access.yaml
jhutar: I see! Looks like I'm not able to fork the repo to be able to create a PR.

hugares: try now, create a PR from a branch in the repo, not a fork

jhutar:
As pointed out by Hugo, this needs to be solved in another repo.

components/perf-team-prometheus-reader/production/kflux-fedora-01/fedora-access.yaml
jhutar: Thank you. Created https://github.com/redhat-appstudio/internal-infra-deployments/pull/84

PR #11983, opened 2026-05-26
Allow konflux-performance group to create SA tokens in perf-team-prometheus-reader

jhutar:
## Summary
- Add Role and RoleBinding granting the `konflux-performance` group permission to create tokens for `perf-team-prometheus-reader-cluster-sa` and `perf-team-prometheus-reader-oomcrash-sa` service accounts
- Added to `base/core` which is used by development and staging overlays only (production has its own standalone resources and is intentionally excluded until soak time validates the change)

## Risk Assessment
- **Level:** Low
- **Description:** Adds narrowly scoped RBAC (create on serviceaccounts/token for two specific SAs) to a single namespace. No existing permissions are modified.
- **Rollback plan:** Revert the PR

## Validation
- `kustomize build` passes for development and staging overlays
- Production overlay confirmed unaffected (0 references to the new resources)

### DIFF
--- a/components/perf-team-prometheus-reader/base/core/kustomization.yaml
+++ b/components/perf-team-prometheus-reader/base/core/kustomization.yaml
@@ -5,3 +5,4 @@ resources:
   - serviceaccount.yaml
   - serviceaccount-oomcrash.yaml
   - perf-team.yaml
+  - sa-token-rbac.yaml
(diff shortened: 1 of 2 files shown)


filariow (approved):
/lgtm

jhutar:
## Qodo response

* **Group-wide SA impersonation** - we need these permissions to be able to rotate the secrets. Also the role is locked down to two specific SAs only.

PR #11995, opened 2026-05-26
Switch back to ppc in us-east

hugares:
We are done with testing in us-south, go back to us-east to be able to clean up IBMC resources in us-south.

Also clean up ssh key ExternalSecrets, since we no longer have 2 for ppc we can use standard name for the one left.

[KFLUXINFRA-3315](https://redhat.atlassian.net/browse/KFLUXINFRA-3315)

[KFLUXINFRA-3315]: https://redhat.atlassian.net/browse/KFLUXINFRA-3315?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/components/multi-platform-controller/staging-downstream/external-secrets.yaml
+++ b/components/multi-platform-controller/staging-downstream/external-secrets.yaml
@@ -80,29 +80,6 @@ metadata:
 spec:
   dataFrom:
     - extract:
-        key: staging/infrastructure/multi-platform-controller/stone-stage-p01/ibm-ppc64le-ssh-key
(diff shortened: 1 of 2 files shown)


hugares:
/hold
I am validating that we have right ssh key

hugares:
> /hold I am validating that we have right ssh key

/unhold

All clear, the ssh key is the right one

sadlerap (approved):

PR #12001, opened 2026-05-26
feat(KFLUXVNGD-918): add changelog generation library and client extension

kelchen123:
For [KFLUXVNGD-918](https://redhat.atlassian.net/browse/KFLUXVNGD-918) (generate changelog for infra-deployments PRs).

Adds the core library under `infra-tools/internal/changelog/`:
- `extractor.go` — reads the operator image SHA from `kustomization.yaml` at two git refs via a worktree
- `bumps.go` — detects sub-service `?ref=` changes from a GitHub compare patch
- `upstream.go` — fetches commit comparisons from GitHub (network behind an interface, so tests need no network)
- `filter.go` — conventional commits filter and two-tier operator commit splitter (notable vs collapsible remaining)
- `formatter.go` — renders everything into the final markdown comment
Also adds `UpsertCommentByMarker` to `internal/github/comments.go` so the binary (PR 2) can post idempotent comments with a custom marker.

[KFLUXVNGD-918]: https://redhat.atlassian.net/browse/KFLUXVNGD-918?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/infra-tools/internal/changelog/bumps.go
+++ b/infra-tools/internal/changelog/bumps.go
@@ -1,0 +1,125 @@
+package changelog
+
+import (
+	"regexp"
+	"strings"
+)
+
+// ServiceBump describes a sub-service whose pinned commit SHA changed between
+// the old and new operator versions.
+type ServiceBump struct {
+	// Owner is the GitHub organisation that owns the service repo (e.g. "konflux-ci").
+	Owner string
+	// Repo is the GitHub repository name (e.g. "build-service", "integration-service").
+	Repo string
+	// OldSHA is the previously pinned commit.
+	OldSHA string
+	// NewSHA is the newly pinned commit.
+	NewSHA string
+}
+
+// refPattern matches a ?ref=<40-char-sha> in a unified diff line.
+var refPattern = regexp.MustCompile(`\?ref=([0-9a-f]{40})`)
+
+// githubRepoPattern extracts the owner and repo from a github.com URL.
+// Matches: https://github.com/{owner}/{repo}/...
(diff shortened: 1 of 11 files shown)


yftacherzog (changes requested):
I think the PR is too big. I'm unable to understand the context of things.

I suggest finding a way we can introduce a minimal framework that will allow us to see the comment on PRs such as https://github.com/redhat-appstudio/infra-deployments/pull/12000 (can also use it as test cases material), and then enhance the framework over multiple PRs until we have the functionality we're after.

I'd imagine something like having the following PRs:

1. Creating the workflow that will have the relevant trigger so it comments only on the relevant PR + minimal call to placeholder go code.
2. Extract operator changes and put them in the comment.
3. Find the upstream refs and put them in the comment.
4. Use the upstream refs to get the upstream changes and put them in the comment (maybe replacing what was extracted on point 3).
5. Additional enhancements.

Ideally, we should also have test cases that will demonstrate the evolving overall behavior in each iteration, and also once merged, we can get feedback from PRs such as the one I mentioned above.

Overall, this will allow us to better understand what we're doing and get feedback and make corrections faster.

We can (we should) break KFLUXVNGD-918 into smaller tickets that will allow us to do this incremental work.
I suggest creating additional tickets in the epic per the steps that you think will allow us to fulfill the original ticket. This way, you'd also be able to show progress, and not feel like the work is stuck, because we're trying to land something very big in one go.

PR #12018, opened 2026-05-27
KFLUXVNGD-1000 Restructure squid production overlay - ring 1

hmariset:
## Summary
- Add per-cluster production overlays for set 1 clusters: kflux-ocp-p01, kflux-prd-rh02, stone-prod-p01
- Each overlay is self-contained with its own helm generator and credentials (follows kyverno pattern)
- Existing flat production files are kept — unmigrated clusters continue deploying from them
- Update ApplicationSet to route set 1 clusters to their per-cluster overlays

Other clusters update will follow in subsequent PRs. No functional change — set 1 clusters deploy identical config via their own overlay instead of the shared flat directory.

Assisted-by: Claude Code

## Risk Assessment
- **Level:** Low
- **Description:** Set 1 clusters switch from `production/` to `production/<cluster>/` but the rendered manifests are identical (verified via kustomize build checksum). Remaining clusters are unaffected.
- **Rollback plan:** Revert the PR

## Validation
- `kustomize build --enable-helm` verified for all 3 set 1 overlays — identical output to flat production (md5: `3d15b6f7b3e884a2b60c3f94e20a1603`)
- Flat `production/` still builds correctly for unmigrated clusters

### DIFF
--- a/argo-cd-apps/base/member/infra-deployments/squid/squid.yaml
+++ b/argo-cd-apps/base/member/infra-deployments/squid/squid.yaml
@@ -17,7 +17,13 @@ spec:
                 environment: staging
                 clusterDir: ""
           - list:
-              elements: []
+              elements:
+                - nameNormalized: kflux-ocp-p01
+                  values.clusterDir: kflux-ocp-p01
+                - nameNormalized: kflux-prd-rh02
+                  values.clusterDir: kflux-prd-rh02
+                - nameNormalized: stone-prod-p01
+                  values.clusterDir: stone-prod-p01
   template:
     metadata:
       name: squid-{{nameNormalized}}
--- a/components/squid/production/kflux-ocp-p01/artifact-registry-credentials.yaml
+++ b/components/squid/production/kflux-ocp-p01/artifact-registry-credentials.yaml
@@ -1,0 +1,22 @@
+apiVersion: external-secrets.io/v1
+kind: ExternalSecret
+metadata:
+  name: artifact-registry-credentials
+  namespace: caching
+  annotations:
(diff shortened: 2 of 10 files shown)


argo-cd-apps/overlays/production-downstream/kustomization.yaml
```diff
@@ -255,7 +255,7 @@ patches:
       kind: ApplicationSet
       version: v1alpha1
       name: pulp-access-controller
-  - path: production-overlay-patch.yaml
```
gbenhaim, line 258: why do we need a special overlay for squid?

gbenhaim:
@hmariset is it possible to split this PR? right now if I merged it I affect all the konflux cluster's at once.

argo-cd-apps/overlays/production-downstream/kustomization.yaml
hmariset: the shared patch only sets `environment: production`, but squid also needs `clusterDir: empty-base` to route unlisted clusters to the fallback directory. The base ApplicationSet defaults `clusterDir: ""` which works for staging/dev, but production needs it set to empty-base.

hmariset:
> @hmariset is it possible to split this PR? right now if I merged it I affect all the konflux cluster's at once.

Updated this PR to ring 1 only — flat production files are kept so unmigrated clusters are unaffected. Please re-review.

gbenhaim (approved):

gbenhaim:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @gbenhaim |
| **Timestamp** | 2026-05-28T14:20:46.682Z |

PR #12030, opened 2026-05-28
feat(KFLUXVNGD-1022): add operator changelog workflow and placeholder comment

kelchen123:
Adds the changelog-generator workflow and a minimal binary that posts a placeholder comment on any PR that bumps the operator SHA, proving the trigger-to-comment pipeline before changelog logic is added in subsequent PRs.

### DIFF
--- a/.github/workflows/operator-changelog.yaml
+++ b/.github/workflows/operator-changelog.yaml
@@ -1,0 +1,39 @@
+name: Operator Changelog
+
+on:
+  pull_request_target:
+    types: [opened, synchronize, reopened]
+    paths:
+      - 'components/konflux-operator/development/invariant/kustomization.yaml'
+
+jobs:
+  changelog:
+    runs-on: ubuntu-latest
+    concurrency:
+      group: operator-changelog-${{ github.event.pull_request.number }}
+      cancel-in-progress: true
+    permissions:
+      pull-requests: write
+      contents: read
+    steps:
+      # Check out the BASE branch (trusted code) so we never execute fork code.
+      # The binary is built here with the elevated GITHUB_TOKEN — a malicious
+      # fork cannot inject code into this step.
+      - uses: actions/checkout@v6
+
+      - uses: actions/setup-go@v5
+        with:
+          go-version-file: infra-tools/go.mod
+          cache-dependency-path: infra-tools/go.sum
+
+      - name: Build changelog-generator
+        working-directory: infra-tools
+        run: go build -o bin/changelog-generator ./cmd/changelog-generator
+
+      - name: Run changelog-generator
+        working-directory: infra-tools
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_REPOSITORY: ${{ github.repository }}
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+        run: ./bin/changelog-generator
--- a/infra-tools/cmd/changelog-generator/main.go
+++ b/infra-tools/cmd/changelog-generator/main.go
@@ -1,0 +1,81 @@
+// Command changelog-generator posts a changelog comment on infra-deployments PRs
+// that bump the Konflux operator SHA.
+//
+// This is the initial placeholder (KFLUXVNGD-1022). It proves the workflow
+// trigger and comment pipeline work end-to-end on real PRs before any
+// changelog logic is added. SHA extraction, upstream comparison, and full
+// formatting are introduced in subsequent PRs.
+package main
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"log/slog"
+	"os"
+	"os/signal"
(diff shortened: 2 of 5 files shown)


.github/workflows/operator-changelog.yaml
```diff
@@ -0,0 +1,36 @@
...
+  changelog:
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
+      contents: read
+    steps:
+      # Check out the BASE branch (trusted code) so we never execute fork code.
+      # The binary is built here with the elevated GITHUB_TOKEN — a malicious
+      # fork cannot inject code into this step.
+      - uses: actions/checkout@v4
```
yftacherzog: Not sure why, but agents tend to think 4 is the latest, but v6 does exist.
```suggestion
      - uses: actions/checkout@v6
```

infra-tools/cmd/changelog-generator/main.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+	prNumber := 0
+	if _, err := fmt.Sscanf(prStr, "%d", &prNumber); err != nil || prNumber == 0 {
+		slog.Error("invalid PR number — printing to stdout", "pr", prStr)
+		fmt.Print(body)
+		return
+	}
+
+	client, err := ghclient.NewCommentClient(token, repo)
```
yftacherzog, line 64: We should inject the client into the function. This way we can replace it with a fake client and have unit tests for it with the fake client.

infra-tools/cmd/changelog-generator/main.go
```diff
@@ -0,0 +1,71 @@
...
+
+	body := commentMarker + "\n### Operator Changelog\n\n" +
+		"_This comment will show operator commits and upstream service changes in upcoming PRs._\n"
+
+	postOrPrint(ctx, *dryRun, token, repo, prStr, body)
+}
+
+// postOrPrint posts body as a PR comment or falls back to stdout when
+// dry-run is set or any required CI identifier is missing.
+func postOrPrint(ctx context.Context, dryRun bool, token, repo, prStr, body string) {
```
yftacherzog: I'd just name it `post` or `comment`.

infra-tools/internal/github/comments.go
```diff
@@ -44,14 +44,20 @@ func NewCommentClient(token, repoFullName string) (*CommentClient, error) {
...
 func (c *CommentClient) UpsertComment(ctx context.Context, prNumber int, body string) error {
-	// Find existing comment
-	existingID, err := c.findMarkedComment(ctx, prNumber)
+	return c.UpsertCommentByMarker(ctx, prNumber, body, CommentMarker)
+}
+
+// UpsertCommentByMarker creates or updates a PR comment identified by the
+// given marker string. This allows multiple tools to each maintain their own
+// idempotent comment on the same PR without interfering with each other.
+func (c *CommentClient) UpsertCommentByMarker(ctx context.Context, prNumber int, body, marker string) error {
```
yftacherzog, line 54: We should add unit tests here (e.g. empty marker, marker in body, marker not in body).

.github/workflows/operator-changelog.yaml
```diff
@@ -0,0 +1,36 @@
+name: Operator Changelog
+
+on:
+  pull_request_target:
+    types: [opened, synchronize, reopened]
+    paths:
+      - 'components/konflux-operator/development/invariant/kustomization.yaml'
+
```
yftacherzog, line 8: maybe we should add concurrency group here so that only one run can live at a time so they don't race for the comment.

yftacherzog (changes requested):
The direction is very good and will make the work much faster. Work towards making the code more testable (e.g. inject dependencies to functions to make them testable) and add unit tests.

yftacherzog (changes requested):
Almost ready as far as I see it. One comment I still have is that we should not fall back to stdout. If we fail to post or to create the client then we should fail the workflow.

infra-tools/cmd/changelog-generator/main.go
```diff
@@ -0,0 +1,78 @@
...
+	if err := post(ctx, client, prNumber, body); err != nil {
+		slog.Error("posting PR comment — printing to stdout as fallback", "err", err)
+		fmt.Print(body)
+	}
+}
+
+// post delivers body as a PR comment using the provided commenter.
+// Keeping the commenter as an injected interface allows tests to verify
+// the correct body and PR number are passed without making real API calls.
+func post(ctx context.Context, c commenter, prNumber int, body string) error {
```
yftacherzog, line 79: Are we going to extend this function in next PRs? If not, maybe we don't need it as a wrapper to UpsertCommentByMarker().

kelchen123: Plan to extend in the next prs

gbenhaim (approved):

PR #12056, opened 2026-05-29
[KONFLUX-14048] Allow hotfixes to target both staging and prod

p8r-the-gr8:
If the PR has the `skip-ring-deployment/hotfix` label applied, the "enforce-ring-deployments" Github Action will automatically pass.
This is to avoid blocking critical hotfixes that need to reach both staging and production environments as quickly as possible.

Example: https://github.com/p8r-the-gr8/infra-deployments/pull/5


filariow (approved):
/lgtm

PR #12152, opened 2026-06-02
Teach Qodo about ring deployment pattern

oswcab:
Qodo was flagging production ring PRs as bugs because only a subset of cluster directories were updated. Add extra_instructions to .pr_agent.toml so Qodo understands that production rollouts are intentionally split across multiple PRs (ring1, ring2, ring3).

Assisted-by: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/.pr_agent.toml
+++ b/.pr_agent.toml
@@ -7,3 +7,30 @@ push_commands = [
 
 # Removes internal Jira links
 require_ticket_analysis_review = false
+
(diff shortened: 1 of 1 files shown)


hugares (approved):
/lgtm

hugares:
failure not due to change for sure, merging

PR #12157, opened 2026-06-02
[KONFLUX-14240] env-detector: Calculate merge base

p8r-the-gr8:
The merge base by Github can become stale, leading to incorrect output.
The git module already contained a function (MergeBase) to calculate the effective merge base and was used by the 'render-diff' tool but not by 'env-detector'.

This change updates the behaviour of 'env-detector' so issues caused by stale ref can be avoided.


flacatus:
Cc @oswcab

enkeefe00 (approved):

PR #12177, opened 2026-06-02
feat(KFLUXVNGD-1023): extract operator SHA and post compare link in changelog comment

kelchen123:
Reads the operator image SHA from kustomization.yaml at both the base branch and PR head, then updates the changelog comment with a compare link (old → new SHA) or a "no change" message. Introduces internal/changelog/extractor.go and updates the workflow to use full git history and the PR merge ref.

### DIFF
--- a/infra-tools/internal/changelog/extractor.go
+++ b/infra-tools/internal/changelog/extractor.go
@@ -1,0 +1,68 @@
+// Package changelog provides logic for generating a human-readable changelog
+// when the Konflux operator ref is bumped in infra-deployments.
+package changelog
+
+import (
+	"fmt"
+	"os"
+	"strings"
+
+	"gopkg.in/yaml.v3"
+)
+
+// operatorResourceURL is the prefix of the GitHub URL used in the resources
+// block to pin the operator source. The ?ref= query parameter holds the git ref.
+const operatorResourceURL = "https://github.com/konflux-ci/konflux-ci/"
+
+// kustomization is a minimal representation of the fields we need from the
+// operator kustomization.yaml. Only the resources block is parsed.
+type kustomization struct {
+	Resources []string `yaml:"resources"`
+}
+
+// ExtractRef reads the kustomization.yaml at the given path and returns the
+// git ref from the ?ref= query parameter of the operator resource URL.
+//
+// Returns an error if the file cannot be read, the YAML cannot be parsed, or
+// no operator resource URL is found. The ref value is returned as-is — format
+// validation happens later at the point of use.
+func ExtractRef(path string) (string, error) {
+	data, err := os.ReadFile(path)
+	if err != nil {
+		return "", fmt.Errorf("reading %s: %w", path, err)
+	}
+
+	var k kustomization
+	if err := yaml.Unmarshal(data, &k); err != nil {
+		return "", fmt.Errorf("parsing %s: %w", path, err)
+	}
+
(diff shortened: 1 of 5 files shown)


infra-tools/internal/changelog/extractor.go
```diff
@@ -0,0 +1,75 @@
...
+		return "", fmt.Errorf("reading %s: %w", path, err)
+	}
+
+	var k kustomization
+	if err := yaml.Unmarshal(data, &k); err != nil {
+		return "", fmt.Errorf("parsing %s: %w", path, err)
+	}
+
+	for _, img := range k.Images {
+		if img.Name == operatorImageName {
```
yftacherzog: Considering we're iterating on something like this:
```
images:
  - name: localhost/konflux-operator
    newName: quay.io/konflux-ci/konflux-operator
    newTag: f8db856bdbc0326f3de2c8249e7a46453e1486c1
```

To make it less eyebrow-raising, can we instead compare `img.newName`, and then we compare to a constant that will get the value of  `quay.io/konflux-ci/konflux-operator`?

yftacherzog: On a second thought, considering this is the file:
```yaml
resources:
  # Upstream operator config (konflux-ci builds install.yaml from this layout).
  - https://github.com/konflux-ci/konflux-ci/operator/config/default?ref=f8db856bdbc0326f3de2c8249e7a46453e1486c1
  - konflux.yaml

patches:
  - path: release-config.yaml

images:
  - name: localhost/konflux-operator
    newName: quay.io/konflux-ci/konflux-operator
    newTag: f8db856bdbc0326f3de2c8249e7a46453e1486c1
```

 We currently assume newTag equals the `resources` ref, but we do not have to make that assumption. E.g. suppose that the pinning logic changes, we need to change this function.

Would it make better sense to make some extra work and parse the git ref under resources, which we know should be valid git ref (regardless of its format)?

If we go this way, I don't think it should be the responsibility of this function to assert the validity of the git ref (currently it's a commit sha, but could be a branch or tag, I suppose). Maybe we should just let it fail at the point where we try later to fetch it (with retries).

Also, if we go this way, the functions should be named ExtractRefs() or something along those lines, rather than ExtractSHAs()

yftacherzog (changes requested):
I think this PR is a good choice for next-step for implementation.

I left a comment about how I think we should extract the sha.

Other than that, I think we need a way to test the functions implemented in main.go. This is our main logic and nothing tests it.

Either it should be much thinner, or we need a way to test it.

yftacherzog (commented):
Looks better. Testing suggestion:

The split into `computeBody` + `ExtractRef` is in good shape and well covered. The remaining gap is git glue: `buildBody` and auto-detect in `resolveRepoRoot`.

Recommendation:

1.Keep path/format logic in unit tests via `computeBody` (as today).

2. Add one integration test with a temp git repo (git init, two commits with different `?ref=` in the invariant kustomization path, call `buildBody` with `--repo-root` and a base commit SHA).
That validates worktree + base/head without mocking git.

3. Possibly, inject a small interface around the git/worktree steps (or extract the “given repo root + base ref, return base and head kustomization file paths” logic) so tests can supply fixed file paths without running git.

yftacherzog (approved):

gbenhaim (approved):

PR #12194, opened 2026-06-03
chore: remove `edit-pipeline-from-build-templates` rolebinding and update test-runner clusterrole

tisutisu:
- Removing the `edit-pipeline-from-build-templates` rolebinding, since service accounts had `pod/exec` priviledges
- Adding the required privileges to run e2e-tests in build-definitions CI to `test-runner` cluster role


mshaposhnik (approved):

components/build-templates/base/e2e/rolebinding.yaml
```diff
@@ -1,22 +1,6 @@
...
-roleRef:
-  apiGroup: rbac.authorization.k8s.io
-  kind: ClusterRole
-  name: edit
-subjects:
-- kind: ServiceAccount
-  name: konflux-ci-sa
-  namespace: konflux-ci
-- kind: ServiceAccount
-  name: konflux-integration-runner
```
mmorhun, line 15: We don't own this one, please do not remove it without Integration team confirmation.
cc @dirgim @jsztuka

dirgim: This change seems to apply to just the `konflux-integration-runner` account on the `rhtap-build-tenant` namespace, so I'm fine with the change.

dirgim (approved):
/lgtm

mmorhun (approved):

jsztuka (approved):
lgtm

enkeefe00:
@tisutisu Can you move this out of draft so we can merge it?

tisutisu:
> @tisutisu Can you move this out of draft so we can merge it?

I think its better to test this change before merging, since it could break the build-definitions CI. My previous assessment was to merge and see if anything breaks, which is not right approach, now I am trying to test this change in development environment (which is not straight forward, taking more time), see if we can eliminate possibility of breaking build-definitions CI before merging. Once the testing completes, will open it for review.

tisutisu:
@enkeefe00 we have validated this change in development environment, its ready for review/merge, PTAL.

Regarding ring deployment warning, we only use `build-templates` component in production env, even if we split, there is no test that we run in stage, so stage soak time does not make sense for this component.

enkeefe00 (approved):

PR #12241, opened 2026-06-04
Upgrade kubearchive to v1.22.2 on development/staging

olegbet:
## Summary
- Upgrade kubearchive from v1.22.1 to v1.22.2 in the development cluster
- Update vacuum image to v1.22.2
- Add `BATCH_SIZE=10000` to the migration Job
- Improved performance of reads with the new DB normalized schema introduced in v1.21

## Test plan
- [ ] ArgoCD sync completes successfully for stone-prod-p02
- [ ] Schema migration job runs and completes
- [ ] KubeArchive API, operator, and sink pods are running
- [ ] Verify vacuum CronJob uses the correct image

## Risk assessment
- Low. Same upgrade that was reverted previously due to migration issues, now with `BATCH_SIZE` tuning from staging to protect deployments during migration. A rollback can be done by reverting this PR.


components/kubearchive/development/kubearchive.yaml
```diff
@@ -1232,7 +1232,7 @@ spec:
               valueFrom:
                 resourceFieldRef:
                   resource: limits.cpu
-          image: quay.io/kubearchive/operator:v1.22.1@sha256:8e50b5bb60058722910a7b073b80c92635683eb50a86191dc5994c4ecea0cbbf
+          image: quay.io/kubearchive/operator@sha256:5f35ebd89eb8c7637d2b45d9587fa8920192258c14e78b8c23574375fdbb1303
```
mafh314: why the reference for the new image doesn't include now the `:v1.22.2` ?

components/kubearchive/development/kustomization.yaml
mafh314: In the PR summary I see:
`Add BATCH_SIZE=10000 to the migration Job`
But here the value is `1000`

components/kubearchive/development/kubearchive.yaml
```diff
@@ -1422,7 +1422,7 @@ metadata:
     app.kubernetes.io/component: kubearchive
     app.kubernetes.io/name: kubearchive-schema-migration
     app.kubernetes.io/part-of: kubearchive
-    app.kubernetes.io/version: v1.22.1
+    app.kubernetes.io/version: v1.22.2
```
maruiz93, line 1425: We need to make an extra change, based on this https://gitlab.cee.redhat.com/konflux/docs/sop/-/blob/main/kubearchive/architecture.md?ref_type=heads#schema-migration

The thing is that the Job changes due to the new kubearchive version, both labels and image and the Jobs in Argo can't be updated (just replaced) so we need to modify in the kustomization the resource to have a suffix that includes both the schema version and the kubearchive version.

For this PR it would be:

`kubearchive-schema-migration-v13-v1.22.2`

maruiz93 (changes requested):

components/kubearchive/development/kubearchive.yaml
olegbet: I believe this is handled in the patch

mafh314 (approved):

maruiz93:
now that we are adding the kubearchive version as suffix in the job, can you please update the upgrade.sh script for taking this under consideration?

PR #12262, opened 2026-06-05
ci(pipelines): Skip production all-deploys-in-sync check for Ring PRs

aThorp96:
Any Pull Request which has a head commit mentioning something like "Ring 1" is expected not to update all production files, making this check a false positive.


mathur07 (approved):
/lgtm

aThorp96:
Closing this as a proper approach would be to add per-ring enforcement

PR #12305, opened 2026-06-10
feat(KFLUXVNGD-1079): detect and list upstream service bumps in changelog comment

kelchen123:
- Calls the GitHub compare API on `konflux-ci/konflux-ci` between the old and new operator refs
- Scans changed files under `operator/upstream-kustomizations/` for `?ref=` SHA changes
- Lists each bumped service with old→new SHA and a compare link in the PR comment
In case of API failure the operator compare link is still posted but the service bumps section notes detection was unavailable.

### DIFF
--- a/infra-tools/cmd/changelog-generator/main.go
+++ b/infra-tools/cmd/changelog-generator/main.go
@@ -1,9 +1,10 @@
 // Command changelog-generator posts a changelog comment on infra-deployments PRs
-// that bump the Konflux operator SHA.
+// that bump the Konflux operator ref.
 //
-// This is step 2 (KFLUXVNGD-1023): it reads the operator SHA at the base branch
-// and PR head, then posts a comment with the compare link. Upstream service
-// changes are introduced in subsequent PRs.
+// This is step 3 (KFLUXVNGD-1024): it reads the operator ref at the base branch
+// and PR head, posts a compare link, and lists which upstream sub-services had
+// their pinned SHA bumped inside the operator repo. Commit-level details per
+// service are introduced in the next PR.
 package main
 
 import (
@@ -14,6 +15,7 @@ import (
 	"os"
 	"os/signal"
 	"path/filepath"
+	"strings"
 	"syscall"
 
 	"github.com/redhat-appstudio/infra-deployments/infra-tools/internal/changelog"
@@ -22,7 +24,7 @@ import (
 )
 
 // kustomizationPath is the path, relative to the repo root, of the
-// kustomization.yaml file that pins the operator SHA.
+// kustomization.yaml file that pins the operator ref.
 const kustomizationPath = "components/konflux-operator/development/invariant/kustomization.yaml"
 
 // commentMarker identifies the changelog comment for idempotent updates.
@@ -46,7 +48,9 @@ func main() {
 	repo := os.Getenv("GITHUB_REPOSITORY")
 	prStr := os.Getenv("PR_NUMBER")
 
-	body, err := buildBody(ctx, *repoRoot, *baseRef)
+	comparer := changelog.NewRepoComparer(token)
+
+	body, err := buildBody(ctx, *repoRoot, *baseRef, comparer)
 	if err != nil {
 		slog.Error("building changelog body", "err", err)
 		os.Exit(1)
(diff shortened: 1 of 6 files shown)


yftacherzog (commented):
I didn't look for where would be the best place(s) to have it, but we need to start to look into API call retries.

The process we're working on has multiple API calls (and will have many more). We should surround them with retries, so we don't fail for transient stuff.

infra-tools/internal/changelog/bumps.go
```diff
@@ -0,0 +1,100 @@
...
+		switch line[0] {
+		case '-':
+			if oldSHA == "" {
+				oldSHA = m[1]
+			}
+		case '+':
+			if newSHA == "" {
+				newSHA = m[1]
+			}
+		}
```
yftacherzog: If I understand correctly, there's an underlying assumption here that if we find a plus and a minus for a ref, it must be the same ref. 

If so, I'm not sure we want to make that assumption. I think it would be best if we verify that the plus and minus are for the same ref (e.g. match them when iterating, or searching for a plus for a particular ref after finding its minus, etc.).

infra-tools/internal/changelog/upstream.go
```diff
@@ -0,0 +1,72 @@
...
+	if token != "" {
+		client = client.WithAuthToken(token)
+	}
+	return client.Repositories
+}
+
+// FetchOperatorCompare calls the GitHub compare API for konflux-ci/konflux-ci
+// between oldRef and newRef and returns the changed files.
+func FetchOperatorCompare(ctx context.Context, c RepoComparer, oldRef, newRef string) (*OperatorCompare, error) {
+	comparison, _, err := c.CompareCommits(ctx, "konflux-ci", "konflux-ci", oldRef, newRef, nil)
```
yftacherzog: Single `CompareCommits` call with `nil` opts; `*gh.Response` is discarded. GitHub may return a truncated compare for large diffs — consider inspecting `resp` (and paging if supported) before treating the result as complete.

— Yftach's Cursor

infra-tools/internal/changelog/upstream.go
```diff
@@ -0,0 +1,72 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("comparing %s...%s in konflux-ci/konflux-ci: %w",
+			refShort(oldRef), refShort(newRef), err)
+	}
+	return &OperatorCompare{Files: convertFiles(comparison.Files)}, nil
```
yftacherzog: Only `comparison.Files` from that one response is used. If the compare is paginated or truncated, upstream kustomization changes can be missed silently.

— Yftach's Cursor

infra-tools/cmd/changelog-generator/main.go
```diff
@@ -104,56 +109,93 @@ func buildBody(ctx context.Context, repoRoot, baseRef string) (string, error) {
...
-func selectBody(oldRef, newRef string) string {
-	if oldRef == newRef {
-		return formatNoChange()
+	compare, err := changelog.FetchOperatorCompare(ctx, comparer, oldRef, newRef)
+	if err != nil {
+		slog.Warn("fetching operator compare; service bumps unavailable", "err", err)
+		return formatCompare(oldRef, newRef, nil), nil
+	}
+
+	bumps := changelog.ExtractServiceBumps(compare.Files)
```
yftacherzog: Assumes `compare.Files` is the full file list. Worth degrading (like API failure) when compare results are incomplete — otherwise we may report "no upstream service refs changed" incorrectly.

— Yftach's Cursor

yftacherzog:
@kelchen123 not for this PR but for awareness: not sure if you're aware or not, but there are upstream components ([can think of UI](https://github.com/konflux-ci/konflux-ci/blob/main/operator/upstream-kustomizations/ui/kustomization.yaml). Don't recall if there are others) which don't have upstream kustomizations -- just image references.

The upstream release notes process have a heuristic for tracking that, and we can consider using something similar. This is something to look into in next PRs. For now, just see that the code is written in a way that can be extended to include something like that.

Other than that, I think Qodo's feedback has good points still.

kelchen123:
applied the actionable items from qodo

yftacherzog (approved):
/lgtm

I see [this one](https://github.com/redhat-appstudio/infra-deployments/pull/12305#issuecomment-4701086456) as the next thing we need to support.

Edit: on the other hand, maybe worth implementing the next visible bit (so we see the change in the PR comments) and do this one later. I leave it up to you.

gbenhaim (approved):

PR #12313, opened 2026-06-11
feat: split IS into per-environment overlays

seanconroy2021:
IS had one kustomization.yaml for all environments so the CRD ref was always updated everywhere at once. This PR split the deployment into 3 env.

Assisted-by: Cursor

### DIFF
--- a/components/internal-services/base/internal_service_service_account_token.yaml
+++ b/components/internal-services/base/internal_service_service_account_token.yaml
@@ -1,3 +1,4 @@
+---
 apiVersion: v1
 kind: Secret
 metadata:
--- a/components/internal-services/base/kustomization.yaml
+++ b/components/internal-services/base/kustomization.yaml
@@ -1,12 +1,11 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
 resources:
 - internal_service_request_role.yaml
 - internal_service_request_role_binding.yaml
 - internal_service_request_service_account.yaml
 - internal_service_service_account_token.yaml
 - internal-services.yaml
-- https://github.com/konflux-ci/internal-services/config/crd?ref=bbba2e6340a3c014adf32c770db98e9237543da6
-
-apiVersion: kustomize.config.k8s.io/v1beta1
-kind: Kustomization
 
 namespace: internal-services
--- a/components/internal-services/development/kustomization.yaml
+++ b/components/internal-services/development/kustomization.yaml
@@ -1,0 +1,6 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - ../base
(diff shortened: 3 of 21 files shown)


components/internal-services/base/kustomization.yaml
```diff
@@ -1,12 +1,11 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
 resources:
 - internal_service_request_role.yaml
 - internal_service_request_role_binding.yaml
 - internal_service_request_service_account.yaml
 - internal_service_service_account_token.yaml
 - internal-services.yaml
-- https://github.com/konflux-ci/internal-services/config/crd?ref=bbba2e6340a3c014adf32c770db98e9237543da6
```
mmalina, line 7: The RPA will need to be modified accordingly, right? To modify the one in dev + staging?

seanconroy2021: Yeah this will need to be updated to make it work https://gitlab.cee.redhat.com/releng/konflux-release-data/-/blob/main/config/stone-prd-rh01.pg1f.p1/service/ReleasePlanAdmission/rhtap-release-2/internal-services.yaml?ref_type=heads#L19

mmalina (approved):

seanconroy2021:
@gbenhaim can you take a look plz :)

seanconroy2021:
Has to rebase due to conflicts :)

components/internal-services/base/internal_service_service_account_token.yaml
```diff
@@ -1,3 +1,4 @@
+---
```
filariow, line 1: Be aware that we are deprecating long-lived service accounts in favor of short lived ones. Please plan migrating to `serviceaccounts/token`.

components/internal-services/development/kustomization.yaml
```diff
@@ -0,0 +1,6 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - ../base
```
filariow, line 5: the `../base` folder is shared among all environments. Please make sure that production has its own `base` folder that's not shared with other environments.

filariow (changes requested):

components/internal-services/development/kustomization.yaml
seanconroy2021: Done production now has its own production/base with its own copy of the resources.

components/internal-services/base/internal_service_service_account_token.yaml
seanconroy2021: Thanks for the heads up. Can you link me a ticket for this so I can create a followup for our team to track it?

gbenhaim (approved):
/approve
/lgtm

filariow (approved):

PR #12337, opened 2026-06-11
Remove ring deployment check

hugares:
We all agree we want to deploy in rings and we want to enforce that, this why we have a feature being worked on. In the meantime, the check we added is creating a lot of confusion in different teams and even delayed OSP upgrade. Infra team, mainly me is doing the damage control with all the teams caught by surprise about this and are asking when this was announce and I need to tell them that the concept is being worked on. This make no sense. It gives the impression that infra team is improvising and I do not want that.

Up until now, we let each team decide their roll out strategy, this will change once ring feature is completed, i.e. all the team are migrated to the new way of doing things so until then, we can advocate for the ring by suggesting it in review but we cannot enforce it.


flacatus:
LGTM

flacatus (approved):

sadlerap (approved):

sadlerap (approved):

PR #12349, opened 2026-06-11
fix(KAR-893) remove gateway from loadbalancer on stone-prod-p02

olegbet:
## Summary
- Removes loki-gateway endpoint from external load-balancer

## Risk Assessment

**Risk Level:** Low
**Rollback:** Revert this PR.
**Description:** These values have been validated in staging. No impact on current cluster network is expected.

## Validation
- Staging validated with the same configuration (PR #12340)

Jira: https://redhat.atlassian.net/browse/KAR-893

### DIFF
--- a/components/vector-kubearchive-log-collector/production/stone-prod-p02/loki-helm-prod-values.yaml
+++ b/components/vector-kubearchive-log-collector/production/stone-prod-p02/loki-helm-prod-values.yaml
@@ -5,7 +5,7 @@ global:
 
 gateway:
   service:
-    type: LoadBalancer
+    type: ClusterIP
   resources:
     requests:
       cpu: 100m


maruiz93 (commented):
Is this supposed to be the change for prod? The title and the cluster where the actual change is done don't match, but I'm wondering why this change isn't done for all prod clusters at the same time (once staging is approved and merged as a way of testing it)

olegbet:
> Is this supposed to be the change for prod? The title and the cluster where the actual change is done don't match, but I'm wondering why this change isn't done for all prod clusters at the same time (once staging is approved and merged as a way of testing it)

Fixed the title, thanks! Current policy is to update cluster one by one as far as I know. This is what I was asked to do when updating otel-collector version.

maruiz93:
> > Is this supposed to be the change for prod? The title and the cluster where the actual change is done don't match, but I'm wondering why this change isn't done for all prod clusters at the same time (once staging is approved and merged as a way of testing it)
> 
> Fixed the title, thanks! Current policy is to update cluster one by one as far as I know. This is what I was asked to do when updating otel-collector version.

oh! I wasn't aware of that, I was doing it on kubearchive but wasn't aware of this being a policy, thanks for the clarification

/lgtm

eedri:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @eedri |
| **Timestamp** | 2026-06-14T14:46:23.505Z |
| **Risk Level** | low |
| **Criteria Override** | AI review findings not addressed |

Qodo repsonse is FP, approving

PR #12332, opened 2026-06-11
Dev+staging Patch pipelines resolver container image for KONFLUX-14376

aThorp96:
Override the Resolvers image to include this change: https://github.com/tektoncd/pipeline/commit/829429b251abc0dd2dd055b893300f505780b70b

<details>
<summary>Full list of commits between images</summary>

```
commit 829429b251abc0dd2dd055b893300f505780b70b
Author: Andrew Thorp <athorp@redhat.com>
Date:   Wed Jun 10 11:52:24 2026 -0400

    fix(resolvers): Allow ResolutionRequests to resolve all Tekton kinds
    
    Tekton controllers only create ResolutionRequests for Pipelines, Tasks,
    and StepActions, however other cluster services only use Tekton
    Resolvers to fetch things like PipelineRun definitions. This is arguably
    a reasonable use-case and previously-supported (though undocumented)
    behaviour.
    
    The security model of the original commit was to ensure irrelevant
    objects or sensitive data (such as source code, Secrets on disk, or
    binaries) would not be written into the ResolutionRequest's status.data
    field. This change preserves that security model while allowing the
    resolvers to remain somewhat flexible for synergistic use-cases.
    
    See also: 0c1a6dea9bf5ba6533fa6ddc45f9e5d1cefc1d1f
    
    (cherry picked from commit 7a6c28c678839f22d273b84d829a7eab9c848b59)

commit 2046a4c4c6cff75a9940c97055328deaa0fb08f7
Author: dependabot[bot] <49699333+dependabot[bot]@users.noreply.github.com>
Date:   Tue Jun 9 20:04:13 2026 +0000

    build(deps): bump the all group in /tekton with 4 updates
    
    Bumps the all group in /tekton with 4 updates: [chainguard/crane](https://github.com/chainguard-images/images), [chainguard/go](https://github.com/chainguard-images/images), [tektoncd/plumbing/ko](https://github.com/tektoncd/plumbing) and [tektoncd/plumbing/koparse](https://github.com/tektoncd/plumbing).
    
    
    Updates `chainguard/crane` from `7469bb7` to `0cb513f`
    - [Commits](https://github.com/chainguard-images/images/commits)
    
    Updates `chainguard/go` from `df345ed` to `3cb640a`
    - [Commits](https://github.com/chainguard-images/images/commits)
    
    Updates `tektoncd/plumbing/ko` from `cee71b8` to `0e959bc`
    - [Commits](https://github.com/tektoncd/plumbing/commits)
    
[...]

PR #12350, opened 2026-06-11
[test] Prod Ring 1: Patch pipeline resolver container image for KONFLUX-14376

mathur07:
Add IMAGE_PIPELINES_CONTROLLER override to patch the pipelines resolver on Ring 1 production clusters (stone-prod-p01, kflux-osp-p01, kflux-fedora-01). Same pattern as the staging fix in #12332.

Resolvers image to include this change: https://github.com/tektoncd/pipeline/commit/829429b251abc0dd2dd055b893300f505780b70b


mathur07:
not sure if this is how its expected.

PR #12351, opened 2026-06-11
(KAR-885): fix minIO misconfiguration

olegbet:


maruiz93:
@olegbet I don't understand why is this necessary and how it relates with removing the load balancer? Also, the development environment is ephermeral so we shouldn't be using actual s3 buckets, but the feeling I have is that you are configuring this to do it otherwise

olegbet:
> @olegbet I don't understand why is this necessary and how it relates with removing the load balancer? Also, the development environment is ephermeral so we shouldn't be using actual s3 buckets, but the feeling I have is that you are configuring this to do it otherwise

This is not related to removing the load balancer. The problem is that, first of all, we mirrored all the images for loki stack to quay.io except minIO image and sometimes it fails to fetch it which causes timouts syncing e2e test instances(they use our development configuration).

maruiz93:
> > @olegbet I don't understand why is this necessary and how it relates with removing the load balancer? Also, the development environment is ephermeral so we shouldn't be using actual s3 buckets, but the feeling I have is that you are configuring this to do it otherwise
> 
> This is not related to removing the load balancer. The problem is that, first of all, we mirrored all the images for loki stack to quay.io except minIO image and sometimes it fails to fetch it which causes timouts syncing e2e test instances(they use our development configuration).

I see! I probably misschecked the jira ticket, that's why I was asking! Thanks for providing the extra context!
/lgtm

PR #12279, opened 2026-06-08
Standardize tekton-events-controller memory to 2048Mi across all production clusters

mathur07:
The events-controller was temporarily bumped to 4096Mi during the v5.0.5-830
rollout. Now that all clusters are healthy,
reduce to 2048Mi as a safe baseline that avoids OOMKills without over-provisioning.

PR #12384, opened 2026-06-14
convert pipelines production tekton-events-controller units from Mi to Gi

ab-ghosh:
4096Mi was not syncing when the value was 4Gi, however it was still being considered "out of sync" and causing an alert to fire
image


infernus01 (commented):
@ab-ghosh Could you please also check #12279 as I believe this PR overlaps with it, which modifies the same memory fields. That PR reduces the memory from 4096Mi → 2048Mi and the 4096Mi was a temporary bump during the v5.0.5-830 rollout (introduced in #12266).

PR #12437, opened 2026-06-16
PVO11Y-5265 Build users.json from raw vault token at runtime

FaisalAl-Rayes:
## Summary
- The perf&scale team stores a raw service account token in Vault (key: `token`) rather than a pre-built `users.json`
- Align the kanary tekton task with the [Jenkins approach](https://gitlab.cee.redhat.com/redhat-performance/ci-configs/-/blob/master/vars/runKonfluxProbeTest.groovy#L125-134): read the raw token from the mounted secret and build `users.json` at runtime using `jq`
- Add `tenant-namespace` param to pipeline and task to specify which namespace the test runs against

Jira: https://redhat.atlassian.net/browse/PVO11Y-5265

## Risk Assessment
**Risk Level:** Low
**Description:** Changes how the kanary task reads user credentials from Vault — aligns with existing Jenkins pattern used by perf&scale
**Rollback:** Revert PR

## Validation
- [x] `kustomize build components/monitoring/kanary/staging/stone-stage-p01/` passes
- [x] `kustomize build components/monitoring/kanary/staging/stone-stg-rh01/` passes

### DIFF
--- a/components/monitoring/kanary/staging/base/pipelines/kanary-otel-pipeline.yaml
+++ b/components/monitoring/kanary/staging/base/pipelines/kanary-otel-pipeline.yaml
@@ -40,6 +40,9 @@ spec:
       default: "https://gitlab.cee.redhat.com/api/v4"
     - name: member-cluster
       type: string
+    - name: tenant-namespace
+      type: string
+      default: "konflux-perfscale-4-tenant"
     # RPM test type only. Leave empty for container test types.
     - name: component-repo-revision
       type: string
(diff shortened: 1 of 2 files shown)


FaisalAl-Rayes:
made 2 reviews by accident on the same code. I will be responding to one and the other if it has something new.

FaisalAl-Rayes:
Review response:
1. Yes jq is part of the load test image.
2. the path of where the token is can be anywhere but when running the jq command to construct the users.json it will be in the working directory where .run-stage.sh is and expects users.json to be
3. there is just one namespace dedicated for kanary at the moment so this is not a concern

jhutar (approved):
/lgtm

PR #12318, opened 2026-06-11
refactor(KONFLUX-13283): use task-runner for claim cleaner in prod

yftacherzog:
Swap appstudio-utils for task-runner:v1 (digest-pinned) via kustomize images in production/ kustomization.yaml.

This is following the same pattern as in dev and stage.

Next steps (not to break enforce-ring-deployment):

PR 3: set base/cronjob.yaml to task-runner with digest pinned (kustomizations will be noops).
PR 4: set dev and stage kustomizations to use
name: quay.io/konflux-ci/task-runner.
PR 5: apply the same to production + remove digest from base.

Assisted-by: Cursor

The change was verified in staging after [this PR](https://github.com/redhat-appstudio/infra-deployments/pull/12312) was merged.

I verified that the cronjob is running with the new image and providing the expected results for when claims are to be deleted and when they are not. I create [this PR](https://github.com/yftacherzog/testrepo-staging/pull/54) which triggered the creation of an orphan claim (so it will not be garbage collected) and watched the cronjob being triggered every hour and the claim deleted after the 4th hour.

Risk assessment: **low**. No logic changes. The image is pulled and behaving correctly in staging. the cronjob is a backup mechanism for deleting claims, so suppose that something is not working, we'll have some time to evaluate and fix it before this becomes an issue (no apparent reason it should behave differently than in stage).

PR #12462, opened 2026-06-16
chore(KONFLUX-13283): update claim cleaner image

yftacherzog:
The image was updated from appstudio-utils to task-runner with kustomize image overrides in all envs. This change now aligns the image in the base manifest to point to the task-runner as well.

This temporarily makes the previous overrides no-ops.

This is the third PR in this effort. Next steps:

PR 4: set dev and stage kustomizations to use
name: quay.io/konflux-ci/task-runner.
PR 5: apply the same to production + remove digest from base.

Assisted-by: Cursor

Risk assessment: No manifest changes are expected in this change. I verified that `kustomize build` generates the same outputs for all envs before and after the change.

## Risk Assessment

**Risk Level:** Low

**Description:** Previous PRs have applied per-overlay (identical) overrides for the image being used for the claim cleanup cronjob. This was done in order to replace a deprecated image with a new one. This PR is changing the image in the base manifest from the deprecated one to the new one. This will make the override into a no-op, both because it will make the override redundant and because the image in base manifest will no longer be matched by the override. Later PRs will restore the overrides functionality, but instead of overriding `old-image -> new-image+digest`, it will be `new-image -> new-image+digest` which is the common pattern we use for overrides.

I ran `kustomize build` for all overlay types before and after the change and the result was that before and after are identical (as expected), but there's a slight chance that I (and the approvers) did not take into account some reasoning for which this will stop working.

Having said that, this change is in the path of making future changes more straightforward and clear. Moreover, when testing the new image on stage and prod, I noticed that it's serving as a backup mechanism that is not actually utilized most of the time, as claims are usually garbage-collected when their owning pipelineruns are removed. I had to specifically create a claim without owner reference for it to endure until it had to be deleted by the job.

So, if the cronjob is broken after the change, we will most likely be able to tell that it's not able to run before it is actually required, and even when required, it will take time for things to start failing due to claims not being deleted.

**Rollback:** click revert on this PR and merge the reverting PR

## Validation

Staging PR: https://github.com/redhat-appstudio/infra-deployments/pull/12312
Production PR: https://github.com/redhat-appstudio/infra-deployments/pull/12318

## Environment Separation

The actual changes in manifests reaching ArgoCD were made in the PRs above. The current PR should be a no-op in terms of manifests pushed to the clusters. An alternative that will allow us not touching the common base will require duplicating the base which will be a riskier change (while I expect should be possible to split between the overlays).

## Qodo Response

**Kustomize override no-op** — This is an expected risk and this is what this PR is about. Next PRs will change the image overrides, one overlay at a time -- again, with no change in the manifests pushed to the clusters. This will make the overrides functional again.

### DIFF
--- a/components/crossplane-control-plane/base/cronjob.yaml
+++ b/components/crossplane-control-plane/base/cronjob.yaml
@@ -11,7 +11,7 @@ spec:
         spec:
           containers:
             - name: namespace-claim-cleaner
-              image: quay.io/konflux-ci/appstudio-utils:e9578e46aefbe58bf77de1f154fbb846fcbdacec@sha256:2094a997c17bc168264209deb43504adcda4b97b64602041e4886f086856301d
+              image: quay.io/konflux-ci/task-runner:v1@sha256:b9ef0479b57a494368a12a02886c650314c9ac7391c6228f21065865aab71c9e
               command:
                 - /bin/bash
                 - /scripts/namespace-claim-cleaner.sh


amisstea (approved):

svatares:
@yftacherzog the following items are needed before this PR can be approved for production:



### - [ ] Risk Assessment

Add a `## Risk Assessment` section to the PR description:

```markdown
## Risk Assessment

**Risk Level:** <Low / Medium / High / Very High>
**Description:** <Brief description of what could go wrong>
**Rollback:** <How to revert if needed>
```

Replace the `<…>` placeholders with your actual values.

### - [ ] Staging Validation

Add a comment with the staging PR URL in this format:

```markdown
## Validation

Staging PR: <your-staging-pr-url>
```

The dashboard needs the full GitHub PR URL to check if it has been merged and calculate soak time.
Replace the `<…>` placeholder with the actual staging PR link.

### - [ ] Environment Separation

This PR modifies files that affect multiple environments simultaneously. Changes must be merged in order:

1. **development** — open a PR with only the development environment file changes
2. **staging** — once merged, open a PR for the staging environment changes and let it soak for at least 1 day
3. **production** — only then update this PR to contain only the production environment changes

The `environment/*` labels are generated automatically from the files changed.

### - [ ] Qodo Findings Response

The Qodo AI review flagged potential bugs or security concerns. Acknowledge each finding by adding a comment:

```markdown
## Qodo Response

**<Finding title>** — <Your explanation (false positive / fixed in commit / accepted risk)>
```

Replace the `<…>` placeholders with the actual finding titles from the Qodo review and your responses.



---

> Posted by the Konflux approval dashboard.

yftacherzog:
@svatares I updated the PR description.

svatares:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @svatares |
| **Timestamp** | 2026-06-16T17:28:03.015Z |
| **Risk Level** | low |
| **Criteria Override** | PR updates multiple environments, AI review findings not addressed |

Approved

PR #12519, opened 2026-06-17
Add ibm-public-key secret access to release pipeline ClusterRole

rhartman93:
## Summary
- Adds `get` access for the `ibm-public-key` secret to the `release-pipeline-resource-role` ClusterRole
- Release pipelines need this secret to pull images signed with the IBM key for releasing to downstream registries
- Access is tightly scoped via `resourceNames` to only the `ibm-public-key` secret with `get`-only permission

## Context
This access was previously defined as per-cluster `Role` resources in [konflux-release-data](https://gitlab.cee.redhat.com/releng/konflux-release-data) under `tenants-config/cluster/*/managed/rhtap-releng-tenant/rb/rb-release-developer-portal-secret-access.yaml`, but Role definitions are not allowed in managed namespace configs there. Moving the permission to this ClusterRole is the correct approach since it's already bound to all release pipeline ServiceAccounts in managed namespaces.

## Test plan
- [ ] Verify the ClusterRole change deploys successfully
- [ ] Verify release pipelines can access the `ibm-public-key` secret in managed namespaces
- [ ] Remove the per-cluster Role/RoleBinding files from konflux-release-data after this merges


rhartman93:
Worth noting if I understand the implications of this change correctly, all release SAs will have access to the secret, pretty sure the secret is just a public key so it shouldn't be that big of a deal but want to call that out in case I'm wrong

seanconroy2021:
<del>Hey, I dont really understand why this secret is being defined for the base binding provided to SA during a release. Normally a secret is given to an SA in releng and that said SA is passed into the release to have access to such a secret. Maybe I am missing something here.</del>

I have found the original thread; So I understand why it can't be in an SA like it's normally done, but I still have a question about how the normal public key works without being defined in the release base, whereas this one needs to be. Have added it here:  https://redhat-internal.slack.com/archives/C04PZ7H0VA8/p1781869789335419?thread_ts=1780919280.075869&cid=C04PZ7H0VA8


**Resolved in slack thread**

konflux-approval-tool:
## Risk Assessment
> *AI calculated Risk* — based on code changes and PR description.
> Override this by adding your own `## Risk Assessment` section to the PR description.

**Risk Level:** Low
**Description:** This change adds tightly scoped read-only access to a specific `ibm-public-key` secret for release pipelines, migrating an existing permission from per-cluster Roles to a central ClusterRole. The risk is low due to its limited scope and read-only nature.
**Rollback:** Revert this PR to remove the added ClusterRole permission.

lkoscova:
@rhartman93 this PR modifies files that affect multiple environments simultaneously, which blocks production approval.

Changes must be merged in order:

1. **development** — open a PR with only the development environment file changes
2. **staging** — once merged, open a PR for the staging environment changes and let it soak for at least 1 day before proceeding
3. **production** — only then update this PR to contain only the production environment changes

The `environment/*` labels are generated automatically from the files changed — splitting the files across separate PRs will result in each PR having only its target environment label.

sadlerap (approved):

pbabinca:
@filariow suggested changes outside of the PR:

> yeah, there's a change to all environments indeed. You need to remove the [clusterrole patch](https://github.com/redhat-appstudio/infra-deployments/blob/main/components/release/staging/ibm-public-key-patch.yaml) from development and staging overlay
> otherwise your PR will create the following change to the ClusteRroles in the other envs

(and there he showed how subset of `kustomize build components/release/staging` would look like).

I'm not collaborator of this PR nor maintainer of the repo so I can't alter this PR.

@rhartman93 I have prepared a patch that you might want to review and apply:

[0001-Remove-ibm-public-key-overlay-patches-superseded-by-.patch](https://github.com/user-attachments/files/29796690/0001-Remove-ibm-public-key-overlay-patches-superseded-by-.patch)

lkoscova:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @lkoscova |
| **Timestamp** | 2026-07-23T09:39:30.643Z |
| **Risk Level** | low |

tested

PR #11854, opened 2026-05-18
fix: token resource sanitized name

filariow:
Tekton-Kueue sanitizes and normalizes the resource names as it needs to use them in an annotation's key.
The normalized name needs to be used in the ClusterQueue.

Fixes: https://github.com/redhat-appstudio/infra-deployments/pull/10850

Signed-off-by: Francesco Ilario <filario@redhat.com>

rh-pre-commit.version: 2.3.2
rh-pre-commit.check-secrets: ENABLED

PR #12529, opened 2026-06-17
KFLUXINFRA-3973: add token resource to kueue in production (ring 1)

oswcab:
## What

Add the `konflux-ci.dev/token` ephemeral resource to ring 1 production clusters: stone-prod-p01, kflux-ocp-p01, kflux-prd-rh02.

Token quotas: stone-prod-p01=300, kflux-ocp-p01=250, kflux-prd-rh02=300.

[KFLUXINFRA-3973](https://redhat.atlassian.net/browse/KFLUXINFRA-3973)

## Why

Introduce a token resource as a secondary constraint alongside `tekton.dev/pipelineruns`  to enable future per-cluster tuning of PipelineRun admission. This commit deploys safe  no-op defaults (1 token per PLR, quota = PLR limit) — current behavior is unchanged.

## Validation

  - Already deployed to dev and staging since PR #10850 (May 15) and fix #11854
  - `kustomize build` passes for all affected overlays
  - All tekton-kueue CEL expression tests pass (including new ClusterQueueTokenResourceTest)

## Risk Assessment

  **Risk Level:** Low
  **Description:** Adds a new ephemeral resource with quota equal to the PLR limit. Each PLR requests 1 token, so the token resource never constrains before the PLR limit — no behavioral change.
  **Rollback:** Revert PR and redeploy previous configuration.

Also adds ClusterQueueTokenResourceTest to validate cluster-queue manifests have the correct token resource and quota.

Contributes to: [KFLUXINFRA-3973](https://redhat.atlassian.net/browse/KFLUXINFRA-3973)

[KFLUXINFRA-3973]: https://redhat.atlassian.net/browse/KFLUXINFRA-3973?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[KFLUXINFRA-3973]: https://redhat.atlassian.net/browse/KFLUXINFRA-3973?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/hack/test-tekton-kueue-config.py
+++ b/hack/test-tekton-kueue-config.py
@@ -1118,6 +1118,16 @@ CONFIG_COMBINATIONS: Dict[str, ConfigCombination] = {
         "name": "Production p02 config",
         "config_file": "components/kueue/production/stone-prod-p02/config.yaml",
         "kustomization_file": "components/kueue/production/base/tekton-kueue/kustomization.yaml"
+    },
+    "production-stone-prod-p01": {
+        "name": "Production p01 config",
+        "config_file": "components/kueue/production/stone-prod-p01/config.yaml",
+        "kustomization_file": "components/kueue/production/base/tekton-kueue/kustomization.yaml"
+    },
+    "production-kflux-prd-rh02": {
+        "name": "Production rh02 config",
+        "config_file": "components/kueue/production/kflux-prd-rh02/config.yaml",
+        "kustomization_file": "components/kueue/production/base/tekton-kueue/kustomization.yaml"
     }
 }
 
@@ -1408,6 +1418,7 @@ TEST_COMBINATIONS: Dict[str, TestCombination] = {
         "config_key": "production-kflux-ocp-p01",
         "expected": {
             "annotations": {
+                "kueue.konflux-ci.dev/requests-konflux-ci-dev-token": "1",
                 "kueue.konflux-ci.dev/requests-linux-amd64": "1",
                 "kueue.konflux-ci.dev/requests-linux-s390x": "1",
                 "kueue.konflux-ci.dev/requests-linux-arm64": "1",
@@ -1425,6 +1436,7 @@ TEST_COMBINATIONS: Dict[str, TestCombination] = {
         "config_key": "production-kflux-ocp-p01",
         "expected": {
             "annotations": {
+                "kueue.konflux-ci.dev/requests-konflux-ci-dev-token": "1",
                 "kueue.konflux-ci.dev/requests-konflux-release": "1",
             },
             "labels": {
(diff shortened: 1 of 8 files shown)


components/kueue/production/base/tekton-kueue/config.yaml
```diff
@@ -92,6 +92,12 @@ cel:
     - |
         plrNamespace == 'mintmaker' ? [resource('mintmaker', 1)] : []
 
+    # The token mechanism allows us to balance admitted PipelineRuns.
```
filariow: > Add the konflux-ci.dev/token ephemeral resource to ring 1 production clusters: stone-prod-p01, kflux-ocp-p01, kflux-prd-rh02.

From `render-diff`'s [output](https://github.com/redhat-appstudio/infra-deployments/pull/12529#issuecomment-4735368687), I see unexpected clusters are impacted by this change. For them, the change is incomplete as the ClusterQueue is not modified.

The reason should be that this `base` file is used by them too.

filariow (changes requested):
from `render-diff`'s output I see this some clusters are not getting

components/kueue/production/base/tekton-kueue/config.yaml
oswcab: Indeed. Fixed.

hack/test-tekton-kueue-config.py
```diff
@@ -1562,9 +1607,163 @@ def check_prerequisites(should_print: bool = True) -> Dict[str, TestConfig]:
...
+                    str(plr_res["nominalQuota"]), expected["plr"],
+                    f"{cluster}: unexpected tekton.dev/pipelineruns quota"
+                )
+
+                token_res = self._find_flavor_resource(rg, "konflux-ci-dev-token")
+                self.assertIsNotNone(token_res, f"{cluster}: konflux-ci-dev-token not in flavor resources")
+                self.assertEqual(
+                    str(token_res["nominalQuota"]), expected["token"],
+                    f"{cluster}: expected token quota {expected['token']}, got {token_res['nominalQuota']}"
+                )
```
filariow: I was not expecting this. Why is it needed?

filariow (changes requested):

hack/test-tekton-kueue-config.py
oswcab: Is not strictly needed, is just an additional test that will allow us to validate the cluster-queue YAML manifests (coveredResources + nominalQuota), which are not covered by the existing mutation tests.  The existing _TektonKueueMutationTest_ validates that the CEL expressions produce the correct annotations on   PipelineRuns but it doesn't check the ClusterQueue manifests at all. We could have a correct CEL expression creating _konflux-ci-dev-token_ annotations, but if the ClusterQueue is missing konflux-ci-dev-token from coveredResources or has the wrong nominalQuota, Kueue won't match the resource and workloads won't be admitted as intended.
In consequence, the tests catch:
- If a cluster is missing konflux-ci-dev-token from its coveredResources
- or if a cluster is missing the nominalQuota entry in the flavor resources
- or if the quota doesn't match the expected value (e.g., someone sets 600 instead of 300)

It's only YAML parsing, not podman required, and tries to catch similar mistakes to what happened in teh original PR in staging and that had to be fixed in a follow up commit, the one where the resource name was _konflux-ci.dev/token_ in the ClusterQueue but _konflux-ci-dev-token_ in the CEL expression.

The maintenance overhead from this new test will be that we need to update EXPECTED_TOKEN_QUOTAS every time a cluster's PLR quota changes, so if this feels like too much, we can remove it.

filariow: I think this is a great idea, but I'd propose to have dedicated PR for it. Also I don't like the fact that's scoped to a specific resource (tokens). I'd propose to have it checking that any resource added via tekton-kueue has a counterpart in the ClusterQueue.

> or if the quota doesn't match the expected value (e.g., someone sets 600 instead of 300)
this could introduce more burden than real advantages to me

oswcab: Removed.

oswcab:
**Token request applied too broadly**  Fixed.

filariow (approved):

meyrevived (approved):
/lgtm

mcada:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @mcada |
| **Timestamp** | 2026-06-22T09:11:59.847Z |
| **Risk Level** | low |

PR #12553, opened 2026-06-18
fix: use newTag instead of digest for image references

oswcab:
The kustomization files for external-secrets-operator and oauth2-proxy were using digest: fields to pin container images. The release automation (infra-deployment-update-script in the RPAs) writes {{ revision }} — a git commit SHA — into these fields, producing invalid image references (a bare 40-char hex string instead of sha256:<64-char>). This caused kube-linter latest-tag failures on every release PR.

Switch both to newTag: with the git SHA tag that points to the same image, consistent with all other Konflux infra components.

No image version change: the newTag values resolve to the exact same container images as the previous digests:
- external-secrets-operator: 81c0813f... -> sha256:993ef0ae...
- oauth2-proxy: 7857e195... -> sha256:87a8794d...

Related MR on konflux-release-data: https://gitlab.cee.redhat.com/releng/konflux-release-data/-/merge_requests/19254

### DIFF
--- a/components/external-secrets-operator/staging/kustomization.yaml
+++ b/components/external-secrets-operator/staging/kustomization.yaml
@@ -22,4 +22,4 @@ helmCharts:
 images:
 - name: ghcr.io/external-secrets/external-secrets
   newName: quay.io/konflux-ci/external-secrets-operator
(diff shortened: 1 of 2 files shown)


filariow (approved):

hugares (approved):
/lgtm

PR #12601, opened 2026-06-19
PVO11Y-5268 Forward tekton_pipelines_controller metrics via tekton-ms

kubasikus:
## Summary
- Configure the tekton-ms MonitoringStack to remote-write `tekton_pipelines_controller_*` metrics directly to RHOBS, replacing the federation pipeline for these metrics
- Remove the 10 `tekton_pipelines_controller_*` entries from the staging federation endpoints-params to avoid duplicate series
- Only forward the specific metrics that RHOBS dashboards and alerts depend on, with a minimal label allowlist
- Increase retention from 3h to 7d and add a 10Gi gp3 PVC for persistent storage
- Staging only (stone-stg-rh01, stone-stage-p01)

## Test plan
- [ ] Verify `rhobs` secret appears in `appstudio-tekton` namespace after sync
- [ ] Verify tekton-ms MonitoringStack has remoteWrite and PVC configured
- [ ] Verify `tekton_pipelines_controller_*` metrics appear in RHOBS with correct `source_cluster` label
- [ ] Verify no duplicate series from both federation and tekton-ms
- [ ] Verify `TaskControllerDeadlock` alert continues to evaluate correctly


components/monitoring/prometheus/base/tekton-ms/monitoringstack.yaml
```diff
@@ -8,7 +8,8 @@ spec:
     matchLabels:
       monitoring.rhobs/stack: appstudio-tekton-ms
   logLevel: info
-  retention: 3h
+  retention: 7d
+  retentionSize: 9GB
```
TominoFTW: Is this some educated guess on the 7 days/9Gigs combination or any particular reason for those values with regards to amount of metrics?

components/monitoring/prometheus/staging/tekton-ms/rhobs-secret-path.yaml
```diff
@@ -0,0 +1,7 @@
+---
+- op: replace
+  path: /apiVersion
+  value: external-secrets.io/v1
```
TominoFTW: Probably copy-paste?

The version is already in the target file correctly (I know historically it might have been here beacause of ExternalSecret Operator updates, but now it is just noop

TominoFTW (commented):
otherwise lgtm

components/monitoring/prometheus/base/tekton-ms/monitoringstack.yaml
kubasikus: It's a guesstimate, I will monitor the usage and adjust based on usage

components/monitoring/prometheus/staging/tekton-ms/rhobs-secret-path.yaml
kubasikus: Good catch - removed.

components/monitoring/prometheus/base/tekton-ms/monitoringstack.yaml
raks-tt: Based on some of the past findings, weren't we targeting to increase the retention of prometheus to 14d? Just curious if that requirement still stands

kubasikus: @raks-tt We can increase to any value. This is what I chose for testing in stage. We can increase as necessary.

TominoFTW (approved):

raks-tt (approved):
/lgtm

PR #12694, opened 2026-06-23
konflux-ui: replace static dex-client SA token with dynamic projected SA token (staging-p01 only)

mshaposhnik:
## What

- Switch dex from a static long-lived \`dex-client\` SA token to a short-lived projected ServiceAccount token (600s TTL)
- Replace dex image with \`quay.io/konflux-ci/dex:latest\` (bundles \`run-dex.sh\`)
- Switch dex entrypoint to \`run-dex.sh\`: reads the projected SA token as \`OPENSHIFT_OAUTH_CLIENT_SECRET\`, auto-restarts dex on token rotation
- Add \`TOKEN_PATH\`, \`NAMESPACE\`, \`POD_SERVICE_ACCOUNT\` env vars; add \`tmp\`, \`dex-token\`, \`run-dex\` volumes
- Change dex connector \`clientID\` to \`\$OPENSHIFT_OAUTH_CLIENT_ID\` (dynamically resolved from the pod SA)
- Redirect the \`oauth-redirecturi\` annotation patch to the \`dex\` SA
- Remove the obsolete \`dex-client\` SA and static token Secret from the staging base

Clusters affected: stone-stage-p01

## Why

Static long-lived SA tokens are a security concern and require manual rotation. Projected SA tokens rotate automatically every 600s with no manual intervention, and tie the OAuth client identity to the pod's own SA.

## Validation

- \`kustomize build --enable-helm components/konflux-ui/staging/stone-stage-p01/\` passes
- \`kustomize build --enable-helm components/konflux-ui/staging/stone-stg-rh01/\` passes
- \`dex-client\` SA and Secret absent from rendered output for both clusters

### DIFF
--- a/components/konflux-ui/staging/base/kustomization.yaml
+++ b/components/konflux-ui/staging/base/kustomization.yaml
@@ -1,7 +1,6 @@
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
-  - dex
   - proxy
   - route-and-oauth.yaml
   - ../../base
--- a/components/konflux-ui/staging/stone-stage-p01/base/dex/dex.yaml
+++ b/components/konflux-ui/staging/stone-stage-p01/base/dex/dex.yaml
@@ -1,0 +1,164 @@
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
+    app: dex
+  name: dex
+  annotations:
+    ignore-check.kube-linter.io/no-anti-affinity: "Using topologySpreadConstraints"
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: dex
+  template:
+    metadata:
+      labels:
+        app: dex
+    spec:
+      serviceAccountName: dex
+      topologySpreadConstraints:
+      - maxSkew: 1
+        topologyKey: topology.kubernetes.io/zone
+        whenUnsatisfiable: ScheduleAnyway
+        labelSelector:
+          matchLabels:
+            app: dex
+      containers:
+      - image: quay.io/konflux-ci/dex:latest
+        name: dex
+        command: ["/usr/local/bin/run-dex.sh"]
+        securityContext:
+          readOnlyRootFilesystem: true
+          runAsNonRoot: true
+        resources:
+          limits:
+            cpu: 100m
+            memory: 256Mi
+          requests:
+            cpu: 100m
+            memory: 256Mi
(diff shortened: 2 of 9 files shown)


components/konflux-ui/staging/base/dex/kustomization.yaml
```diff
@@ -1,4 +1,11 @@
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
-  - dex.yaml
+- dex.yaml
+
```
filariow: we should overwrite `quay.io/konflux-ci/dex:latest`'s tag here

components/konflux-ui/staging/base/dex/kustomization.yaml
filariow, line 1: we should also remove [the dex-client ServiceAccount and the Long-Lived Token Secret](https://github.com/redhat-appstudio/infra-deployments/blob/406513025ac95d3cc0d0c5ec771b9efa7a086686/components/konflux-ui/staging/base/route-and-oauth.yaml#L39-L52)

filariow (changes requested):
as this is not installed in the development overlay -hence its installation is not tested, please add it to development or restrict to a single staging cluster

cc: @gbenhaim

components/konflux-ui/staging/base/dex/kustomization.yaml
mshaposhnik: Doesnt get it, sorry..

components/konflux-ui/staging/base/dex/kustomization.yaml
mshaposhnik: ✔️

components/konflux-ui/staging/base/dex/kustomization.yaml
mshaposhnik: @filariow can you elaborate the idea more ?

filariow: the `dex` Deployment is using the image `quay.io/konflux-ci/dex:latest`. we need to use a digest instead. We are not allowed to use the `latest` tag. I'd expect the tag to be overwritten in this kustomization file.

```bash
kustomize build components/konflux-ui/staging/stone-stg-rh01 | yq 'select((.. | select(has("image"))) .image == "quay.io/konflux-ci/dex:latest")'
```

filariow (approved):
/lgtm

great work on this! 😄

PR #12702, opened 2026-06-23
MPC wait for Nvidia CDI to be ready in stg

filariow:
We are noticing some flakiness due to Nvidia CDI not being fully up before the task is put in execution.

This change introduces a wait loop for the CDI to be fully up before completing the provisioning phase.

Assisted-by: Gemini

### DIFF
--- a/components/multi-platform-controller/staging/base/host-config-chart/files/linux-g64xlarge-amd64-init.sh
+++ b/components/multi-platform-controller/staging/base/host-config-chart/files/linux-g64xlarge-amd64-init.sh
@@ -1,4 +1,6 @@
-#!/bin/bash -ex
+#!/bin/bash
+
+set -xeuo pipefail
 
 # Format and mount NVMe disk
 mkfs -t xfs /dev/nvme1n1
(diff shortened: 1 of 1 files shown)


components/multi-platform-controller/staging/base/host-config-chart/files/linux-g64xlarge-amd64-init.sh
```diff
@@ -1,4 +1,6 @@
-#!/bin/bash -ex
+#!/bin/bash
```
hugares, line 1: we are removing the -e which change the behavior, if any command fails, the script will keep going, is this what we want?

filariow: we are not, the very next line is

```
set -xeuo pipefail
```

hugares: LOL

hugares (approved):
/lgtm

PR #12758, opened 2026-06-24
Add RISC-V static host config to MPC staging upstream

wbrefvem:
This is necessary for RISC-V support in MPC.


sadlerap (approved):

components/multi-platform-controller/staging/host-values.yaml
sadlerap, line 1: I think this needs to be in `components/multi-platform-controller/staging-downstream/host-values.yaml`.  The address listed is internal, and stone-stg-rh01 won't be able to connect to it.

sadlerap (changes requested):

components/multi-platform-controller/staging/host-values.yaml
sadlerap: Commit 54d54aa adds it to the configuration for `staging-downstream`, but doesn't remove it from `staging`'s configuration.

wbrefvem: This was very confusing, because I thought I'd removed it from staging. Turns out I forgot save the file so it didn't get picked up in the commit.

mshaposhnik (approved):
/lgtm

sadlerap (approved):

sadlerap (approved):

PR #12794, opened 2026-06-25
feat(KFLUXVNGD-1094): fetch and display per-service feat/fix commits in changelog comment

kelchen123:
For each upstream service bumped by the operator, fetch and display feat/fix conventional commits in the PR comment; also fix a bug where only the first ref change in a kustomization file was detected, causing additional services in the same file to be silently dropped

### DIFF
--- a/infra-tools/cmd/changelog-generator/main.go
+++ b/infra-tools/cmd/changelog-generator/main.go
@@ -1,10 +1,9 @@
 // Command changelog-generator posts a changelog comment on infra-deployments PRs
 // that bump the Konflux operator ref.
 //
-// This is step 3 (KFLUXVNGD-1024): it reads the operator ref at the base branch
-// and PR head, posts a compare link, and lists which upstream sub-services had
-// their pinned SHA bumped inside the operator repo. Commit-level details per
-// service are introduced in the next PR.
+// This is step 4 (KFLUXVNGD-1079): it reads the operator ref, lists upstream
+// sub-services whose pinned SHA changed, and fetches per-service feat/fix
+// conventional commits to display under each bump entry.
 package main
 
 import (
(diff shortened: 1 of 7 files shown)


infra-tools/cmd/changelog-generator/main.go
```diff
@@ -203,6 +230,26 @@ func formatCompare(oldRef, newRef string, bumps []changelog.ServiceBump) string
...
+				fmt.Fprintln(&b)
+				for _, c := range bwc.Commits {
+					if c.Scope != "" {
+						fmt.Fprintf(&b, "- %s(%s): %s\n", c.Type, c.Scope, c.Subject)
+					} else {
+						fmt.Fprintf(&b, "- %s: %s\n", c.Type, c.Subject)
+					}
+				}
+				if bwc.Truncated {
+					fmt.Fprintf(&b, "\n_Showing first 250 commits only — results may be incomplete._\n")
```
yftacherzog: There's a const somewhere in this commit for this number. Let's use consts instead of magic numbers.

infra-tools/internal/changelog/bumps.go
```diff
@@ -25,8 +25,10 @@ var (
 // for kustomization files under operator/upstream-kustomizations/ that have
 // a ?ref=<sha> change, and returns one ServiceBump per bumped service.
 //
-// Duplicate (owner, repo) pairs are deduplicated — only the first occurrence
-// is kept, since the same service can appear in multiple kustomization files.
+// A single kustomization file can reference multiple upstream repositories;
```
yftacherzog, line 28: Kustomization files can also reference images without referencing upstream git resources (e.g. the UI component).

Can we work in parallel on a PR for finding upstream git commits based on images? Would it cause conflicts?

I think it would be beneficial for finding some logic issues we might missed.

infra-tools/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/redhat-appstudio/infra-deployments/infra-tools
 
-go 1.24.5
+go 1.24.4
```
yftacherzog: what's the reason for this change?

yftacherzog (commented):
@kelchen123 Nice work!
Left a few minor comments.

infra-tools/internal/changelog/bumps.go
kelchen123: Sure, I can look into opening a pr for this.

yftacherzog (approved):

PR #12866, opened 2026-06-29
KFLUXVNGD-985 Update konflux-operator and enable kite/kubearchive proxy endpoints in dev

gbenhaim:
## What

- Update konflux-operator to `72d73f696e8847edfd53bbf9d302d6f2f703cfef` in development invariant overlay
- Enable `kite` and `kubearchive` proxy endpoints in the development UI config

Clusters affected: development

[KFLUXVNGD-985](https://redhat.atlassian.net/browse/KFLUXVNGD-985)

## Why

Bump the konflux-operator to the latest version and enable kite and kubearchive proxy endpoints in the development environment UI.

Made with [Cursor](https://cursor.com)

[KFLUXVNGD-985]: https://redhat.atlassian.net/browse/KFLUXVNGD-985?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/components/konflux-kite/development/kustomization.yaml
+++ b/components/konflux-kite/development/kustomization.yaml
@@ -21,7 +21,7 @@ configMapGenerator:
     literals:
       - KITE_PROJECT_ENV=development
       - KITE_HOST=0.0.0.0
-      - KITE_PORT=8080
+      - KITE_PORT=8443
       - KITE_DB_SSL_MODE=disable
       - KITE_LOG_LEVEL=debug
       - KITE_LOG_FORMAT=text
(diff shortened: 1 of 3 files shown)


yftacherzog (approved):
/lgtm

avi-biton (approved):
/approve
/lgtm

components/konflux-kite/development/kustomization.yaml
```diff
@@ -19,9 +19,9 @@ configMapGenerator:
   - name: kite-config
     namespace: konflux-kite
     literals:
-      - KITE_PROJECT_ENV=development
+      - KITE_PROJECT_ENV=staging
```
JoaoPedroPP, line 22: This configuration suppose to be only for local development. Can we keep it?

JoaoPedroPP (commented):
LGTM

Just a quick question below.

components/konflux-kite/development/kustomization.yaml
gbenhaim: It doesn't support tls so it's not suitable here. We should test a configuration which is similar as possible to production.

PR #12874, opened 2026-06-29
feat(PVO11Y-5321): Forward KAExporter labels stage

pacho-rh:
Enable forwarding of labels used in KAExporter to RHOBS. These labels are necessary to get an accurate tenant-level metrics.


raks-tt (approved):
/lgtm

PR #12875, opened 2026-06-30
fix(KONFLUX-13202): Upsize cost-management-operator pod resource limits

pumahaka:
This is for fixing the OOM errors we get on the cost-management-operator pods. Upsizing the default Memory from 500Mi to 1Gi.


pumahaka:
Validation performed on stone-stage-p01 cluster - 
% kustomize build --enable-helm components/cost-management/staging/stone-stage-p01/ | \
    yq 'select(.kind == "Subscription" and .metadata.name == "costmanagement-metrics-operator")'
apiVersion: operators.coreos.com/v1alpha1
kind: Subscription
metadata:
  annotations:
    argocd.argoproj.io/sync-wave: "-1"
  labels:
    operators.coreos.com/costmanagement-metrics-operator.costmanagement-metrics-operator: ""
  name: costmanagement-metrics-operator
  namespace: costmanagement-metrics-operator
spec:
  channel: stable
  config:
    resources:
      limits:
        cpu: 500m
        memory: 1Gi
      requests:
        cpu: 100m
        memory: 100Mi

components/cost-management/staging/base/cost-management-operator-resources-patch.yaml
```diff
@@ -0,0 +1,14 @@
...
+  namespace: costmanagement-metrics-operator
+spec:
+  config:
+    resources:
+      limits:
+        cpu: "500m"
+        memory: "1Gi"
+      requests:
+        cpu: "100m"
+        memory: "100Mi"
```
kubasikus: How were these requests and limits chosen?

pumahaka: Below were the default values: 
        cpu: "500m"
        memory: "500Mi"
      requests:
        cpu: "100m"
        memory: "20Mi"

Based on the logs data we observed that the pod faced OOM errors multiple times in a day. The operator failed 3 times out of the 4 runs. We will increase the memory to 1GB and then observe the behaviour. In case we notice the pod still faces OOM errors, we will upsize the memory to 2GB.

kubasikus: Did you look at the baseline memory/cpu usage when setting the requests?

pumahaka: I checked the pod stats on both staging and production clusters. Below is the average Memory utilisation across the pods.

Staging env:
Usage ranges between 220Mi - 400Mi. We can see occasional spikes close to 500Mi. Drops to 220-260Mi during low usage.

Prod env:
Usage is consistently between 330Mi - 500Mi. We can see multiple spikes close to 500Mi. Typical range is around 380-450Mi. 

As per the above stats, I feel we can go ahead with below values:
Staging env:
cpu: "500m"
memory: "800Mi"
requests:
cpu: "100m"
memory: "300Mi"

Production env:
cpu: "500m"
memory: "1Gi"
requests:
cpu: "100m"
memory: "450Mi"

kubasikus (approved):
/lgtm

components/cost-management/staging/base/cost-management-operator-resources-patch.yaml
kubasikus: Thanks for investigating. This looks good!

PR #12880, opened 2026-06-30
PVO11Y-5342 Clean up kanary staging params and add pre-cleanup step

FaisalAl-Rayes:
## Summary

- Remove cluster-specific defaults from task/pipeline params (tested-cluster, component-repo, user-prefix, quay-repo, fork-target, gitlab-api-url, member-cluster) so cronjobs must set them explicitly
- Remove redundant `--param release-managed-namespace` from all cronjobs (already has a default in the task)
- Add `single-arch-` prefix to single-arch cronjob `--prefix-name` for consistency with multi-arch
- Add pre-cleanup step that removes orphaned resources older than 12h (ReleasePlans, IntegrationTestScenarios, Applications, RPAs) to prevent resource quota exhaustion

## Validation

- Staging only — no production changes
- `kustomize build` passes for both staging clusters (stone-stage-p01, stone-stg-rh01)

## JIRA

[PVO11Y-5342](https://redhat.atlassian.net/browse/PVO11Y-5342)

[PVO11Y-5342]: https://redhat.atlassian.net/browse/PVO11Y-5342?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/components/monitoring/kanary/staging/base/tasks/run-kanary-tests.yaml
+++ b/components/monitoring/kanary/staging/base/tasks/run-kanary-tests.yaml
@@ -6,19 +6,15 @@ spec:
   params:
     - name: tested-cluster
       type: string
-      default: stone-stage-p01
     - name: test-type
       type: string
       default: container-single-arch
     - name: component-repo
       type: string
-      default: "https://gitlab.cee.redhat.com/jhutar/nodejs-devfile-sample4"
     - name: user-prefix
       type: string
-      default: "stagep01cs"
     - name: quay-repo
       type: string
-      default: "redhat-user-workloads-stage"
     - name: pipeline-repo-templating-source
       type: string
       default: "https://github.com/rhtap-perf-test/konflux-probe-test-templates"
@@ -34,13 +30,12 @@ spec:
     # Private clusters only (GitLab-based repos). Leave empty for public clusters.
     - name: fork-target
       type: string
-      default: "jhutar"
+      default: ""
     - name: gitlab-api-url
       type: string
-      default: "https://gitlab.cee.redhat.com/api/v4"
+      default: ""
     - name: member-cluster
       type: string
-      default: "https://api.stone-stage-p01.hpmt.p1.openshiftapps.com:6443"
     - name: tenant-namespace
       type: string
       default: "konflux-perfscale-4-tenant"
@@ -67,6 +62,40 @@ spec:
           - key: token
             path: token
   steps:
+    - name: pre-cleanup
+      image: quay.io/redhat-user-workloads/konflux-perfscale-tenant/loadtest
+      volumeMounts:
+        - name: users-secret
+          mountPath: /opt/users-secret
+          readOnly: true
+      env:
+        - name: SERVER_API_URL
+          value: $(params.member-cluster)
+        - name: TENANT_NAMESPACE
+          value: $(params.tenant-namespace)
+        - name: RELEASE_MANAGED_NAMESPACE
+          value: $(params.release-managed-namespace)
+        - name: RELEASE_MANAGED_TOKEN
+          valueFrom:
+            secretKeyRef:
+              name: kanary-probe-prometheus-credentials
+              key: managed_sa_token
+      script: |
+        #!/usr/bin/env bash
+        set -uo pipefail
+        MAX_AGE=43200
+        echo "Pre-cleanup: removing orphaned resources older than $((MAX_AGE / 3600))h"
+        oc login --token="$(cat /opt/users-secret/token)" --server="$SERVER_API_URL" --insecure-skip-tls-verify 2>/dev/null
+        for resource in ReleasePlan IntegrationTestScenario Application; do
+          oc -n "$TENANT_NAMESPACE" get "$resource" -o json 2>/dev/null \
+            | jq -r --argjson max "$MAX_AGE" '.items[] | select((now - (.metadata.creationTimestamp | fromdateiso8601)) > $max) | .metadata.name' \
+            | xargs -r oc -n "$TENANT_NAMESPACE" delete "$resource" --wait=false 2>/dev/null || true
+        done
+        oc login --token="$RELEASE_MANAGED_TOKEN" --server="$SERVER_API_URL" --insecure-skip-tls-verify 2>/dev/null
+        oc -n "$RELEASE_MANAGED_NAMESPACE" get ReleasePlanAdmission -o json \
+          | jq -r --arg ns "$TENANT_NAMESPACE" --argjson max "$MAX_AGE" '.items[] | select(.spec.origin == $ns) | select((now - (.metadata.creationTimestamp | fromdateiso8601)) > $max) | .metadata.name' \
+          | xargs -r oc -n "$RELEASE_MANAGED_NAMESPACE" delete ReleasePlanAdmission --wait=false 2>/dev/null || true
+        echo "Pre-cleanup done"
     - name: run-test
       image: quay.io/redhat-user-workloads/konflux-perfscale-tenant/loadtest
       volumeMounts:
--- a/components/monitoring/kanary/staging/base/tasks/run-kanary-tests.yaml
+++ b/components/monitoring/kanary/staging/base/tasks/run-kanary-tests.yaml
@@ -82,19 +82,35 @@ spec:
               key: managed_sa_token
       script: |
         #!/usr/bin/env bash
-        set -uo pipefail
-        MAX_AGE=43200
-        echo "Pre-cleanup: removing orphaned resources older than $((MAX_AGE / 3600))h"
-        oc login --token="$(cat /opt/users-secret/token)" --server="$SERVER_API_URL" --insecure-skip-tls-verify 2>/dev/null
-        for resource in ReleasePlan IntegrationTestScenario Application; do
-          oc -n "$TENANT_NAMESPACE" get "$resource" -o json 2>/dev/null \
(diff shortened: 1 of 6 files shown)


components/monitoring/kanary/staging/base/tasks/run-kanary-tests.yaml
```diff
@@ -67,6 +62,40 @@ spec:
...
+          value: $(params.release-managed-namespace)
+        - name: RELEASE_MANAGED_TOKEN
+          valueFrom:
+            secretKeyRef:
+              name: kanary-probe-prometheus-credentials
+              key: managed_sa_token
+      script: |
+        #!/usr/bin/env bash
+        set -uo pipefail
+        MAX_AGE=43200
```
kubasikus:
```suggestion
        MAX_AGE_SECONDS=43200
```

components/monitoring/kanary/staging/base/tasks/run-kanary-tests.yaml
```diff
@@ -67,6 +62,40 @@ spec:
...
+        echo "Pre-cleanup: removing orphaned resources older than $((MAX_AGE / 3600))h"
+        oc login --token="$(cat /opt/users-secret/token)" --server="$SERVER_API_URL" --insecure-skip-tls-verify 2>/dev/null
```
kubasikus: Why is the stderr routed to /dev/null? Doesn't that hide the possible errors? Also, if the login fails, then all subsequent commands will fail. Maybe we should check the exit code and only continue if login was successful?

kubasikus (changes requested):

components/monitoring/kanary/staging/base/tasks/run-kanary-tests.yaml
```diff
@@ -67,6 +62,40 @@ spec:
...
+        for resource in ReleasePlan IntegrationTestScenario Application; do
+          oc -n "$TENANT_NAMESPACE" get "$resource" -o json 2>/dev/null \
+            | jq -r --argjson max "$MAX_AGE" '.items[] | select((now - (.metadata.creationTimestamp | fromdateiso8601)) > $max) | .metadata.name' \
+            | xargs -r oc -n "$TENANT_NAMESPACE" delete "$resource" --wait=false 2>/dev/null || true
```
kubasikus: Why is the `|| true` needed? AFAICS, you want the script to keep running even if something fails in the pipe? But why is the pipefail option set when you don't seem care about anything in the pipe actually failing?

components/monitoring/kanary/staging/base/tasks/run-kanary-tests.yaml
```diff
@@ -67,6 +62,40 @@ spec:
...
+              key: managed_sa_token
+      script: |
+        #!/usr/bin/env bash
+        set -uo pipefail
+        MAX_AGE=43200
+        echo "Pre-cleanup: removing orphaned resources older than $((MAX_AGE / 3600))h"
+        oc login --token="$(cat /opt/users-secret/token)" --server="$SERVER_API_URL" --insecure-skip-tls-verify 2>/dev/null
+        for resource in ReleasePlan IntegrationTestScenario Application; do
+          oc -n "$TENANT_NAMESPACE" get "$resource" -o json 2>/dev/null \
+            | jq -r --argjson max "$MAX_AGE" '.items[] | select((now - (.metadata.creationTimestamp | fromdateiso8601)) > $max) | .metadata.name' \
```
kubasikus: IHMO, this script needs some debug output, at least a count of deleted resources or something. Otherwise, it's going to be hard to debug.

components/monitoring/kanary/staging/base/tasks/run-kanary-tests.yaml
FaisalAl-Rayes: ack, will apply this.

components/monitoring/kanary/staging/base/tasks/run-kanary-tests.yaml
FaisalAl-Rayes: fair, will handle the oc login outcome explicitly

components/monitoring/kanary/staging/base/tasks/run-kanary-tests.yaml
FaisalAl-Rayes: good catch the pipefail should be dropped since the failure of this would not not stop from moving on to the next tekton step

components/monitoring/kanary/staging/base/tasks/run-kanary-tests.yaml
FaisalAl-Rayes: agreed, will update it.

components/monitoring/kanary/staging/base/tasks/run-kanary-tests.yaml
```diff
@@ -67,6 +62,56 @@ spec:
...
+          echo "WARNING: failed to login as tenant SA, skipping tenant cleanup"
+        else
+          for resource in ReleasePlan IntegrationTestScenario Application; do
+            names=$(oc -n "$TENANT_NAMESPACE" get "$resource" -o json 2>/dev/null \
+              | jq -r --argjson max "$MAX_AGE_SECONDS" '.items[] | select((now - (.metadata.creationTimestamp | fromdateiso8601)) > $max) | .metadata.name' || true)
+            count=$(echo "$names" | grep -c . || true)
+            if [[ -n "$names" ]]; then
+              echo "Deleting $count orphaned $resource in $TENANT_NAMESPACE:"
+              echo "$names" | sed 's/^/  /'
+              echo "$names" | xargs -r oc -n "$TENANT_NAMESPACE" delete "$resource" --wait=false 2>/dev/null || true
```
kubasikus: how do we know if anything was actually deleted? The stderr is routed to /dev/null, effectively being silenced. And we don't check the exit code of the deletion command. So while the script will pretend like it's deleting things, it might not be deleting anything at all.

components/monitoring/kanary/staging/base/tasks/run-kanary-tests.yaml
```diff
@@ -67,6 +62,56 @@ spec:
...
+        if ! oc login --token="$RELEASE_MANAGED_TOKEN" --server="$SERVER_API_URL" --insecure-skip-tls-verify; then
+          echo "WARNING: failed to login as managed SA, skipping RPA cleanup"
+        else
+          names=$(oc -n "$RELEASE_MANAGED_NAMESPACE" get ReleasePlanAdmission -o json \
+            | jq -r --arg ns "$TENANT_NAMESPACE" --argjson max "$MAX_AGE_SECONDS" '.items[] | select(.spec.origin == $ns) | select((now - (.metadata.creationTimestamp | fromdateiso8601)) > $max) | .metadata.name' || true)
+          count=$(echo "$names" | grep -c . || true)
+          if [[ -n "$names" ]]; then
+            echo "Deleting $count orphaned ReleasePlanAdmission in $RELEASE_MANAGED_NAMESPACE:"
+            echo "$names" | sed 's/^/  /'
+            echo "$names" | xargs -r oc -n "$RELEASE_MANAGED_NAMESPACE" delete ReleasePlanAdmission --wait=false 2>/dev/null || true
```
kubasikus: Do we have to use xargs? Can't we just run `oc delete $names`?

components/monitoring/kanary/staging/base/tasks/run-kanary-tests.yaml
```diff
@@ -67,6 +62,56 @@ spec:
...
+        #!/usr/bin/env bash
+        set -u
+        MAX_AGE_SECONDS=43200
+        echo "Pre-cleanup: removing orphaned resources older than $((MAX_AGE_SECONDS / 3600))h"
+        if ! oc login --token="$(cat /opt/users-secret/token)" --server="$SERVER_API_URL" --insecure-skip-tls-verify; then
+          echo "WARNING: failed to login as tenant SA, skipping tenant cleanup"
+        else
+          for resource in ReleasePlan IntegrationTestScenario Application; do
+            names=$(oc -n "$TENANT_NAMESPACE" get "$resource" -o json 2>/dev/null \
+              | jq -r --argjson max "$MAX_AGE_SECONDS" '.items[] | select((now - (.metadata.creationTimestamp | fromdateiso8601)) > $max) | .metadata.name' || true)
```
kubasikus: What is the purpose of `|| true` ? Since the script does not `set -x`, a failing command will not cause the script to exit. So I think it's not needed.

components/monitoring/kanary/staging/base/tasks/run-kanary-tests.yaml
FaisalAl-Rayes: valid, moving it to oc delete, it is very unlikely to reach the max limit of argument length.

components/monitoring/kanary/staging/base/tasks/run-kanary-tests.yaml
FaisalAl-Rayes: Valid, removing it.

components/monitoring/kanary/staging/base/tasks/run-kanary-tests.yaml
FaisalAl-Rayes: valid, removed the `|| true`

kubasikus (approved):

PR #12893, opened 2026-06-30
[KFLUXINFRA-3980] Add type guards to build-platforms CEL expressions

oswcab:
First in dev/staging

The `tekton-kueue` webhook CEL expressions for multi-platform resource requests call `.value.map()` on the `build-platforms` parameter, assuming it is an array. When a PipelineRun submits build-platforms as a string, the CEL runtime crashes with "got 'types.String', expected iterable type", causing HTTP 500 and rejecting the PipelineRun entirely (`failurePolicy: Fail`). This was observed on `kflux-prd-rh02` on 2026-06-09.

This commit applies the fix to development and staging environments:
- Adds `type(value) == list` guards to both `build-platforms` CEL expressions (resource request and AWS IP), so they return [] instead of crashing when the value is a string.
- Adds a new CEL expression that detects wrong-typed `build-platforms` and sets a `tekton-kueue.konflux-ci.dev/validation-error` annotation with an error message for operator visibility.
- Adds test cases for string-typed `build-platforms` to the test script.

Production promotion in a follow-up PR after staging validation.

JIRA: [KFLUXINFRA-3980](https://redhat.atlassian.net/browse/KFLUXINFRA-3980)


components/kueue/development/tekton-kueue/config.yaml
```diff
@@ -36,6 +38,14 @@ cel:
           resource('aws-ip', 1)
         ) : []
 
+    # Flag validation error when build-platforms is present but not a list
+    - |
+        has(pipelineRun.spec.params) &&
+        pipelineRun.spec.params.exists(p, p.name == 'build-platforms') &&
+        type(pipelineRun.spec.params.filter(p, p.name == 'build-platforms')[0].value) != list ?
+        [annotation('tekton-kueue.konflux-ci.dev/validation-error',
+          'build-platforms parameter must be an array of platform strings, got a string value')] : []
```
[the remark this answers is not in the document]
oswcab: Done.

filariow (changes requested):
without this change the malformed PLRs are rejected with a 500 return code. With this change they'll be admitted without any `resource` annotation. So they won't be queued correctly and they may be able to consume resources leaving kueue unaware of it.

Until we promote the `no-skip.kueue` ValidatingAdmissionPolicy to production I propose to temporarily introduce a VAP to check if the `tekton-kueue.konflux-ci.dev/validation-error` annotation is set. If it is, then we deny the request. This way the PLR creation will still be denied.

When the `no-skip.kueue` VAP will be in production, we'll remove this temporary VAP. Users will see the PipelineRun being created and failing because they are malformed.

oswcab:
> without this change the malformed PLRs are rejected with a 500 return code. With this change they'll be admitted without any `resource` annotation. So they won't be queued correctly and they may be able to consume resources leaving kueue unaware of it.
> 
> Until we promote the `no-skip.kueue` ValidatingAdmissionPolicy to production I propose to temporarily introduce a VAP to check if the `tekton-kueue.konflux-ci.dev/validation-error` annotation is set. If it is, then we deny the request. This way the PLR creation will still be denied.
> 
> When the `no-skip.kueue` VAP will be in production, we'll remove this temporary VAP. Users will see the PipelineRun being created and failing because they are malformed.

Done.

components/policies/development/kueue/deny-on-validation-error/.chainsaw-test/chainsaw-assert-policy.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingAdmissionPolicy
+metadata:
+  name: deny-invalid-build-platforms.kueue.konflux-ci.dev
+status:
+  observedGeneration: 1
+  typeChecking: {}
```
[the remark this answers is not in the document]
oswcab: The assert files follow the same pattern used by the existing no-skip kueue policy tests in this repo  (`components/policies/development/kueue/no-skip/.chainsaw-test/`), which assert on `status.observedGeneration` and `status.typeChecking` for the VAP and `status.acceptedNames` for the CRD. These tests run successfully in CI.

Chainsaw's assert step retries with a configurable timeout until the assertion matches, so it naturally waits for  the controller to populate the status fields — it does not fail immediately if the fields are absent. This is
the intended mechanism for waiting on controller-managed state.

Changing to spec-based assertions would diverge from the established pattern in this repo without a demonstrated flakiness issue.

components/policies/development/kueue/deny-invalid-build-platforms/deny-invalid-build-platforms-validatingadmissionpolicy.yaml
```diff
@@ -0,0 +1,17 @@
...
+  matchConstraints:
+    resourceRules:
+      - apiGroups: ["tekton.dev"]
+        apiVersions: ["v1"]
+        operations: ["CREATE"]
+        resources: ["pipelineruns"]
+        scope: "Namespaced"
+  validations:
+    - expression: "!has(object.metadata.annotations) || !('tekton-kueue.konflux-ci.dev/validation-error' in object.metadata.annotations)"
+      message: "PipelineRun rejected: build-platforms parameter must be an array of platform strings, got a non-array value"
```
filariow: the validation error message is expected to be stored in the message. We should not hardcode it here.

components/policies/development/kueue/deny-invalid-build-platforms/deny-invalid-build-platforms-validatingadmissionpolicy.yaml
```diff
@@ -0,0 +1,17 @@
+---
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingAdmissionPolicy
+metadata:
+  name: deny-invalid-build-platforms.kueue.konflux-ci.dev
```
filariow: we can make it more flexible and name it after a generic validation error

components/policies/development/kueue/deny-invalid-build-platforms/deny-invalid-build-platforms-validatingadmissionpolicy.yaml
```diff
@@ -0,0 +1,17 @@
...
+  failurePolicy: Fail
+  matchConstraints:
+    resourceRules:
+      - apiGroups: ["tekton.dev"]
+        apiVersions: ["v1"]
+        operations: ["CREATE"]
+        resources: ["pipelineruns"]
+        scope: "Namespaced"
+  validations:
+    - expression: "!has(object.metadata.annotations) || !('tekton-kueue.konflux-ci.dev/validation-error' in object.metadata.annotations)"
```
filariow: I would suggest to use the `kueue.konflux-ci.dev` domain instead.

```suggestion
    - expression: "!has(object.metadata.annotations) || !('kueue.konflux-ci.dev/validation-error' in object.metadata.annotations)"
```

filariow (changes requested):

sadlerap (approved):

PR #12901, opened 2026-07-01
PVO11Y-5323 Fix --namespaces flag to use repeated args

kubasikus:
## Summary

- Split `--namespaces=openshift-monitoring,appstudio-tekton` into two separate `--namespaces` flags. The go-flags library parses `[]string` slices by repeating the flag, not comma-separated — so the single arg was treated as one namespace `"openshift-monitoring,appstudio-tekton"` which doesn't exist, breaking all discovery.

## Risk Assessment

- **Level:** Critical fix
- **Description:** Discovery is completely broken on main — the exporter finds zero Prometheus instances. This restores Platform Prometheus scraping and enables Tekton Prometheus discovery.

## Validation

- Verified locally by building the exporter and testing flag parsing: comma-separated produces 1 namespace, repeated flags produces 2

### DIFF
--- a/components/monitoring/cardinality/base/resources/deployment.yaml
+++ b/components/monitoring/cardinality/base/resources/deployment.yaml
@@ -32,7 +32,8 @@ spec:
             # blocks /api/v1/status/tsdb (OBSDA-1356)
             - "--namespaces=openshift-monitoring"
             - "--namespaces=appstudio-tekton"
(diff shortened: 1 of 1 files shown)


TominoFTW (approved):

martysp21 (approved):
/lgtm

TominoFTW (approved):

PR #12921, opened 2026-07-02
feat(PVO11Y-5279): Update kaexporter ref for startupProbe

pacho-rh:
Updating kaexporter git ref to include changes where startupProbe is set. The startupProbe is needed to give the exporter time to do a cold start-up.


raks-tt (approved):
/lgtm

PR #12925, opened 2026-07-03
PVO11Y-5323 Forward Tekton Prometheus TSDB head metrics to RHOBS

kubasikus:
## Summary

- Add `prometheus_tsdb_head_series` and `prometheus_tsdb_head_chunks` to the Tekton MS remote-write Keep allowlist so we can monitor the Tekton Prometheus cardinality from RHOBS. No new labels needed — these metrics only use the base labels already in the LabelKeep regex.

## Risk Assessment

- **Level:** Low
- **Description:** Adds 2 metric names to the existing allowlist. Minimal cardinality increase (2 series per metric).


TominoFTW (approved):

FaisalAl-Rayes (approved):
/lgtm

PR #12928, opened 2026-07-03
removing konflux_pipelinerun_status metric - KFLUXINFRA-3577

celeztyne:
Removing the metric konflux_pipelinerun_status added in https://github.com/redhat-appstudio/infra-deployments/pull/11933  - [KFLUXINFRA-3577](https://redhat.atlassian.net/browse/KFLUXINFRA-3577):
- O11Y have concerns of this metric saturating unnecessarily Prometheus server, so it would need to be refined a lot, the problem is that at this point was not possible
- The metric is not working anyway
- The kube_resourcequota configured to restrict the amount of PLRs activates a metric that provides the set quota (hard) and current value (used) that serve perfectly for this purpose

[KFLUXINFRA-3577]: https://redhat.atlassian.net/browse/KFLUXINFRA-3577?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/components/monitoring/custom-kube-state-metrics/staging/custom-resource-state-config.yaml
+++ b/components/monitoring/custom-kube-state-metrics/staging/custom-resource-state-config.yaml
@@ -66,7 +66,6 @@ data:
             namespace: [metadata, namespace]
             pipeline: [metadata, labels, tekton.dev/pipeline]
             finalizers: [metadata, finalizers]
-            status: [status, conditions, "[type=Succeeded]", reason]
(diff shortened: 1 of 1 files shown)


celeztyne:
cc @glevi-rh

glevi-rh (approved):
/lgtm

PR #12979, opened 2026-07-09
priority lane for signing releases to dev and stg

filariow:
Signing PLRs are created by the Release Pipelines. Them competing for the same resource can cause deadlocks when burst of Release Pipelines are admitted.

We want to temporarily use tokens to create a priority lane for Signing PLRs. 
We'll remove this when Signing PLRs are moved to common clusters.

rh-pre-commit.version: 2.4.0

### DIFF
--- a/components/kueue/development/queue-config/cluster-queue.yaml
+++ b/components/kueue/development/queue-config/cluster-queue.yaml
@@ -36,9 +36,9 @@ spec:
       - name: default-flavor
         resources:
         - name: tekton.dev/pipelineruns
-          nominalQuota: "500"
-        - name: konflux-ci-dev-token
           nominalQuota: "600"
+        - name: konflux-ci-dev-token
+          nominalQuota: "500"
         - name: linux-arm64
           nominalQuota: "10"
         - name: linux-amd64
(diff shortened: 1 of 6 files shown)


components/kueue/development/queue-config/cluster-queue.yaml
```diff
@@ -51,6 +52,8 @@ spec:
           nominalQuota: "250"
         - name: mintmaker
           nominalQuota: "150"
+        - name: konflux-sign
+          nominalQuota: '50'
```
hugares: I do not know how, but we need to allow way more than this. 

On external cluster, these signing PLR are not happening in the cluster so not kueued. To have similar behavior, I would like to  allow like 300 signing (or more)and make sure these do not chew up the 300 PLR allowed for other PLR, the normal one (build,mintmaker, release,...)

filariow: agreed. This is development and staging, so we can have lower numbers here. Let's discuss it async for prod. We can use `pipelineruns` and `tokens` to allow more PLRs and still have a balanced load

hugares (approved):
/lgtm

PR #12980, opened 2026-07-09
Update kaexporter ref to prevent truncation of resources

raks-tt:
Updating kaexporter git ref to include changes for the increase in Max Items during Cold Start to avoid truncation of PLRs and Resource. This setting change is required to accommodate namespaces with high pipeline volume during the initial 30-day backfill.


pacho-rh (approved):
/lgtm

PR #12989, opened 2026-07-09
Allow 25 more PLR on stone-prod-p02

hugares:
This is a change part of a series to tune kueue in order to reduce the time we are at PLR limit, mainly caused by Mintmaker, thus reducing the admission wait time of all the priority classes other than konflux-dependency-update.

The changes will be small, one at the time to make sure we improve situation while not filling ETCD.

## Risk Assessment
**Risk Level:** Low
**Description:** P02 kueue tuning to improve admission wait time
**Rollback:** Revert PR


hugares:
/hold

ready to review/approve but I want to be there when this change goes in so I can keep an eye on it

PR #13028, opened 2026-07-14
PVO11Y-5094 Move Tekton MS base config into per-environment overlays

kubasikus:
## Summary

Move the shared `base/tekton-ms/` resources (MonitoringStack, RBAC, ServiceMonitors, namespace) into `staging/tekton-ms/base/` and `production/tekton-ms/base/` so each environment can independently tune replicas, resources, and retention without JSON patches.

This is a prerequisite for the production Tekton Prometheus deployment, which needs 2 replicas and potentially different resource limits than staging.

## Risk Assessment

- **Level:** Low
- **Description:** File moves only. The rendered Kubernetes resources for staging are identical — verified with `kustomize build`. Production overlay is new but not yet referenced by any cluster kustomization.

## Validation

- `kustomize build components/monitoring/prometheus/staging/stone-stg-rh01/` renders the Tekton MS MonitoringStack correctly after the move


FaisalAl-Rayes (approved):
/lgtm

PR #13039, opened 2026-07-14
KFLUXINFRA-4007: Combine ring overlays

enkeefe00:
This combines the ring overlays into rd-production and rd-staging so that one Application manages all ApplicationSets on a give ArgoCD instance.


hugares (approved):
/lgtm

PR #13074, opened 2026-07-16
[KFLUXINFRA-3003] Replace DR sleep-pipeline with nightly sanity test pipeline

vsokolen-redhat:
**Summary**
  
Replaces the placeholder `sleep-pipeline` in the `konflux-disaster-recovery` namespace with a real DR sanity test pipeline that runs the proven manual drill against `releng-test-product-tenant` on a nightly schedule.

  
**Changes**

- **pipeline.yaml** — 6 new Tekton Tasks + `dr-sanity-pipeline` implementing: preflight checks → baseline build/integration/release→ Velero backup → disaster simulation (NetworkPolicy + namespace deletion) → restore → post-restore build/integration/release verification → cleanup (openshift-adp), tenant namespace operations, release pipeline polling (rhtap-releng-tenant), and NetworkPolicy management (image-controller)
- **cronjob.yaml** — schedule changed from hourly to daily at 02:00 UTC, added `activeDeadlineSeconds: 7200` and  `startingDeadlineSeconds: 300`
- **triggertemplate.yaml** — references new pipeline, uses `dr-sanity-runner` SA, 2h timeout
- **triggerbinding.yaml** — removed unused `duration` param

**Context**

The full-chain DR drill was validated manually on 2026-07-15 ([KFLUXINFRA-3914](https://redhat.atlassian.net/browse/KFLUXINFRA-3914)). This PR automates the same procedure. The Go e2e test suite (Meirav/Manish) will complement this from CI when ready ([Option C, confirmed in [KFLUXINFRA-3003](https://redhat.atlassian.net/browse/KFLUXINFRA-3003) comments](https://redhat.atlassian.net/browse/KFLUXINFRA-3003?focusedCommentId=17616945)).
  
**Test plan**
  
- [ ] Kustomize build passes locally (`kustomize build components/disaster-recovery/staging/stone-stg-rh01/`)
- [ ] ArgoCD syncs successfully after merge
- [ ] First nightly run completes (monitor `oc get pipelineruns -n konflux-disaster-recovery`)
- [ ] All 5 pipeline tasks pass: preflight, baseline-run, backup, disaster-and-restore, post-restore-verify
- [ ] Cleanup task removes test Backup/Restore/Release CRs

**Jira**
https://redhat.atlassian.net/browse/KFLUXINFRA-3003

[KFLUXINFRA-3914]: https://redhat.atlassian.net/browse/KFLUXINFRA-3914?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/components/disaster-recovery/development/pipeline.yaml
+++ b/components/disaster-recovery/development/pipeline.yaml
@@ -2,23 +2,692 @@
 apiVersion: tekton.dev/v1
 kind: Task
 metadata:
-  name: sleep
+  name: dr-preflight
 spec:
+  params:
+  - name: TENANT_NAMESPACE
+    type: string
+  - name: COMPONENT_NAME
+    type: string
+  - name: RELEASE_PLAN_NAME
+    type: string
   steps:
-  - name: sleep
+  - name: check
+    image: quay.io/konflux-ci/tools@sha256:b846eb68ab06f4c61c987bd1053e0e2dda13e18ecd5a75c945a18f7d78743bd8
+    command: ["bash", "-eu", "-o", "pipefail", "-c"]
+    args:
+    - |
+      echo "=== DR Preflight Check ==="
+      NS="$(params.TENANT_NAMESPACE)"
+      COMP="$(params.COMPONENT_NAME)"
+      RP="$(params.RELEASE_PLAN_NAME)"
+
+      echo "Checking Velero deployment..."
+      READY=$(oc get deploy velero -n openshift-adp -o jsonpath='{.status.readyReplicas}' 2>/dev/null || echo "0")
+      if [ "$READY" != "1" ]; then
+        echo "FAIL: Velero deployment not ready (readyReplicas=$READY)"
+        exit 1
+      fi
+      echo "OK: Velero ready"
+
+      echo "Checking BackupStorageLocation..."
+      BSL_PHASE=$(oc get backupstoragelocations -n openshift-adp -o jsonpath='{.items[0].status.phase}' 2>/dev/null || echo "Unknown")
+      if [ "$BSL_PHASE" != "Available" ]; then
+        echo "FAIL: BSL not Available (phase=$BSL_PHASE)"
+        exit 1
+      fi
+      echo "OK: BSL Available"
+
+      echo "Checking tenant namespace $NS..."
+      if ! oc get namespace "$NS" &>/dev/null; then
+        echo "FAIL: Namespace $NS does not exist"
+        exit 1
+      fi
+      echo "OK: Namespace exists"
+
+      echo "Checking Component $COMP..."
+      if ! oc get components.appstudio.redhat.com "$COMP" -n "$NS" &>/dev/null; then
+        echo "FAIL: Component $COMP not found in $NS"
+        exit 1
+      fi
+      echo "OK: Component exists"
+
+      echo "Checking ReleasePlan $RP..."
+      if ! oc get releaseplans.appstudio.redhat.com "$RP" -n "$NS" &>/dev/null; then
+        echo "FAIL: ReleasePlan $RP not found in $NS"
+        exit 1
+      fi
+      echo "OK: ReleasePlan exists"
+
+      echo "=== Preflight PASSED ==="
+    computeResources: {}
+---
+apiVersion: tekton.dev/v1
+kind: Task
+metadata:
+  name: dr-baseline-run
+spec:
+  params:
+  - name: TENANT_NAMESPACE
+    type: string
+  - name: MANAGED_NAMESPACE
+    type: string
+  - name: COMPONENT_NAME
+    type: string
+  - name: APP_NAME
+    type: string
+  - name: RELEASE_PLAN_NAME
+    type: string
+  results:
+  - name: snapshot-name
+    description: The Snapshot name created by the successful integration test
+  steps:
+  - name: run-baseline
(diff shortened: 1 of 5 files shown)


components/disaster-recovery/development/triggertemplate.yaml
```diff
@@ -1,22 +1,18 @@
...
       pipelineRef:
-        name: sleep-pipeline
-      params:
-      - name: duration
-        value: $(tt.params.duration)
+        name: dr-sanity-pipeline
+      taskRunTemplate:
+        serviceAccountName: dr-sanity-runner
+      timeouts:
+        pipeline: 2h
```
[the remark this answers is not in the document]
meyrevived: The e2e tests also set the timeout at 3h, so this is a valid point

components/disaster-recovery/development/pipeline.yaml
```diff
@@ -2,23 +2,700 @@
...
+  - name: cleanup
     taskRef:
       kind: Task
-      name: sleep
+      name: dr-cleanup
+    params:
+    - name: TENANT_NAMESPACE
+      value: $(params.TENANT_NAMESPACE)
+    - name: BACKUP_NAME
+      value: $(tasks.backup.results.backup-name)
```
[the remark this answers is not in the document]
meyrevived: A good point

gbenhaim:
@flacatus @enkeefe00 @p8r-the-gr8 do you think this test pipeline should be integrated with ring deployments?

enkeefe00:
> @flacatus @enkeefe00 @p8r-the-gr8 do you think this test pipeline should be integrated with ring deployments?

In the next fixVersion, yes. This is out of scope for our current one.

components/disaster-recovery/development/rbac.yaml
```diff
@@ -24,8 +24,88 @@ metadata:
...
   name: tekton-triggers-eventlistener-clusterroles
+---
+# Pipeline ServiceAccount — used by PipelineRuns to execute DR sanity tests
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: dr-sanity-runner
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
```
filariow, line 40: are all these permissions required with cluster scope?

components/disaster-recovery/development/pipeline.yaml
```diff
@@ -2,23 +2,692 @@
...
+    type: string
+  results:
+  - name: backup-name
+    description: The name of the Velero Backup CR created
+  steps:
+  - name: backup
+    image: quay.io/konflux-ci/tools@sha256:b846eb68ab06f4c61c987bd1053e0e2dda13e18ecd5a75c945a18f7d78743bd8
+    command: ["bash", "-eu", "-o", "pipefail", "-c"]
+    args:
+    - |
```
filariow, line 223: I'd suggest we move scripts to kustomize generated ConfigMaps so we can easily run shellcheck on them

filariow (dismissed):

components/disaster-recovery/development/pipeline.yaml
vsokolen-redhat: Agree with you, that standalone .sh files would be better for shellcheck. Just want to confirm scope — this would mean:
- Creating 6 script files under scripts/
- Using kustomize configMapGenerator to produce a ConfigMap
- Refactoring all 6 Tasks to mount the ConfigMap and pass Tekton params as env vars (since $(params.*)substitution doesn't work inside ConfigMap data)

Happy to do it, but it's a non-trivial change, IMO. Want to confirm this is a blocker for 'lgtm' or if we can track it as a follow-up?

components/disaster-recovery/development/rbac.yaml
vsokolen-redhat: Yes — the pipeline operates across multiple NS: openshift-adp (Velero backup/restore/BSL), image-controller (NetworkPolicy to block image-controller during disaster simulation), plus dynamically parameterized $(TENANT_NAMESPACE) and $(MANAGED_NAMESPACE). Namespace resources themselves are also cluster-scoped. A ClusterRole is the only practical option here since the target namespaces are not all known at deploy time.

I can consolidate the duplicate rules (imagerepositories/components/applications appear in multiple entries) and add comments grouping them by purpose — would that address the concern?

components/disaster-recovery/development/pipeline.yaml
filariow: I thought it was easier 😄 we can do that in a follow-up PR.

components/disaster-recovery/development/rbac.yaml
filariow: got it, I think we can scope them down a bit, to have access to Velero resources in the openshift-adp ns only, NetworkPolicy in the image-controllr ns only. It's not much gain though. In case, we can punt that to a follow-up PR.

Thanks for the explanation 😄

vsokolen-redhat:
@filariow would be great to get 'lgtm' as soon as you confirm the changes, thanks

vsokolen-redhat:
CI is green except appstudio-konflux-disaster-recovery which is a known infra flake — ephemeral cluster bootstrap gets stuck on OADP/Kueue sync (our pipeline code is never reached). This test is not required for merge. Could someone /lgtm please?

meyrevived (approved):
/lgtm

PR #13076, opened 2026-07-16
set stage pod requests same as limits for image-controller & build-se…

rcerven:
## What:                                                                                  
stage set requests same as limits for image-controller & build-service pods
                                                                                          
## Why:                                                                                   
so that pods won't have qos Burstable but Guaranteed 
                                                                                          
## Validation:                                                                            
CI
                                                                                          
## Risk Assessment                                                                        
**Risk Level:** Low                                                                       
**Description:** stage image-controller & build-service pods requests same as limits
**Rollback:** Revert PR


hugares (approved):
/lgtm

PR #13080, opened 2026-07-16
fix(KONFLUX-14971, kaexporter): Avoid OOMKilled

pacho-rh:
In attempt to avoid the kaexporter from being OOMKilled:
- Exclude more tenants which will reduce the exporter's memory consumption some.
- Increase memory request and limit to 2Gi.


raks-tt (approved):
/lgtm

FaisalAl-Rayes (approved):
/lgtm

PR #13135, opened 2026-07-20
KFLUXVNGD-1097: add risk-assessment skill

kelchen123:
Teach coder/review agents and interactive workflows how to draft and review production risk assessment sections, and correct ring/hotfix docs.

Assisted-by: Cursor Auto (Composer)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -26,7 +26,7 @@ GitOps monorepo deploying 50+ Kubernetes components across multiple clusters via
 - Promotion order: development/staging → production; changes must be validated in dev/staging before promoting to production
 - Production has per-cluster overlay directories; rollouts must be split into rings (subsets of clusters), not applied to all at once
 - All changes via PR; CODEOWNERS approval required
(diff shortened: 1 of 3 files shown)


.claude/skills
```diff
@@ -0,0 +1 @@
+../.agents/skills
```
glevi-rh: We are not going to have `.claude/` in our repos - it was flagged as an attack vector, please delete it. To notify Claude that we have a new skill, we're just adding it under the skills section and telling the agent how and when to use it in the `AGENTS.md`, it's also not agent agnostic.

.agents/skills/risk-assessment/SKILL.md
```diff
@@ -0,0 +1,137 @@
+---
```
glevi-rh: Please move the skill under the skills folder.

glevi-rh (changes requested):
Please don't change the agent structure of the repo, it's better to keep it agent agnostic and avoid attack vectors.

glevi-rh (approved):
/lgtm

PR #13138, opened 2026-07-21
KFLUXINFRA-3980: add type guards to build-platforms CEL expressions (…

oswcab:
…ring-3)

## What

Applies the build-platforms CEL type guard fix to the final production ring  and consolidates the temporary per-cluster overrides created in rings 1–2  into the shared base/policy.

Clusters affected: `kflux-fedora-01`, `kflux-rhel-p01`, `stone-prd-rh01` (ring 3)

Also touches (consolidation only, no functional change): `stone-prod-p01`,  `kflux-ocp-p01`, `kflux-prd-rh02`, `kflux-osp-p01`, `kflux-prd-rh03`,  `stone-prod-p02` (rings 1–2)

- Adds `type(value) == list` guards and the `validation-error` annotation   expression directly to `production/base/tekton-kueue/config.yaml`.   `kflux-fedora-01` and `stone-prd-rh01` inherit base, so they pick up the  fix without a per-cluster override. `kflux-rhel-p01` has a permanent  override (extra `konflux-release` logic), updated directly.
- Folds the `deny-on-validation-error` ValidatingAdmissionPolicy into the  shared `components/policies/production/policies/kueue` kustomization, so  all 9 production clusters now enforce it.
- Removes the temporary per-cluster `tekton-kueue` config overrides created  in ring 1 (`stone-prod-p01`, `kflux-prd-rh02`) and ring 2  (`kflux-osp-p01`, `kflux-prd-rh03`) — confirmed byte-identical to base  before deletion — and the explicit `deny-on-validation-error` references  in their and `kflux-ocp-p01`/`stone-prod-p02`'s policy kustomizations,  since the fold-in now covers them.
- Updates `hack/test-tekton-kueue-config.py`: removes the now-obsolete  per-cluster config combinations for the four rolled-back overrides, adds   a base `production` test case and a `kflux-rhel-p01` case for  string-typed `build-platforms`.

[KFLUXINFRA-3980](https://redhat.atlassian.net/browse/KFLUXINFRA-3980)

## Why

Completes the production rollout of the fix for the tekton-kueue webhook crash (HTTP 500, `failurePolicy: Fail`) when a PipelineRun submits `build-platforms` as a string instead of an array. Folding into base now that all 9 clusters need the fix keeps the config from permanently accumulating one-off per-cluster overrides.

## Validation

- `kustomize build` passes for all 9 production `kueue` and `policies`  overlays; rendered diff confirms all 9 clusters now carry the CEL guard  and the VAP (previously only rings 1–2's 6 clusters had it via temporary  wiring).
- `python hack/test-tekton-kueue-config.py` passes, including the base  `production` and `kflux-rhel-p01` string-typed `build-platforms` cases.
- Already deployed to dev/staging, ring 1 (#13059), and ring 2 (#13096) with no issues.

## Risk Assessment

**Risk Level:** Low
**What could go wrong:** The guard only changes behavior for the previously-crashing case (string-typed `build-platforms`); array-typed values take the same code path as before, on all 9 clusters, matching production behavior validated in rings 1 and 2. The VAP is scoped to tenant namespaces and only denies PipelineRuns already carrying the
validation-error annotation, so it cannot affect any currently-passing PipelineRun. The override removals were verified byte-identical to base before deletion, so no cluster's effective config changes as a side effect of the consolidation.

**Rollback:** Revert this PR.
**Blast radius:** All 9 production clusters (final ring), but the change itself is a no-op for well-formed PipelineRuns; only previously-crashing malformed requests are affected.

Closes: [KFLUXINFRA-3980](https://redhat.atlassian.net/browse/KFLUXINFRA-3980)


meyrevived (approved):
/lgtm

mcada:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @mcada |
| **Timestamp** | 2026-07-21T08:04:37.628Z |
| **Risk Level** | low |

hack/test-tekton-kueue-config.py
```diff
@@ -1679,41 +1659,13 @@ def check_prerequisites(should_print: bool = True) -> Dict[str, TestConfig]:
...
 
-    # KFLUXINFRA-3980 ring 1: build-platforms type guard
-    # Ring-1 production configs emit kueue.konflux-ci.dev/validation-error
-    # (matching the VAP), unlike the dev/staging default which still emits
-    # tekton-kueue.konflux-ci.dev/validation-error.
-    "multiplatform_new_string_param_production-kflux-ocp-p01": {
+    # KFLUXINFRA-3980 ring 3: build-platforms type guard folded into base
+    # (and kflux-rhel-p01's permanent override). Ring 1's stone-prod-p01 /
+    # kflux-prd-rh02 and ring 2's kflux-osp-p01 / kflux-prd-rh03 temporary
+    # overrides are removed; those clusters now inherit this base coverage.
```
filariow: if ring 3 is the last one, I think we can clean this up

oswcab: Done.

oswcab:
Qodo:
1. production/base changes all clusters; This is ring 3 and the goal is exactly that, to deploy to all production cluster.

filariow (approved):

oswcab:
I removed the `infra-hold` label as Michael has already approved this PR and the only change after his approval was removing a comment in the tests.

PR #13140, opened 2026-07-21
feat(KFLUXVNGD-1154): update konflux-operator image to acf4c5d7 with gap fixes

avi-biton:
Updated the operator image and upstream manifests ref to `acf4c5d7ada0ffaeda6e7d9784893464e4c745ab` in ring-0 invariant.

Added `trusted-ca` ConfigMap resources to the build-service and integration-service operator components.

The new operator image includes fixes for gaps identified during staging migration testing:

- **trusted-ca**: The operator now mounts the OpenShift cluster CA bundle into build-service and integration-service controller containers, matching the legacy `rh-certs` component behavior
- **CONSOLE_URL_TASKLOG**: The operator now auto-discovers the konflux-ui Route and injects `CONSOLE_URL` and `CONSOLE_URL_TASKLOG` env vars into the integration controller, eliminating the need for manual ConfigMap configuration
- **info enhancements**: Improvements to the konflux-info component handling

- `rings/ring-0/base/invariant/kustomization.yaml` — bumped ref and image tag
- `rings/ring-0/base/cr/build/` — added `trusted-ca-configmap.yaml`
- `rings/ring-0/base/cr/integration/` — added `trusted-ca-configmap.yaml`

### DIFF
--- a/components/konflux-operator/rings/ring-0/base/cr/build/trusted-ca-configmap.yaml
+++ b/components/konflux-operator/rings/ring-0/base/cr/build/trusted-ca-configmap.yaml
@@ -1,0 +1,8 @@
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: trusted-ca
+  namespace: build-service
+  labels:
+    config.openshift.io/inject-trusted-cabundle: "true"
--- a/components/konflux-operator/rings/ring-0/base/cr/integration/trusted-ca-configmap.yaml
+++ b/components/konflux-operator/rings/ring-0/base/cr/integration/trusted-ca-configmap.yaml
@@ -1,0 +1,8 @@
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: trusted-ca
+  namespace: integration-service
+  labels:
+    config.openshift.io/inject-trusted-cabundle: "true"
--- a/components/konflux-operator/rings/ring-0/base/cr/build/kustomization.yaml
+++ b/components/konflux-operator/rings/ring-0/base/cr/build/kustomization.yaml
@@ -1,4 +1,6 @@
 apiVersion: kustomize.config.k8s.io/v1alpha1
 kind: Component
+resources:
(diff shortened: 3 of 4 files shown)


components/konflux-operator/rings/ring-0/base/cr/build/trusted-ca-configmap.yaml
```diff
@@ -0,0 +1,8 @@
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: trusted-ca
+  namespace: build-service
```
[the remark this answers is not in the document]
avi-biton: Its not an issue.
The namespaces are created by the operator as part of its reconciliation loop. 
The ConfigMaps may briefly fail on first sync but will succeed on the next ArgoCD self-heal cycle, this is the standard pattern.

components/konflux-operator/rings/ring-0/base/cr/integration/trusted-ca-configmap.yaml
yftacherzog, line 1: Is it the same today for each environment?
should it be maintained per ring or copied as part of the invariant?

avi-biton: Good point, I think its cross rings, so maybe we can add it to invariant.

avi-biton: Thinking about it again, I think we should leave it as it is.
- each configmap is in the specific component folder
- OWNERS file is different for each and moving them will shift ownership

yftacherzog: ok. Let's see where it leads. If we get too much duplicates for similar things over time, maybe we'll have some sort of base after all, that is invariant across envs and revisions (we can have ownerfiles there as well).

yftacherzog (approved):
/approve
/lgtm

PR #13202, opened 2026-07-27
ci: kustomize build with retries for hack/forbid-clusterpolicies.sh

yftacherzog:
This script was failing in CI on transient issues. Rerunning it requires privileges that are not available for most PR owners, which means they either have to rebase the PR and trigger all CI jobs again, or ask someone with the right privileges to rerun the job.

Adding retries to the script should hopefully reduce the number of times this is needed.

Assisted-by: Cursor

### DIFF
--- a/hack/forbid-clusterpolicies.sh
+++ b/hack/forbid-clusterpolicies.sh
@@ -5,6 +5,7 @@ set -e -o pipefail
 # inputs
 declare OUTPUT
 folder="${1}"
+script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 
 error_handler() {
   local exit_code=$?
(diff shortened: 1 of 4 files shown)


hack/forbid-clusterpolicies.sh
```diff
@@ -23,8 +23,40 @@ error_handler() {
 # Trap the ERR signal and call the error_handler function
 trap error_handler ERR
 
+kustomize_build_with_retry() {
+  local max_attempts="${KUSTOMIZE_RETRIES:-5}"
+  local attempt=1
+  local kustomize_output=""
+
+  trap - ERR
```
avi-biton: ### ERR trap not restored on unexpected failures

The function disables the ERR trap at the top (`trap - ERR`) and only restores it on the two explicit `return` paths. If something else in the loop body fails unexpectedly — `echo`, `yq`, `sleep`, or the arithmetic — the function exits without re-arming the trap. From that point on, the rest of the script runs without `error_handler` protection.

Consider using `trap 'trap error_handler ERR' RETURN` at the top of the function so the ERR trap is unconditionally restored when the function exits, regardless of how it exits.

_Posted by Avi's Cursor_

yftacherzog: done

yftacherzog: I moved to a different approach after fixing.

avi-biton (approved):
/approve
/lgtm

PR #13227, opened 2026-07-28
fix(): update namespace-lister resources

avi-biton:
update namespace-lister resources.


yftacherzog (approved):
/approve
/lgtm

PR #13237, opened 2026-07-28
RELEASE-2681: fix staging Vault path for release-service GitHub token

happybhati:
## Summary
Follow-up to #13196. ExternalSecret could not sync because the Vault secret lives under the `stonesoup` mount at `staging/release/...`, not `release/github/...` under `konflux`.
## Change
- Update `remoteRef.key` to `staging/release/github/konflux-release-service`


johnbieren (approved):
and this is why we do staging first =)

PR #13371, opened 2026-08-06
Switch Tekton Results from standalone to operator-managed deployment

enarha:
For development and staging only. Enables result in TektonConfig, removes standalone manifests, moves resources from tekton-results to openshift-pipelines namespace and align the configuration.

Assisted-by: Claude Code (claude-opus-4-6)

### DIFF
--- a/components/konflux-operator/rings/ring-0/base/cr/konflux-ui/ui.yaml
+++ b/components/konflux-operator/rings/ring-0/base/cr/konflux-ui/ui.yaml
@@ -9,6 +9,8 @@ spec:
         nodePortService:
           httpsPort: 30011
       proxy:
+        tektonResults:
+          hostname: tekton-results-api-service.openshift-pipelines.svc.cluster.local
         endpoints:
           kite:
             enabled: true
--- a/components/pipeline-service/development/dev-only-pipeline-service-storage-configuration.yaml
+++ b/components/pipeline-service/development/dev-only-pipeline-service-storage-configuration.yaml
@@ -1,4 +1,4 @@
-# Dev-only storage: PostgreSQL + MinIO (S3) for Tekton Results. MinIO runs with reduced resources.
+# Dev-only storage: MinIO (S3) for Tekton Results. MinIO runs with reduced resources.
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
@@ -46,131 +46,29 @@ subjects:
   name: openshift-gitops-argocd-application-controller
   namespace: openshift-gitops
 ---
-apiVersion: argoproj.io/v1alpha1
-kind: Application
-metadata:
-  annotations:
-    argocd.argoproj.io/sync-wave: "0"
-  name: postgres
-  namespace: openshift-gitops
-spec:
-  destination:
-    namespace: tekton-results
-    server: https://kubernetes.default.svc
-  project: default
-  source:
-    chart: postgresql
-    helm:
-      parameters:
-      - name: tls.enabled
-        value: "true"
-      - name: tls.certificatesSecret
-        value: postgresql-tls
-      - name: tls.certFilename
-        value: tls.crt
-      - name: tls.certKeyFilename
-        value: tls.key
-      - name: auth.database
-        value: tekton_results
-      - name: auth.username
-        value: tekton
-      - name: auth.existingSecret
-        value: tekton-results-database
-      - name: auth.secretKeys.userPasswordKey
-        value: db.password
-      - name: auth.secretKeys.adminPasswordKey
-        value: db.password
-      - name: primary.resources.requests
-        value: "null"
-      - name: primary.podSecurityContext.fsGroup
-        value: "null"
-      - name: primary.podSecurityContext.seccompProfile.type
-        value: RuntimeDefault
-      - name: primary.containerSecurityContext.runAsUser
-        value: "null"
-      - name: primary.containerSecurityContext.allowPrivilegeEscalation
-        value: "false"
-      - name: primary.containerSecurityContext.runAsNonRoot
-        value: "true"
-      - name: primary.containerSecurityContext.seccompProfile.type
-        value: RuntimeDefault
-      - name: primary.containerSecurityContext.capabilities.drop[0]
-        value: ALL
-      - name: volumePermissions.enabled
-        value: "false"
-      - name: shmVolume.enabled
-        value: "false"
-      - name: volumePermissions.image.registry
-        value: quay.io
-      - name: volumePermissions.image.repository
-        value: openshift-pipeline/os-shell
-      - name: volumePermissions.image.tag
-        value: latest
-      - name: image.registry
-        value: quay.io
-      - name: image.repository
-        value: openshift-pipeline/postgresql
-      - name: image.tag
-        value: latest
-      - name: global.security.allowInsecureImages
-        value: "true"
-      releaseName: postgres
-    repoURL: registry-1.docker.io/bitnamichartssecure
-    targetRevision: 17.0.2
-  syncPolicy:
-    automated:
-      prune: true
-      selfHeal: true
-    retry:
-      backoff:
-        duration: 10s
-        factor: 2
-        maxDuration: 3m
-      limit: -1
-    syncOptions:
-    - CreateNamespace=false
-    - Validate=false
----
-apiVersion: v1
-type: Opaque
-kind: Secret
-metadata:
-  name: repo-bitnami-postgresql
-  namespace: openshift-gitops
-  labels:
-    argocd.argoproj.io/secret-type: repository
-data:
-  enableOCI: dHJ1ZQ==
-  name: Yml0bmFtaWNoYXJ0c3NlY3VyZQ==
-  project: ZGVmYXVsdA==
-  type: aGVsbQ==
-  url: cmVnaXN0cnktMS5kb2NrZXIuaW8vYml0bmFtaWNoYXJ0c3NlY3VyZQ==
----
-# MinIO Tenant root credentials. Same credentials are used in tekton-results-s3 for Tekton Results API.
 apiVersion: v1
 kind: Secret
 metadata:
   name: minio-storage-configuration
-  namespace: tekton-results
+  namespace: openshift-pipelines
 type: Opaque
 stringData:
   config.env: |
-    export MINIO_ROOT_USER=minioadmin
-    export MINIO_ROOT_PASSWORD=minioadmin
+    export MINIO_ROOT_USER=minioadmin # gitleaks:allow
+    export MINIO_ROOT_PASSWORD=minioadmin # gitleaks:allow
 ---
-# S3 credentials for Tekton Results (API and watcher). Points at in-cluster MinIO tenant "storage".
 apiVersion: v1
 kind: Secret
 metadata:
   name: tekton-results-s3
-  namespace: tekton-results
+  namespace: openshift-pipelines
 type: Opaque
 stringData:
-  aws_access_key_id: minioadmin
-  aws_secret_access_key: minioadmin
+  aws_access_key_id: minioadmin # gitleaks:allow
+  aws_secret_access_key: minioadmin # gitleaks:allow
   aws_region: us-east-1
   bucket: tekton-results
-  endpoint: https://minio.tekton-results.svc.cluster.local
+  endpoint: https://minio.openshift-pipelines.svc.cluster.local
 ---
 apiVersion: minio.min.io/v2
 kind: Tenant
@@ -184,7 +82,7 @@ metadata:
   labels:
     app: minio
   name: storage
-  namespace: tekton-results
+  namespace: openshift-pipelines
 spec:
   certConfig: {}
   configuration:
@@ -240,3 +138,89 @@ spec:
     minioServiceAnnotations: {}
     minioServiceLabels: {}
   subPath: ""
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
+  annotations:
+    argocd.argoproj.io/sync-options: SkipDryRunOnMissingResource=true
+    argocd.argoproj.io/sync-wave: "1"
+    argocd.argoproj.io/hook: Sync
+    argocd.argoproj.io/hook-delete-policy: BeforeHookCreation
+  name: create-minio-bucket
+  namespace: openshift-pipelines
+spec:
+  backoffLimit: 3
+  ttlSecondsAfterFinished: 300
+  template:
+    spec:
+      restartPolicy: OnFailure
+      containers:
+      - name: mc
+        image: quay.io/minio/mc:RELEASE.2023-01-28T20-29-38Z
+        command:
+        - /bin/sh
+        - -c
+        - |
+          set -e
+          echo "Waiting for MinIO to become ready..."
+          for i in $(seq 1 30); do
+            if mc --insecure --config-dir /tmp config host add minio "$S3_ENDPOINT" "$S3_ACCESS_KEY_ID" "$S3_SECRET_ACCESS_KEY" >/dev/null 2>&1; then
+              echo "MinIO is ready."
+              break
+            fi
+            if [ "$i" -eq 30 ]; then
+              echo "MinIO did not become ready in time." >&2
+              exit 1
+            fi
+            echo "MinIO not ready yet (attempt $i/30), retrying in 10s..."
+            sleep 10
+          done
+          if mc --insecure --config-dir /tmp ls minio | grep -q "$S3_BUCKET_NAME"; then
+            echo "Bucket $S3_BUCKET_NAME already exists."
+          else
+            mc --insecure --config-dir /tmp mb --with-lock --region "$S3_REGION" minio/"$S3_BUCKET_NAME"
+            echo "Bucket $S3_BUCKET_NAME created."
+          fi
+        env:
+        - name: S3_ACCESS_KEY_ID
+          valueFrom:
+            secretKeyRef:
+              key: aws_access_key_id
+              name: tekton-results-s3
+        - name: S3_SECRET_ACCESS_KEY
+          valueFrom:
+            secretKeyRef:
+              key: aws_secret_access_key
+              name: tekton-results-s3
+        - name: S3_REGION
+          valueFrom:
+            secretKeyRef:
+              key: aws_region
+              name: tekton-results-s3
+        - name: S3_BUCKET_NAME
+          valueFrom:
+            secretKeyRef:
+              key: bucket
+              name: tekton-results-s3
+        - name: S3_ENDPOINT
+          valueFrom:
+            secretKeyRef:
+              key: endpoint
+              name: tekton-results-s3
+        resources:
+          limits:
+            cpu: 100m
+            memory: 128Mi
+          requests:
+            cpu: 5m
+            memory: 32Mi
+        securityContext:
+          readOnlyRootFilesystem: true
+          runAsNonRoot: true
+        volumeMounts:
+        - mountPath: /tmp
+          name: tmp-mc-volume
+      volumes:
+      - emptyDir: {}
+        name: tmp-mc-volume
--- a/components/pipeline-service/development/kustomization.yaml
+++ b/components/pipeline-service/development/kustomization.yaml
@@ -11,7 +11,9 @@ resources:
   - github.com/minio/operator?ref=v5.0.15  # when updating the tag, update also the patches below
   - main-pipeline-service-configuration.yaml
   - dev-only-pipeline-service-storage-configuration.yaml
-  - ../base/rbac
+  - ../base/rbac/cluster-role
+  - ../base/rbac/openshift-pipelines
+  - ../base/rbac/tekton-logging
 
 patches:
   - target:
--- a/components/pipeline-service/development/main-pipeline-service-configuration.yaml
+++ b/components/pipeline-service/development/main-pipeline-service-configuration.yaml
@@ -9,16 +9,6 @@ metadata:
   name: openshift-pipelines
 ---
 apiVersion: v1
-kind: Namespace
-metadata:
-  annotations:
-    argocd.argoproj.io/sync-options: SkipDryRunOnMissingResource=true
-    argocd.argoproj.io/sync-wave: "-1"
-  labels:
-    argocd.argoproj.io/managed-by: openshift-gitops
-  name: tekton-results
----
-apiVersion: v1
 kind: ServiceAccount
 metadata:
   annotations:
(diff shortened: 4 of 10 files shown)


yftacherzog:
This PR migrates Tekton Results to operator-managed deployment in the `openshift-pipelines` namespace for development and staging pipeline-service. A few other infra-deployments components still reference the previous standalone deployment in the `tekton-results` namespace and were not updated in this diff.

**Legacy Konflux UI proxy (nginx-based)**

These configs still proxy `/api/k8s/plugins/tekton-results/` to `tekton-results-api-service.tekton-results.svc.cluster.local:8080`:

- `components/konflux-ui/staging/base/proxy/tekton-results.conf`
- `components/konflux-ui/production/base/proxy/tekton-results.conf`
- `components/konflux-ui/production/base/proxy/tekton-results-workspaces.conf`
- `components/konflux-ui/production/kflux-fedora-01/kustomization.yaml` (`TEKTON_RESULTS_URL`)

If merged as-is for environments that use this nginx proxy path, the UI Tekton Results plugin route is expected to fail once the standalone `tekton-results` namespace is removed (502/connection errors from the proxy, or missing Results data in the UI).

**Telemetry (KonfluxSegmentBridge)**

`components/konflux-operator/rings/ring-1/base/cr/telemetry/telemetry.yaml` sets `TEKTON_RESULTS_API_ADDR` to `tekton-results-api-service.tekton-results.svc.cluster.local:8080`, with a comment noting Results lives in its own namespace. On clusters where this PR's pipeline-service changes are applied, the telemetry CronJob may be unable to reach the Results API at that address.

**New-cluster bootstrap template**

`hack/new-cluster/templates/konflux-ui/kustomization.yaml` sets `TEKTON_RESULTS_URL` to the `tekton-results` namespace address. New clusters bootstrapped from this template may get a UI proxy init config that does not match operator-managed Results in `openshift-pipelines`.

---

These are separate from the operator-overlay e2e failure (konflux-operator Caddy proxy, init-time DNS discovery). They concern the legacy nginx UI proxy path, telemetry wiring, and cluster bootstrap templates that still assume the old namespace layout.

— Yftach's Cursor

yftacherzog:
> **Telemetry (KonfluxSegmentBridge)**
> 
> `components/konflux-operator/rings/ring-1/base/cr/telemetry/telemetry.yaml` sets `TEKTON_RESULTS_API_ADDR` to `tekton-results-api-service.tekton-results.svc.cluster.local:8080`, with a comment noting Results lives in its own namespace. On clusters where this PR's pipeline-service changes are applied, the telemetry CronJob may be unable to reach the Results API at that address.

Human says: this is only relevant for clusters using the Konflux Operator.

avi-biton:
The pipeline-service half of this migration looks right, but staging UI on `stone-stg-rh01` is still wired to the old Results Service. If this merges as-is, Argo will prune `tekton-results` and that UI will lose logs/results.

`components/konflux-ui/staging/base/proxy/tekton-results.conf` still does:
```
proxy_pass https://tekton-results-api-service.tekton-results.svc.cluster.local:8080/;
```

That hostname is hardcoded. There is no DNS probe or fallback. After this PR, operator-managed Results lives at `tekton-results-api-service.openshift-pipelines.svc.cluster.local:8080`, and the `tekton-results` namespace goes away.

This only bites clusters that still run the **legacy nginx UI**. `argo-cd-apps/overlays/staging-downstream/exclude-operator-owned-clusters.yaml` drops `stone-stage-p01`, `lightwell-dev`, and `kflux-lw-p01` from the `konflux-ui` ApplicationSet - those clusters use the operator/Caddy proxy, which already probes `openshift-pipelines`. `stone-stg-rh01` is not in that list, so it still gets this nginx config **and** the pipeline-service change.

On that cluster the failure mode is: browser → `/api/k8s/plugins/tekton-results/` → nginx → NXDOMAIN/connection refused → **502**. Pipelines can still run; the UI just cannot read stored results.

Production proxies can wait — production Results stays standalone in `tekton-results`.

Please either:

1. Change the staging `proxy_pass` to `https://tekton-results-api-service.openshift-pipelines.svc.cluster.local:8080` in this PR, or
2. Land that UI change first and block this PR on it.

Merging this first and fixing the UI later leaves a staging UI outage in between.

---
*Posted by Avi's Cursor*

components/pipeline-service/staging/lightwell-dev/deploy.yaml
```diff
@@ -26,6 +26,23 @@ metadata:
 ---
```
avi-biton, line 26: `lightwell-dev` is a cluster that **does not run Tekton Results**. 
It was removed fro it in this PR: https://github.com/redhat-appstudio/infra-deployments/pull/13457

The current PR turns Results on there without giving it the secrets it needs.

`components/pipeline-service/staging/lightwell-dev/resources/kustomization.yaml` still says Results is not deployed, and it still deletes the `tekton-results-database` and `tekton-results-s3` secrets.

`deploy.yaml` (what Argo syncs) now sets `result.disabled: false` and expects those same secrets in `openshift-pipelines`.

So the operator will try to start Results, the secrets will not be there, and `TektonConfig` will not go Ready.

There is a discussion about adding `tekton-results` to `lightwell-dev` cluster: 
https://redhat-internal.slack.com/archives/C0BP4GYQJPQ/p1786638749650029

avi-biton:
`hack/new-cluster/templates/konflux-ui/kustomization.yaml` still sets

`TEKTON_RESULTS_URL=https://tekton-results-api-service.tekton-results.svc.cluster.local:8080`

That does not break any live cluster. It will bite the next cluster bootstrapped from this template if Results is operator-managed there.

avi-biton:
`components/konflux-operator/rings/ring-1/base/cr/telemetry/telemetry.yaml`
Still refers to `tekton-results`:
```
        env:
          # tekton-results is deployed in its own namespace on this cluster
          # (see components/pipeline-service), not in openshift-pipelines as
          # the operator's default assumes.
          - name: TEKTON_RESULTS_API_ADDR
            value: tekton-results-api-service.tekton-results.svc.cluster.local:8080
```

enarha:
> The pipeline-service half of this migration looks right, but staging UI on `stone-stg-rh01` is still wired to the old Results Service. If this merges as-is, Argo will prune `tekton-results` and that UI will lose logs/results.
> 
> `components/konflux-ui/staging/base/proxy/tekton-results.conf` still does:
> 
> ```
> proxy_pass https://tekton-results-api-service.tekton-results.svc.cluster.local:8080/;
> ```
> 
> That hostname is hardcoded. There is no DNS probe or fallback. After this PR, operator-managed Results lives at `tekton-results-api-service.openshift-pipelines.svc.cluster.local:8080`, and the `tekton-results` namespace goes away.
> 
> This only bites clusters that still run the **legacy nginx UI**. `argo-cd-apps/overlays/staging-downstream/exclude-operator-owned-clusters.yaml` drops `stone-stage-p01`, `lightwell-dev`, and `kflux-lw-p01` from the `konflux-ui` ApplicationSet - those clusters use the operator/Caddy proxy, which already probes `openshift-pipelines`. `stone-stg-rh01` is not in that list, so it still gets this nginx config **and** the pipeline-service change.
> 
> On that cluster the failure mode is: browser → `/api/k8s/plugins/tekton-results/` → nginx → NXDOMAIN/connection refused → **502**. Pipelines can still run; the UI just cannot read stored results.
> 
> Production proxies can wait — production Results stays standalone in `tekton-results`.
> 
> Please either:
> 
> 1. Change the staging `proxy_pass` to `https://tekton-results-api-service.openshift-pipelines.svc.cluster.local:8080` in this PR, or
> 2. Land that UI change first and block this PR on it.
> 
> Merging this first and fixing the UI later leaves a staging UI outage in between.
> 
> _Posted by Avi's Cursor_

Updated the URL to point to the new Service endpoint in the `openshift-pipelines` namespace. That's easy to do for staging because we switch the staging clusters at once. For prod, we'll need to use some kustomizations per cluster, but that won't be strictly required as Tekton Results is not production service in Konflux. No one should be using it at the moment.

components/pipeline-service/staging/lightwell-dev/deploy.yaml
enarha: I understand we need both at the moment (Results deployed, it'll collect data, but no one should be querying it). So we do not need to restore all the manifests #13457 removed, but just the external DB secrets.

enarha:
> `hack/new-cluster/templates/konflux-ui/kustomization.yaml` still sets
> 
> `TEKTON_RESULTS_URL=https://tekton-results-api-service.tekton-results.svc.cluster.local:8080`
> 
> That does not break any live cluster. It will bite the next cluster bootstrapped from this template if Results is operator-managed there.

I do not think we can change that at the moment because it'll affect any new cluster bootstrapped, including production clusters. Once I start switching the prod clusters, hopefully immediately after this PR is merged, I'll change the bootstrap kustomization as well. Again, the current plan for Results is to update the existing deployment so it start collect data, but no one (UI, etc.) should be using it.

enarha:
> `components/konflux-operator/rings/ring-1/base/cr/telemetry/telemetry.yaml` Still refers to `tekton-results`:
> 
> ```
>         env:
>           # tekton-results is deployed in its own namespace on this cluster
>           # (see components/pipeline-service), not in openshift-pipelines as
>           # the operator's default assumes.
>           - name: TEKTON_RESULTS_API_ADDR
>             value: tekton-results-api-service.tekton-results.svc.cluster.local:8080
> ```

I assume this is prod, isn't it? We'll need to change it once we start updating prod.

avi-biton:
> > `components/konflux-operator/rings/ring-1/base/cr/telemetry/telemetry.yaml` Still refers to `tekton-results`:
> > ```
> >         env:
> >           # tekton-results is deployed in its own namespace on this cluster
> >           # (see components/pipeline-service), not in openshift-pipelines as
> >           # the operator's default assumes.
> >           - name: TEKTON_RESULTS_API_ADDR
> >             value: tekton-results-api-service.tekton-results.svc.cluster.local:8080
> > ```
> 
> I assume this is prod, isn't it? We'll need to change it once we start updating prod.

ring-1 is staging, not prod

components/konflux-operator/rings/ring-0/base/cr/konflux-ui/ui.yaml
```diff
@@ -9,6 +9,8 @@ spec:
         nodePortService:
           httpsPort: 30011
       proxy:
+        tektonResults:
+          hostname: tekton-results-api-service.openshift-pipelines.svc.cluster.local
```
aThorp96, line 13: This is a bit surprising to me as a new key instead of a changed line. Does the UI hardcode the previous URL somewhere?

components/pipeline-service/development/dev-only-pipeline-service-storage-configuration.yaml
```diff
@@ -46,131 +46,29 @@ subjects:
...
+  namespace: openshift-pipelines
 type: Opaque
 stringData:
   config.env: |
-    export MINIO_ROOT_USER=minioadmin
-    export MINIO_ROOT_PASSWORD=minioadmin
+    export MINIO_ROOT_USER=minioadmin # gitleaks:allow
+    export MINIO_ROOT_PASSWORD=minioadmin # gitleaks:allow
 ---
-# S3 credentials for Tekton Results (API and watcher). Points at in-cluster MinIO tenant "storage".
```
aThorp96, line 161: any reason this comment was deleted?

components/pipeline-service/development/dev-only-pipeline-service-storage-configuration.yaml
```diff
@@ -240,3 +138,89 @@ spec:
     minioServiceAnnotations: {}
     minioServiceLabels: {}
   subPath: ""
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
```
aThorp96, line 144: Possibly something I'll learn later in the review:
Why are we creating the minio bucket here? Were we not creating it before? 

TODO for me: confirm the argocd annotations cause this to re-run when it changes or when it's necessary, but wont rerun every sync attempt since it may be cleaned up very often. It looks like it's idempotent so that's good

components/pipeline-service/development/kustomization.yaml
```diff
@@ -11,7 +11,9 @@ resources:
   - github.com/minio/operator?ref=v5.0.15  # when updating the tag, update also the patches below
   - main-pipeline-service-configuration.yaml
   - dev-only-pipeline-service-storage-configuration.yaml
-  - ../base/rbac
+  - ../base/rbac/cluster-role
+  - ../base/rbac/openshift-pipelines
+  - ../base/rbac/tekton-logging
```
aThorp96, line 16: I assume this is because we can't delete `base/rbac/tekton-results` until the migration is complete?

components/pipeline-service/development/main-pipeline-service-configuration.yaml
```diff
@@ -525,325 +238,13 @@ metadata:
...
   name: pipeline-service-exporter-reader-binding
 roleRef:
-  apiGroup: rbac.authorization.k8s.io
-  kind: ClusterRole
-  name: pipeline-service-exporter-reader
-subjects:
-- kind: ServiceAccount
-  name: pipeline-service-exporter
-  namespace: openshift-pipelines
----
```
aThorp96, line 535: ~I think this was deleted by mistake~ _it's still there but the git diff renders it as (re)moved_

components/pipeline-service/development/main-pipeline-service-configuration.yaml
```diff
@@ -525,325 +238,13 @@ metadata:
...
-roleRef:
-  apiGroup: rbac.authorization.k8s.io
-  kind: ClusterRole
-  name: tekton-results-admin
-subjects:
-- kind: ServiceAccount
-  name: tekton-results-watcher
-  namespace: tekton-results
----
-apiVersion: rbac.authorization.k8s.io/v1
```
aThorp96, line 604: I assume all of these ClusterRoleBindings are created by the Operator?

components/pipeline-service/development/main-pipeline-service-configuration.yaml
```diff
@@ -1020,666 +333,6 @@ spec:
...
-            - ALL
-          readOnlyRootFilesystem: true
-          runAsNonRoot: true
-          seccompProfile:
-            type: RuntimeDefault
-      initContainers:
-      - args:
-        - -c
-        - |
-          mc --config-dir /tmp config host add minio "$S3_ENDPOINT" "$S3_ACCESS_KEY_ID" "$S3_SECRET_ACCESS_KEY" &&
```
aThorp96, line 1399: Ah, we used to create the bucket in this initcontainer

components/pipeline-service/development/main-pipeline-service-configuration.yaml
```diff
@@ -2199,7 +806,92 @@ spec:
...
+              spec:
+                volumes:
+                  - name: ca-s3
+                    secret:
+                      items:
+                        - key: public.crt
+                          path: s3-cert.crt
+                      secretName: storage-tls
+                containers:
+                  - name: api
```
aThorp96, line 825: Could we convert this configuration to be done via the normal TektonConfig settings such as `result.secret_name` to configure the AWS settings, and `results.logs_api: "true"` instead of `result.options.deployments.tekton-results-api.spec.template.containers[0].env["LOGS_API"]: "true"`?  If you'd rather keep the same config for 1:1 parity during the migration no objection from me, just wanted to confirm

components/pipeline-service/development/main-pipeline-service-configuration.yaml
```diff
@@ -2199,7 +806,92 @@ spec:
...
+                        cpu: 100m
+                        memory: 512Mi
+        tekton-results-watcher:
+          spec:
+            template:
+              spec:
+                containers:
+                  - name: watcher
+                    args:
+                      - -completed_run_grace_period=2h
```
aThorp96, line 876: Did we intentionally drop the flags `-api_addr <addr>`, `-auth_mode token`, and `-check_owner=false`?

aThorp96: I guess `-api_addr` probably isn't necessary anymore

components/pipeline-service/development/main-pipeline-service-configuration.yaml
```diff
@@ -2336,3 +1002,35 @@ volumes:
 - persistentVolumeClaim
 - projected
 - secret
+---
```
aThorp96, line 1005: ~Hmmm, if we're adding this back, what's the reason for dropping `base/rbac/tekton-results` from the kustomization?~ Ah we have to add this because `base/rbac/tekton-results` uses the different namespace. Could we use a kustomize patch for that instead of copying the whole thing over? We're in the process(ish) of refactoring `components/pipeline-service` into `components/pipeline-service-rd` and adding more resources to the `pipeline-service-configurtion.yaml` makes it a lot more challenging to validate the migration

components/pipeline-service/staging/base/kustomization.yaml
```diff
@@ -12,5 +12,33 @@ resources:
...
+  - ../../base/rbac/cluster-role
+  - ../../base/rbac/openshift-pipelines
+  - ../../base/rbac/tekton-logging
   - ../../base/certificates
+
+patches:
+  - target:
+      kind: ExternalSecret
+      name: tekton-results-database
+      namespace: tekton-results
```
aThorp96, line 24: Doing this for the `base/rbac/tekton-results` roles and bindings too would be great

aThorp96 (commented):
This is a huge help. Thanks @enarha !

Working through staging now, but wanted to share my comments on the development overlay in case I don't finish reviewing staging today.

components/pipeline-service/development/dev-only-pipeline-service-storage-configuration.yaml
aThorp96: Answered my own question here: https://github.com/redhat-appstudio/infra-deployments/pull/13371#discussion_r4019197628

Still need to confirm the argocd annotations though

components/pipeline-service/staging/base/main-pipeline-service-configuration.yaml
```diff
@@ -2039,7 +714,148 @@ spec:
...
+                        cpu: 5m
+                        memory: 64Mi
+        tekton-results-watcher:
+          spec:
+            template:
+              spec:
+                containers:
+                  - name: watcher
+                    args:
+                      - -check_owner=false
```
aThorp96, line 821: Any reason we dropped `-auth_mode token` here?

aThorp96 (commented):
Wrapped up reviewing staging. I'll get the hack secret creator reviewed ASAP.

aThorp96:
One concern I have is that we're losing having dedicated API instances for the watcher and for the UI/scripters. If the Konflux UI goes under heavy load, I worry it may cause API latency which would impact the watcher. Though I don't suppose there's an easy way to setup independent API instances via the TektonConfig

enarha:
> One concern I have is that we're losing having dedicated API instances for the watcher and for the UI/scripters. If the Konflux UI goes under heavy load, I worry it may cause API latency which would impact the watcher. Though I don't suppose there's an easy way to setup independent API instances via the TektonConfig

We just added support in Results for horizontal scaling. We tested with 3 Watchers and 3 API servers and it worked well. We still have to test it end-to-end when it's set through the TektonConfig, but that should be the way forward instead of the "workaround" we used earlier.

components/konflux-operator/rings/ring-0/base/cr/konflux-ui/ui.yaml
enarha: I believe that was suggested to me, but after further investigation it does not really connects to anything (it's a no-op). Removed.

components/pipeline-service/development/dev-only-pipeline-service-storage-configuration.yaml
enarha: Deleted by mistake, restored.

components/pipeline-service/development/dev-only-pipeline-service-storage-configuration.yaml
enarha: Is there still anything for me to answer/verify here?

components/pipeline-service/development/kustomization.yaml
enarha: Yes. We'll remove it with the switch in prod.

components/pipeline-service/staging/base/kustomization.yaml
enarha: Addressed.

components/pipeline-service/staging/base/main-pipeline-service-configuration.yaml
enarha: The same value is set by the operator. I checked the values from the ephemeral cluster used by the CI. BTW we recently exposed all the Watcher args through the tektonconfig. I' not sure if we using recent enough version in Konflux which includes that change, if so we'll move all args to the tektonconfig. It'll also avoid duplication of those args because args set through options are actually added to the config, not replacing the values from the config. anyway, I'll check again the actual configuration from the CI cluster and adjust further if required.

components/pipeline-service/development/main-pipeline-service-configuration.yaml
enarha: I'm switching logs_api/logs_type for now to configuration through the tektonconfig. Those are easy because the values are strings. Changing the secret_name is trickier because we also need to change the actual secret to strictly match the field names. I prefer to do that as a separate change.

components/pipeline-service/development/main-pipeline-service-configuration.yaml
enarha: I already answered that above, but I re-added -check_owner=false and the other two were already good.

components/pipeline-service/development/main-pipeline-service-configuration.yaml
enarha: I believe I did that. Please check the newest update and let me know if that's the case.

components/pipeline-service/staging/base/main-pipeline-service-configuration.yaml
enarha: Addressed above, the value is properly set by the operator.

components/pipeline-service/development/dev-only-pipeline-service-storage-configuration.yaml
aThorp96: > Is there still anything for me to answer/verify here?

Nope this looks good.

One possible race condition may be that this code used to run before the Tekton Results API started, but now it could in theory run after the API has started. It doesn't appear like that will be an issue though unless user workloads start doing stuff before the argo sync completes

aThorp96: Github isn't showing me a "Resolve conversation" button but you can consider this resolved 👍🏼

aThorp96 (approved):
/lgtm
/retest

components/konflux-operator/rings/ring-0/base/cr/konflux-ui/ui.yaml
enarha: My assumption was premature. This is actually needed. It's an attempt to fix the race condition we experience with the check for proxying the requests to the Results API. Here is the PR which introduced that field: https://github.com/konflux-ci/konflux-ci/pull/8839
I'm re-adding this and will retry to ensure it consistently fixes the issue. I believe the test was failing even with that, but let me confirm.

PR #13412, opened 2026-08-10
feat(KFLUXVNGD-1149): Add independent caching ArgoCD app and component definitions

hmariset:
Create independent artifact-registry-proxy (nginx-only) and container-image-proxy (squid-only) components to replace the monolithic squid component. Includes base, development, staging, and production/empty-base overlays with ArgoCD ApplicationSet definitions. Overlay wiring deferred to the deployment task.

Jira-Url: https://redhat.atlassian.net/browse/KFLUXVNGD-1149
Assisted-by: Claude Code

### DIFF
--- a/argo-cd-apps/overlays/rd-dev/artifact-registry-proxy/artifact-registry-proxy-appset.yaml
+++ b/argo-cd-apps/overlays/rd-dev/artifact-registry-proxy/artifact-registry-proxy-appset.yaml
@@ -1,0 +1,45 @@
+apiVersion: argoproj.io/v1alpha1
+kind: ApplicationSet
+metadata:
+  name: artifact-registry-proxy
+spec:
+  generators:
+    - merge:
+        mergeKeys:
+          - nameNormalized
+        generators:
+          - clusters:
+              selector:
+                matchLabels:
+                  appstudio.redhat.com/member-cluster: "true"
+              values:
+                sourceRoot: components/artifact-registry-proxy
+                ring: "ring-0"
+                clusterDir: base
+          - list:
+              elements: []
+
+  template:
+    metadata:
+      name: artifact-registry-proxy-{{nameNormalized}}
+    spec:
+      project: default
+      source:
+        path: '{{values.sourceRoot}}/rings/{{values.ring}}/{{values.clusterDir}}'
+        repoURL: https://github.com/redhat-appstudio/infra-deployments.git
+        targetRevision: main
+      destination:
+        namespace: artifact-registry-proxy
+        server: '{{server}}'
+      syncPolicy:
+        automated:
+          prune: true
+          selfHeal: true
+        syncOptions:
+          - CreateNamespace=true
+        retry:
+          limit: -1
+          backoff:
+            duration: 10s
+            factor: 2
+            maxDuration: 3m
--- a/argo-cd-apps/overlays/rd-dev/artifact-registry-proxy/kustomization.yaml
+++ b/argo-cd-apps/overlays/rd-dev/artifact-registry-proxy/kustomization.yaml
@@ -1,0 +1,5 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+resources:
+  - artifact-registry-proxy-appset.yaml
--- a/argo-cd-apps/overlays/rd-dev/container-image-proxy/container-image-proxy-appset.yaml
+++ b/argo-cd-apps/overlays/rd-dev/container-image-proxy/container-image-proxy-appset.yaml
@@ -1,0 +1,45 @@
+apiVersion: argoproj.io/v1alpha1
+kind: ApplicationSet
+metadata:
+  name: container-image-proxy
+spec:
+  generators:
+    - merge:
+        mergeKeys:
+          - nameNormalized
+        generators:
+          - clusters:
+              selector:
+                matchLabels:
+                  appstudio.redhat.com/member-cluster: "true"
+              values:
+                sourceRoot: components/container-image-proxy
+                ring: "ring-0"
+                clusterDir: base
+          - list:
+              elements: []
+
+  template:
+    metadata:
+      name: container-image-proxy-{{nameNormalized}}
+    spec:
+      project: default
+      source:
+        path: '{{values.sourceRoot}}/rings/{{values.ring}}/{{values.clusterDir}}'
+        repoURL: https://github.com/redhat-appstudio/infra-deployments.git
+        targetRevision: main
+      destination:
+        namespace: container-image-proxy
+        server: '{{server}}'
+      syncPolicy:
+        automated:
+          prune: true
+          selfHeal: true
+        syncOptions:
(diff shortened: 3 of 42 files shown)


argo-cd-apps/base/member/infra-deployments/artifact-registry-proxy/kustomization.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- artifact-registry-proxy.yaml
```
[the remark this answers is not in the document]
hmariset: Yes, it's not rendered/applied intentionally since we would be doing that in the next PR

argo-cd-apps/base/member/infra-deployments/artifact-registry-proxy/artifact-registry-proxy.yaml
```diff
@@ -0,0 +1,46 @@
...
+      annotations:
+        argocd.argoproj.io/sync-wave: "1"
+    spec:
+      project: default
+      source:
+        path: '{{values.sourceRoot}}/{{values.environment}}/{{values.clusterDir}}'
+        repoURL: https://github.com/redhat-appstudio/infra-deployments.git
+        targetRevision: main
+      destination:
+        namespace: nginx-proxy
```
amisstea:
```suggestion
        namespace: artifact-registry-proxy
```

`nginx-proxy` is too generic. There are multiple instances of nginx running in a Konflux cluster.

Other references will need updating as well.

argo-cd-apps/base/member/infra-deployments/container-image-proxy/container-image-proxy.yaml
```diff
@@ -0,0 +1,46 @@
...
+      annotations:
+        argocd.argoproj.io/sync-wave: "1"
+    spec:
+      project: default
+      source:
+        path: '{{values.sourceRoot}}/{{values.environment}}/{{values.clusterDir}}'
+        repoURL: https://github.com/redhat-appstudio/infra-deployments.git
+        targetRevision: main
+      destination:
+        namespace: squid-proxy
```
amisstea:
```suggestion
        namespace: container-image-proxy
```

Similar thing here. `squid-proxy` is too generic.

amisstea (changes requested):

argo-cd-apps/base/member/infra-deployments/artifact-registry-proxy/artifact-registry-proxy.yaml
hmariset: updated

argo-cd-apps/base/member/infra-deployments/container-image-proxy/container-image-proxy.yaml
hmariset: updated

components/artifact-registry-proxy/staging/caching-helm-generator.yaml
```diff
@@ -0,0 +1,57 @@
...
+    enabled: false
+  selfsigned-bundle:
+    enabled: false
+  mirrord:
+    enabled: false
+  test:
+    enabled: false
+  environment: release
+  prometheus:
+    serviceMonitor:
```
amisstea, line 54: I found a couple issues with the servicemonitors when building this overlay.

```
kustomize build --enable-helm components/artifact-registry-proxy/staging/
```

1. There are two servicemonitors when there should only be one. There's one named `squid` which shouldn't be deployed for artifact-registry-proxy.
2. In each of these SMs, there's a namespaceSelector still referencing the `caching` namespace so the SM won't be able to scrape metrics.
```
  namespaceSelector:
    matchNames:
    - caching
```

components/artifact-registry-proxy/staging/caching-helm-generator.yaml
```diff
@@ -0,0 +1,57 @@
+apiVersion: builtin
+kind: HelmChartInflationGenerator
+metadata:
+  name: caching-helm
+name: caching-helm
+repo: oci://quay.io/konflux-ci/caching
+version: 0.1.1775+d865e45
```
amisstea: Why not use the newest version available with all the namespace separation enhancements?

components/artifact-registry-proxy/staging/caching-helm-generator.yaml
```diff
@@ -0,0 +1,57 @@
...
+valuesInline:
+  squid:
+    enabled: false
+  nginx:
```
amisstea, line 11: I think we should explicitly define `nginx.namespace`/`squid.namespace` in each of the helm chart values. It works right now but relies on the implicit kustomize namespace transformer to post-process all the resources that get inflated.

amisstea (changes requested):

components/artifact-registry-proxy/staging/caching-helm-generator.yaml
hmariset: now it's fixed after updating the version, thought of updating the version once squid update is merged.

components/artifact-registry-proxy/staging/caching-helm-generator.yaml
hmariset: thought of updating the version once squid update is merged.

components/artifact-registry-proxy/staging/caching-helm-generator.yaml
hmariset: fixed

amisstea (approved):

filariow (commented):
there's no rendered diff: https://github.com/redhat-appstudio/infra-deployments/pull/13412#issuecomment-5242947069

you need to add the components to this [kustomization file](https://github.com/redhat-appstudio/infra-deployments/blob/main/argo-cd-apps/base/member/infra-deployments/kustomization.yaml) and patch-delete them from the environments in which you don't want them (prod).

@enkeefe00 does it make sense to onboard them in ring-deployments already?

enkeefe00:
> there's no rendered diff: [#13412 (comment)](https://github.com/redhat-appstudio/infra-deployments/pull/13412#issuecomment-5242947069)
> 
> you need to add the components to this [kustomization file](https://github.com/redhat-appstudio/infra-deployments/blob/main/argo-cd-apps/base/member/infra-deployments/kustomization.yaml) and patch-delete them from the environments in which you don't want them (prod).
> 
> @enkeefe00 does it make sense to onboard them in ring-deployments already?

I think so since they'll just have to reorganize these new components soon-ish otherwise.

flacatus:
are we going to deploy only to stage? I see also rd production here... If we target prod then we need ring 2,3,4 like we have in etcd-shield or authentication components

@hmariset @amisstea 

/hold

amisstea:
@hmariset let's drop the prod deployments and focus on rd-dev and rd-staging only for now.

enkeefe00 (approved):

PR #13430, opened 2026-08-11
feat(KFLUXVNGD-1230): Add lightwell-dev to internal-tenant staging cluster list

avi-biton:
The konflux-operator ApplicationSet uses a merge generator that requires an explicit list entry to map each cluster to its ring and config directory. Without this entry lightwell-dev falls through to the empty-base path and the operator is never deployed.

Add lightwell-dev with ring-1 targeting so the rendered Application points to components/konflux-operator/rings/ring-1/lightwell-dev/.


gbenhaim (approved):

yftacherzog (approved):
/approve
/lgtm

PR #13453, opened 2026-08-12
Add standalone playbook for Tekton Chains key creation

gbenhaim:
## What

- Add `hack/new-cluster/chains-key-playbook.yaml` — a standalone playbook that creates a Tekton Chains cosign signing keypair and stores it in Vault at a user-specified path
- Update `hack/new-cluster/README.md` with usage instructions for the new playbook

## Why

The full `playbook.yaml` requires many variables and runs numerous steps (GitHub app creation, ApplicationSet patching, component overlay generation). When you only need to provision a chains signing key for a new or existing cluster path, a lightweight standalone playbook avoids unnecessary complexity.

## Validation

- `ansible-playbook --syntax-check hack/new-cluster/chains-key-playbook.yaml` passes
- README updated with prerequisites and usage examples

Made with [Cursor](https://cursor.com)


gbenhaim:
I used the playbook included in this PR for creating the tekton chains key.

PR #13464, opened 2026-08-12
fix(KFLUXVNGD-1234):Add knative-eventing deployment for lightwell-dev staging cluster

avi-biton:
The knative-eventing ApplicationSet defaults to clusterDir "base" which has no corresponding directory, causing the knative-eventing-lightwell-dev application to fail with "app path does not exist". Add lightwell-dev to the per-cluster list and create its staging overlay referencing the shared base, matching the pattern used by stone-stage-p01.

Assisted-by: Cursor + Opus4.6


gbenhaim:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @gbenhaim |
| **Timestamp** | 2026-08-12T10:21:15.698Z |
| **Criteria Override** | Risk assessment not provided, Staging validation not found, PR updates multiple environments |

This PR affect only lightwell-dev
Multi env change labels are expected when changing ArgoCD app sets

PR #13500, opened 2026-08-13
fix(env-detector): attribute overlay diffs to AppSet generator environment

flacatus:
## Problem

`detectOverlayDiffs` compared full rendered kustomize YAML blobs and marked the overlay's environment whenever they differed. Since all overlays include `argo-cd-apps/base/`, adding a new ApplicationSet to base (e.g. a staging-only AppSet) caused dev, staging, and production overlays to all show a diff — spuriously labeling the PR as `multi-env` even when the change only affects staging.

## Solution

For each overlay diff, parse both HEAD and base ApplicationSets by name. For new or modified ApplicationSets, extract the `values.environment` field from the cluster generator and use that as the affected environment instead of the overlay's environment. Falls back to the overlay's environment when no explicit environment is found (preserving the `multi-env` label for truly global AppSets with `clusters: {}`).

## Test plan

- [ ] `TestDetectOverlayDiffs_BaseAppSetWithExplicitEnv`: staging AppSet added to base → only `environment/staging` marked across all overlays
- [ ] `TestDetectOverlayDiffs_BaseAppSetWithProductionEnv`: production AppSet added to base → only `environment/production` marked
- [ ] `TestDetectOverlayDiffs_MultiEnvWarning`: global AppSet (`clusters: {}`, no explicit env) → all environments marked (multi-env preserved)
- [ ] All existing detector and appset tests pass

### DIFF
--- a/infra-tools/internal/appset/parse.go
+++ b/infra-tools/internal/appset/parse.go
@@ -4,6 +4,7 @@ package appset
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"io"
 	"strings"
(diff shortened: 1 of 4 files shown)


flacatus:
cc @filariow @enkeefe00

PR #13503, opened 2026-08-13
PVO11Y-5445 Exclude mintmaker from kube_pod_* federation filters in staging

kubasikus:
## What

Exclude the `mintmaker` namespace from broad kube_pod_* federation match filters in staging.

Metrics excluded:
- `kube_pod_container_status_waiting_reason`
- `kube_pod_container_status_restarts_total`
- `kube_pod_container_info`
- `kube_pod_status_unschedulable`
- `kube_pod_status_phase`
- `kube_pod_container_status_terminated_reason`

Clusters affected: staging

[PVO11Y-5445](https://redhat.atlassian.net/browse/PVO11Y-5445)

## Why

MintMaker runs high-volume, short-lived Renovate bot pods that generate significant series churn for kube_pod_* metrics forwarded to RHOBS. No alerts or dashboards consume these pod-level metrics for the mintmaker namespace — most pod alerts (`PodNotReady`, `PodOOMKilled`, `ControllerPodOOMKilled`) already explicitly exclude it.

**Note:** Two alerts — `UnschedulablePods` and `CrashLoopBackOff` — do not currently exclude mintmaker. Both are `severity: warning` routed to the mintmaker team via `alert_routing_key`. This change will prevent those alerts from firing for mintmaker pods. MintMaker has its own dedicated controller availability alerts, so the impact should be minimal, but the mintmaker team should be aware.

## Validation

- `kustomize build --enable-helm` passes for staging federation overlay

[PVO11Y-5445]: https://redhat.atlassian.net/browse/PVO11Y-5445?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


FernandesMF:
@kubasikus, thanks for bringing this up! I would like to propose a slight change to the current solution.

We do have two groups of pods in MintMaker, the ones relative to the core service, and those from the temporary workloads. While removing the second group from the metrics won't be a problem for us, we'd prefer to keep the metrics for the first group (which fortunately is much much smaller).

Since the temporary pods always have names starting with `renovate`, could we take a more targeted approach? The idea would be to keep the namespace-level exclusion you've added, but add a companion `match[]` selector for each metric that re-includes the `mitnmaker` namespace while filtering out the `renovate` pods by name:

Container Metrics:
  - `{__name__="kube_pod_container_status_waiting_reason", namespace="mintmaker", pod!~"renovate.*"}`
  - `{__name__="kube_pod_container_status_restarts_total", namespace="mintmaker", pod!~"renovate.*"}`
  - `{__name__="kube_pod_container_info", namespace="mintmaker", pod!~"renovate.*"}`
  - `{__name__="kube_pod_container_status_terminated_reason", namespace="mintmaker", pod!~"renovate.*"}`

Pod Metrics:
  - `{__name__="kube_pod_status_unschedulable", namespace="mintmaker", pod!~"renovate.*"}`
  - `{__name__="kube_pod_status_phase", namespace="mintmaker", pod!~"renovate.*"}`

Each of these would be placed right after the corresponding modified line. Since `match[]` entries are OR'd, this brings back only the non-`renovate` pods from our namespace, and the cardinatily would still be reduced greatly.

Do you think this would work, @kubasikus? I'll be happy to adjust the approach if needed!


**Disclaimer 1**: it's important to mention that part of the load that MintMaker causes is not directlty from it's own resources, but from the CIs that it triggers in the component repos after proposing updates.

**Disclaimer 2**: assisted by Claude Opus 4.6

pacho-rh:
Including `mintmaker` but filtering out the `renovate` pods makes sense to me. Checking the metric in-cluster, doing so will be still be saving a ton on cardinality. `renovate` pods account for several thousands of series while non-`renovate` pod account for a few hundred series.

I tried updating this branch in include the additional match entries but unfortunately I do not have permission. We'll wait for @kubasikus to be back to update this PR.

kubasikus:
Ran a query to see how much reduction will happen by filtering out the renovate pod series:

image

Looks like most of it, as @pacho-rh says. So let's do it this way :)

rh-rsaar:
Will this still monitor `container_*` metrics, e.g. about mem/cpu usage and `container_cpu_cfs_throttled_periods_total` and `container_cpu_cfs_periods_total`?

kubasikus:
@rh-rsaar as mentioned in slack, yes, those metrics will still be available.

pumahaka (approved):

PR #13512, opened 2026-08-13
KFLUXINFRA-4375: Add buildah-oci-ta-min to Kyverno seccomp policy

oswcab:
The Kyverno policy was missing coverage for the buildah-oci-ta-min task variant, used by the docker-build-oci-ta-min pipeline. Without this rule, builds using the minimal pipeline would not get the seccomp, /dev/fuse, and SYS_CHROOT mutations injected.

Tested on lightwell-dev ROKS cluster — pod with label tekton.dev/task=buildah-oci-ta-min received all mutations and buildah bud completed successfully.

Assisted-by: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/components/policies/staging/lightwell-dev/buildah-roks-seccomp.yaml
+++ b/components/policies/staging/lightwell-dev/buildah-roks-seccomp.yaml
@@ -11,12 +11,13 @@ metadata:
       policy injects the Localhost unshare.json seccomp profile, /dev/fuse
       access for fuse-overlayfs, and SETFCAP + SYS_CHROOT capabilities on
       buildah task pods to enable rootless builds while keeping seccomp
-      enforcement enabled.
+      enforcement enabled. Covers buildah-oci-ta, buildah-oci-ta-min, and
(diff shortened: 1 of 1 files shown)


components/policies/staging/lightwell-dev/buildah-roks-seccomp.yaml
```diff
@@ -46,6 +47,36 @@ spec:
...
+              kinds:
+                - Pod
+              operations:
+                - CREATE
+              selector:
+                matchLabels:
+                  tekton.dev/task: buildah-oci-ta-min
+              namespaceSelector:
+                matchLabels:
+                  konflux-ci.dev/type: tenant
```
sadlerap: Since the patch we're applying is identical no matter which type of pipelinerun we're matching, we could merge these three rules into a single rule by combining their `match` sections.

oswcab: Fixed.

sadlerap (approved):

PR #13541, opened 2026-08-16
fix(KFLUXVNGD-1267): Move disaster-recovery empty-base under staging/ to match ApplicationSet path

avi-biton:
The disaster-recovery ApplicationSet resolves paths as components/disaster-recovery/{{environment}}/{{clusterDir}}, defaulting to environment=staging and clusterDir=empty-base for clusters not explicitly listed. The empty-base directory was at the wrong level (components/disaster-recovery/empty-base/) so the fallback path components/disaster-recovery/staging/empty-base/ did not exist.

This caused ArgoCD to report errors for clusters where disaster-recovery should not be deployed (e.g. lightwell-dev). Moving empty-base into the staging/ directory aligns the file layout with the ApplicationSet path template, giving non-targeted clusters a valid empty kustomization that deploys nothing.

Checks:
- `empty-base` directory lives at `components/disaster-recovery/staging/empty-base/`
- No per-cluster overlay exists at `components/disaster-recovery/staging/lightwell-dev/`
- `kustomize build components/disaster-recovery/staging/empty-base/` succeeds with zero resources

Assisted-by: Cursor (claude-4.6-opus)


manish-jangra (approved):
/approve
/lgtm

PR #13383, opened 2026-08-06
KAR-935: upgrade loki helm stack on staging

olegbet:
## Summary
-  propagates updates of the loki stack components helm chart versions
[KAR-935](https://redhat.atlassian.net/browse/KAR-935)

## Risk assessment
- Low. No actual disruption to be expected by this change.

Closes: [KAR-935](https://redhat.atlassian.net/browse/KAR-935)

## Validation
Validated on development: https://github.com/redhat-appstudio/infra-deployments/pull/13185

[KAR-935]: https://redhat.atlassian.net/browse/KAR-935?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[KAR-935]: https://redhat.atlassian.net/browse/KAR-935?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #13559, opened 2026-08-17
KONFLUX-14937: upgrade Vector tekton logs collector to 0.57.0

mathur07:
## What

- Upgrade Vector Helm chart from 0.41.0 → 0.57.0
- Upgrade Vector image from 0.45.0-distroless-libc → 0.57.0-distroless-libc
- Handle Vector 0.55+ breaking changes (remove deprecated `api.playground`)
- Handle Vector 0.57 breaking change: re-enable `${VAR}` interpolation for staging S3 sink credentials

Clusters affected: development, staging

[KONFLUX-14937](https://redhat.atlassian.net/browse/KONFLUX-14937)

## Why

CVE-2026-39197 — Denial of Service via crafted request to Vector's HTTP endpoint. While the current version (0.45.0) does not expose the vulnerable HTTP ingest sources, upgrading to the patched release (0.57.0) as defense-in-depth.

## Validation

- `kustomize build --enable-helm` passes for both development and staging overlays
- Similar upgrade validated in [#13383](https://github.com/redhat-appstudio/infra-deployments/pull/13383) for `vector-kubearchive-log-collector`

[KONFLUX-14937]: https://redhat.atlassian.net/browse/KONFLUX-14937?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/components/vector-tekton-logs-collector/development/vector-helm-values.yaml
+++ b/components/vector-tekton-logs-collector/development/vector-helm-values.yaml
@@ -12,7 +12,6 @@ customConfig:
   api:
     enabled: true
     address: 127.0.0.1:8686
-    playground: false
   sources:
     kubernetes_logs:
       type: kubernetes_logs
@@ -74,7 +73,7 @@ tolerations:
     value: konflux-tenants
 image:
   repository: quay.io/openshift-pipeline/vector
-  tag: 0.45.0-distroless-libc
+  tag: 0.57.0-distroless-libc
 securityContext:
   allowPrivilegeEscalation: false
   capabilities:
--- a/components/vector-tekton-logs-collector/staging/vector-helm-values.yaml
+++ b/components/vector-tekton-logs-collector/staging/vector-helm-values.yaml
@@ -12,7 +12,6 @@ customConfig:
   api:
     enabled: true
     address: 127.0.0.1:8686
-    playground: false
   sources:
     kubernetes_logs:
       type: kubernetes_logs
@@ -77,6 +76,10 @@ customConfig:
       filename_time_format: "-%s"
       filename_append_uuid: false
 env:
+  # Vector 0.57 disables ${VAR} interpolation by default. Restore it for
+  # BUCKET/ENDPOINT sourced from the tekton-results-s3 secret.
+  - name: VECTOR_DANGEROUSLY_ALLOW_ENV_VAR_INTERPOLATION
+    value: "true"
   - name: AWS_ACCESS_KEY_ID
     valueFrom:
       secretKeyRef:
@@ -111,7 +114,7 @@ tolerations:
     value: konflux-tenants
 image:
   repository: quay.io/openshift-pipeline/vector
-  tag: 0.45.0-distroless-libc
+  tag: 0.57.0-distroless-libc
 securityContext:
   allowPrivilegeEscalation: false
   capabilities:
--- a/components/vector-tekton-logs-collector/development/vector-helm-generator.yaml
+++ b/components/vector-tekton-logs-collector/development/vector-helm-generator.yaml
@@ -6,9 +6,9 @@ name: vector
 repo: https://helm.vector.dev
 # We mirror the image from Docker to Quay. To update the version, mirror the respective
 # image. You can use skopeo e.g
-# `skopeo copy docker://timberio/vector:0.45.0-distroless-libc docker://quay.io/openshift-pipeline/vector:0.45.0-distroless-libc`
+# `skopeo copy docker://timberio/vector:0.57.0-distroless-libc docker://quay.io/openshift-pipeline/vector:0.57.0-distroless-libc`
 # The tag obviously will differ.
-version: 0.41.0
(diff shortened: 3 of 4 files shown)


components/vector-tekton-logs-collector/development/vector-helm-values.yaml
```diff
@@ -12,7 +12,6 @@ customConfig:
   api:
     enabled: true
     address: 127.0.0.1:8686
-    playground: false
```
mathur07, line 15: Vector's API had a built-in GraphQL playground (a web UI for querying Vector's internal metrics). The `playground: false` setting was explicitly disabling it.

In Vector 0.55, the `api.playground` option was removed entirely — the playground feature no longer exists in the codebase. Keeping `playground: false` in the config would cause a validation warning or error on startup with Vector 0.57, so we removed it.

mathur07 (commented):
fyi

components/vector-tekton-logs-collector/staging/vector-helm-values.yaml
```diff
@@ -77,6 +76,10 @@ customConfig:
       filename_time_format: "-%s"
       filename_append_uuid: false
 env:
+  # Vector 0.57 disables ${VAR} interpolation by default. Restore it for
+  # BUCKET/ENDPOINT sourced from the tekton-results-s3 secret.
+  - name: VECTOR_DANGEROUSLY_ALLOW_ENV_VAR_INTERPOLATION
+    value: "true"
```
aThorp96, line 82: Minor request to tweak the comment here. Mainly just adding the reference. 

```suggestion
  # Vector 0.57 disables ${VAR} interpolation by default. Restore it for
  # ${BUCKET} and ${ENDPOINT}, which are sourced from the tekton-results-s3 secret.
  # See: https://vector.dev/docs/reference/environment_variables/
  - name: VECTOR_DANGEROUSLY_ALLOW_ENV_VAR_INTERPOLATION
    value: "true"
```

For posterity: This stuck out as a red flag but after looking into it, I believe it's the right move. The alternative is to use a ["secret" backend](https://vector.dev/docs/reference/configuration/secrets/) which doesn't seem to be very fully-featured. The documentation for how the `command` field needs to work, but it's explained in [this post](https://vector.dev/highlights/2022-07-07-secrets-management/) and [this GH issue](https://github.com/vectordotdev/vector/discussions/17255). TL;DR: we'd need to inline a script that takes a structured json object on stdin, resolves a number of secrets, then outputs a structured json blob with the result (and any errors). It seems silly complex with no provided solutions from the maintainers for common usecases like a `secretRef`. 
The [documentation also states](https://vector.dev/docs/reference/environment_variables/#security-restrictions):
>  Only enable [env interpolation] with `--dangerously-allow-env-var-interpolation` if you fully control every environment variable accessible to the Vector process and accept that environment variables may leak to users that have access to the Vector process.

Given the above, and since we do control the environment variables and the vector process is restricted, I think this is fine? The only thing i'm not sure about is "you fully control every environment variable accessible to the Vector process". Does this process run in a sidecar on to the TaskRun pods? If so, then I think users may be able to manipulate things using sidecartemplates, but I'm not sure.

mathur07:
cc @enarha

aThorp96 (commented):
/approve

olegbet (approved):
/lgtm

aThorp96:
@mathur07 given how many times tests have failed, have you had a chance to look into the failures and confirm they're not related to the change?

mathur07:
E2E is failing because quay.io/openshift-pipeline/vector:0.57.0-distroless-libc hasn't been mirrored; the DaemonSet can't pull the image, so the ArgoCD app never becomes healthy and the test times out. I'm blocked on Quay auth to push to the openshift-pipeline org

components/vector-tekton-logs-collector/staging/vector-helm-values.yaml
mathur07: No, vector-tekton-logs-collector runs as a DaemonSet, with its own pods on each node. It's not a sidecar on TaskRun pods. It collects logs by reading container log files from the node filesystem via the kubernetes_logs source. Users can't inject or manipulate env vars in the Vector pods, so "you fully control every environment variable accessible to the Vector process" holds here.

infernus01 (commented):
/lgtm

aThorp96 (approved):
/lgtm

mathur07:
Seems like it is missing approvals from
@eisraeli @oswcab

mathur07:
Or maybe @enkeefe00 
Can you please take a look? TY!

enarha (approved):
/approve

mathur07:
Follow-up: add OWNERS file pr: https://github.com/redhat-appstudio/infra-deployments/pull/13722

aThorp96 (approved):
/lgtm

PR #13585, opened 2026-08-18
fixup(PVO11Y-5384): Correct placement of managed PaC OTLP env vars

ci-operator:
Move OTEL_EXPORTER_OTLP_ENDPOINT and OTEL_TRACES_SAMPLER for
pac-controller and pac-watcher to
spec.platforms.openshift.pipelinesAsCode.options.deployments in
stone-stg-rh01, stone-stage-p01, and lightwell-dev.

### DIFF
--- a/components/pipeline-service/staging/base/main-pipeline-service-configuration.yaml
+++ b/components/pipeline-service/staging/base/main-pipeline-service-configuration.yaml
@@ -1804,11 +1804,6 @@ spec:
               spec:
                 containers:
                   - name: pac-controller
-                    env:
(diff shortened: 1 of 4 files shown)


aThorp96 (approved):
/lgtm

I think at some point we must have made a change so that `options.deployments.*` only applies to deployments owned by the respective TektonInstallerSet. In turn, this also means the resources in these templates isn't applied either 🤔 We'll need to fix that as well as a followup

image

mathur07 (approved):
/lgtm

PR #13624, opened 2026-08-20
fix(pipelines-staging): use correct resolver bucket count

aThorp96:
After https://github.com/tektoncd/pipeline/pull/10480, when resolvers are configured as a statefulset the bucket count must match the replicacount

See also: https://redhat-internal.slack.com/archives/C02CTEB3MMF/p1786608615401769
See also: https://github.com/tektoncd/operator/issues/3957


mathur07 (approved):
/lgtm

infernus01 (commented):
/lgtm

PR #13632, opened 2026-08-20
fixup(PVO11Y-5072): match MTTB spans and derive service.name from instrumentation scope

ci-operator:
Match filter/mttb on instrumentation_scope.name instead of
resource.attributes["service.name"], and add transform/mttb-service-name
to backfill service.name from the scope when it's missing or unknown.


raks-tt (approved):
/lgtm

PR #13657, opened 2026-08-21
feat(KFLUXVNGD-1289): rewrite Cargo config.json URLs to cluster proxy

kelchen123:
Extend staging nginx.subFilters on cargo-proxy config.json to replace Nexus host in dl/api with artifact-registry-proxy.caching.svc.cluster.local. Set sub_filter_once off so both URLs are rewritten.

### DIFF
--- a/components/squid/staging/caching-helm-generator.yaml
+++ b/components/squid/staging/caching-helm-generator.yaml
@@ -32,9 +32,12 @@ valuesInline:
       - location: '^/repository/cargo-proxy/config\.json$'
         types:
           - application/json
+        once: false
         filters:
           - string: '"auth-required":true'
             replacement: '"auth-required":false'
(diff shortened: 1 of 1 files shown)


components/squid/staging/caching-helm-generator.yaml
```diff
@@ -32,9 +32,14 @@ valuesInline:
       - location: '^/repository/cargo-proxy/config\.json$'
         types:
           - application/json
+        once: false
         filters:
           - string: '"auth-required":true'
             replacement: '"auth-required":false'
+          - string: '"auth-required": true'
+            replacement: '"auth-required": false'
```
amisstea: We should be able to combine these into a single rule since `string` takes a regex.

amisstea (approved):

PR #13704, opened 2026-08-25
chore(KFLUXVNGD-1158): federate squid_up from container-image-proxy namespace in staging

hmariset:
Add squid_up metric federation for the new container-image-proxy namespace, matching the existing pattern for the caching namespace. The container-image-proxy ring-1 rollout deploys squid to a dedicated namespace; without this entry the central Prometheus does not scrape its health metric.

Jira-Url: https://redhat.atlassian.net/browse/KFLUXVNGD-1158
Authored-by: Claude Code (claude-sonnet-4-6)

### DIFF
--- a/components/monitoring/prometheus/staging/base/federation/endpoints-params.yaml
+++ b/components/monitoring/prometheus/staging/base/federation/endpoints-params.yaml
@@ -116,6 +116,9 @@
     # Namespace: caching
     - '{__name__="squid_up", namespace="caching"}'
 
+    # Namespace: container-image-proxy
+    - '{__name__="squid_up", namespace="container-image-proxy"}'
+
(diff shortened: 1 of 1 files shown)


pacho-rh:
Hello, this change is not needed since looks like the `squid_up` series is collected by User Workload Monitoring (UWM) Prometheus which automatically forwards any metrics it collects to RHOBS. This also applies to `'{__name__="squid_up", namespace="caching"}'` but no need to clean that up in this PR. The configuration file is for Platform Prometheus which uses it as a whitelist for series it should forward to RHOBS.

hmariset:
this change is not needed to forward squid_up metric to RHOBS, so closing the PR

PR #13745, opened 2026-08-27
KFLUXINFRA-4375 allow 172.24.0.0/14 pod CIDR in ESO webhook network p…

oswcab:
…olicies

Add 172.24.0.0/14 to the allow-webhook-ingress and allow-health-checks-ingress NetworkPolicies in both staging and production so that clusters using this pod CIDR can reach the ESO validating webhook.


hugares (approved):
/lgtm

oswcab:
qodo: ESO bypasses rollout rings 
On purpose on this PR, not an issue

hugares:
This change is targeting both staging and prod and this is intentional, this change is very low risk and we want this on prod ASAP, needed for LW

PR #13748, opened 2026-08-27
(onboarding) Migrate production etcd-defrag AppSet to ring deployments ArgoCD

celeztyne:
### Summary of Changes
- Creation of rings structure for production clusters according [directory structure documentation](https://github.com/redhat-appstudio/infra-deployments/blob/main/docs/ring-deployments/directory-layout.md) and copying/migration of etcd-defrag cluster configuration

This is affecting rings 2,3 and 4 as defined in the [directory structure documentation](https://github.com/redhat-appstudio/infra-deployments/blob/main/docs/ring-deployments/directory-layout.md)
** For ring 2 kflux-fedora-01, kflux-lw-p01, kflux-ocp-p01, kflux-osp-p01k, flux-prd-rh03, kflux-rhel-p01 and stone-prod-p01

### Validation
- kustomize build has been performed on the current configuration (config/etcd-defrag/production) the new (config/etcd-defrag-rd/rings/rings{2,3,4}) directories to render config files, giving the same output for all clusters, with exception of the cronjob that has some improvements suggested by reviewers (mostly security-related)
- Also kustomize build argo-cd-apps/overlays/rd-production output ApplicationSet with the rings list elements ring-{2,3,4} for etcd-defrag

### Verification
The changes have been implemented in Staging in PR https://github.com/redhat-appstudio/infra-deployments/pull/13482

### Risk Level
**Risk Level**: Medium
**Reasoning**: No changes in current platforms, the new directories are a new structure. Changes will be done in a controlled form and have been already tested in staging. In any case, the rendering (using oc kustomize) shown that the output is identical to current one (see Validation section above) 
**Rollback Strategy**: Reversing this PR + manual sync in the current/old repository-validator ApplicationSet will make the structure back as before changes

[KFLUXINFRA-4290](https://redhat.atlassian.net/browse/KFLUXINFRA-4290)

[KFLUXINFRA-4290]: https://redhat.atlassian.net/browse/KFLUXINFRA-4290?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #13773, opened 2026-08-31
KFLUXINFRA etcd-defrag-rd: add kargo-promotion-image for ring-1

flacatus:
The etcd-defrag CronJob uses the quay.io/konflux-ci/etcd-defrag image as an env var (IMAGE), not as a container image. Kustomize's images: transformer only works on container image: fields, so kustomize-set-image can't update it.

This adds a kargo-promotion-image.yaml file — a strategic merge patch that overrides the IMAGE env var in the CronJob. Kargo updates this file via yaml-update during ring promotions, while humans continue to modify the CronJob structure through base-snapshot/ as usual.

Why a separate file instead of inline patch? Kargo's yaml-update can target fields in a real YAML file but can't modify values inside an inline YAML string (patch: |-).

### DIFF
--- a/configs/etcd-defrag-rd/rings/ring-1/base/kargo-promotion-image.yaml
+++ b/configs/etcd-defrag-rd/rings/ring-1/base/kargo-promotion-image.yaml
@@ -1,0 +1,14 @@
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: etcd-maintenance
+spec:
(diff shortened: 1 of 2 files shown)


enkeefe00:
I think this can be closed in favor of the comment I put [here](https://github.com/redhat-appstudio/infra-deployments/pull/13748) regarding images.

PR #13787, opened 2026-09-01
feat(PVO11Y-5503): Push kanary probe-run artifacts to S3 after each run

FaisalAl-Rayes:
# PVO11Y-5503: Push kanary probe-run artifacts to S3

Upload kanary probe test results to S3 after each run to enable downstream analysis and long-term storage.

## Changes (staging only)

- Add S3 upload step to `run-kanary-tests` task
- Upload artifacts tarball to `s3://konflux-perfscale-artifacts/run-probe/<cluster>/<test-type>/run-probe-<cluster>-<test-type>-<timestamp>.tar.gz`
- Add `s3-upload-failed` result and metric to track upload failures
- Configure all staging cronjobs to use `konflux-perfscale-artifacts` bucket
- AWS credentials sourced from existing `kanary-probe-credentials` secret (keys: `aws_access_key_id`, `aws_secret_access_key`, `aws_region`)
- Remove empty external-secrets kustomization from stone-stage-p01

## S3 Path Structure

```
s3://konflux-perfscale-artifacts/
  └── run-probe/                                    # Loadtest type
      └── <cluster>/                                # Cluster name
          └── <test-type>/                          # Test type (container-single-arch, etc.)
              └── run-probe-<cluster>-<test-type>-<timestamp>.tar.gz
```

**Example:**
```
s3://konflux-perfscale-artifacts/run-probe/lightwell-dev/container-single-arch/run-probe-lightwell-dev-container-single-arch-20260914T103045Z.tar.gz
```

The filename is self-documenting with all context embedded, making it easy to organize downloads locally.

## Behavior

- Upload is **non-fatal** - task succeeds even if S3 upload fails
- Upload failures are tracked via `s3-upload-failed` result and `kanary_error` metric
- S3 upload can be disabled per-cronjob by omitting `--param s3-bucket`

## Validation

- Staging kanary runs will upload artifacts to S3 on next execution
- Monitor `kanary_error{phase="upload-results-to-s3"}` for upload failures

## Risk Assessment

**Level:** Low

**What could go wrong:**
- S3 upload failures could fill TaskRun logs, but upload is non-fatal so kanary tests continue normally
- Missing/invalid AWS credentials would cause upload failures tracked via `kanary_error` metric
- Large artifact volumes could increase S3 storage costs (mitigated by lifecycle policies on bucket)

**Rollback plan:**
- Revert PR - removes S3 upload step from pipeline
- Or disable per-cluster by removing `--param s3-bucket` from specific cronjobs

**Blast radius:**
- Staging only (lightwell-dev, stone-stage-p01, stone-stg-rh01)
- Kanary probe functionality unaffected - upload is non-fatal
- No impact to existing metrics or test results


FaisalAl-Rayes:
@jhutar PTAL

components/monitoring/kanary/staging/base/tasks/run-kanary-tests.yaml
```diff
@@ -346,3 +353,72 @@ spec:
...
+          value: $(params.tested-cluster)
+        - name: TEST_TYPE
+          value: $(params.test-type)
+        - name: S3_BUCKET
+          value: $(params.s3-bucket)
+        - name: AWS_ACCESS_KEY_ID
+          valueFrom:
+            secretKeyRef:
+              name: kanary-probe-credentials
+              key: aws_access_key_id
```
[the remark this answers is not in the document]
FaisalAl-Rayes: This is fine. the keys should be there. the failing of upload should be due to reasons other than "keys not present in the secrets" (since we should have them there, they are provided by perf&scale team)

components/monitoring/kanary/staging/base/tasks/run-kanary-tests.yaml
```diff
@@ -346,3 +353,72 @@ spec:
         else
           echo "WARN: load-test.json not found or empty, failure-cause and failure-phase will be empty" >&2
         fi
+    - name: upload-to-s3
+      image: quay.io/redhat-user-workloads/konflux-perfscale-tenant/loadtest
```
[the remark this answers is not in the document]
FaisalAl-Rayes: this is true, fixing it by allowing the failure of the test specifically (`|| true`)and proceeding so that we can upload whatever artifacts (if any) that were created during the test run.

We keep `set -e` in case other failures happen that we should consider as failures in the kanary pipeline (this would be emitted in the kanary_error metric as part of the finally task of pushing metrics)

components/monitoring/kanary/staging/base/tasks/run-kanary-tests.yaml
```diff
@@ -346,3 +353,72 @@ spec:
...
+        echo "Creating tarball from /tmp/artifacts/results/"
+        if ! tar -czf "${tarball}" -C /tmp/artifacts/results .; then
+          fail_upload "failed to create tarball ${tarball}"
+        fi
+
+        echo "Uploading to s3://${S3_BUCKET}/${s3_key}"
+        if ! python3 ci-scripts/s3-artifacts.py push \
+          --bucket "${S3_BUCKET}" \
+          --local "${tarball}" \
+          --remote "${s3_key}"; then
```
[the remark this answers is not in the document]
FaisalAl-Rayes: this seems to be not correct. the file that i am using is [s3-artifacts.py](https://github.com/konflux-ci/loadtest/blob/df3a8bca08bdeb6bb600b6cbee5cff4d2f3e492f/ci-scripts/s3-artifacts.py)

components/monitoring/kanary/staging/base/tasks/run-kanary-tests.yaml
```diff
@@ -294,6 +300,7 @@ spec:
         echo -n "unknown" > $(results.test-outcome.path)
         echo -n "" > $(results.failure-cause.path)
         echo -n "" > $(results.failure-phase.path)
+        echo -n "false" > $(results.s3-upload-failed.path)
```
kubasikus: Why is this result defined here and not at the start of the `upload-to-s3` step?

components/monitoring/kanary/staging/base/tasks/run-kanary-tests.yaml
```diff
@@ -310,8 +310,8 @@ spec:
           --arg NAMESPACE "$TENANT_NAMESPACE" \
           --rawfile USER_TOKEN /opt/users-secret/token \
           '[{apiurl: $SERVER_API_URL, namespace: $NAMESPACE, token: ($USER_TOKEN | rtrimstr("\n")), verified: true}]' > users.json
-        ./ci-scripts/run-probe/run.sh
-        cp -f started ended /tmp/artifacts/
+        ./ci-scripts/run-probe/run.sh || true
+        cp -f started ended /tmp/artifacts/ || true
```
kubasikus, line 313: Could we put `onError: continue` on this step? That way, we could guarantee that subsequent steps always complete, no matter what happened here.

FaisalAl-Rayes: if i understand correctly that would mean anything failing in the step it would continue nonetheless.

if `jq` step fails before it should fail the step and move to the finally task but if the running of the test itself fails that is what we want to not fail the step on.

components/monitoring/kanary/staging/base/tasks/run-kanary-tests.yaml
FaisalAl-Rayes: so that we can always have it as false but this is actually fair since i did set the default to false when defined the parameter. I will do this change while still keeping it in`upload-to-s3` but moving it to the beginning of that step. (for the sake of clarity despite the default value in the params)

change done in: 7c40c850fdc71921419a031bbf1b0658c68d0b71

components/monitoring/kanary/staging/base/tasks/run-kanary-tests.yaml
kubasikus: Oh, I see, then my comment doesn't make sense :)

PR #13794, opened 2026-09-01
PVO11Y-5384 bump release-service dev and staging to 8666f413

ci-operator:
Bumps both development and staging to include the tracing code from
konflux-ci/release-service#1544. Dev overlay added per review feedback.

### DIFF
--- a/components/release/development/kustomization.yaml
+++ b/components/release/development/kustomization.yaml
@@ -3,13 +3,13 @@ kind: Kustomization
 resources:
   - ../base
   - ../base/monitor/development
-  - https://github.com/konflux-ci/release-service/config/default?ref=0168a88a95b51319a514dedd495cd8fa1e6eaea0
+  - https://github.com/konflux-ci/release-service/config/default?ref=8666f413d8c787ac69d32a861cf855b3b7d58449
   - release_service_config.yaml
 
 images:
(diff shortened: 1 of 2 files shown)


seanconroy2021:
@johnbieren Can we not do this with the promotion script, as we don't know what is being promoted? 
https://github.com/konflux-ci/release-service-utils/tree/main/ci/promote-overlay

The dev environment hasn't been updated in 2 months, so we don't even have https://github.com/konflux-ci/release-service/commit/8666f413d8c787ac69d32a861cf855b3b7d58449 in development so we need promoted into that first. https://github.com/ci-operator/infra-deployments/blob/main/components/release/development/kustomization.yaml

johnbieren:
> @johnbieren Can we not do this with the promotion script, as we don't know what is being promoted?
> 
> The dev environment hasn't been updated in 2 months, so we don't even have [konflux-ci/release-service@8666f41](https://github.com/konflux-ci/release-service/commit/8666f413d8c787ac69d32a861cf855b3b7d58449) in development so we need promoted into that first.

This obviously comes from outside our team, so I get why the script wasn't used. Good catch though that development should definitely be promoted first

johnbieren:
Asked on slack for the development overlay to be promoted first

PR #13808, opened 2026-09-01
KFLUXINFRA-4426 Distribute RH internal CA to kflux-lw-p01 via trust-m…

oswcab:
…anager

## Summary

- Add a `roks-trusted-ca` component that uses trust-manager to distribute the Red Hat internal CA bundle to tenant namespaces on ROKS HCP clusters
- Patch the `ca-bundle` ApplicationSet for production to enable per-cluster overlays and route `kflux-lw-p01` to its overlay in `internal-infra-deployments`
- Disable the Kyverno `create-trusted-ca-configmap` rule on kflux-lw-p01 (changes selector to `disabled-on-roks`)

### Why

On ROKS HCP clusters the Proxy CR `trustedCA` field cannot be set (blocked by ValidatingAdmissionPolicy), so OpenShift's `inject-trusted-cabundle` mechanism does not distribute the RH internal CA. Build tasks that access internal RH services (e.g. `gitlab.cee.redhat.com`) over TLS fail certificate verification.

### What this does

A trust-manager `Bundle` resource creates a ConfigMap named `trusted-ca` with key `ca-bundle.crt` in every namespace labeled `konflux-ci.dev/type=tenant`. This matches the build task's default `caTrustConfigMapName` parameter so no workload changes are needed.

The Bundle also sets `config.openshift.io/inject-trusted-cabundle: "false"` on managed ConfigMaps to prevent OpenShift's CA injection from overwriting trust-manager's content. This handles existing ConfigMaps created by the Kyverno rule (which has `synchronize: false`, so the old ConfigMaps remain after the rule is disabled). trust-manager's continuous reconciliation loop converges within seconds of deployment — any brief race with OpenShift injection is transient and self-healing.

### How the pieces fit together

| Component | Repo | What it does |
|-----------|------|--------------|
| `ca-bundle` overlay (`production/kflux-lw-p01/`) | `internal-infra-deployments` ([PR #139](https://github.com/redhat-appstudio/internal-infra-deployments/pull/139)) | Deploys the existing `user-ca-bundle` ConfigMap to `cert-manager` namespace (renamed `rh-internal-ca`, Proxy CR removed) |
| `ca-bundle` ApplicationSet patch | this PR | Routes `kflux-lw-p01` to its per-cluster overlay, sets `clusterDir: base` for production |
| `roks-trusted-ca` component | this PR | trust-manager Bundle that references `rh-internal-ca` and distributes it to tenant namespaces |
| Kyverno policy patch | this PR | Disables the `create-trusted-ca-configmap` rule on `kflux-lw-p01` |

The `ca-bundle` ApplicationSet patch uses a dedicated `ca-bundle-production-overlay-patch.yaml` instead of the shared `production-overlay-patch.yaml` because it needs additional ops (`clusterDir: base` and list element) that are specific to `ca-bundle` and would break other ApplicationSets.

A standalone `roks-trusted-ca` component is used instead of extending the `trust-manager` component to avoid Helm chart download conflicts in CI when multiple kustomize overlays reference the same `HelmChartInflationGenerator` base.

### Dependency

[internal-infra-deployments#139](https://github.com/redhat-appstudio/internal-infra-deployments/pull/139) must merge first as the source ConfigMap (`rh-internal-ca` in `cert-manager` namespace) must exist before the trust-manager Bundle references it.

### Render-diff noise

The render-diff report shows build errors for `empty-base` and diffs for `kflux-lw-p01` tagged with all three environments. This is expected for a new component:

- **`empty-base` BUILD ERROR**: Intentionally empty (`resources: []`) exists so non-ROKS clusters get a valid no-op ArgoCD Application. The render-diff tool reports "does not exist on either ref" because there is no rendered output.
- **`kflux-lw-p01` showing for development/staging**: The render-diff tool renders every discovered `kustomization.yaml` for all environments. Only the production path is actually deployed.
- **`policies/production/kflux-lw-p01`**: The real change: Kyverno rule selector from `tenant` to `disabled-on-roks`.

### Kyverno rule clarification

The policy `bootstrap-tenant-namespace-rbcm` contains a single rule (`create-trusted-ca-configmap`) that generates an empty `trusted-ca` ConfigMap so it is not an RBAC policy despite the name. The `rbcm` suffix refers to the original multi-rule policy it was split from. trust-manager's Bundle replaces this rule's function.

## Risk Assessment

**Risk Level:** Low
**What could go wrong:** The trust-manager Bundle fails to reconcile, leaving tenant namespaces without the `trusted-ca` ConfigMap (or with the old empty Kyverno-generated one). Build tasks on kflux-lw-p01 would fail TLS verification against `gitlab.cee.redhat.com` — which is the current broken state already. This change cannot make things worse; it can only fix or leave unchanged.
**Rollback:** Revert this PR. The Kyverno rule re-enables (`disabled-on-roks` → `tenant`), Kyverno recreates empty `trusted-ca` ConfigMaps (`generateExisting: true`), and the cluster returns to the current (broken) state. trust-manager Bundle deletion is handled by ArgoCD prune. No manual cleanup needed.
**Blast radius:** Production kflux-lw-p01 only. No other clusters are affected — the `roks-trusted-ca` ApplicationSet only lists `kflux-lw-p01`, and the `ca-bundle` patch only routes `kflux-lw-p01` to its per-cluster overlay. All other production clusters continue using `ca-bundle/production/base/` (unchanged behavior).

## Test plan

- [X] Validated trust-manager Bundle + CA distribution end-to-end on throwaway ROKS cluster (`da6nnjlw00gukspnspd0`)
- [ ] Verify `rh-internal-ca` ConfigMap exists in `cert-manager` namespace (from [internal-infra-deployments#139](https://github.com/redhat-appstudio/internal-infra-deployments/pull/139))
- [ ] Verify `trusted-ca` ConfigMap appears in tenant namespaces on kflux-lw-p01 with RH internal CA content
- [ ] Verify `config.openshift.io/inject-trusted-cabundle` label is set to `false` on managed ConfigMaps
- [ ] Verify build task can mount `trusted-ca` and `update-ca-trust` makes the RH Internal Root CA a trusted anchor
- [ ] Verify Kyverno no longer generates empty `trusted-ca` ConfigMaps in new tenant namespaces
- [ ] Post-rollout: assert every tenant namespace has a trust-manager-managed ConfigMap with the correct label and content:
  ```bash
  TENANT_NS=$(oc get ns -l konflux-ci.dev/type=tenant -o jsonpath='{.items[*].metadata.name}')
  MANAGED_NS=$(oc get cm trusted-ca -A -l trust.cert-manager.io/bundle=trusted-ca \
    -o jsonpath='{.items[?(@.metadata.labels.config\.openshift\.io/inject-trusted-cabundle=="false")].metadata.namespace}')
  diff <(echo "$TENANT_NS" | tr ' ' '\n' | sort) <(echo "$MANAGED_NS" | tr ' ' '\n' | sort)
  ```
  Empty diff = all tenant namespaces have the correctly labeled, trust-manager-managed ConfigMap.

Contributes to: [KFLUXINFRA-4426](https://redhat.atlassian.net/browse/KFLUXINFRA-4426)

[KFLUXINFRA-4426]: https://redhat.atlassian.net/browse/KFLUXINFRA-4426?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Depends on: https://github.com/redhat-appstudio/infra-deployments/pull/13833

### DIFF
--- a/argo-cd-apps/base/ca-bundle/ca-bundle.yaml
+++ b/argo-cd-apps/base/ca-bundle/ca-bundle.yaml
@@ -17,7 +17,9 @@ spec:
                 matchLabels:
                   appstudio.redhat.com/internal-member-cluster: "true"
           - list:
-              elements: []
+              elements:
+                - nameNormalized: kflux-lw-p01
+                  values.clusterDir: kflux-lw-p01
   template:
     metadata:
       name: ca-bundle-{{nameNormalized}}
--- a/argo-cd-apps/base/cert-manager-trusted-ca/cert-manager-trusted-ca.yaml
+++ b/argo-cd-apps/base/cert-manager-trusted-ca/cert-manager-trusted-ca.yaml
@@ -1,0 +1,57 @@
+# Distributes the RH internal CA bundle to tenant namespaces on HCP
+# clusters via trust-manager. Only deployed to clusters listed below —
+# clusters where the Proxy CR trustedCA cannot be set.
+#
+# This is a standalone component (not part of trust-manager) to avoid
+# Helm chart download conflicts in CI when multiple kustomize overlays
+# reference the same base with a HelmChartInflationGenerator.
+#
+# To add a new cluster: add an entry to the list generator below
+# and create a matching directory under components/cert-manager-trusted-ca/.
+apiVersion: argoproj.io/v1alpha1
+kind: ApplicationSet
+metadata:
+  name: cert-manager-trusted-ca
+spec:
+  generators:
+    - merge:
+        mergeKeys:
+          - nameNormalized
+        generators:
+          - clusters:
+              selector:
+                matchLabels:
+                  appstudio.redhat.com/internal-member-cluster: "true"
+              values:
+                sourceRoot: components/cert-manager-trusted-ca
+                environment: staging
+                clusterDir: empty-base
+          # Explicit list of clusters that need the CA bundle.
+          - list:
+              elements:
+                - nameNormalized: kflux-lw-p01
+                  values.clusterDir: kflux-lw-p01
+  template:
+    metadata:
+      name: cert-manager-trusted-ca-{{nameNormalized}}
+    spec:
+      project: default
+      source:
+        path: '{{values.sourceRoot}}/{{values.environment}}/{{values.clusterDir}}'
+        repoURL: https://github.com/redhat-appstudio/infra-deployments.git
+        targetRevision: main
+      destination:
+        namespace: cert-manager
+        server: '{{server}}'
+      syncPolicy:
+        automated:
+          prune: true
+          selfHeal: true
+        syncOptions:
+          - CreateNamespace=false
+        retry:
+          limit: -1
+          backoff:
+            duration: 10s
+            factor: 2
+            maxDuration: 3m
--- a/components/cert-manager-trusted-ca/production/kflux-lw-p01/rh-internal-ca-bundle.yaml
+++ b/components/cert-manager-trusted-ca/production/kflux-lw-p01/rh-internal-ca-bundle.yaml
@@ -1,0 +1,43 @@
+# Workaround for ROKS HCP where the Proxy CR trustedCA cannot be set.
+#
+# On ROSA clusters, OpenShift's inject-trusted-cabundle mechanism distributes
+# the RH internal CA via the Proxy CR. On ROKS HCP this is blocked, so we use
+# trust-manager to distribute the CA bundle to tenant namespaces instead.
+#
+# The Bundle creates a ConfigMap named "trusted-ca" with key "ca-bundle.crt"
+# in every namespace labeled konflux-ci.dev/type=tenant — matching the default
+# caTrustConfigMapName parameter in build tasks. No workload changes needed.
+#
+# The source ConfigMap (rh-internal-ca) containing the CA certificates is
+# deployed from internal-infra-deployments (private repo) via the ca-bundle
+# component's kflux-lw-p01 overlay.
+#
+# The corresponding Kyverno rule (create-trusted-ca-configmap in
+# bootstrap-tenant-namespace-rbcm) is disabled on this cluster via a Kustomize
+# patch in the policies overlay to avoid conflict.
+#
+# Can be removed once IBM supports setting trustedCA on HostedCluster or the
+# ibmcloud CLI supports trusted CA bundles (IDEA-I-4688).
+apiVersion: trust.cert-manager.io/v1alpha1
+kind: Bundle
+metadata:
+  name: trusted-ca
+spec:
+  sources:
+    - configMap:
+        name: rh-internal-ca
+        key: ca-bundle.crt
+  target:
+    configMap:
+      key: ca-bundle.crt
+      metadata:
+        labels:
+          # Disable OpenShift's inject-trusted-cabundle on ROKS. The existing
+          # Kyverno-generated ConfigMaps have this label set to "true", which
+          # causes OpenShift to overwrite ca-bundle.crt with system CAs only
+          # (Proxy CR trustedCA is not set on HCP). Setting it to "false"
+          # stops the injection and lets trust-manager manage the content.
+          config.openshift.io/inject-trusted-cabundle: "false"
+    namespaceSelector:
+      matchLabels:
+        konflux-ci.dev/type: tenant
--- a/argo-cd-apps/base/cert-manager-trusted-ca/kustomization.yaml
+++ b/argo-cd-apps/base/cert-manager-trusted-ca/kustomization.yaml
@@ -1,0 +1,4 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- cert-manager-trusted-ca.yaml
--- a/argo-cd-apps/overlays/production-downstream/kustomization.yaml
+++ b/argo-cd-apps/overlays/production-downstream/kustomization.yaml
@@ -8,6 +8,7 @@ resources:
   - ../../base/monitoring-blackbox
   - ../../base/smee-client
   - ../../base/ca-bundle
+  - ../../base/cert-manager-trusted-ca
   - ../../base/repository-validator
   - ../../base/cluster-secret-store-rh
 patchesStrategicMerge:
@@ -401,3 +402,8 @@ patches:
       kind: ApplicationSet
       version: v1alpha1
       name: release
+  - path: production-overlay-patch.yaml
+    target:
+      kind: ApplicationSet
+      version: v1alpha1
+      name: cert-manager-trusted-ca
--- a/components/cert-manager-trusted-ca/production/empty-base/kustomization.yaml
+++ b/components/cert-manager-trusted-ca/production/empty-base/kustomization.yaml
@@ -1,0 +1,4 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+# No resources — this component only applies to ROKS clusters.
+resources: []
--- a/components/cert-manager-trusted-ca/production/kflux-lw-p01/kustomization.yaml
+++ b/components/cert-manager-trusted-ca/production/kflux-lw-p01/kustomization.yaml
@@ -1,0 +1,4 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- rh-internal-ca-bundle.yaml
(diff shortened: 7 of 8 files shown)


components/cert-manager-trusted-ca/development/empty-base/kustomization.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
```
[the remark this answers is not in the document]
oswcab: `development/empty-base/` is currently unreachable since no development overlay patches the roks-trusted-ca ApplicationSet. It's intentional defensive scaffolding: when a development overlay is eventually added (setting environment: development), the path components/roks-trusted-ca/development/empty-base/ will resolve correctly without requiring a follow-up PR to create the directory. This follows the same pattern used by other components in this repo (e.g., kyverno/staging/empty-base/, konflux-ui/production/empty-base/). Removing it would save one 4-line file but create a future breakage if someone adds the overlay without remembering to also create the directory.

oswcab:
qodo: Default overlay path remains unset 🐞 Bug ≡ Correctness

The base ca-bundle ApplicationSet already has clusterDir: base (set by PR #13833, merged). Non-overridden clusters resolve to components/ca-bundle/production/base/ which is the correct path after the production restructuring in internal-infra-deployments#139. The production overlay patch only needs to set environment: production and add kflux-lw-p01 to the list as clusterDir is inherited from the base.

argo-cd-apps/base/ca-bundle/ca-bundle.yaml
```diff
@@ -17,7 +17,10 @@ spec:
                 matchLabels:
                   appstudio.redhat.com/internal-member-cluster: "true"
           - list:
-              elements: []
+              elements:
+                - nameNormalized: kflux-lw-p01
+                  values.clusterDir: kflux-lw-p01
+                  values.environment: production
```
hugares: no need for "environement: production", remove that

argo-cd-apps/base/member/infra-deployments/cert-manager-trusted-ca/cert-manager-trusted-ca.yaml
```diff
@@ -0,0 +1,58 @@
...
+spec:
+  generators:
+    - merge:
+        mergeKeys:
+          - nameNormalized
+        generators:
+          - clusters:
+              selector:
+                matchLabels:
+                  appstudio.redhat.com/member-cluster: "true"
```
hugares: I think here we can target only internal cluster


```suggestion
                  appstudio.redhat.com/internal-member-cluster: "true"
```

argo-cd-apps/base/member/infra-deployments/cert-manager-trusted-ca/cert-manager-trusted-ca.yaml
```diff
@@ -0,0 +1,58 @@
...
+              values:
+                sourceRoot: components/cert-manager-trusted-ca
+                environment: staging
+                clusterDir: empty-base
+          # Explicit list of clusters that need the CA bundle.
+          - list:
+              elements:
+                - nameNormalized: kflux-lw-p01
+                  values.clusterDir: kflux-lw-p01
+                  values.environment: production
```
hugares: no need for "environement: production", remove that

argo-cd-apps/base/cert-manager-trusted-ca/cert-manager-trusted-ca.yaml
hugares, line 1: all other applicationSet that we target only internal member are not in the member folder. Let's do the same here so move this "cert-manager-trusted-ca" folder to be a sibling of ca-bundle, i.e. github.com/redhat-appstudio/infra-deployments/argo-cd-apps/base/cert-manager-trusted-ca

you will also need to include this directly in the production-downstream overlay file:

```
diff --git a/argo-cd-apps/overlays/production-downstream/kustomization.yaml b/argo-cd-apps/overlays/production-downstream/kustomization.yaml
index 824937a1e..d8ff699e6 100644
--- a/argo-cd-apps/overlays/production-downstream/kustomization.yaml
+++ b/argo-cd-apps/overlays/production-downstream/kustomization.yaml
@@ -10,6 +10,7 @@ resources:
   - ../../base/ca-bundle
   - ../../base/repository-validator
   - ../../base/cluster-secret-store-rh
+  - ../../base/cert-manager-trusted-ca
 patchesStrategicMerge:
   - delete-applications.yaml
 namespace: argocd
@@ -299,6 +300,11 @@ patches:
       kind: ApplicationSet
       version: v1alpha1
       name: monitoring-cardinality
+  - path: production-overlay-patch.yaml
+    target:
+      kind: ApplicationSet
+      version: v1alpha1
+      name: cert-manager-trusted-ca
   # Exclude operator-managed clusters from Argo generation for services owned
   # by the konflux-operator (ring-2). Orphan (do not prune) when Applications go away.
   - path: exclude-operator-owned-clusters.yaml
```

components/cert-manager-trusted-ca/development/empty-base/kustomization.yaml
hugares, line 1: since we move the appset out of member, this development folder is not needed

components/cert-manager-trusted-ca/production/kflux-lw-p01/rh-internal-ca-bundle.yaml
```diff
@@ -0,0 +1,43 @@
...
+# ibmcloud CLI supports trusted CA bundles (IDEA-I-4688).
+apiVersion: trust.cert-manager.io/v1alpha1
+kind: Bundle
+metadata:
+  name: trusted-ca
+spec:
+  sources:
+    - configMap:
+        name: rh-internal-ca
+        key: ca-bundle.crt
```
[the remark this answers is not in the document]
oswcab: The Bundle intentionally contains only the RH internal CAs, not system roots. Build tasks mount `trusted-ca`, copy it to `/etc/pki/ca-trust/source/anchors/`, and run `update-ca-trust extract`, which merges the custom CA with the system CA bundle already present in the container image. The result is a trust store containing both system roots (for public endpoints like GitHub) and the RH internal CA (for `gitlab.cee.redhat.com`). This is the same mechanism used on ROSA clusters today and was validated  on a throwaway ROKS cluster.

components/cert-manager-trusted-ca/staging/empty-base/kustomization.yaml
hugares, line 1: since we move the appset out of member, this staging folder is not needed

argo-cd-apps/base/ca-bundle/ca-bundle.yaml
oswcab: Done.

argo-cd-apps/base/member/infra-deployments/cert-manager-trusted-ca/cert-manager-trusted-ca.yaml
oswcab: Done.

argo-cd-apps/base/member/infra-deployments/cert-manager-trusted-ca/cert-manager-trusted-ca.yaml
oswcab: Done.

components/cert-manager-trusted-ca/development/empty-base/kustomization.yaml
oswcab: Done.

components/cert-manager-trusted-ca/staging/empty-base/kustomization.yaml
oswcab: Done.

argo-cd-apps/base/cert-manager-trusted-ca/cert-manager-trusted-ca.yaml
oswcab: Done

hugares (approved):
/lgtm

manish-jangra (approved):
/approve
/lgtm

eedri:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @eedri |
| **Timestamp** | 2026-09-07T11:17:57.845Z |
| **Risk Level** | low |

PR #13721, opened 2026-08-26
fixup(PVO11Y-5384): move PaC resource limits to pipelinesAsCode.options

ci-operator:
## What

Add resource limits for pac-controller, pac-watcher, and pac-webhook
to pipelinesAsCode.options.deployments where the operator applies them.
Remove the identical entries from pipeline.options.deployments where
they had no effect.

Clusters affected: all staging

[PVO11Y-5384](https://issues.redhat.com/browse/PVO11Y-5384)

## Why

PaC deployments are managed under pipelinesAsCode.options, not
pipeline.options. The resource limits from #7262 were in the wrong
section and silently dropped by the operator.

## Validation

- kustomize build passes for all staging overlays

[PVO11Y-5384]: https://redhat.atlassian.net/browse/PVO11Y-5384?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #13827, opened 2026-09-02
Increase pac-controller memory limit to 4Gi

infernus01:
After PR #13721, the pipelines-as-code-controller pods are hitting OOM kills with spikes reaching ~2.8Gi against the 2560Mi limit, causing frequent restarts. Bump memory limits and requests from 2560Mi to 4Gi across all staging clusters.


mathur07 (approved):
/lgtm

aThorp96 (approved):
/lgtm

PR #13841, opened 2026-09-02
fixup(PVO11Y-4877): fix spanmetrics ServiceMonitor scrape regex

ci-operator:
ServiceMonitor regex for spanmetrics scraping used `traces_spanmetrics_` but
the connector emits `traces_span_metrics_` (upstream renamed the component from
`spanmetrics` to `span_metrics`). Metrics were not being scraped as a result.


pacho-rh (approved):
/lgtm

PR #13988, opened 2026-09-08
Drop maven-zip-build from default pipelines

chmeliik:
The build-maven-zip tasks are getting deprecated in https://github.com/konflux-ci/build-definitions/pull/3827

This effectively deprecates the entire maven-zip-build pipeline. As of today, nobody has used the pipeline for at least 3 months. Remove it from the default onboarding options to make sure nobody uses it by accident.

## Risk Assessment

Low (None)

These pipelines are not used by any users today and haven't been for a while. Not to mention that removing them here only affects the default pipeline choices for onboarding, nothing else.

## Validation

Not a single Component referenced `maven-zip-build-oci-ta` in at least the last 3 months, so this config change will have no discernible effect in any environment. There's no staging validation that can be done.

### DIFF
--- a/components/build-service/base/build-pipeline-config/build-pipeline-config.yaml
+++ b/components/build-service/base/build-pipeline-config/build-pipeline-config.yaml
@@ -31,10 +31,6 @@ data:
       detail: Build multi-platform container images with secure data transfer for the pipelines. Pipeline can be customized while maintaining trust.
       additional-params:
       - build-platforms
-    - name: maven-zip-build-oci-ta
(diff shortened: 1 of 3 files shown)


fmudry (approved):

ILpinto:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @ILpinto |
| **Timestamp** | 2026-09-22T11:30:28.126Z |
| **Risk Level** | low |

PR #14026, opened 2026-09-09
Add ansible-tenant specific schedule for MintMaker

rh-rsaar:


KristianTkacik (approved):

components/mintmaker/production/stone-prod-p02/create-dependency-update-check-ansible-tenant.yaml
```diff
@@ -0,0 +1,48 @@
...
+                  metadata:
+                    labels:
+                      app.kubernetes.io/name: dependencyupdatecheck
+                      app.kubernetes.io/part-of: mintmaker
+                      app.kubernetes.io/managed-by: kustomize
+                      app.kubernetes.io/created-by: mintmaker
+                    generateName: dependencyupdatecheck-ansible-tenant-
+                  spec:
+                    namespaces:
+                    - namespace: ansible-tenant
```
FernandesMF, line 36: ~I'm thinking that the controller might actually miss this DUC. It uses a cache to watch for resources that are scoped to the mintmaker namespace:
https://github.com/konflux-ci/mintmaker/blob/main/cmd/manager/main.go#L168-L189~

~So I guess we would also have to unscope that cache, for the controller to pick this DUC. I think it wouldn't be a big issue, since we are the only service creating with DUCs.~

~Assisted-by: Claude Opus 4.8~

Edit: I was mistaken, nevermind this

eedri:
## Risk Assessment
> *AI calculated Risk* — based on code changes and PR description.
> Override this by adding your own `## Risk Assessment` section to the PR description.

**Risk Level:** Low
**Description:** This PR adds a new Kubernetes CronJob to a specific production environment, scheduled to run every 4 hours. The CronJob creates a DependencyUpdateCheck custom resource specifically for the ansible-tenant namespace, introducing a new, isolated scheduled task without modifying existing critical components.
**Rollback:** Revert this PR to remove the CronJob from the cluster.

components/mintmaker/production/stone-prod-p02/create-dependency-update-check-ansible-tenant.yaml
FernandesMF: I realized I was not thinking correctly in my previous question.

The DUC itself will reside in the mintmaker namespace, and will be picked by the cache and later the controller.

The spec.namespace field will scope the scannig of components just right.

FernandesMF (approved):
/approve
/lgtm

eedri:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @eedri |
| **Timestamp** | 2026-09-10T07:45:55.107Z |
| **Risk Level** | low |
| **Criteria Override** | Qodo review findings not addressed |

components/mintmaker/production/stone-prod-p02/create-dependency-update-check-ansible-tenant.yaml
```diff
@@ -0,0 +1,48 @@
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: create-dependencyupdatecheck-ansible-tenant
+  namespace: mintmaker
+spec:
+  schedule: "0 */4 * * *" # every 4 hours
```
filariow, line 7: @rh-rsaar will this conflict with the base DependencyUpdateCheck's [cron schedule](https://github.com/redhat-appstudio/infra-deployments/blob/0ccca36d4c9948c3fc7ad927b1843fb478907887/components/mintmaker/production/stone-prod-p02/patch_create-dependency-update-check_cron.yaml#L7)?

Should we use instead:

```suggestion
  schedule: "0 16,20 * * *" # at 16:00 and 20:00 UTC 
```

rh-rsaar: It should not, the logic in the controller is: if a PLR for a repo+branch exists, it won't create a new one, regardless of whether it came from the regular cronjob or the new one.

rh-rsaar: Unless there would be some race condition on the k8s API level when retrieving the PLR list, I'm not that familiar with the inner mechanics of that though, unfortunately.

PR #14048, opened 2026-09-10
add VAP to deprecate Konflux ClusterRoles - prod

filariow:
We want to flag some ClusterRoles as deprecated. We don't want users to use them in new RoleBindings.
At the same time we don't want to break users which already bound to them. The ones who already bound to them will be informed and asked to migrate away.

rh-pre-commit.version: 2.4.0


filariow:
/hold

we are reconsidering the approach

filariow:
we decided not to deprecate the ClusterRoles

PR #14055, opened 2026-09-10
Revert "add VAP to deprecate ClusterRoles"

filariow:
Reverts redhat-appstudio/infra-deployments#13799

We agreed not to deprecate these ClusterRoles.


oswcab (approved):

hugares (approved):
/lgtm

PR #14098, opened 2026-09-11
Increase PaC controller CPU and Memory limits in staging

aThorp96:
It appears that the CPU is being throttled when memory is thrashing, causing the GC cycles to stack up and the pod eventually OOMs

Increasing the CPU limit should help address throttling. Increasing the memory limit shouldn't be necessary once GC is succeeding quickly, but in case there are other memory issues going on it will help


mathur07 (approved):
/lgtm

enarha (commented):
/lgtm

PR #14112, opened 2026-09-14
enable g6.xlarge instances in rh02

filariow:
We want our users to be able to select a VM flavor smaller than `g64.xlarge`.

This change mimics what already provided to `ring-2/kflux-osp-p01` but for `kflux-prod-rh02`.


components/kueue/rings/ring-4/kflux-prd-rh02/cluster-queue.yaml
```diff
@@ -127,6 +127,8 @@ spec:
         nominalQuota: '250'
       - name: linux-g64xlarge-amd64
         nominalQuota: '250'
+      - name: linux-g6xlarge-amd64
+        nominalQuota: '2'
```
oswcab, line 131: Just for my understanding: why do we have such a large difference on the quota, 2 to 250 for other flavors? Some kind of restriction? 
If we want to give users an alternative to bigger machines, shouldn't we provide a (maybe) similar number for these smaller ones?

filariow: it's very easy to incur in EC2's InsufficientInstanceCapacity errors for this flavor and we don't require a lot of concurrency as per our today's usage. We'd likely need to reduce the `nominalQuota` for `linux-g64xlarge-amd64` too, but I'll leave that for a future PR

oswcab: Understood, thanks

mcada:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @mcada |
| **Timestamp** | 2026-09-14T14:32:36.035Z |
| **Risk Level** | very low |
| **Criteria Override** | Staging validation not found |

gcpsoares (approved):
/lgtm

PR #14163, opened 2026-09-16
chore(konflux-info): add jnkubica to production banner OWNERS

bhills-rh:
Add jnkubica as approver and reviewer for banner-content.yaml files across production clusters so core user support can manage UI banners directly in EMEA TZ as well.

Signed-off-by: Ben Hills <bhills@redhat.com>

Assisted-by: OpenCode <opencode@redhat.com>


rrosatti (approved):
LGTM

PR #14165, opened 2026-09-16
fixup(PVO11Y-4877): drop delivery.tekton.dev.application from MTTB span metrics

ci-operator:
Removes `delivery.tekton.dev.application` from the spanmetrics connector's
promoted labels and from the federation allowlist (along with `service_name`).
Adds `series_expiration: 1h` to expire stale label combinations individually
(`metrics_expiration` only expires entire metric names, which never fires
because the cluster always has active spans). Replaces the ServiceMonitor's
blanket drop rules with a keep filter for `traces_span_metrics_` series and a
`service_name` labeldrop.


raks-tt (approved):
/lgtm

raks-tt (approved):

PR #14168, opened 2026-09-16
KFLUXINFRA-4509 Add rotation banner to stg-p01

oswcab:
As part of the Service Account rotation preparations, add banner to the stg-rh01 cluster so users are aware of the upcoming change.


filariow (approved):
/lgtm

PR #14197, opened 2026-09-17
PVO11Y-5575: Deploy Alloy in stage

martysp21:
This is part of https://redhat.atlassian.net/browse/PVO11Y-5575 and deploying Alloy alongside current stack to get resource effectiveness and grasp of how easy it is to deploy.
The deployment is using helm chart as the proposed alloy operator is not subscription based and the deployment proved to be more problematic since we cannot use helm install but only helm template.

The alloy is set up in HA with 3 replicas and respective PVCs. Although not necessary for stage it is for testing. 

It is deployed under Prometheus only in this POC because creating new directory with envs would have bigger impact than I am aiming for. In overall deployment this will move under its own directory.


components/monitoring/prometheus/staging/stone-stg-rh01/poc/alloy/namespace.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: appstudio-alloy-poc
+  annotations:
+    argocd.argoproj.io/sync-wave: "-3"
```
kubasikus: Why is this -3? AFAIK, you just want to create the ns before other resources. In that case, using `-1` should be enough, because other resources use value `0`
```suggestion
    argocd.argoproj.io/sync-wave: "-1"
```

components/monitoring/prometheus/staging/stone-stg-rh01/poc/alloy/namespace.yaml
```diff
@@ -0,0 +1,11 @@
...
+  labels:
+    app.kubernetes.io/part-of: alloy-poc
+    pod-security.kubernetes.io/audit: restricted
+    pod-security.kubernetes.io/enforce: restricted
+    pod-security.kubernetes.io/warn: restricted
```
kubasikus: I haven't seen this used in other namespaces. So why is this needed?

components/monitoring/prometheus/staging/stone-stg-rh01/poc/alloy/namespace.yaml
martysp21: It was a leftover from my tries to deploy operator alloy. Changed to -1

components/monitoring/prometheus/staging/stone-stg-rh01/poc/alloy/namespace.yaml
martysp21: They are not. removed

components/monitoring/prometheus/staging/stone-stg-rh01/poc/alloy/alloy-values.yaml
```diff
@@ -0,0 +1,115 @@
...
+  type: statefulset
+  replicas: 3
+  podLabels:
+    app.kubernetes.io/part-of: alloy-poc
+  podDisruptionBudget:
+    enabled: true
+    minAvailable: 2
+  affinity:
+    podAntiAffinity:
+      preferredDuringSchedulingIgnoredDuringExecution:
```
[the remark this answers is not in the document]
martysp21: For PoC it is acceptable. 
For deployment a topic for discussion.

raks-tt (approved):
/lgtm

ci-operator (approved):
/lgtm

PR #14212, opened 2026-09-18
(onboarding) Remove production external-secrets-operator apps from AppSRE ArgoCD

flacatus:
## Summary
- Add `$patch: delete` entries for `external-secrets-operator` ApplicationSet in both `production-downstream` and `konflux-public-production` overlays
- Part 5 of the ring deployment migration SOP — removing old production AppSets now that rd-production is deployed and healthy
- Follows merge of #14044 (production ring deployment)

## Risk Assessment
- **Level:** Low
- **What could go wrong:** Old ESO ApplicationSet gets removed before rd-production is confirmed healthy
- **Rollback plan:** Revert this PR to restore old AppSet
- **Blast radius:** Production ESO — but rd-production already manages the same workloads

## Validation
- `kustomize build --enable-helm` for both production-downstream and konflux-public-production returns zero `external-secrets-operator` references
- Both overlays build successfully


enkeefe00 (commented):
I think you'll need to delete the entries in both the overlays' Kustomize files too. They have patches applied to them.

enkeefe00 (approved):

PR #14249, opened 2026-09-21
KFLUXVNGD-1360 Retain staging CA Bundles during ownership transfer

hmariset:
Protect the legacy Bundles from ArgoCD pruning and application deletion before the staged handover.

Jira-Url: https://redhat.atlassian.net/browse/KFLUXVNGD-1360
Assisted-by: OpenAI Codex


components/container-image-proxy/rings/ring-1/stone-stage-p01/kustomization.yaml
```diff
@@ -2,3 +2,13 @@ apiVersion: kustomize.config.k8s.io/v1beta1
...
 resources:
 - ../base
+
+patches:
+# Retain the Bundle and its tenant ConfigMaps during the application handover.
+- patch: |-
+    apiVersion: trust.cert-manager.io/v1alpha1
+    kind: Bundle
+    metadata:
+      name: container-image-proxy-ca-bundle
```
amisstea: Do we really need to safeguard this bundle as well? I thought it was only necessary for `caching-ca-bundle`.

hmariset: `container-image-proxy-ca-bundle` is protected because overriding the new chart’s Bundle name to `caching-ca-bundle` removes its original Bundle from the rendered manifests. Keeping both protected preserves any consumers of either name during migration. Once we confirm nothing uses `container-image-proxy-ca-bundle`, we can remove it.

amisstea: It's a new service. Nothing should be using `container-image-proxy-ca-bundle` or else we wouldn't have encountered the trust issue when switching to this service.

hmariset: updated it

amisstea (approved):

PR #14302, opened 2026-09-23
(onboarding) Create new development pvsec AppSet in rd-dev overlay

sguarin:
[PVSEC-13310](https://redhat.atlassian.net/browse/PVSEC-13310)

[PVSEC-13310]: https://redhat.atlassian.net/browse/PVSEC-13310?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sguarin:
> The OWNERS file contains untrusted users, which makes it INVALID. The following users are mentioned in OWNERS file(s) but are untrusted for the following reasons. One way to make the user trusted is to add them as [members](https://github.com/orgs/redhat-appstudio/people) of the redhat-appstudio org. You can then trigger verification by writing `/verify-owners` in a comment.
> 
> * sguarin
>   
>   * User is not a member of the org. User is not a collaborator. Satisfy at least one of these conditions to make the user trusted.
> * jboggs
>   
>   * User is not a member of the org. User is not a collaborator. Satisfy at least one of these conditions to make the user trusted.

@enkeefe00 any advise on this? should I remove us from the OWNERS file, or can we be members of the org?

hugares:
> > The OWNERS file contains untrusted users, which makes it INVALID. The following users are mentioned in OWNERS file(s) but are untrusted for the following reasons. One way to make the user trusted is to add them as [members](https://github.com/orgs/redhat-appstudio/people) of the redhat-appstudio org. You can then trigger verification by writing `/verify-owners` in a comment.
> > 
> > * sguarin
> >   
> >   * User is not a member of the org. User is not a collaborator. Satisfy at least one of these conditions to make the user trusted.
> > * jboggs
> >   
> >   * User is not a member of the org. User is not a collaborator. Satisfy at least one of these conditions to make the user trusted.
> 
> @enkeefe00 any advise on this? should I remove us from the OWNERS file, or can we be members of the org?

Invite sent
image

